PR #2, opened 2022-07-28
Update to latest Mach version

slimsag:
This fixes the issues I noted in Matrix. The only thing I couldn't figure out was the macOS natural scrolling issue (scrolling is both laggy/glitchy, and inverted - I'm not completely sure why)


david-vanderson:
Thank you very much!

PR #3, opened 2022-09-15
Update to latest mach

desttinghim:
This updates the library to work on top of mach after the gpu rewrite.


david-vanderson:
Thank you!

PR #5, opened 2022-09-16
README: update Standalone Mach App build instruction

alichraghi:
`--recurse-submodules` will clones unneeded submodules like examples assets, Spirv-Tools, etc.


david-vanderson:
Thanks!  I was wondering about that.

PR #8, opened 2023-06-22
Fix building on latest version of Zig

AnErrupTion:
This PR fixes building on the latest version of Zig (on master).


AnErrupTion:
@david-vanderson I'm not sure if everything I did is actually correct (like if I chose the correct replacements for methods for example), could you please review?

david-vanderson:
I've merged and will fixup any things I see, but it looks great!  Thanks!

PR #9, opened 2023-07-13
Do not allow scrollbar to scroll beyond content

iacore:
No more glitching

### DIFF
--- a/src/gui.zig
+++ b/src/gui.zig
@@ -5996,6 +5996,12 @@ pub const ScrollInfo = struct {
             .horizontal => self.viewport.x = f * self.scroll_max(dir),
         }
     }
+
+    /// clip viewport within virtual rect
+    pub fn clip_viewport(self: *ScrollInfo) void {
+        self.viewport.x = std.math.clamp(self.viewport.x, 0, self.scroll_max(.horizontal));
+        self.viewport.y = std.math.clamp(self.viewport.y, 0, self.scroll_max(.vertical));
+    }
 };
 
 pub const ScrollContainerWidget = struct {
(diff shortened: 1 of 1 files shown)


david-vanderson:
Thank you for the patch!  Can I ask what you mean by glitching?

The scrolling was intended to be able to scroll a little past the content and bump back to give visual feedback that the user hit the edge.

Is that not working for you, or do you just want the option to turn that off?

iacore:
https://github.com/david-vanderson/gui/assets/74560659/913ed763-89a5-4df0-a541-b44cd45622d6

david-vanderson:
Good video - I guess you are using the mouse wheel right?

I think you are right that for mouse wheel events it should not do that.  When I get to touch events I'll revisit it.  Give me some time to figure out how to integrate your patch.

Thanks!

david-vanderson:
Based on your patch I clamped during event processing (see latest HEAD).  Does that work for you?

iacore:
It works! Thanks!

david-vanderson:
There are 2 ways.  You can increase the font size in the code (either specifically or by creating a different theme or modifying the default theme).  Something along these lines:
```zig
if (try gui.button(@src(), "Increase Font", .{})) {
    themeGet().font_body = themeGet().font_body.resize(@round(themeGet().font_body.size * 1.1));
}
```

Is that what you want?

If you want to scale everything there is a ScaleWidget, which is what drives the "zoom in" and "zoom out" buttons in the GUI Demo window.

iacore:
This doesn't increase all font size.

image

david-vanderson:
Yes - you'd have to do the same thing with each font in the theme.  I'm not sure if there should be a global font scale or not.  Kind of the same situation with the sliders - their size is configured programatically, so (I think?) it's up to the application whether to scale them along with the font size.

The name is a good point.  I don't have a good one, but also am not sure how much it matters.  I guess for searchability a name would help but keeping the library import short is important.  Could keep the library as "gui" but have a longer official name?

iacore:
Name-wise we have imgui libui libui-ng. Mabe dvgui?

PR #10, opened 2023-07-13
Add entypo icons

iacore:
The icons are in SVG format. Icons are from http://entypo.com/.

A word about tinyvg: the converter is in C#, and needs libssl on Linux to build. It doesn't seem to be developed as twice failed to implement the converter.

It might be better for the project to read SVG directly. You only need to support a subset of SVG for icons.


david-vanderson:
Thank you!  I've been meaning to look back at the icon situation.  I still like the tinyvg concept.  I'll see how many of these can be converted.

david-vanderson:
How do you download the icon set?  The download link on entypo.com doesn't work for me.

iacore:
The icons are available as files in this PR. I downloaded it with `wget2 -rk https://entypo.com/`

david-vanderson:
Aha - thanks!  I've converted them and replaced the papirus icons.  Look good?

david-vanderson:
Me neither.  The icon set has a bunch of icons for social media and other internet companies, so I figured it was something like that, but 500px.com doesn't have an infinity sign, so maybe it's old.

iacore:
Do you have a community forum? You can enable Github Discussion

david-vanderson:
Thanks - didn't know about Github Discussions but have turned them on.

PR #13, opened 2023-07-26
Quick fix for building on Windows. Link needed libs

wilsonk:
This is needed to build on my Win10 machine. With this I have tested the three examples on win10, linux (Arch) and macos (aarch64 Monterey). Good stuff.


david-vanderson:
Thank you!

PR #14, opened 2023-07-26
Quick fix for a crash in sdl-test.

wilsonk:
Small fix for a underflow (in debug or ReleaseSafe)...so I guess not really a crash. Use inc and dec functions for uniformity. One could also just use the '**%=' operators to avoid the underflow in the example, but this leads to some strange looking/interacting with the text widget, as there is a wrap to the end of the input with the 'start/end/cursor', and it isn't very consistent with the way I think the widget should work. My opinion only...so take with a grain of salt :)


david-vanderson:
Thank you!

PR #15, opened 2023-07-27
Small nit-pick. I think the arrow_all is more appropriate for the cor…

wilsonk:
…ner of the FloatingWindowWidget. The other arrow looks 'backwards' to me. You don't have to apply if you feel differently :)

Hmm, I will use smaller commit messages in the future and comment more in depth in these notes. Sorry.

### DIFF
--- a/src/gui.zig
+++ b/src/gui.zig
@@ -3792,7 +3792,7 @@ pub const FloatingWindowWidget = struct {
                     if (me.kind == .press and me.kind.press == .left) {
                         // capture and start drag
                         self.captured = captureMouse(self.wd.id);
-                        dragStart(me.p, .arrow_nw_se, Point.diff(rs.r.bottomRight(), me.p));
+                        dragStart(me.p, .arrow_all, Point.diff(rs.r.bottomRight(), me.p));
                         e.handled = true;
                     } else if (me.kind == .release and me.kind.release == .left) {
                         // stop drag and capture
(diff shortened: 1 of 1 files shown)


wilsonk:
Sorry, this seems like it was only an issue on Linux. Windows and Mac cursors look ok (reasonable orientation). They still look reasonable with this change, but it may not be your taste.

I also noticed that the cursors don't seem to be updating properly on Linux/Mac after dragging a corner of the FloatingWindowWidget. Works perfectly on win10, but not on Linux/Mac. I will file an issue.

wilsonk:
Woops, sorry again, I jumped the gun on those cursor issues on Linux/Mac. It ends up it is my VNC software causing the problems and things work fine when logged in directly. Sorry for the spam.

david-vanderson:
No problem at all!

david-vanderson:
Sorry - just to be sure, did you still want this change?  I'm unsure what the best cursor to use is, I tried to match the behavior I normally see on my machines.

wilsonk:
I do still want this change, as otherwise, I see a strange cursor on my Linux box. 

The cursor looks like an arrow head that faces the top-left of the screen when hovering on the bottom-right corner of the floatingWindowWidget ... whereas the cursor looks like an arrow head facing the bottom-right of the screen when I hover over the Main SDL windows bottom-right corner to drag it.  This seems quite incongruous to me to have the arrow face the top-left.

With this change then the cursor on Windows will be the four pointed cross (signifying that one could drag the corner in all directions), on Mac it would be a similar cursor that is four arrows with a dot in the middle. Linux is similar to Mac. So with this change then all three cursors will be pretty similar looking.

david-vanderson:
Sounds good to me.  On my Linux machine it was already showing a four-arrow cursor.  Thanks!

PR #17, opened 2023-07-28
Test proportional window scaling on startup

wilsonk:
Just trying to make the startup of this example a little more readable on my 4k displays (old pet peeve of mine). Only do a proportional scaling if the resolution is above 1080p. Not sure if this should be added as a general option/function in 'gui' or not.

### DIFF
--- a/sdl-test.zig
+++ b/sdl-test.zig
@@ -27,6 +27,9 @@ pub fn main() !void {
     var scale_val: f32 = 1.0;
     var scale_mod: gui.enums.Mod = .none;
 
+    // This will get updated during window startup on displays larger than 1080p resolution
+    scale_val = SDLBackend.proportionalScale;
+
     //var rng = std.rand.DefaultPrng.init(0);
 
     main_loop: while (true) {
--- a/src/SDLBackend.zig
+++ b/src/SDLBackend.zig
@@ -4,6 +4,8 @@ pub const c = @cImport({
     @cInclude("SDL2/SDL.h");
 });
 
+pub var proportionalScale: f32 = 1.0;
+
 const SDLBackend = @This();
 
 window: *c.SDL_Window,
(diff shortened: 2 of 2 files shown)


wilsonk:
Seems to work on 4k win/linux and my Mac at 2k. Haven't tested an actual 1080p output, but it should be fine.

david-vanderson:
Shoot - adjusting the scale_val shouldn't be necessary.  gui.Window already tracks the scale between pixels and "logical pixels" (meaning 1/72 of an inch).  On your 4k display, what does SDLBackend.windowSize() and SDLBackend.pixelSize() report?  Is the scale between those not correct?

wilsonk:
Hmm, on Win10/Mac they both report 800x600 but on my Linux box it is 1600X1200? Not sure why. The sdl-test window looks fine on Win/Mac (with or without the proportional code), but it is almost literally unusable/unreadable on Linux without the proportional code. 

With the ability to scale all the widgets and fonts in the window means this isn't a big deal...unless one doesn't know that is an option, like the very first time someone opens the window. With the ctrl+scroll working this is probably also not as big of  a deal, as most browser users probably default to that action.

I can just deal with this locally and this code doesn't need to be picked up in the main code base. Especially if it might cause issues elsewhere.

wilsonk:
Woops, sorry I was still on the proportional branch for that 1600X1200 above...it says 800x600 on Linux also, for both windowSize() and pixelSize().

david-vanderson:
Hmm okay looks like I didn't understand what SDL what doing.  I can reproduce the issue on my Linux box where changing the display scaling doesn't update gui.Window's natural_scale.  Let me look at the code and see where to integrate that.

wilsonk:
Just as an aside, my Mac mini reports '800x600 and 800x600 natural scale 1', but I am connected to an external 2k monitor, so perhaps that is the difference there. I can test on either of my daughters macbook air or pro, if needed at some point.

Might also try on my Win10 laptop with the builtin screen and the external monitor...might be interesting to see what it reports on each monitor.

david-vanderson:
https://github.com/libsdl-org/SDL/blob/main/docs/README-highdpi.md

Looks like we will have to wait to use SDL3 to get proper support for display scaling.  I'll test to see if compiling with SDL3 fixes this for me.

wilsonk:
Ah yes, well that looks like the issue and fix. I can just deal with it locally until SDL3 stabilizes/releases (ooof, SDL3 is only %36 done since November...so it may be a year or more yet, I suppose).

david-vanderson:
I haven't been able to get any content scale out of SDL3 other than 1.0 on Linux Mint (x11).  I'll try on Mac and Win next.  src/SDLBackend now has a sdl3 flag to help test that.

In the meantime, I've added Window.content_scale that can be set directly.  This should address the subwindow issue.  Does it do what you want?

wilsonk:
Yes, this seems to work as I would assume the scaling stuff would. The only difference is that I had the main window scaling also...this keeps me from having to drag the bottom right out when I scale things a lot. This is just a personal preference, though, I would guess.

david-vanderson:
Do you mean the initial size of the main window?  Hope to address that, but in the meantime since you have extra code to set the content_scale you can multiply the width/height in the call to SDLBackend.init - does that not work for you?

wilsonk:
Actually I believe this PR can be closed as it is pretty much taken care of. The scaling of the main window is actually part of the other open PR. I will move over there.

david-vanderson:
I was finally able to find an SDL2 function that returned a different dpi depending on my linux gnome x11 content scale setting.  Doesn't seem to work on my Mac or Windows (but they have pretty old displays).  I've only integrated it in sdl-test.zig for now.

Does that help at all for you?  What does running sdl-test output for you now?

wilsonk:
Yep, that works fine here on Linux. Thanks.

PR #18, opened 2023-07-29
Example scrolling of demo window

wilsonk:
Just a test of this possible modification. Also added font size changes for the theme (may be useful). The only issue is that I would like for the sub-windows (floatingWindow, debug window, etc.) to scale also, but I couldn't get it working right. I see there is a note in the todo.txt file about this, but I wasn't sure how to go about it effectively.

This code may be dropped, or just parts may be used...up to you David :)

### DIFF
--- a/sdl-test.zig
+++ b/sdl-test.zig
@@ -6,9 +6,11 @@ var gpa_instance = std.heap.GeneralPurposeAllocator(.{}){};
 const gpa = gpa_instance.allocator();
 
 pub fn main() !void {
+    const winWidth = 800;
+    const winHeight = 600;
     var win_backend = try SDLBackend.init(.{
-        .width = 800,
-        .height = 600,
+        .width = winWidth,
+        .height = winHeight,
         .vsync = false,
         .title = "GUI SDL test",
     });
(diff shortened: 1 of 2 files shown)


david-vanderson:
Very cool!  I think what you really want it to modify gui.Window.natural_scale, but there's currently no code to make that work.  Subwindows should already use natural_scale to work the way you want.

I'll see if I can make that work - maybe an extra scale value that is multiplied into natural_scale?

wilsonk:
I think this title should have be 'Example scaling...' but anyways. The only issue I have left after the ProportionalWindowScaling PR was closed is that all windows don't scale when I use the ctrl+scroll action. The scaling works fine with the buttons you added, however, I like to use the ctrl+scroll to scale faster. 

This is really just because of the very small size of the opening fonts/widgets on my Linux box. 

So I can just add in the little bit of code from this PR where I call ``` SDLBackend.c.SDL_SetWindowSize(win_backend.window, @as(c_int, @as(c_int, @intFromFloat(winWidth * scale_val))), @as(c_int, @as(c_int, @intFromFloat(winHeight * scale_val)))); ```, but substitute window.content_scale. It should work fine.

wilsonk:
This is pretty much taken care of (I can modify the one liner locally, if needed) and I will close this stale PR. Thanks.

PR #26, opened 2023-09-03
`TextEntryWidget`: InvalidUtf8 error fixes

r4gus:
This adds a new example and a few bug fixes related to the `TextEntryWidget`:

* Removing a char using backspace now decrements the `TextLayoutWidget` cursor by the size of the utf8 char, i.e. removing the char `a` would decrement it by 1 byte and removing the char `€` would decrement it by 3.
* Same goes for moving the cursor left/ right using the arrow keys.

## Issue

Precondition: The user entered a character that is encoded using more than one byte, e.g., `€`.

Deleting the char using backspace or moving the cursor left/right "over" the char crashes the application with the following trace:

```
error: InvalidUtf8
/usr/local/bin/zig/lib/std/unicode.zig:231:13: 0x2d2521 in init (password)
            return error.InvalidUtf8;
            ^
/home/sugar/dvui/src/dvui.zig:644:21: 0x2d3224 in textSizeRaw (password)
        var utf8 = (try std.unicode.Utf8View.init(text)).iterator();
                    ^
/home/sugar/dvui/src/dvui.zig:617:19: 0x2d3bae in textSizeEx (password)
        const s = try sized_font.textSizeRaw(text, max_width_sized, end_idx, end_metric);
                  ^
/home/sugar/dvui/src/dvui.zig:592:23: 0x2d403e in textSize (password)
            const s = try self.textSizeEx(text[end..], null, &end_idx, .before);
                      ^
/home/sugar/dvui/src/dvui.zig:5282:30: 0x2de609 in addText (password)
                const size = try options.fontGet().textSize(txt[0 .. self.selection.cursor - self.bytes_seen]);
                             ^
/home/sugar/dvui/src/dvui.zig:8094:9: 0x2e28f9 in install (password)
        try self.textLayout.addText(self.init_opts.text[0..self.len], self.wd.options.strip());
        ^
/home/sugar/dvui/src/dvui.zig:7986:5: 0x2e4985 in textEntry (password)
    try ret.install(.{});
    ^
/home/sugar/dvui/examples/password.zig:68:18: 0x2e6327 in password_prompt (password)
        var te = try dvui.textEntry(@src(), .{
                 ^
/home/sugar/dvui/examples/password.zig:37:9: 0x30007a in main (password)
        try password_prompt();
        ^

```


david-vanderson:
This is great - thank you!  I'll figure out how to do similar utf8 multibyte handling for TextLayoutWidget.  Once the password is working, an example should go into the demo window in `basicWidgets()`.

r4gus:
since you're bringing it up: Maybe the `cursor` shouldn't be a offset in bytes but instead a offset in utf8 chars. This offset could then be translated into the correct byte offset for the underlying `u8` slice. Widgets like `TextEntryWidget` could then just do something like `sel.cursor -|= 1;` without problems (or even better, make it a function where the user can just call `sel.cursorLeft(<n-bytes>)`). One special case would be the replacement of one text with a completely other text (where some or all chars have a different encoding), e.g., `€€€€` becomes `abcd`, but this shouldn't be a problem if you always translate the cursor index into the corresponding byte index before accessing/ manipulating the underlying slice.

Just some thoughts

david-vanderson:
I thought about storing it in char offsets a while ago but decided against it.  It makes it more annoying for the widgets themselves, but makes it much easier to interoperate with the rest of the program using normal zig slicing and byte manipulation.

r4gus:
I see

david-vanderson:
I cherry-picked these commits into master, but left out the moving files.  I (currently) expect all examples of widgets to be added to the demo window and not to have separate executable files.

Hope I did this right - I'm not very experienced at merging pull requests yet.

Thanks very much!

r4gus:
I'll close this and open a new pull request once I have something to show for #27

Issue #45, opened 2023-11-05
Feature request: float/vector draggable text field


KnockerPulsar:
Hello, thanks for the work on this library so far!

I think the inclusion of some float/vector widgets might be a big boon to the library in terms of ease of use. What I'm aiming for is something similar to ImGui's `SliderFloat` (correct me if I'm wrong), where you pass in some floats, and the data is updated as you drag or modify the text field. This would mix the functionality of sliders and text fields, with the added bonus of perhaps not needing a min/max like a real slider, just some increment speed. 

So it would be something like:
```zig
const foo = @Vector(3, f32) { 420, 1337, 0 };
dvui.sliderFloat(@src(), f32, 3, &foo, ..); 
```
Where `..` denotes any extra needed parameters (init options and so on).

Forgive me if I got some details wrong, can't quite fire up ImGui's demo right now. I also understand if you're not interested in emulating ImGui. If so, feel free to close this issue.

david-vanderson:
Thank you!

That is a good idea.  I looked at ImGui's slider float and how it works - they have it mode switch to text input when you ctrl-click (or double-click for SliderFloat).  I'm thinking instead we could mode switch on any key event.

They also distinguish between a SliderFloat (has a min/max, single click sets the value) and a DragFloat (optional min/max, only dragging changes the number).  Maybe we can combine them somehow (if you give min/max, do the slider thing, otherwise do the drag?).

There are a lot of details to get right, so it'll take a few iterations of testing.  Maybe we can use zig comptime to automatically handle both floats and ints depending on the type of pointer passed in?

I'll start with a single float, and post again when there's something to test.  Thanks!

KnockerPulsar:
Thanks a lot! I wanted to put together a prototype before posting the issue, but got a bit intimidated by the source code of the library. I'll be closely following your works so I perhaps get to help later on :)

david-vanderson:
Totally understand, and that's something that needs to be improved.  What can be done to make it more approachable?  Is the primary barrier lack of high-level docs talking about how things are put together?  Or is it more lack of source-level comments?  Something else?  Any feedback would be very useful, because it's hard for me to see it with fresh eyes.  Thanks!

KnockerPulsar:
Well, mostly my lack of expertise with Zig. I've written a bit of code with it but I'm just using it as C+ basically. I guess a good way to introduce someone to the codebase would be to show a brief example of how to add an "idiomatic" but simple widget. 

[...]

PR #46, opened 2023-11-10
Draft: Add `sliderVector` widget.

KnockerPulsar:
***NOTE: Work in progress.***
As mentioned in https://github.com/david-vanderson/dvui/issues/45#issuecomment-1803095238, this widget extends `sliderEntry` to allow for vector/array types.


KnockerPulsar:
@david-vanderson What do you think so far? I'm kinda lost on how to pass the options, I went with adding another struct `SliderVectorInitOptions` that's very similar to `SliderEntryInitOptions`. The main difference is that the struct doesn't have a pointer to the data as that wouldn't allow the usage of the widget with both arrays (`[3]f32`) and vectors `@Vector(3, f32)` easily (Would need some casting around). 

I decided to pull the pointer out to its own parameter and do some comptime checking there. I'm pretty sure this won't work with slices but I'm curious for your feedback as well. Other than this, the widget is pretty trivial so I don't have much else in mind to add.

david-vanderson:
Looking pretty good!  Nice comptime logic!

Do you have a motivating example for someone using this?  That might help answer some of these questions.

Thoughts/Questions:
* SliderVectorInitOptions seems good to me
* yes about pulling out the value param to do comptime on it - I ran into the same problem with the label_fmt param
* you are right about slices - they can't do comptime length, but I guess since we have `num_components` we could accept a slice and grab that many elements?
* maybe we could use the same init options for each dimension?  I'm having trouble thinking of a situation where you want different options per dimension.
* usually for compound widgets, most of the Options only apply to the outermost widget, and you would call `.strip()` to get options to pass to the inner components
  * probably `.strip().override()` with some margin or something?
  * this prevents stuff like `.min_size_content` from applying to the individual sliderEntry widgets
* can you explain the "catch false" comment?  It seems to work using `try` for me, but what problem did you see?

KnockerPulsar:
So sorry for disappearing for so long, IRL stuff got the better of me :)

> * you are right about slices - they can't do comptime length, but I guess since we have `num_components` we could accept a slice and grab that many elements?

I think this is doable, would be a bit unsafe if the user passed a slice of incorrect length, but we can't do much on our end.

> * maybe we could use the same init options for each dimension?  I'm having trouble thinking of a situation where you want different options per dimension.

I had this thought before, I guess we can keep things simple and re-add stuff back if anyone needs it.

> * usually for compound widgets, most of the Options only apply to the outermost widget, and you would call `.strip()` to get options to pass to the inner components
>   * probably `.strip().override()` with some margin or something?
>   * this prevents stuff like `.min_size_content` from applying to the individual sliderEntry widgets

Oh I remember  seeing thatt in other widgets. Will do!

> * can you explain the "catch false" comment?  It seems to work using `try` for me, but what problem did you see?

I remember that while manipulating a slider, the following ones didn't get rendered unless I had `catch false`. I do remember trying to remove it right after reading your comment and things seemed to work fine, perhaps something changed between my first prototype and then. 

Anyway, I'll try wrapping this up ASAP since its so trivial.

KnockerPulsar:
Alright, that should be all of the previous feedback. The only change I'm not 100% sure about is the modification of `SliderEntryInitOptions` by pulling out `min`, `max`, and `interval` into `RangeOptions`. Feels a bit like a premature optimization, but I want to hear other point of views as well.

david-vanderson:
No worries, thanks for the work!

I'll try to take it from here:
* I don't know why but the pull request is including some commits that are already in main, so I'll try to merge directly from your branch.
* Looking at it, I think I led you astray with the Options.strip() - it removed the background and padding.  This is the first of this kind of compound widget (that's more of a collection of widgets rather than wrapping them).  I'll play with it.
* I agree about the RangeOptions.  It's not a big deal either way, but let's leave it out for now.

I'll post back after I've merged.

david-vanderson:
Okay I think I've merged it correctly.  Does it look okay?

Thanks very much!

KnockerPulsar:
> Okay I think I've merged it correctly. Does it look okay?
> 
> Thanks very much!

Just tried it out, fantastic! Thanks so much for your patience and help throughout this.

Here's a thing we could create an issue to discuss further: Why not give the user the option to control the size of the slider? Right now, I think it might be too wide for my usage (3D editor).

david-vanderson:
Good point - usually that would be controlled by `.min_size_content` on the containing box, and then we would divide up the given space among the sliders, by passing `.expand = .both` into each sliderEntry.  Want to give that a try?

KnockerPulsar:
> Good point - usually that would be controlled by `.min_size_content` on the containing box, and then we would divide up the given space among the sliders, by passing `.expand = .both` into each sliderEntry. Want to give that a try?

Tried that, seems to shrink the slider to the minimum size that fits the text inside. I was thinking more of modifying this block inside `sliderEntry`:
```zig
if (options.min_size_content == null) {
    const msize = options.fontGet().textSize("M") catch unreachable;
    options.min_size_content = .{ .w = msize.w * 10, .h = msize.h };
}
```
For example, I wanted to make the slider a little thinner when testing the original `sliderEntry` implementation and figured out that the width multiplier is the thing responsible for that. I'm not sure how we can make this configurable, and it's really not that big of an issue. And as you mentioned, the user can probably set the slider as a child of some thinner box and it'll do its thing.

david-vanderson:
I've committed what I hope addresses some of this.  It seemed natural to forward all the options to the sliderEntries.  In the demo now the sliderVector is expanded horizontally.  Is it doing what you want?

Not sure I understand which direction you wanted thinner.  I've also passed a smaller `.font` to make the demo vertically smaller.

In general it doesn't work to cram a widget inside a smaller box, because then you only see part of it.  You usually need to pass `.min_size_content = .{}` or some small value to override the default min size.  Does that make sense?

PR #51, opened 2023-11-18
Add source info to each widget

iacore:
Remember `.id_extra`? Now the error message includes source location as well!

todo

- [ ] discard source info in release mode

Also, instead of `std.debug.print`, this might be better.

```zig
const logger = std.log.scoped(.dvui);
logger.warn(...);
logger.info(...);
logger.debug(...);
```

We can also just have a function like `log(..., .{...})` that handles prefixing "dvui:" to messages and add color.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -319,6 +319,12 @@ pub fn demo() !void {
         try dvui.debugFontAtlases(@src(), .{});
     }
 
+    if (try dvui.expander(@src(), "(Debug) duplicate id", .{ .expand = .horizontal })) {
+        for (0..2) |_| {
+            try dvui.label(@src(), "i should be highlighted", .{}, .{});
+        }
+    }
+
     if (show_dialog) {
         try dialogDirect();
     }
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -1298,12 +1298,17 @@ pub fn minSizeGet(id: u32) ?Size {
     }
 }
 
-pub fn minSizeSet(src: std.builtin.SourceLocation, id: u32, s: Size) !void {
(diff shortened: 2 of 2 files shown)


david-vanderson:
All great additions!  I should have been using the logging stuff already but haven't gotten to it.  Let me know when this is ready to merge.  Thanks!

david-vanderson:
Also for discarding in release mode, you can look at how I did that by searching for `std.debug.runtime_safety`.  I think that's the right way?

iacore:
> All great additions! I should have been using the logging stuff already but haven't gotten to it. Let me know when this is ready to merge. Thanks!

I was hoping you can do the new logging API (in a separate branch), since you have some idea what each `std.debug.print` call is about.

david-vanderson:
Looks good!  You are right probably I should be using `builtin.mode != .Debug` in the other places.  I'll do that.

I'll do the logging, but I don't know how to add style to it (and how to know when we can use color).  But I can do the first step.

david-vanderson:
I just pushed a commit that included reversing some of the .Debug if tests - it looked like they were backwards.  Can you check me on this?

iacore:
> I just pushed a commit that included reversing some of the .Debug if tests - it looked like they were backwards. Can you check me on this?

You are right. I was not being in the right mind :face_exhaling:

> I'll do the logging, but I don't know how to add style to it (and how to know when we can use color). But I can do the first step.

I can add the style. Do you want style? If so, how to style the log text?

david-vanderson:
I've done the logging conversion.  Honestly I don't know if putting in the effort to style the log messages is worth it.  Are you thinking of more than coloring the terminal output?  That's all I can think of.

iacore:
Probably not worth it now, since there are only a few logs.

PR #60, opened 2023-11-25
SDLBackend now allows you to set the minimum and maximum size of the …

r4gus:
…main window

### DIFF
--- a/src/backends/SDLBackend.zig
+++ b/src/backends/SDLBackend.zig
@@ -28,6 +28,16 @@ pub const initOptions = struct {
     /// content of a PNG image (or any other format stb_image can load)
     /// tip: use @embedFile
     icon: ?[]const u8 = null,
+    /// Set the minimum size of the window
+    min_size: ?struct {
+        width: u32,
+        height: u32,
+    } = null,
+    /// Set the maximum size of the window
+    max_size: ?struct {
+        width: u32,
+        height: u32,
+    } = null,
 };
 
 pub fn init(options: initOptions) !SDLBackend {
@@ -179,6 +189,14 @@ pub fn init(options: initOptions) !SDLBackend {
 
     if (options.icon) |bytes| {
         back.setIconFromFileContent(bytes);
+    }
+
+    if (options.min_size) |size| {
+        _ = c.SDL_SetWindowMinimumSize(window, @as(c_int, @intFromFloat(back.initial_scale * @as(f32, @floatFromInt(size.width)))), @as(c_int, @intFromFloat(back.initial_scale * @as(f32, @floatFromInt(size.height)))));
+    }
+
+    if (options.max_size) |size| {
+        _ = c.SDL_SetWindowMaximumSize(window, @as(c_int, @intFromFloat(back.initial_scale * @as(f32, @floatFromInt(size.width)))), @as(c_int, @intFromFloat(back.initial_scale * @as(f32, @floatFromInt(size.height)))));
     }
 
     return back;
--- a/src/backends/SDLBackend.zig
+++ b/src/backends/SDLBackend.zig
@@ -21,23 +21,18 @@ cursor_backing_tried: [@typeInfo(dvui.Cursor).Enum.fields.len]bool = [_]bool{fal
 arena: std.mem.Allocator = undefined,
 
 pub const initOptions = struct {
-    width: u32,
-    height: u32,
+    /// The initial size of the application window
+    size: dvui.Size,
+    /// Set the minimum size of the window
+    min_size: ?dvui.Size = null,
+    /// Set the maximum size of the window
+    max_size: ?dvui.Size = null,
     vsync: bool,
+    /// The application title to display
     title: [:0]const u8,
     /// content of a PNG image (or any other format stb_image can load)
     /// tip: use @embedFile
     icon: ?[]const u8 = null,
-    /// Set the minimum size of the window
-    min_size: ?struct {
-        width: u32,
-        height: u32,
-    } = null,
-    /// Set the maximum size of the window
-    max_size: ?struct {
-        width: u32,
-        height: u32,
-    } = null,
 };
 
 pub fn init(options: initOptions) !SDLBackend {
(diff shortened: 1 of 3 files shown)


iacore:
it seems like every field of `InitOptions` can be updated after the window is created.

idea:

```
pub fn setOptions(self: *@This(), options: InitOptions) void {
    ...
}
```

src/backends/SDLBackend.zig
```diff
@@ -28,6 +28,16 @@ pub const initOptions = struct {
...
+    /// Set the minimum size of the window
+    min_size: ?struct {
+        width: u32,
+        height: u32,
+    } = null,
+    /// Set the maximum size of the window
+    max_size: ?struct {
+        width: u32,
+        height: u32,
+    } = null,
```
AnErrupTion, line 40: I'd recommend you use the ``Size`` struct (in ``src/Size.zig``) here, and maybe even in the "normal" width/height too.

AnErrupTion:
> it seems like every field of `InitOptions` can be updated after the window is created.
> 
> idea:
> 
> ```
> pub fn setOptions(self: *@This(), options: InitOptionsButEveryFieldIsOptional) void {
>     ...
> }
> ```

What's the reasoning behind this? Setting the min/max size of the window requires specific backend API functions, not just updating the width/height of the window (unless I'm misunderstanding something here).

iacore:
> What's the reasoning behind this?

uniform API for setting window options

> Setting the min/max size of the window requires specific backend API functions, not just updating the width/height of the window (unless I'm misunderstanding something here).

all need to call backend API functions

AnErrupTion:
> uniform API for setting window options

In my opinion, we shouldn't assume a backend allows setting min/max dimensions. Also, what would be other options included in this uniform API? We'd need to ensure compatibility with other backends too (not just the ones we already have).

src/backends/SDLBackend.zig
david-vanderson: I thought that too initially, but there's no way to ask for a fractional size right?

david-vanderson (approved):
Looks good - thanks!

src/backends/SDLBackend.zig
r4gus: But then I'd turn the `width` and `height` at the top also into a `Size` otherwise we have two distinct ways to express the window size.

david-vanderson:
@r4gus Am I supposed to merge it?  Or just mark it "Ready for review"?  I'm pretty new to the github reviewing stuff and don't know what a pull request being marked as "draft" means.

I'm good with the addition, but for my own learning, can you say why you want this feature?  I don't yet understand the usecase.  Thanks!

AnErrupTion:
@david-vanderson A draft PR generally means it's incomplete/WIP. However, I don't think it is either here.

david-vanderson:
> > uniform API for setting window options
> 
> In my opinion, we shouldn't assume a backend allows setting min/max dimensions. Also, what would be other options included in this uniform API? We'd need to ensure compatibility with other backends too (not just the ones we already have).

I think you both have good points, but we don't need to stress about it too much at this point.  Each backend can have it's own specific functions, but we'll try to make them relatively uniform if we can.

src/backends/SDLBackend.zig
david-vanderson: Either way is fine (pair of u32 or Size), but it probably makes sense to use the same thing for min/max/initial size.  @r4gus how about you go ahead with what you think is best, and we can change it later if needed?

r4gus:
> @r4gus Am I supposed to merge it? Or just mark it "Ready for review"? I'm pretty new to the github reviewing stuff and don't know what a pull request being marked as "draft" means.
> 
> I'm good with the addition, but for my own learning, can you say why you want this feature? I don't yet understand the usecase. Thanks!

Thanks for the review! Draft means (at least for me) a pull request that is open for discussion and/ or incomplete. It's up to you to accept it, reject it, or merge it.

The use case for me: by limiting the window to a minimum value I can prevent the application from looking cramped which adds to the overall experience. One could also say that it's easier to write a application where you know that it always "looks good" (whatever that means).

AnErrupTion:
> The use case for me: by limiting the window to a minimum value I can prevent the application from looking cramped which adds to the overall experience. One could also say that it's easier to write a application where you know that it always "looks good" (whatever that means).

In theory, a maximum size doesn't really make sense if you make a good application, since it should scale well on any size. This is also partially true for minimum sizes, except of course at some point if the size is too small it's just silly, so it can make sense.

iacore:
by uniform I meant not uniform between backends, but like this:


```
//not uniform
backend.setIconFromFile(...);
backend.setWindowSize(x, y);
//uniform
backend.setOptions(.{.icon = ...});
backend.setOptions(.{.width = x, .height = y});
```

AnErrupTion:
> by uniform I meant like this
> 
> ```
> //not uniform
> backend.setIconFromFile(...);
> backend.setWindowSize(x, y);
> //uniform
> backend.setOptions(.{.icon = ...});
> backend.setOptions(.{.width = x, .height = y});
> ```

I see. However, in this case, why not just put these in the init options of the Backend struct directly, instead of in a separate function?

r4gus:
> > The use case for me: by limiting the window to a minimum value I can prevent the application from looking cramped which adds to the overall experience. One could also say that it's easier to write a application where you know that it always "looks good" (whatever that means).
> 
> In theory, a maximum size doesn't really make sense if you make a good application, since it should scale well on any size. This is also partially true for minimum sizes, except of course at some point if the size is too small it's just silly, so it can make sense.

That's generally true for a maximum value but a minimum value can be quite important. Personally, I find it annoying if you can (accidentally) make the whole application "disappear" (except for the title bar) by dragging it.

src/backends/SDLBackend.zig
AnErrupTion: > But then I'd turn the `width` and `height` at the top also into a `Size` otherwise we have two distinct ways to express the window size.

Yes, that's the intent; this way it's uniform. However I agree with what David said, go with what you think is best.

iacore:
> In theory, a maximum size doesn't really make sense if you make a good application

In practice, (this is Virtual Machine Manager)

image

src/backends/SDLBackend.zig
r4gus: Ok then lets make it a `Size`

r4gus:
So for example this is a good window size for my application:
Screenshot_20231125_161422

But when shrinking the window it starts getting ugly:
noo

david-vanderson (approved):
Looks great - thank you!

david-vanderson:
@r4gus Just to clear, my expectation at this point is that you will merge it.  Let me know if I'm wrong!

AnErrupTion (approved):

PR #68, opened 2024-03-22
Use OKHSL for colors

### LINKED ISSUE #56: use proper colorspace for color operations
cannot lighten black

`0 * (1 + y) == 0`

https://github.com/david-vanderson/dvui/blob/811e8d93f9f55c0a16d18d45e61a2db646198b52/src/Color.zig#L28-L35

iacore:
Second try's a charm.

Fixes #56.


iacore:
Oh no. The window borders are now gone :scream:

See the `ok_color` branch in this repo.

iacore:
I'm now getting NaN. Great.

david-vanderson:
Sorry I haven't gotten to this yet, but I did think that we could use the header as-is instead of translating it to zig.  We could bundle it into the same static lib that the stb_* stuff goes into.  Does that help?

iacore:
> Sorry I haven't gotten to this yet, but I did think that we could use the header as-is instead of translating it to zig. We could bundle it into the same static lib that the stb_* stuff goes into. Does that help?

`Adwaita.zig` uses that at comptime. not good.

david-vanderson:
Ah right.  Well hang on a little longer and hopefully I can figure something out here.

PR #70, opened 2024-05-20
A fix for text entry widgets crashing after upgrading to zig-0.12

wilsonk:
Not sure if this is the way to do it, but this seems to fix the problems I am seeing. Any letter entered into a text entry was crashing with "@memcpy arguments have non-equal lengths". Then the enter key or hitting the end of a buffer was also causing the same crash, so I just changed these occurrences of @memcpy.


david-vanderson:
Sorry about that.  You and I ran into this at the same time, and had essentially the same fix.  Hopefully this is working for you now.  If not, reopen this please.

PR #84, opened 2024-08-01
Moved test files to subdirectory to clean up root directory

VisenDev:
I didn't like how messy the root directory was. To improve this, I moved test files to their own directory. 

Feel free to disregard this PR, organization of files is a personal preference.


david-vanderson:
I agree with the motivation, but these are more like examples, so I've used that name.  I've also taken this opportunity to rename the files to put the backend first, and remove sdl-test (it was really just left over, I should use sdl-standalone or similar for experiments).  Does that help?

VisenDev:
Sounds good

VisenDev:
close

PR #87, opened 2024-08-08
Initial raylib event support

VisenDev:
Super experimental raylib event handling.

Currently the raylib backend is logging events to stderr but dvui is not responding to the events so I'm not sure if I missed something.

### DIFF
--- a/examples/raylib-standalone.zig
+++ b/examples/raylib-standalone.zig
@@ -45,7 +45,6 @@ pub fn main() !void {
 
         // send all SDL events to dvui for processing
         const quit = try backend.addAllEvents(&win);
-        //const quit = try backend.addAllEvents(&win);
         if (quit) break :main_loop;
 
         // if dvui widgets might not cover the whole window, then need to clear
--- a/examples/raylib-standalone.zig
+++ b/examples/raylib-standalone.zig
@@ -28,6 +28,7 @@ pub fn main() !void {
         .icon = window_icon_png, // can also call setIconFromFileContent()
     });
     defer backend.deinit();
+    backend.log_events = true;
 
     // init dvui Window (maps onto a single OS window)
     var win = try dvui.Window.init(@src(), 0, gpa, backend.backend());
@@ -43,7 +44,7 @@ pub fn main() !void {
         try win.begin(std.time.nanoTimestamp());
 
         // send all SDL events to dvui for processing
-        const quit = c.WindowShouldClose();
+        const quit = try backend.addAllEvents(&win);
         //const quit = try backend.addAllEvents(&win);
         if (quit) break :main_loop;
 
--- a/src/backends/RaylibBackend.zig
+++ b/src/backends/RaylibBackend.zig
@@ -159,20 +159,15 @@ pub fn drawClippedTriangles(self: *RaylibBackend, texture: ?*anyopaque, vtx: []c
     const EBO = c.rlLoadVertexBufferElement(idx.ptr, @intCast(idx.len * @sizeOf(u32)), false);
     c.rlEnableVertexBufferElement(EBO);
 
-    const OffsetType = @typeInfo(@TypeOf(c.rlSetVertexAttribute)).Fn.params[5].type.?;
-    _ = OffsetType; // autofix
-    //const expects_ptr = @typeInfo(OffsetType) == .Pointer;
-    const expects_ptr = true;
-
-    const pos = if (expects_ptr) @as(?*anyopaque, @ptrFromInt(@offsetOf(dvui.Vertex, "pos"))) else @offsetOf(dvui.Vertex, "pos");
+    const pos = @as(?*anyopaque, @ptrFromInt(@offsetOf(dvui.Vertex, "pos")));
     c.rlSetVertexAttribute(@intCast(shader.locs[c.RL_SHADER_LOC_VERTEX_POSITION]), 2, c.RL_FLOAT, false, @sizeOf(dvui.Vertex), pos);
     c.rlEnableVertexAttribute(@intCast(shader.locs[c.RL_SHADER_LOC_VERTEX_POSITION]));
 
-    const col = if (expects_ptr) @as(?*anyopaque, @ptrFromInt(@offsetOf(dvui.Vertex, "col"))) else @offsetOf(dvui.Vertex, "col");
+    const col = @as(?*anyopaque, @ptrFromInt(@offsetOf(dvui.Vertex, "col")));
     c.rlSetVertexAttribute(@intCast(shader.locs[c.RL_SHADER_LOC_VERTEX_COLOR]), 4, c.RL_UNSIGNED_BYTE, false, @sizeOf(dvui.Vertex), col);
     c.rlEnableVertexAttribute(@intCast(shader.locs[c.RL_SHADER_LOC_VERTEX_COLOR]));
 
-    const uv = if (expects_ptr) @as(?*anyopaque, @ptrFromInt(@offsetOf(dvui.Vertex, "uv"))) else @offsetOf(dvui.Vertex, "uv");
+    const uv = @as(?*anyopaque, @ptrFromInt(@offsetOf(dvui.Vertex, "uv")));
     c.rlSetVertexAttribute(@intCast(shader.locs[c.RL_SHADER_LOC_VERTEX_TEXCOORD01]), 2, c.RL_FLOAT, false, @sizeOf(dvui.Vertex), uv);
     c.rlEnableVertexAttribute(@intCast(shader.locs[c.RL_SHADER_LOC_VERTEX_TEXCOORD01]));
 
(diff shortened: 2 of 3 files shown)


VisenDev:
I'm having some issues with detecting key input events.

Raylib does not have a generic way to get all key events since the last frame. It can only give you a list of which keys have been pressed - not specific key_up or key_down events.

This creates issues when attempting to use the `GetKeyPressed()` function for modifier keys, since we don't know when the modifer key gets released. 

I will probably have to manually check the status of all modifer keys to be accurate here.

Additionally, it is not clear the difference between a text input event and a key press event. For example - if I press the "a" key, do I send both a text input event and a key press event? 

If I press a modifer key, do I send that as its own keypress event? Or do I cache it and use that modifier key to modify the next key press event? Or do both?

How do I handle multiple modifer keys being pressed? The eventKey function only lets you select a single modifer key?

VisenDev:
The alternative to using `GetKeyPressed` is to keep track of the state of every single key, and then iterate over all keys and check for any changes. 

This would require a lot more boilerplate - but might be necessary if I can't make `GetKeyPressed` work correctly with dvui events.

david-vanderson:
I'm thinking about this too.  I guess we have to map from the current state into events, like you are trying.  I can't think of anything else to do.

My first guess here is on a frame where we have a key pressed, send a key press event to dvui, and record that we saw that key down (only have to store the ones we saw down).

On each frame, first we look at the stored list of keys pressed on last frame, and if any of them are up, send a key release event to dvui and remove it from our stored list.  I think that should cover the modifier keys as well.  A modifier key would also send key press/release events for itself, as well as modifying others - so both.

I think for raylib each key press should generate both a key press event and a text input event.  Since there's no ordering, there's no way to coalesce multiple text input events together, so we just have to rely on the framerate being high enough that there's only ever one per frame.

Does that help?

iacore:
> The alternative to using GetKeyPressed is to keep track of the state of every single key, and then iterate over all keys and check for any changes.

Never use `GetKeyPressed`. I got burned by this.

If you press two keys in one frame, or you press a key and release it and press it again in one frame, functions like `GetKeyPressed` won't reflect that.

Input are events, so treat input as events.

On XInput/X11, if you have key repeat enabled, and you press down the X key, the events it generates look like this:

```
X KeyPress
X KeyPress flags:repeat
X KeyPress flags:repeat
X KeyPress flags:repeat
...
X KeyPress flags:repeat
X KeyRelease
```

Try running `xinput test-xi2 --root` and press some keys.

david-vanderson:
Oh that is interesting, looks like raylib is maintaining some sort of queue of events internally?  I'll have to look more closely at the raylib code.

VisenDev:
> Oh that is interesting, looks like raylib is maintaining some sort of queue of events internally? I'll have to look more closely at the raylib code.

Yes raylib has its own internal queue

VisenDev:
> ## Revised answer
> One approach:
> 
> Call `GetCharPressed()` repeatly until it returns 0. This is the text input event. `GetKeyPressed()` for key event. I think you generate one keypress and one keyrelease together.
> 
> raylib's input system is really bad...

I've almost got something working for GetKeyPressed(). I'm not sure if using GetCharPressed() will fix the problem of a key being pressed and released in a single frame. This is still experimental. 

Here is the implementation of the two raylib functions from `rcore.c`

```c
// Get the last key pressed
int GetKeyPressed(void)
{
    int value = 0;

    if (CORE.Input.Keyboard.keyPressedQueueCount > 0)
    {
        // Get character from the queue head
        value = CORE.Input.Keyboard.keyPressedQueue[0];

        // Shift elements 1 step toward the head
        for (int i = 0; i < (CORE.Input.Keyboard.keyPressedQueueCount - 1); i++)
            CORE.Input.Keyboard.keyPressedQueue[i] = CORE.Input.Keyboard.keyPressedQueue[i + 1];

        // Reset last character in the queue
        CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount - 1] = 0;
        CORE.Input.Keyboard.keyPressedQueueCount--;
    }

    return value;
}

// Get the last char pressed
int GetCharPressed(void)
{
    int value = 0;

    if (CORE.Input.Keyboard.charPressedQueueCount > 0)
    {
        // Get character from the queue head
        value = CORE.Input.Keyboard.charPressedQueue[0];

        // Shift elements 1 step toward the head
        for (int i = 0; i < (CORE.Input.Keyboard.charPressedQueueCount - 1); i++)
            CORE.Input.Keyboard.charPressedQueue[i] = CORE.Input.Keyboard.charPressedQueue[i + 1];

        // Reset last character in the queue
        CORE.Input.Keyboard.charPressedQueue[CORE.Input.Keyboard.charPressedQueueCount - 1] = 0;
        CORE.Input.Keyboard.charPressedQueueCount--;
    }

    return value;
}
```

VisenDev:
And here is where raylib adds key press events to the queue

```c
// Play a recorded automation event
void PlayAutomationEvent(AutomationEvent event)
{
#if defined(SUPPORT_AUTOMATION_EVENTS)
    // WARNING: When should event be played? After/before/replace PollInputEvents()? -> Up to the user!

    if (!automationEventRecording)      // TODO: Allow recording events while playing?
    {
        switch (event.type)
        {
            // Input event
            case INPUT_KEY_UP: CORE.Input.Keyboard.currentKeyState[event.params[0]] = false; break;             // param[0]: key
            case INPUT_KEY_DOWN: {                                                                              // param[0]: key
                CORE.Input.Keyboard.currentKeyState[event.params[0]] = true;

                if (CORE.Input.Keyboard.previousKeyState[event.params[0]] == false)
                {
                    if (CORE.Input.Keyboard.keyPressedQueueCount < MAX_KEY_PRESSED_QUEUE)
                    {
                        // Add character to the queue
                        CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount] = event.params[0];
                        CORE.Input.Keyboard.keyPressedQueueCount++;
                    }
                }
                //...
```

VisenDev:
Alternatively, we could override raylib's glfw `KeyCallback` with our own.

But we would probably only want to do that if we could also trigger the normal raylib `KeyCallback` inside our owns - because raylib would probably break otherwise :/

(Maybe breaking raylib's key input system is okay? dvui users would just have to use the dvui keyboard input system instead of raylib's in their application)

```c
 // Set input callback events
    glfwSetKeyCallback(platform.handle, KeyCallback);
```

VisenDev:
If we implement our own KeyCallback, we could also update raylib's internal state if raylib will let us access `CORE`

```c
// GLFW3 Keyboard Callback, runs on key pressed
static void KeyCallback(GLFWwindow *window, int key, int scancode, int action, int mods)
{
    if (key < 0) return;    // Security check, macOS fn key generates -1

    // WARNING: GLFW could return GLFW_REPEAT, we need to consider it as 1
    // to work properly with our implementation (IsKeyDown/IsKeyUp checks)
    if (action == GLFW_RELEASE) CORE.Input.Keyboard.currentKeyState[key] = 0;
    else if(action == GLFW_PRESS) CORE.Input.Keyboard.currentKeyState[key] = 1;
    else if(action == GLFW_REPEAT) CORE.Input.Keyboard.keyRepeatInFrame[key] = 1;

    // WARNING: Check if CAPS/NUM key modifiers are enabled and force down state for those keys
    if (((key == KEY_CAPS_LOCK) && ((mods & GLFW_MOD_CAPS_LOCK) > 0)) ||
        ((key == KEY_NUM_LOCK) && ((mods & GLFW_MOD_NUM_LOCK) > 0))) CORE.Input.Keyboard.currentKeyState[key] = 1;

    // Check if there is space available in the key queue
    if ((CORE.Input.Keyboard.keyPressedQueueCount < MAX_KEY_PRESSED_QUEUE) && (action == GLFW_PRESS))
    {
        // Add character to the queue
        CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount] = key;
        CORE.Input.Keyboard.keyPressedQueueCount++;
    }

    // Check the exit key to set close window
    if ((key == CORE.Input.Keyboard.exitKey) && (action == GLFW_PRESS)) glfwSetWindowShouldClose(platform.handle, GLFW_TRUE);
}
```

VisenDev:
I'll defer experimenting with overriding `KeyCallback` for a later PR.

Basic functionality should be working now however.

Follow up work: Get the raylib backend to compile on mac and windows (currently only works on linux as far as I can tell)

david-vanderson:
Looks great!  Thank you!

david-vanderson:
Apparently we are using different versions of raylib.  For me, `rlSetVertexAttribute()` last parameter is `int`, but looks like for you it is still `void *`.  What version of raylib are you using?

Here's the commit that changed it:

commit f795941521d038840461e96779aaa097e8a2407a
Author: Ray <raysan5@gmail.com>
Date:   Sat Apr 20 23:37:31 2024 +0200

    REVIEWED: `rlSetVertexAttribute()`, define last parameter as offset #3800

VisenDev:
> Apparently we are using different versions of raylib. For me, `rlSetVertexAttribute()` last parameter is `int`, but looks like for you it is still `void *`. What version of raylib are you using?

I'm not sure since I think I just copied a compiled libraylib.a from some zig cache somewhere. We can probably fix this once we are using a specific raylib version in build.zig.zon.

PR #88, opened 2024-08-09
Build raylib from source using the zig build system

VisenDev:
Raylib now fetched using the zig build rather than relying on system installed libraries. Also removed linux-specific include paths in favor of raylib-bundled headers.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -73,51 +73,20 @@ pub fn build(b: *std.Build) !void {
         }
     }
 
-    // EXPERIMENTAL: for now I make a raylib symlink that points to the raylib git checkout
+    // EXPERIMENTAL raylib backend
     const raylib_mod = b.addModule("RaylibBackend", .{
         .root_source_file = b.path("src/backends/RaylibBackend.zig"),
         .target = target,
         .optimize = optimize,
         .link_libc = true,
     });
-    raylib_mod.addImport("dvui", dvui_mod_raylib);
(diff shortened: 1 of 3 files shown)


david-vanderson:
Thanks!

iacore:
The lazy dependency is required. How does it work in Zig?

VisenDev:
> The lazy dependency (raylib) is required to build dvui. How does it work in Zig?

Could you expand on this question?

iacore:
I don't understand when is `maybe_raylib` null.

VisenDev:
> I don't understand when is `maybe_raylib` null. Can you tell me?

I'm not entirely sure how lazy dependencies work. I just know that if I use `b.lazyDependency` it returns an optional.

https://ziglang.org/documentation/master/std/#std.Build.lazyDependency

david-vanderson:
Took me a while to understand this as well.  The zig build system is a 2 step process:
1. `build.zig` is supposed to build a graph of build steps with all possible top-level steps
2. The top-level steps requested are actually run (with all dependent steps)

`b.lazyDependency` is a way to reference a dependency, but only retrieve that dependency if the build steps requested depend on it.  In that case, if the zig build runner detects that the dependency is actually needed, it downloads the dependency and then stops and reruns the whole `build.zig`.  I'm still unclear as to how exactly the build runner detects when a dependency is required.

https://github.com/ziglang/zig/pull/18778

Issue #89, opened 2024-08-10
Backend functionality for easier integration into existing applications


VisenDev:
The idea is to add functionality to the backends to separate dvui specific initialization - and general window / rendering initialization. 

Existing applications will already have functionality for opening up a window, rendering a frame, etc... with a given backend. 

The dvui user interface code will work the same whether the window was created by dvui or by a user application. So it would be nice to accommodate the use case of desiring this.

This proposal would allow an end user to use dvui only for adding widgets - the end user still manually creates a window and handles rendering frames and so on...

Possible implementation strategy - 

- Separate backend functions that manage the window into two separate functions. For example, `init` becomes `initWindow` and `initDvuiState`. `initWindow` creates the window, and `initDvuiState` initializes backend data fields.

- Other functions like `begin` and `end` would also need to be split up in a similar way

- Thus, users that also want dvui to create a window can call both functions, users that only want to use ui widgets can only initialize the dvui state

- This differentiation could also be accomplished by giving all of the aformentioned functions another parameter to specify which behavior is desired.

david-vanderson:
I agree!  In fact, this is the idea behind having a separate "-ontop" example.  Let's add a "raylib-ontop" example and use that to figure out how this should work (looking at "sdl-ontop" shows a possible way, but it's not great).

I'm pretty sure that only `initWindow` needs a split - `begin` and `end` work on a `dvui.Window` no matter how it was made.  But I could be wrong.

david-vanderson:
Sorry - I should have said.  Can you please add a "raylib-ontop" example (can crib most of it from the sdl-ontop) with your best guess as to the right way to do this?

VisenDev:
> Sorry - I should have said. Can you please add a "raylib-ontop" example (can crib most of it from the sdl-ontop) with your best guess as to the right way to do this?

yes, I'll start experimenting with this

david-vanderson:
Thanks again for working on this.  Sorry - I didn't quite realize how different raylib would be from SDL.  Doesn't look like there is a way to wait with event interrupts.

I've done a pass over it with my own opinions, so what do you think?  I was trying to see how it would work to require more of the underlying raylib API instead of the backend handling that complexity.  If you think that looks okay, I will rework the sdl backend similarly (init vs initWindow like you suggested).

You are the primary driver and user here, so your opinion matters most.  Better or worse?

VisenDev:
[...]

PR #92, opened 2024-08-10
Experimental Implementation of Raylib Ontop

VisenDev:
Experimental implementation of a raylib-ontop example. The `RaylibBackend.init` function is modified to take a parameter which specifies whether the OS window is owner/managed by the end user or by dvui. 

If the OS window is not owned by dvui, then dvui does not create the window or use functions like `BeginDrawing`

See https://github.com/david-vanderson/dvui/issues/89

TODO:
- figure out how to more gracefully control the event system when using the ontop mode. For example, how does dvui tell the user application that a particular mouse click was registered by dvui and does not need to be addressed by the application? Is this important? 
- figure out how to copy once-per-frame data like key-down events, such that the user application can also access such data when using dvui

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -116,28 +116,32 @@ pub fn build(b: *std.Build) !void {
     }
 
     // raylib example
-    {
+    const raylib_examples = [_][]const u8{
+        "raylib-standalone",
+        "raylib-ontop",
+    };
+
+    inline for (raylib_examples) |ex| {
         const exe = b.addExecutable(.{
-            .name = "raylib-standalone",
-            .root_source_file = b.path("examples/raylib-standalone.zig"),
+            .name = ex,
+            .root_source_file = b.path("examples/" ++ ex ++ ".zig"),
             .target = target,
             .optimize = optimize,
         });
 
         exe.root_module.addImport("dvui", dvui_mod_raylib);
         exe.root_module.addImport("RaylibBackend", raylib_mod);
+        if (maybe_ray) |ray| {
+            exe.addIncludePath(ray.path("src"));
+        }
 
-        const exe_install = b.addInstallArtifact(exe, .{});
-
-        const compile_step = b.step("compile-raylib-standalone", "Compile the Raylib standalone example");
-        compile_step.dependOn(&exe_install.step);
-        // not ready for this to be in a standard zig build
-        //b.getInstallStep().dependOn(compile_step);
+        const compile_step = b.step("compile-" ++ ex, "Compile " ++ ex);
+        compile_step.dependOn(&b.addInstallArtifact(exe, .{}).step);
 
         const run_cmd = b.addRunArtifact(exe);
         run_cmd.step.dependOn(compile_step);
 
-        const run_step = b.step("raylib-standalone", "Run the Raylib standalone example");
+        const run_step = b.step(ex, "Run " ++ ex);
         run_step.dependOn(&run_cmd.step);
     }
 
--- a/examples/raylib-ontop.zig
+++ b/examples/raylib-ontop.zig
@@ -1,83 +1,69 @@
 const std = @import("std");
 const dvui = @import("dvui");
 const RaylibBackend = @import("RaylibBackend");
+const ray = @cImport({
+    @cInclude("raylib.h");
+});
+
+var show_dialog_outside_frame: bool = false;
+var vsync = true;
 
 const window_icon_png = @embedFile("zig-favicon.png");
 
-var gpa_instance = std.heap.GeneralPurposeAllocator(.{}){};
-const gpa = gpa_instance.allocator();
-
-const vsync = true;
-
-var show_dialog_outside_frame: bool = false;
-
-pub const c = RaylibBackend.c;
-
-/// This example shows how to use the dvui for a normal application:
-/// - dvui renders the whole application
-/// - render frames only when needed
 pub fn main() !void {
+    var gpa_instance = std.heap.GeneralPurposeAllocator(.{}){};
+    const gpa = gpa_instance.allocator();
+
     defer _ = gpa_instance.deinit();
 
-    // init Raylib backend (creates OS window)
-    var backend = try RaylibBackend.init(.dvui, .{
-        .allocator = gpa,
-        .size = .{ .w = 800.0, .h = 450.0 },
-        .vsync = vsync,
-        .title = "DVUI Raylib Standalone Example",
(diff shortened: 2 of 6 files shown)


VisenDev:
So for some reason objects drawn by raylib are always rendering ontop of objects drawn by dvui. Anyone have any ideas as to why this is happening?

Screenshot_20240810_163217

david-vanderson:
When looking at the raylib code, there was stuff about batching drawing.  Is all the raylib stuff actually drawn at the point that `EndDrawing()` is called?

david-vanderson:
```// By default EndDrawing() does this job: draws everything + SwapScreenBuffer() + manage frame timing + PollInputEvents()```

This comment in the raylib cheatsheet suggests that might be the case.

VisenDev:
> This comment in the raylib cheatsheet suggests that might be the case.

hmm... if so we may need to rework the dvui rendering function so that it integrates with raylibs batching system. I'm not sure if that will be possible or not

david-vanderson:
In `EndDrawing()` the first thing it does is `rlDrawRenderBatchActive();`  If you call that after the raylib drawing and before the dvui stuff, does that change anything?

VisenDev:
> In `EndDrawing()` the first thing it does is `rlDrawRenderBatchActive();`  If you call that after the raylib drawing and before the dvui stuff, does that change anything?

I'm about to go to dinner. I'll test this out when I get back. Thanks!

VisenDev:
The above suggest fixed the layering issue.

However, I have discovered a new issue with the raylib backend, memory usage seems to go up dramatically every time you switch themes. This seems to be caused by the creation of new textures. Whenever I swap themes, dvui uses up another ~250mb of ram. Once I do this enough times my application freezes and I have to kill the process

I think the freezing might be from my mac laptop being low on ram/vram - but the memory issues should probably be addressed.

I also tested with the sdl backend and the memory usage stays around a consistent ~30mb in total when swapping themes.

The way raylib loads and destroys textures may need to be reworked. I'm not completely familar with how texture loading backend functionality is used and managed throughout the rest of dvui

david-vanderson:
I see the memory issues in main, and I also see a slow ramp in memory use even when not doing anything, which probably means we are not freeing something.  Just looking at drawClippedTriangles, it looks suspicious that we call rlLoadVertexArray/Buffer/BufferElement and no matching unload.

Yep - that seems to have fixed it for me.  I've committed that to main.  raylib is now spamming the terminal with INFO messages, so maybe we are still not using these things the right way?

david-vanderson:
Okay got raylib to stop spamming messages by only calling `rlLoadVertexArray()` once during `init()`.

david-vanderson:
Sorry - I forgot to address your TODOs.

For the first, all the `dvui.addEvent...()` functions return a bool - if true it means that dvui thinks that event is for a dvui floating window, and the application should not process it.  I haven't gotten any feedback on this, which either means it works, or that nobody has tried to use it.

I don't quite understand the second.  I expect that when using dvui in "ontop" mode, the underlying game/app would already have some kind of event handling system.  Can you explain more or give an example here?

VisenDev:
> For the first, all the `dvui.addEvent...()` functions return a bool - if true it means that dvui thinks that event is for a dvui floating window, and the application should not process it. I haven't gotten any feedback on this, which either means it works, or that nobody has tried to use it.

Okay this is useful, I will think about how to make this data more exposed to the user.

> I don't quite understand the second. I expect that when using dvui in "ontop" mode, the underlying game/app would already have some kind of event handling system. Can you explain more or give an example here?

Yes, what I meant was for functions like `GetKeyPressed` for raylib or `SDL_PollEvent` - dvui running ontop shouldn't extract all event data internally because that would prevent the application from also using that data for its event processing. 

Knowing this, the sdl-ontop example makes more sense - the user polls the events and then sends them to dvui for extra processing. That approach seems to work for SDL, where all events are processed thru a single api.

However, for raylib, the only functions that give you a stream of events in this manner are `GetKeyPressed` and `GetCharPressed`. 

Perhaps we should implement our own `Event` type that performs a similar function to `SDL_Event` but for Raylib. Our event processing function should split into two functions - one function that produces an array of all events, and another function that registers all of those events.

This would allow for end users to determine whether a specific event was used by dvui, as is done in the SDL example. This would help ease the integration of a floating window for example.

```zig
            if (try backend.addEvent(&win, event)) {
                // dvui handles this event as it's for a floating window
            } else {
                // dvui doesn't handle this event, send it to the underlying application
            }
 ```
 
 (Also - whatever approach ends up being successful, maybe we should standardize it as part of the backend interface so all future backends will also handle ontop applications gracefully)

david-vanderson:
Ooooo I think I get it - the raylib backend is doing all this work to construct this nice event stream, but maybe the app would like to use it outside of dvui?

I think you are on to something there, but let's keep it in the raylib backend until we have another similar backend.  The raylib backend can have lots of functions that are extra on top of the ones needed by dvui.

Maybe the sdl backend way of checking each event is not right for raylib.  For a game using dvui for extra floating windows, maybe we simplify the check to something like "at the beginning of this frame, was a dvui floating window focused?"  If yes, the application ignores input for that frame.  It would have to always send everything to dvui.

Hmmm - I guess it depends on if the app/game is using raylib's event queue or not.  Well in the absence of perfect knowledge, we should strive first to fulfill your usecase.

VisenDev:
I've pretty much done all I want to do with raylib-ontop until #94 and #95 are discussed. 

Follow up work:
- If setting the `window_owner` during initialization is an acceptable way to differentiate ontop and standalone applications - this approach should be standardized across other backends. Variable naming is up for discussion. Perhaps we could use
`enum{ontop, standalone}` instead of `enum{dvui, user}` for this purpose since it more closely reflects the examples.

VisenDev:
As to the best way to handle events, I'm really not sure whats gonna be the best way to do that.
> Maybe the sdl backend way of checking each event is not right for raylib. For a game using dvui for extra floating windows, maybe we simplify the check to something like "at the beginning of this frame, was a dvui floating window focused?" If yes, the application ignores input for that frame. It would have to always send everything to dvui.

I like this idea though because it simplifies things - I'm gonna have to think about the best way to make this simple and consistent for ontop applications. 

Actually I think a really good test bed would be making a theme editor as the raylib-ontop example. I would use the raygui color picker widget along with dvui widgets, which would let me test how I keep raylib and dvui from getting mixed up about event handling.

david-vanderson:
Fantastic work - thank you!

PR #99, opened 2024-08-12
Fill background using color from theme

VisenDev:


david-vanderson:
This brings up the question of what exactly `backend.clear()` should be doing.  For -ontop situations, I see it as a stand-in for the app/games own rendering.  For -standalone situations, dvui widgets should cover the window, and the user should never see this right?

What are you thinking here?

VisenDev:
> This brings up the question of what exactly `backend.clear()` should be doing. For -ontop situations, I see it as a stand-in for the app/games own rendering. For -standalone situations, dvui widgets should cover the window, and the user should never see this right?
> 
> What are you thinking here?

If those are the only two uses of `clear()` then the function could probably be removed. For ontop you can just call the appropriate c function for that backend to accomplish this

VisenDev:
I branched my other PR off of this one so this one is kind of unnecessary

david-vanderson:
Yes I think you are right - I'll remove `clear()` from the backends.

PR #100, opened 2024-08-12
Raylib ontop expansion

VisenDev:


david-vanderson:
Sorry - you want `const rs = overlay.data().contentRectScale()` then `rs.r` gives you a rect in screen pixel coordinates and `rs.s` gives you the scale factor (if you want the raylib drawing to be scaled the same as the dvui stuff).

First, does that work?

Second, if that does work, then I'd like to know where you looked because clearly the documentation has failed here.  Do you remember the first couple of places you looked for this info?

VisenDev:
> Sorry - you want `const rs = overlay.data().contentRectScale()` then `rs.r` gives you a rect in screen pixel coordinates and `rs.s` gives you the scale factor (if you want the raylib drawing to be scaled the same as the dvui stuff).
> 
> First, does that work?

Yes that seems to work, thanks!

> Second, if that does work, then I'd like to know where you looked because clearly the documentation has failed here. Do you remember the first couple of places you looked for this info?

I was primarily looking thru the implementations of BoxWidget, OverlayWidget, and WidgetData to look for any obvious fields or functions like "contentScreenBounds" or something. I also looked at `Example.zig`, particularly the overlay/layout section.

david-vanderson:
Thank you!

Issue #101, opened 2024-08-13
Proposal:  Automatic Struct-To-UI function


VisenDev:
I propose a function with a signature resembling

```zig
pub fn buildStructUI(comptype T: type, opts: StructUIOptions) !T {}
```

This function would accept a struct type as an input. It would then iterate over all fields at comptime and add dvui widgets appropriate for editing all struct fields (number input for `floats/ints`, text input for `[]const u8`, dropdown box for `enums`, etc...)

This would be a very convenient utility function, as it would allow effortless creation of configuration ui elements. It would also be helpful for debugging purposes, as it would allow users to easily edit zig structs at runtime.

Example:

```zig
const Config = struct {
    width: f32,
    height: f32,
    name: []const u8,
    category: enum{old, new},
};
```
`buildStructUI` could automatically generate a ui (or subwindow) that number input fields for `width/height`, a `name` text input field, and a dropdown for `category`

A more advanced use case would be automatically generating an editor for types like `dvui.Theme`

PR #102, opened 2024-08-13
Struct Widget Experiments

VisenDev:
This PR's purpose is to share my experiments addressing https://github.com/david-vanderson/dvui/issues/101 in the hopes that they will be useful

**TODO**
- [x] ~~Swap structEntry to use `Miller Columns` instead of expanders for sub-structs~~     (**Expanders are better**)
- [ ] Add functionality for disabling certain fields 
- [ ] Implement copying and pasting of structEntry sub-entries via std.json serialization
- [x] Implement FieldOptions for all types
- [ ] Implement Arrays, Slices, and Vectors via reorderable lists. Depends on #109


VisenDev:
So here is the idea for the `field_options` parameter. Dvui constructs a struct type based off of the input type, it contains the same field names, but the field types are instead a `TypeOptions` corresponding to the type. Thus, specific options can be passed in to change how different field widget inputs get rendered, like so

```zig
        const TestType = struct {
            a: i8,
            b: u16,
            c: f32,
            g: f64,
            foo: bool,
            bar: dvui.Color,
            bap: enum { red, green, blue, bip },
        };
        var result: TestType = undefined;
        try dvui.structWidget(@src(), TestType, &result, .{
            .a = .{ .widget_type = .slider },
            .foo = .{ .widget_type = .toggle },
            .bar = .{
                .r = .{ .widget_type = .slider },
                .g = .{ .widget_type = .slider },
                .b = .{ .widget_type = .slider },
            },
        });
```

Notably, this does not just accept `anytype`, it actually constructs a corresponding specific `options` type for every struct at comptime.

All of this configuration code is now working - what needs to be done next is actually write the `intInputWidget` and `floatInputWidget` and so on

VisenDev:
Example of simple label layout generated from a `TestType`. No input widgets implemented yet

Screenshot 2024-08-13 at 2 38 20 PM

VisenDev:
More input widgets working
Screenshot 2024-08-17 at 10 25 07 AM

david-vanderson:
This is very cool

VisenDev:
Basic input functionality should be working for most types. I still have a lot of work left on adjusting the UI layout so things look right.

Slice, Array, and Vector types are not really fully developed yet

Sample of a `structEntry` generated from a `dvui.Theme`

Screenshot 2024-08-19 at 2 48 26 PM

VisenDev:
@david-vanderson I've got the basic experimental things working here now. I think what I have is good as an initial implementation and I can finish up adding the features I want in future PRs.

david-vanderson:
Awesome!  Will look at it in the next day or 2.  Thanks!

david-vanderson:
Looks good enough to merge and we can keep working on it in main.  Great work!

david-vanderson:
We might step on each other's toes while working on it, and I might make some changes you don't like.  If that happens please let me know so we can sort it out and make this awesome.

VisenDev:
> We might step on each other's toes while working on it, and I might make some changes you don't like. If that happens please let me know so we can sort it out and make this awesome.

Sounds good

PR #107, opened 2024-08-18
Text entry number improvements

VisenDev:
When working on my struct-ui branch it seemed evident that being able to set a minimum and maximum value for a `textEntryNumber` would be useful

Example use cases,
- A f32 normalized value constrained between 0 and 1
- A f32 percent constrained between 0 and 100
- A f32 "seconds" value constrained between 0 and 60
- An integer "hours" value needing to be constrained between 0 and 12
- An integer "degrees" value needing to be constrained between 0 and 360
- etc...

**Screenshots:**

(Empty)
Screenshot 2024-08-18 at 3 46 18 PM

(With both valid and invalid numbers entered)
Screenshot 2024-08-18 at 3 46 35 PM

I also changed the widget to create its own buffer rather than requiring the user to provide one, as this simplifies usage.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -733,14 +733,29 @@ pub fn textEntryWidgets() !void {
         defer hbox_aligned.deinit();
         left_alignment.record(hbox.data().id, hbox_aligned.data());
 
-        const buf: []u8 = dvui.dataGetSliceDefault(null, hbox.wd.id, "buffer", []u8, &[_]u8{0} ** 20);
-
-        const num = try dvui.textEntryNumber(@src(), T, .{ .text = buf }, .{});
+        const num = try dvui.textEntryNumber(@src(), T, .{}, .{});
 
         if (num) |n| {
             try dvui.label(@src(), "{d}", .{n}, .{ .gravity_y = 0.5 });
-        } else if (std.mem.sliceTo(buf, 0).len == 0) {
-            try dvui.label(@src(), "empty", .{}, .{ .gravity_y = 0.5 });
+        } else {
+            try dvui.label(@src(), "invalid", .{}, .{ .gravity_y = 0.5 });
+        }
+    }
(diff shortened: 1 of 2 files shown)


david-vanderson:
Thanks!  This looks pretty good, but we need 2 more things:

1. Some way to also return the different invalid states, so the code can message the user properly.  In the example, to the right of the text entry we should be able to display "too low", "too high", or "invalid".  What do you think about returning something like `error{TooLow, TooHigh, Invalid}!?f32` where the return is the number if it was valid?

2. Passing in a pointer to a number, because many times an app will need to prepopulate it with a number.  I guess we add an optional pointer into the TextEntryNumberInitOptions?  If it's not null, then it both prepopulates the text with that number and also on a valid change writes the new number back?

VisenDev:
> Thanks! This looks pretty good, but we need 2 more things:
> 
> 1. Some way to also return the different invalid states, so the code can message the user properly.  In the example, to the right of the text entry we should be able to display "too low", "too high", or "invalid".  What do you think about returning something like `error{TooLow, TooHigh, Invalid}!?f32` where the return is the number if it was valid?
> 2. Passing in a pointer to a number, because many times an app will need to prepopulate it with a number.  I guess we add an optional pointer into the TextEntryNumberInitOptions?  If it's not null, then it both prepopulates the text with that number and also on a valid change writes the new number back?

On it

VisenDev:
@david-vanderson How do you feel about how I addressed issues 1 and 2?

david-vanderson:
That looks pretty good!  I like the return strategy.

For #2, can you change `TextEntryNumberInitOptions.initialize` to be a pointer `?*T` and a name like `value` and if the number parses valid, write the number back to that pointer?

VisenDev:
Should I remove`result.Valid` also returning the number?

PR #122, opened 2024-09-10
Added OpenDyslexic

VisenDev:


david-vanderson:
This looks pretty good, some questions:
* Looks like we only need the otf versions, do we need the other versions?  Also the pdf?
* When I'm testing them, the letter spacing seems too large.  Do you think so?
* Should we have have the bold and bold-italic versions loaded by default as well?
* In the pdf, some glyphs look like they have colored parts (like the capital A on page 3).  Am I seeing that correctly?  If so can you add an issue to track supporting colors in font glyphs?

david-vanderson:
I'll take a second look at the letter spacing.

Thank you!

PR #136, opened 2024-09-25
add "backend" module to dvui

marler8997:
Instead of adding an SDLBackend or RaylibBackend module to dvui, this PR always adds it as "backend".  This establishes a simple way to expose it to common application code via `@import("dvui").backend` and more importantly, allows dvui to easily reference the backend directly which enables static comptime interfaces.  This gives dvui more flexibility which will help with the win32 backend I've started.


david-vanderson:
Thank you very much!

iacore:
@marler8997  what the win32 backend is about? it's using OpenGL, right?

marler8997:
Was gonna start with using Direct2D...but, not sure if it has the "triangle drawing ability" that dvui needs out of the box.  If not then I'll likely have to use Direct3D directly.  No plans to use OpenGL.

PR #138, opened 2024-09-26
use Context type directly from the backend

marler8997:
This allows the backend to provide any Context type to dvui and removes the requirement that it be castable to a `usize` value.


david-vanderson:
I merged this, but clearly I didn't do the right thing to show this as being merged.  Sorry, and thanks!

PR #147, opened 2024-10-09
autodoc

VisenDev:
Added autodocs generation

To test autodocs locally, you need to run a http server on localhost and connect in your browser. The easiest way I've discovered to do this is to run `python -m http.server` in `zig-out/docs`. Then you can connect to `http://localhost:8000` in your browser

2024-10-09-092235_hyprshot

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -92,6 +92,22 @@ pub fn build(b: *std.Build) !void {
 
         b.getInstallStep().dependOn(compile_step);
     }
+
+    const docs = b.addObject(.{
+        .name = "dvui",
+        .root_source_file = b.path("src/dvui.zig"),
+        .target = target,
+        .optimize = optimize,
+    });
+
+    const install_docs = b.addInstallDirectory(.{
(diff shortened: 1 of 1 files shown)


iacore:
Nice. Maybe we could copy the `zig std` http server to this repo and run it at `zig build docs`.

VisenDev:
> Or, publish the docs somewhere, like Github pages with CI.

I presumed that is what david would want to do once he was happy with it (similar to how the web example works)

david-vanderson:
Thanks very much!  Yes - I will publish them to github pages alongside the demo.

david-vanderson:
Is there a reason to not build the docs by default with `zig build`?

iacore:
> Is there a reason to not build the docs by default with `zig build`?

it takes time?

VisenDev:
> > Is there a reason to not build the docs by default with `zig build`?
> 
> That it takes time?

Other than that, building it by default should be fine

david-vanderson:
Will have it build by default and see how that affects development speed.

PR #153, opened 2024-10-11
calculator: reset digits_after_dot for second number

cedric-h:
oops, probably should've tested this a bit more thoroughly

Before: Click `9` then `.`, then `9`, click `+`, then `5`

Expected result: `14.9`
Actual result: `9.95`

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -269,10 +269,10 @@ pub fn calculator() !void {
                     digits_after_dot = 0;
                 }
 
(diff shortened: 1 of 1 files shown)


david-vanderson:
For stuff like this, you can directly commit it.  Let me know if that's not working for you, or if you'd prefer to have me merge it.

cedric-h:
It's merged! I'll save MRs for larger initiatives that warrant discussion, rather than quick hotfixes.

PR #154, opened 2024-10-12
prototype demo tabs

cedric-h:


david-vanderson:
I like this a lot, but I will need some time to figure out how to best integrate it:
* I think I want to keep the normal example window and have the demos show in a floating window
* Integrate the toasts
* See if there's a way to add offsets (in addition to the scaling) and see if that helps
* Look into that rectFor thing, you shouldn't need to call it separately
* I want the demo functions to be called directly, not through a function pointer
* I want to add a flexbox layout widget (but can do that after this is integrated

Thanks very much, I'll get started on this!

PR #162, opened 2024-11-08
raylib: implement textureCreateTarget and renderTarget

david-vanderson:
Rendering to the texture produces an upsidedown image.  In the web backend I fixed this by changing the matrix used in the shader.  But that same fix doesn't seem to work in raylib.

The fix is commented out with a TODO

I've also turned off all clipping while testing this.

### DIFF
--- a/src/backends/raylib_backend.zig
+++ b/src/backends/raylib_backend.zig
@@ -12,9 +12,7 @@ pub const c = @cImport({
 const RaylibBackend = @This();
 pub const Context = *RaylibBackend;
 
-var gpa_instance = std.heap.GeneralPurposeAllocator(.{}){};
-const gpa = gpa_instance.allocator();
-
+gpa: std.mem.Allocator = undefined,
 we_own_window: bool = false,
 shader: c.Shader = undefined,
 VAO: u32 = undefined,
@@ -26,8 +24,8 @@ mouse_button_cache: [RaylibMouseButtons.len]bool = .{false} ** RaylibMouseButton
 touch_position_cache: c.Vector2 = .{ .x = 0, .y = 0 },
 dvui_consumed_events: bool = false,
 cursor_last: dvui.enums.Cursor = .arrow,
-frame_buffers: std.AutoArrayHashMap(u32, u32) = std.AutoArrayHashMap(u32, u32).init(gpa),
-texture_sizes: std.AutoArrayHashMap(u32, dvui.Size) = std.AutoArrayHashMap(u32, dvui.Size).init(gpa),
+frame_buffers: std.AutoArrayHashMap(u32, u32) = undefined,
+texture_sizes: std.AutoArrayHashMap(u32, dvui.Size) = undefined,
 fb_width: ?c_int = null,
 fb_height: ?c_int = null,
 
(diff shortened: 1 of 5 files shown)


david-vanderson:
@VisenDev any chance you could take a look at this?

david-vanderson:
@VisenDev Sorry for the noise.  After I walked away I thought "maybe it's our old friend backface culling", and yes.  Cleaning it up now.

VisenDev:
No problem. Want me to still take a look or is it good?

david-vanderson:
Give me a few minutes and then I'd love a review.

david-vanderson:
Okay - how does that look?

src/backends/raylib_backend.zig
VisenDev, line 1: Is there any reason `gpa_instance` was implemented as a global rather than as a field of `RaylibBackend`. Don't we pass an allocator parameter to `initWindow` anyways?

src/backends/raylib_backend.zig
```diff
@@ -279,21 +302,84 @@ pub fn textureCreate(_: *RaylibBackend, pixels: [*]u8, width: u32, height: u32,
...
+    c.rlTextureParameters(texid, c.RL_TEXTURE_WRAP_T, c.RL_TEXTURE_WRAP_CLAMP);
+
+    c.rlFramebufferAttach(id, texid, c.RL_ATTACHMENT_COLOR_CHANNEL0, c.RL_ATTACHMENT_TEXTURE2D, 0);
+
+    // Check if fbo is complete with attachments (valid)
+    if (!c.rlFramebufferComplete(id)) {
+        return error.textureError;
+    }
+
+    self.frame_buffers.put(texid, id) catch unreachable;
```
VisenDev, line 336: Since this function has an error return type, would it not be better to use `try` instead of `catch unreachable`?

src/backends/raylib_backend.zig
```diff
@@ -279,21 +302,84 @@ pub fn textureCreate(_: *RaylibBackend, pixels: [*]u8, width: u32, height: u32,
...
+    self.texture_sizes.put(texid, .{ .w = @floatFromInt(width), .h = @floatFromInt(height) }) catch unreachable;
+
+    self.renderTarget(@ptrFromInt(texid));
+    c.ClearBackground(c.BLANK);
+    self.renderTarget(null);
+
+    return @ptrFromInt(texid);
 }
 
 pub fn renderTarget(self: *RaylibBackend, texture: ?*anyopaque) void {
```
VisenDev, line 346: it might be worthwhile leaving a comment here explaining that passing in a valid texture begins texture mode, and that passing in a null ends it. It took me a second to understand what was happening here.

VisenDev (approved):
Looks good to me, just a few minor suggestions

david-vanderson:
Good points - will address soon.

src/backends/raylib_backend.zig
david-vanderson: Yes it's a bit weird but that's how opengl does it, so it kind of makes sense.

src/backends/raylib_backend.zig
iacore: It is a field.

david-vanderson: The first time I had it as a top-level var inside raylib_backend, and then @VisenDev pointed out that it could be a field so I changed it.

PR #166, opened 2024-11-27
better support focusing inputs in non-fullscreen canvases for the web backend

MichaelBelousov:
- add window scroll and canvas element bounding rect properties to input offset so focus is on the correct position even in non-fullscreen and canvases that must be scrolled to in websites where dvui is rendering into a canvas via the web backend
- without this, you can have a jarring scroll to random parts of the page when clicking on an input in a dvui-controlled canvas while scrolled down


david-vanderson:
Thank you!

Issue #143, opened 2024-10-02
SDL3 Support Status?


LuisJohnsonALR:
Hey,
I just wanted to check in on the status of the SDL3 backend. I noticed a few bits and pieces in the SDL backend file, but it doesn’t seem fully implemented yet. Is SDL3 still on the to-do list, or is it something that’s not being focused on right now?

No rush or pressure at all - I totally get this is a project done in your free time! I’m just asking because I’m planning to use SDL3 for my project, and it would be super helpful if it were supported.

Thanks a lot, and best wishes,
Luis Johnson

david-vanderson:
Thanks for the question!  I think you got the correct impression - at one point I got it working behind the "sdl3" comptime flag but it's not being kept up to date and there's no build integration.

I think that adding build integration for prebuilt sdl3 should be easy - a build.zig -Dsdl3 flag that changes what installed sdl library we look for.  Probably we should not try to compile sdl3 ourselves (we just got sdl2 working in that way).

Does that sound like something you could try making a PR for?

LuisJohnsonALR:
Hey,

Thanks so much for the explanation! That’s more or less what I expected, and it doesn’t sound too difficult to implement. I'm pretty new to this area of programming and not too familiar with SDL3 just yet - I actually just made my GitHub account yesterday so I could ask this question!

I’m still learning about things like Pull Requests, but I’ll see if I can wrap my head around how to approach this. Thanks again for your reply and for all the work you and the contributors have put into this project - it’s really impressive!

One quick question: does this mean we should rely on an SDL3 runtime installed on the system, or is that requirement only for the build process?

Best,
Luis

david-vanderson:
Awesome - welcome!  It can be intimidating when starting, so please ask questions when you get stuck.

Yes - I think to start we should rely on SDL3 runtime.  I'd suggest starting by just hacking build.zig:
```zig
        if (b.systemIntegrationOption("sdl2", .{})) {
                backend_mod.linkSystemLibrary("SDL2", .{}
```
replacing sdl2 and SDL2 with sdl3 and SDL3 (and also setting sdl3=true in sdl_backend) and trying to get that to work on your machine.  Once something is working you can worry about making a Pull Request, or just post the changes you needed.

iacore:
SDL3 is not even out yet!

david-vanderson:
[...]

PR #170, opened 2024-12-22
Add sdl3 support

Deins:
Currently compiles sdl3 from source - build scripts dependency supports windows only. 
I assume for linux using sdl3 from system should be straight forward but haven't implemented build config or looked into.

Backend can be compiled run with:
```
zig build sdl-standalone -Dsdl3
zig build sdl-ontop -Dsdl3 
```
SDL2 config should work as before - unchanged. 

relates to: https://github.com/david-vanderson/dvui/issues/143

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -218,10 +218,22 @@ fn addDvuiModule(
                 dvui_mod.addCSourceFiles(.{ .files = &.{
                     "src/stb/stb_image_impl.c",
                 } });
+                var sdl_options = b.addOptions();
                 if (b.systemIntegrationOption("sdl2", .{})) {
+                    // SDL2 from system
+                    sdl_options.addOption(std.SemanticVersion, "version", .{ .major = 2, .minor = 0, .patch = 0 });
                     backend_mod.linkSystemLibrary("SDL2", .{});
+                } else if (b.option(bool, "sdl3", "Use SDL3 compiled from source") orelse false) {
+                    // SDL3 compiled from source
+                    sdl_options.addOption(std.SemanticVersion, "version", .{ .major = 3, .minor = 0, .patch = 0 });
+                    if (b.lazyDependency("sdl3", .{})) |sdl3| {
+                        backend_mod.linkLibrary(sdl3.artifact("sdl3"));
+                        backend_mod.addImport("sdl3_c", sdl3.module("sdl"));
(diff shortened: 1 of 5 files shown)


david-vanderson:
This is fantastic - thank you!  First question - `@import("sdl3_c")` - what module is that importing?

Deins:
https://github.com/david-vanderson/dvui/blob/53f0379ed43297be03509609e69511321e3fdb4a/build.zig#L231
its https://github.com/swan-www/zig_sdl3/ dependancy, it already provides translated-c module.

david-vanderson:
Thank you - can't believe I missed that!  Give me some time to dig into it.

david-vanderson:
`_ = winapi.AttachConsole(0xFFFFFFFF);` - I can't seem to see how this makes a difference, how do I run sdl-standalone to see what this does?

Otherwise, this looks very good.  Actually compiling the sdl3 doesn't currently work on my windows machine (seems to be missing some windows sdk stuff).  But since it's behind a zig build option I think we merge and continue the work on main.

Thank you!

Deins:
forgot to take it out, `AttachConsole` is unrelated change just for debug, without it stdout/stderr doesn't get printed in console/terminal when program is launched. Can be tested ` with `std.debug.print("TEST\n", .{});` or std.log.

david-vanderson:
Aha got it.  Will think about whether we should keep it in.

For anyone following along, it makes debug output show up in Windows cmd console when running `zig build sdl-standalone` or directly running the binary `.\zig-out\bin\sdl-standalone`.

I normally run the git-bash mingw console on windows.  In that (regardless of whether we run AttachConsole) I never see the output when running `zig build` but do see it when directly running the binary.

Deins:
Also pushed linux support, linking sdl3 from system: https://github.com/david-vanderson/dvui/commit/6bedaf83b9a63a9caf6d195fe0ed479370f2fdf9
Tested on ubuntu 22 in WSL, SDL3 compiled and installed from source on system following SDL3 build instructions. 
Kept flag wise that `-Dsdl3` is for configurations built from source (currently windows only) and `-fsys=sdl3` when linking against system library in case on some platforms bot configurations can be supported.

david-vanderson:
Awesome thank you!

Issue #133, opened 2024-09-20
Figure out accessing system fonts


david-vanderson:
https://github.com/david-vanderson/dvui/issues/121#issuecomment-2362212259 points out that for many languages the default fonts we ship with dvui don't support enough characters.

Unsure how this should work, but it seems like we need some way to pull in system fonts at runtime, and either use them directly or as fallback when the built-in font doesn't have the characters.  This would be optional so apps that only want the fonts they choose to bundle can opt out.

iacore:
We can load ttc/ttf at runtime. That's what nearly every graphical application does, I think.

- rasterization: most type faces are bezier paths. Most, as in not color emoji font, not svg font.
- hinting: opentype text without hinting looks awful.

I feel like using Github issue to track potential features is getting wonky. Since features depend on each other, it might be better to have the textual description of each issue saved somewhere else where we can edit collaboratively. 

Example using git repo itself to store docs: https://codeberg.org/VnPower/PixivFE/src/branch/v2/doc

At the very bare minimum, every feature need to list its dependencies, which are features that may not exist yet.

david-vanderson:
Yes that sounds right - we can have either compiled-in fonts or load them at runtime (or both).  We'll normally have freetype2 compiled in, which will handle rasterization and hinting for us.  In some places (web), we're currently using stb_truetype for rasterization, which doesn't have hinting.

I'm also feeling a bit overwhelmed by the number of open issues we have, but I think it speaks to the growing usefulness of dvui.  Using github issues seems to be what people expect, so let's live with it a bit more and see.  If it is getting out of control we'll figure out something.

VisenDev:
May I suggest either
- Using issue labels to categorize issues better or
- Using **Github Projects** to organize related issues and topics

There are many highly active github repos with thousands of issues - which shows that is it possible to scale the issues system as a project grows

VisenDev:
Perhaps we could start by using some labels like 
- Planned/Accepted (Contributors plan to implement this)
- Bug
- Enhancement
- Low Priority
- High Priority

david-vanderson:
Labels could be useful.  My preference would be for the labels to describe what action we are likely to take.  So maybe like "discussion", and "planned" (meaning enough details are sorted that we could start trying to implement it).  I think that covers all the existing issues?

david-vanderson:
I closed #149 after finding that fonts were the main source of excess size in web-test.wasm.

[...]

PR #172, opened 2025-01-09
SDL2 - Add IME composition text support and example text entry

JaeHeyK:
I tried to add composition support and example for it.
Tested on windows 10 and 11, using windows native IME.

- Korean
dvui_IME_korean

- Japanese
dvui_IME_japanese

- Chinese (Some characters are broken, due to the font)
dvui_IME_chinese

And as for adding fonts, I think the other way would be better, but I've temporarily added it in Font.zig for testing purposes.
It might be better to wait until the discussion below is complete: https://github.com/david-vanderson/dvui/issues/133

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -39,6 +39,7 @@ var slider_entry_max: bool = true;
 var slider_entry_interval: bool = true;
 var slider_entry_vector: bool = false;
 var text_entry_buf = std.mem.zeroes([50]u8);
+var text_entry_ime_buf = std.mem.zeroes([50]u8);
 var text_entry_password_buf = std.mem.zeroes([30]u8);
 var text_entry_password_buf_obf_enable: bool = true;
 var text_entry_multiline_allocator_buf: [1000]u8 = undefined;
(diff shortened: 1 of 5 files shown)


david-vanderson:
Thanks!  It looks like the primary issue here is having enough font support, so I am looking into that.  I don't really want to bundle 10s of megabytes if I can get font loading working.

It looks like I should focus on Google's noto family of fonts, but I'm at a bit of a loss.  Do we have to load a separate font for each language (so 4 separate files to support English, Korean, Japanese, and Chinese)?

JaeHeyK:
>Do we have to load a separate font for each language (so 4 separate files to support English, Korean, Japanese, and Chinese)?

For english, all Noto fonts support english characters by default. So we don't need to think about it seperately.

And if you want to fully support ALL of CJK characters, yes, we need to load all three seperate CJK fonts. Check this site: https://fonts.google.com/knowledge/glossary/cjk

But I think that's too much for single IME composition example. Maybe we can bring single font which partially supports all those CJK characters for demo (e.g., NotoSansKR regular). And if user need more character support, we can let them select their system font for their own language.

david-vanderson:
Thanks that helps.  I'm slowly figuring this out.  We want a good demo where you can enter and see CJK characters (at least some), but also a way to have a slim web demo.

JaeHeyK:
I totally agree with you. And I just checked your recent commits that load the user's font file directly, and that seems like the better approach. I'll close this PR for now.

david-vanderson:
Thank you!

david-vanderson:
@JaeHeyK I've updated the web demo https://david-vanderson.github.io/ so that under the multiline text entry there is a button to download/add a noto font.  Does that work for you?

JaeHeyK:
Yeah, that feature works well on web demo. Great work! Thank you :)

PR #119, opened 2024-09-07
Fix compilation for zig 0.14

raddari:
Compilation succeeds with the command:
`zig build -fsys=sdl2 sdl-standalone`

Only SDL2 backend tested.

Fixes #118.

PR #173, opened 2025-01-10
Update to zig version 0.14.0-dev.2802+257054a14

rohlem:
Continuation of #119 .
Some dependencies also aren't upstream-updated yet, so I created updated forks and linked to them in `build.zig.zon` for now.

With this, on Windows 11 I was able to build and run all examples (`sdl-(standalone|ontop)`, `dx11-(standalone|ontop)`, `raylib-(standalone|ontop)`, `web-test`).
I also tried to get the `-Dsdl3` flag working, and got it from compile errors to "`unable to find artifact 'sdl3'`".
EDIT: However, `web-test` currently requires a workaround that breaks for non-`Debug` optimization modes. (See below.)

Most of the changes should be straightforward, the main motivators being https://github.com/ziglang/zig/pull/21225 and https://github.com/ziglang/zig/issues/16865 / https://github.com/ziglang/zig/pull/21817 .

The only unfortunate change/workaround is in `build.zig`:
https://github.com/ziglang/zig/pull/22240 removed libc support for the target `wasm32-freestanding-musl`.
This means for `web-test`, when compiling module `dvui_mod_web` for this target, libc headers aren't found:

```
...\dvui\src/stb/stb_truetype.h:441:13: error: 'math.h' file not found
   #include <math.h>
            ^~~~~~~~~
...\dvui\src/stb/stb_truetype_impl.c:8:10: note: in file included from ...\dvui\src/stb/stb_truetype_impl.c:8:
#include "stb_truetype.h"
         ^
...\dvui\src/stb/stb_image.h:385:10: error: 'stdlib.h' file not found
#include <stdlib.h>
         ^~~~~~~~~~~
...\dvui\src/stb/stb_image_impl.c:10:10: note: in file included from ...\dvui\src/stb/stb_image_impl.c:10:
#include "stb_image.h"
         ^
src\dvui.zig:63:15: error: C import failed
pub const c = @cImport({
              ^~~~~~~~
...\dvui\src\stb/stb_image.h:370:10: error: 'stdio.h' file not found
#include <stdio.h>
         ^
```

Fwict the primary motivation for this was that `freestanding` means having no system interface, while `musl` as an ABI implies a particular system interface.
However, the headers here are primarily math helper functions. Previously zig would provide these for this target, now it doesn't anymore.

The fix should be to instead specify a libc-available target, such as `wasm32-wasi-musl`, however linking two objects for this target together currently results in a duplicate symbol error:
```
error: wasm-ld: duplicate symbol: __stack_chk_guard
    note: defined in ...\zig\o\19a6733837159280ecaa4b92d137e967\libc.a(>..\zig\o\f831e7109012a241701a8177894b04e1\__stack_chk_fail.o)
    note: defined in ...\dvui\.zig-cache\o\c632e504dcd72bcac6a51b59beb9385f\web-test.wasm.o
error: wasm-ld: duplicate symbol: __stack_chk_fail
    note: defined in ...\zig\o\19a6733837159280ecaa4b92d137e967\libc.a(...\zig\o\f831e7109012a241701a8177894b04e1\__stack_chk_fail.o)
    note: defined in ...\dvui\.zig-cache\o\c632e504dcd72bcac6a51b59beb9385f\web-test.wasm.o
```

I think this might well be an upstream zig issue (those two symbols should be weak or something).
I haven't found mention of this on the GitHub issue tracker (though there's similar issue https://github.com/ziglang/zig/issues/22096 ),
however someone reported it in a `#zig-help` discord thread: https://discord.com/channels/605571803288698900/1302304150792568862

The workaround for now seems to be to target `wasi` for only one object, and `freestanding` for all others.
An upstream report would probably be good, however I don't know much about wasm, so if someone else were able to reduce it to a minimal example, that would be great.

Also note that the target workaround actually **only seems to work for Debug** optimization mode.
With any other optimization mode, zig refuses to link because wasm-ld warns `wasm-ld: warning: Linking two modules of different target triples: '...\dvui\.zig-cache\o\6222e68294243b2dfd92cc297ecb1962\web-test.wasm.o' is 'wasm32-unknown-unknown-musl' whereas 'ld-temp.o' is 'wasm32-unknown-wasi0.1.0-musl'`.
So this really is just a brittle, temporary workaround.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -105,7 +105,7 @@ pub fn build(b: *std.Build) !void {
         const cb = b.addExecutable(.{
             .name = "cacheBuster",
             .root_source_file = b.path("src/cacheBuster.zig"),
-            .target = b.host,
+            .target = b.graph.host,
         });
         const cb_run = b.addRunArtifact(cb);
         cb_run.addFileArg(b.path("src/backends/index.html"));
--- a/build.zig
+++ b/build.zig
@@ -44,7 +44,12 @@ pub fn build(b: *std.Build) !void {
 
     // web test
     {
-        const webtarget = std.Target.Query{
+        const webtarget_library = std.Target.Query{
+            .cpu_arch = .wasm32,
+            .os_tag = .wasi,
+            .abi = .musl,
+        };
+        const webtarget_exe = std.Target.Query{
             .cpu_arch = .wasm32,
             .os_tag = .freestanding,
             .abi = .musl,
@@ -52,7 +57,7 @@ pub fn build(b: *std.Build) !void {
 
         const dvui_mod_web = b.addModule("dvui_web", .{
             .root_source_file = b.path("src/dvui.zig"),
-            .target = b.resolveTargetQuery(webtarget),
+            .target = b.resolveTargetQuery(webtarget_library),
             .optimize = optimize,
         });
 
@@ -69,7 +74,7 @@ pub fn build(b: *std.Build) !void {
         const wasm = b.addExecutable(.{
             .name = "web-test",
             .root_source_file = b.path("examples/web-test.zig"),
-            .target = b.resolveTargetQuery(webtarget),
+            .target = b.resolveTargetQuery(webtarget_exe),
             .optimize = optimize,
             .link_libc = true,
             .strip = if (optimize == .ReleaseFast or optimize == .ReleaseSmall) true else false,
--- a/build.zig
+++ b/build.zig
@@ -51,7 +51,7 @@ pub fn build(b: *std.Build) !void {
         };
         const webtarget_exe = std.Target.Query{
             .cpu_arch = .wasm32,
-            .os_tag = .freestanding, //change to .wasi once, build fails, change back to .freestanding, build succeeds
+            .os_tag = .freestanding, //the initial build of this fails for me. => change to .wasi once, build fails, change back to .freestanding, build succeeds
             .abi = .musl,
         };
 
--- a/build.zig
+++ b/build.zig
@@ -51,7 +51,7 @@ pub fn build(b: *std.Build) !void {
         };
         const webtarget_exe = std.Target.Query{
             .cpu_arch = .wasm32,
-            .os_tag = .freestanding,
+            .os_tag = .freestanding, //change to .wasi once, build fails, change back to .freestanding, build succeeds
             .abi = .musl,
         };
 
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -4,23 +4,22 @@
     .paths = .{""},
     .dependencies = .{
         .sdl = .{
-            .url = "https://github.com/allyourcodebase/SDL/archive/a9107e9ab042a2d6c76c73c0a485352ccb216ffc.tar.gz",
-            .hash = "1220fb0249c15493882350f135bc8b56c389a7f5131cc1b1a5208db472fa00e26b52",
+            .url = "git+https://github.com/allyourcodebase/SDL#a13d7e80e8d847db2b2ff8f334d57eb5bbb489dd",
+            .hash = "12200aa79b05aaeefff144b9e376371e2e7ddc982b9207d146163baf56361331a834",
             .lazy = true,
         },
         .sdl3 = .{
-            .lazy = true,
-            .url = "https://github.com/swan-www/zig_sdl3/archive/refs/tags/3.1.6.tar.gz",
-            .hash = "12203db5d6acf22b69b929ee32194e6352ca9406f0192e7d72dbc2907a1df4e32d8c",
+            .url = "git+https://github.com/rohlem/zig_sdl3-update-zig#8737ef1810285b95b3aa04f534dd85b6e016afbc",
+            .hash = "12207ab67f1a9f7994f5ffcba0b8294360d95ab583026934e750848179fc48dab735",
         },
         .freetype = .{
-            .url = "https://github.com/david-vanderson/freetype/archive/a62b6192d81952e64b8c16504427f3e28ef10e5f.tar.gz",
-            .hash = "12203988c200e30d2fbc2678bb3e49eb3f117cc79faf8d228b63e54d8cdf278378e1",
+            .url = "git+https://github.com/rohlem/freetype-update-zig#a50182dc5787a8d49dd1ad922699a1c8771aa34a",
+            .hash = "12203811b00197ba9f939a8114941bdb18d421c3fa498ec4045bc9ce0c91a5a20b9f",
             .lazy = true,
         },
         .raylib = .{
-            .url = "https://github.com/raysan5/raylib/archive/63ae57d2e31afa8bc8cab7f17f0e44afa4e7552d.tar.gz",
-            .hash = "122078aed9bcba3e2319105d04324bf923a52c65405c3662457cf73e7ef28f445e52",
+            .url = "git+https://github.com/raysan5/raylib#433cc23ea4511fbeddc0fccc597cde91f5756d59",
+            .hash = "1220d18e6fba0a609f3afb055c2876d93e19b3f54f3e2e730c3fd261efd78ccd6002",
             .lazy = true,
         },
         //used in the raylib-ontop example
@@ -31,8 +30,8 @@
         },
         // required for the directx11 backend
         .zigwin32 = .{
-            .url = "git+https://github.com/marlersoft/zigwin32#407a4c7b869ee3d10db520fdfae8b9faf9b2adb5",
-            .hash = "1220cc9c9028e20f4ec2ece1155ae3479acc1cc509f9ab93acb74e8f5bbf8eefd2a9",
+            .url = "git+https://github.com/rohlem/zigwin32-update-zig#406caae350880d15da93d7087f7560897f1a532d",
+            .hash = "1220327265a70f5a41aac138b1571985a7b135c0a3014043be6976a758dfd726fc5d",
             .lazy = true,
         },
     },
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -11,6 +11,7 @@
         .sdl3 = .{
             .url = "git+https://github.com/rohlem/zig_sdl3-update-zig#8737ef1810285b95b3aa04f534dd85b6e016afbc",
             .hash = "12207ab67f1a9f7994f5ffcba0b8294360d95ab583026934e750848179fc48dab735",
+            .lazy = true,
         },
         .freetype = .{
             .url = "git+https://github.com/rohlem/freetype-update-zig#a50182dc5787a8d49dd1ad922699a1c8771aa34a",
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -4,8 +4,8 @@
     .paths = .{""},
     .dependencies = .{
         .sdl = .{
-            .url = "git+https://github.com/allyourcodebase/SDL#a13d7e80e8d847db2b2ff8f334d57eb5bbb489dd",
-            .hash = "12200aa79b05aaeefff144b9e376371e2e7ddc982b9207d146163baf56361331a834",
+            .url = "git+https://github.com/rohlem/SDL#4abe06a7497eea7e5aeb3705951c2fced0491cb4",
+            .hash = "1220647b4508a0226ad9825586e683d68aef6d227fdab84b3c3b029366eec9ef18dd",
             .lazy = true,
         },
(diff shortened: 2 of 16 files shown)


david-vanderson:
This is awesome, I really appreciate it!

MichaelBelousov:
I ran into this earlier (the #zig-help thread is actually mine).
I plan on taking a look at the web workaround this week... hopefully if it's a compiler bug we can get it on the radar for the 0.14.0 release

MichaelBelousov:
@rohlem I can't seem to build your branch on your fork:
```
mike@mikeframework ~/projects/dvui (update-zig-to-0.14.0-dev.2628+5b5c60f43) 11:09:05.874
$ zig build
/home/mike/.cache/zig/p/12207ab67f1a9f7994f5ffcba0b8294360d95ab583026934e750848179fc48dab735/build.zig:101:43: error: expected 1 argument(s), found 2
        const win_sdk = std.zig.WindowsSdk.find(b.allocator, target.result.cpu.arch) catch std.zig.WindowsSdk{ .windows10sdk = null, .windows81sdk = null, .msvc_lib_dir = null }; //try
```
do you know why off the top of your head? Looks potentially related to forks you made of the dependencies.

rohlem:
@MichaelBelousov that function signature was changed from 1 to 2 arguments in https://github.com/ziglang/zig/commit/ba37a4369b936c3fe4ad4f4b1d61bc93d9936018 , so the easiest explanation would be if you're using a Zig version before that.
I just tried cleaning all my caches and rebuilding, and the `web-test` step failed with "file not found" this time (still on version `0.14.0-dev.2628+5b5c60f43` though).
I'll try updating to current downloadable master, getting it to work again, and reporting the steps to get there if simple `zig build` doesn't cut it.

MichaelBelousov:
that makes perfect sense, thank you for pointing that out, I did in fact just try using a an older 0.14.0 I had lying around and hoped it would work. Sorry I should have tried upgrading earlier. I'll take a look later again

rohlem:
Okay, when initially building step `web-test` with zig version `0.14.0-dev.2628+5b5c60f43`, the build for me (on Windows 11) fails with what looks like an linker / file management error internal to zig:
```
error: warning(compilation): failed to delete '...\dvui\.zig-cache\tmp\5a56515740f53e9a-stb_truetype_impl.o.d': FileNotFound
warning(compilation): failed to delete '...\dvui\.zig-cache\tmp\a8a96324d3ea619b-stb_image_impl.o.d': FileNotFound
```
To make these sub-compilations succeed, `webtarget_exe.os_tag` has to be `.wasi` instead of `.freestanding`.
This will fail the combined compilation, but the sub-compilation succeeds, after which combined compilation (with `webtarget_exe.os_tag = .freestanding`) works again.
(I'm adding a commit with that an inline comment in `build.zig`.)
It's possible that the new wasm-linker work in https://github.com/ziglang/zig/pull/22220 could fix / interact with this, no idea.

On the front of updating to current ziglang.org-downloadable master (`0.14.0-dev.2802+257054a14`), that's a bit more messy.
Some translate-c output of an SDL error now generates a type error on bit-or `|` between `c_int` and `c_uint`.
I didn't see any commits referencing translate-c, so maybe this is a side-effect of a change in zig's own type analysis?
I'll try reducing it into an upstream issue.
Since a workaround would probably be too involved I'll keep that version (with broken SDL integration) on a separate branch for now.

david-vanderson:
> Some translate-c output of an SDL header now generates a type error on bit-or | between c_int and c_uint.

https://github.com/ziglang/zig/issues/21220 and I saw somebody on discord work around it by doing an explicit `@as(ENUM_TYPE, ENUM_VALUE)`

rohlem:
> > Some translate-c output of an SDL header now generates a type error on bit-or `|` between `c_int` and `c_uint`.
> 
> [ziglang/zig#21220](https://github.com/ziglang/zig/issues/21220) and I saw somebody on discord work around it by doing an explicit `@as(ENUM_TYPE, ENUM_VALUE)` but unsure if that helps here?

Eh, that's also related to types, but a bit of a different scenario.
I know how I would fix the code (either wrap `X` in `@as(c_uint, @bitCast(X))` in Zig, or explicitly cast the argument in C),
but if the C code is valid (and I believe it is), it's a translate-c issue that should be fixed upstream.

The explicit cast in C would probably be the better workaround in the meantime.
I didn't really want to invest more time, but I guess I'll try it.

rohlem:
> I don't meant to pressure you into sinking more time into it. I very much appreciate the effort you've already put in!

No worries! I was just wary of maybe a dozen more issues creeping up, but that constant was the only place that triggered a compile error.

I rebased onto current `main` branch and added a commit to update to `0.14.0-dev.2802+257054a14`.
(But kept my branch name because iirc renaming can cause issues like the PR closing.)

`zig build` of all steps now works for me, except that `web-test` requires the `webtarget_exe` workaround in `build.zig` for the first build - see [comment here](https://github.com/rohlem/dvui/blob/af5613138eb0b5fff7b399e857ec31f31f25193d/build.zig#L54) and [earlier issue comment](https://github.com/david-vanderson/dvui/pull/173#issuecomment-2599731754).
Since target `.wasi` succeeds with the sub-compilation, compiling both modules to `.wasi` would be the ideal solution (and is probably intended),
the only blocker being the duplicate symbol error on `__stack_chk_guard` and `__stack_chk_fail`. (still unreduced, not yet reported upstream)

david-vanderson:
Thank you!!

david-vanderson:
@rohlem I'm getting this when compiling against sdl2:

/home/dvanderson/.cache/zig/p/12200aa79b05aaeefff144b9e376371e2e7ddc982b9207d146163baf56361331a834/src/stdlib/SDL_string.c:373:12: error: call to undeclared function 'wcslcpy'; ISO C99 and later do not support implicit function declarations
    return wcslcpy(dst, src, maxlen);

I'm going to work around it for now, but does that ring any bells for you?

rohlem:
Nope, haven't run into that. It reads like a missing `#include` in SDL's source?
Apparently it's [a POSIX function](https://pubs.opengroup.org/onlinepubs/9799919799/functions/wcslcpy.html); `#include <wchar.h>` should be the correct include for it, which looks like it's included in `SDL_string.c` via `SDL_stdinc.h`.
Are you compiling with system libc or zig provided? gnu.org says it's ["missing on many systems"](https://www.gnu.org/software/gnulib/manual/html_node//wcslcpy.html).

david-vanderson:
Good question, I'm not sure.  We are setting .link_libc = true in build.zig and I'm getting the error from:

`zig build-lib <many SDL files> -ODebug -I /home/dvanderson/.cache/zig/p/12200aa79b05aaeefff144b9e376371e2e7ddc982b9207d146163baf56361331a834/include -I /home/dvanderson/code/dvui/.zig-cache/o/8baaa61a3c1a2bdde19252fa228d348c/ -I /home/dvanderson/code/dvui/.zig-cache/o/76093cf1769599eadcd71b2e28c303f0/ -DSDL_USE_BUILTIN_OPENGL_DEFINITIONS=1 -DHAVE_GCC_ATOMICS=1 -DHAVE_GCC_SYNC_LOCK_TEST_AND_SET=1 -DUSING_GENERATED_CONFIG_H= -DSDL_VIDEO_RENDER_SW=1 -Mroot -lc --cache-dir /home/dvanderson/code/dvui/.zig-cache --global-cache-dir /home/dvanderson/.cache/zig --name SDL2 -static --zig-lib-dir /home/dvanderson/apps/zig-linux-x86_64-0.14.0-dev.2851+b074fb7dd/lib/ --listen=-`

I always assumed it was the system libc when I'm not cross compiling.  This is on Linux Mint 21.3.

rohlem:
> I always assumed it was the system libc when I'm not cross compiling. This is on Linux Mint 21.3.

I believe that's true. It might change as soon as you specify a `-Dtarget` parameter because that then counts as cross-compilation (doesn't optimize to your specific cpu model)?
What should definitely work is specifying a particular version of `glibc`. If I'm googling correctly they were introduced in glibc 2.38, so `"-Dtarget=x86_64-linux-gnu.2.38"` / `"-Dtarget=native-native-gnu.2.38"` should do the trick.

david-vanderson:
Thanks very much for the info - you are totally right.  Looks like my system is still on glibc 2.35.  I agree - we'll keep them disabled for now.  Thanks!

MichaelBelousov:
am I still building this wrong @rohlem?

```sh
$ cd dvui
$ git checkout e50cfbf7d3cba03552cd85a9b368a0940b1f9d9b # the merge commit from this PR
$ ~/opensource/zig-linux-x86_64-0.14.0-dev.2802+257054a14/zig
0.14.0-dev.2802+257054a14
$ ~/opensource/zig-linux-x86_64-0.14.0-dev.2802+257054a14/zig build web-test
```

My version looks correct according to the PR title, I see it changed once in the convo
I get:

```
error: warning(compilation): failed to delete '/home/mike/projects/dvui/.zig-cache/tmp/9b38c3e5d56fb01f-stb_truetype_impl.o.d': FileNotFound
warning(compilation): failed to delete '/home/mike/projects/dvui/.zig-cache/tmp/712c0570eb366c06-stb_image_impl.o.d': FileNotFound
```

MichaelBelousov:
nevermind, I see the comment now about having to build .wasi first :(

MichaelBelousov:
non-debug mode builds for me with the following patch which I could turn into a PR,
but there is a runtime error which I would guess is a zig compiler bug, and I will try to make an MVR and submit a bug report:

```diff
diff --git a/build.zig b/build.zig
index 2559c8b..21ab254 100644
--- a/build.zig
+++ b/build.zig
@@ -51,7 +51,7 @@ pub fn build(b: *std.Build) !void {
         };
         const webtarget_exe = std.Target.Query{
             .cpu_arch = .wasm32,
-            .os_tag = .freestanding, //the initial build of this fails for me. => change to .wasi once, build fails, change back to .freestanding, build succeeds
+            .os_tag = .wasi, //the initial build of this fails for me. => change to .wasi once, build fails, change back to .freestanding, build succeeds
             .abi = .musl,
         };

diff --git a/examples/web-test.zig b/examples/web-test.zig
index cc3f67a..0b2ba9e 100644
--- a/examples/web-test.zig
+++ b/examples/web-test.zig
@@ -219,3 +219,7 @@ fn dvui_frame() !void {
         win.content_scale = ns;
     }
 }
+
+export fn main(_: u32, _: u32) u32 {
+    return 0;
+}
diff --git a/src/backends/WebBackend.zig b/src/backends/WebBackend.zig
index 832f780..511c02d 100644
--- a/src/backends/WebBackend.zig
+++ b/src/backends/WebBackend.zig
@@ -57,8 +57,8 @@ pub const wasm = struct {
     pub extern fn wasm_add_noto_font() void;
 };

-export const __stack_chk_guard: c_ulong = 0xBAAAAAAD;
-export fn __stack_chk_fail() void {}
+//export const __stack_chk_guard: c_ulong = 0xBAAAAAAD;
+//export fn __stack_chk_fail() void {}

 export fn dvui_c_alloc(size: usize) ?*anyopaque {
     //std.log.debug("dvui_c_alloc {d}", .{size});
```

Basically, in my zig-help thread I encountered a similar issue and noticed that dvui exports its own `__stack_chk_guard` for some reason, which I assume was due to using a freestanding and not a wasi target. Now that we need to ship wasi, I assume this isn't necessary, but I have not `git blame`'d to find out why this was there in the first case.

Also note that I had to add a fake main function for the wasm linker, I assume this zig version's compiler or build system's handling of `exe.entry = .disabled` is broken.

So it builds now but (regardless of optimization setting) at runtime the webassembly fails to import with:

```
Uncaught (in promise) TypeError: WebAssembly.instantiate(): Import #0 "wasi_snapshot_preview1": module is not an object or function
```

I will look into that.

MichaelBelousov:
I have an MVR, I will talk to zig folks, it doesn't like like wasm is in a good state on master

MichaelBelousov:
After some testing, thinking and [asking zig-help](https://discord.com/channels/605571803288698900/1331648530611175477), it looks like the way we _should_ do this is
by linking a non-standard and ideally partial libc to satisfy the C dependencies like `stb_image.h`, and then keep building freestanding. WASI is not really a good option for the browser target.

An alternative is to use something like [`jco`](https://github.com/bytecodealliance/jco) to polyfill the wasi env while loading the web test. I am afraid this may significantly bloat dvui with stuff like file system APIs that are very much unused in most applications.

PR #175, opened 2025-01-18
Better SDL3 integration

mdmrk:
Replace current SDL3 source upstream for [castholm/SDL](https://github.com/castholm/SDL) as it allows for better support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,13 +14,9 @@ How to run the built-in examples:
   - then load `zig-out/bin/index.html`
   - [online demo](https://david-vanderson.github.io)
 - SDL3:
-  - Windows (SDL3 built from source): 
-    - ```zig build sdl-standalone -Dsdl3```
-    - ```zig build sdl-ontop -Dsdl3```
-  - Linux (SDL3 dev must be installed on system):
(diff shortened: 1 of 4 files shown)


Deins:
Just tested on windows.
If this works on linux by compiling from source & linking statically, I suppose we can get rid of `-fsystem` build variant. I suspect this commit breaks it. 
https://github.com/david-vanderson/dvui/blob/de8f7931a2a92f4bb1053835e9124505d251f9b3/build.zig#L237

Otherwise looks good. Thanks.

mdmrk:
Working on Linux x64. I tried cross compilation both to Linux and Windows but there's still some compilation/linking errors.

However I got Linux x86_64 cross compilation working on my machine by using this command:

 ```bash
zig build sdl-standalone -Dsdl3 -Dtarget=x86_64-linux-gnu.2.38 
```

Removing 2.38 suffix or using a lower version causes undefined symbol for multiple __isoc23_* functions.

mdmrk:
This last commit fixes cross compilation issues.

Deins:
Tested linux (wsl) build and also cross compilation works, nice!

david-vanderson:
Thank you very much!

PR #180, opened 2025-01-22
Update sdl3 to stable 3.2.0

mdmrk:


david-vanderson:
I like updating sdl3, but I think the compilation stuff is going backwards?  Can you retry with latest nightly zig?

mdmrk:
So far these are my results compiling this branch against sdl3

`0.14.0-dev.2628+5b5c60f43`: passes
`0.14.0-dev.2577+271452d22`: passes
`0.14.0-dev.2851+b074fb7dd` (master): [fails](https://pastebin.com/C4PsDNuC)

david-vanderson:
I must be missing something.  It looks like your second commit modified things so that they wouldn't work with the latest zig `0.14.0-dev.2851+b074fb7dd`.  Is that correct?  It seems like we should only use the first commit (the one that updates build.zig.zon)?

mdmrk:
zig master's fails to compile sdl3 dep. the other two zig dev versions mentioned worked completely for me applying the second commit patches. I might be wrong somewhere but that's what I found.

david-vanderson:
Okay got it thanks.  Yes things are going to be broken for at least a little bit as we try to sync up to zig master leading up to the 0.14 release.  I believe the sdl3 compilation errors on zig master are fixed in git, but I'm waiting for a new nightly to be posted.  Will report back once that happens.

david-vanderson:
zig 0.14.0-dev.2989+bf6ee7cb3 works and I've updated to that version of SDL3.  Please reopen if that doesn't work for you.  Thank you!

PR #185, opened 2025-02-21
fix for zig 0.14.0-dev.3271+bd237bced

nhanb:
- `Allocator.rawFree()` now takes `alignment` as an enum, probably an example of <https://ziglang.org/devlog/2024/#2024-11-04>
- `popOrNull()` is now just `pop()`

### DIFF
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -2506,7 +2506,11 @@ pub const Window = struct {
 
         pub fn free(self: *const SavedData, allocator: std.mem.Allocator) void {
             if (self.data.len != 0) {
-                allocator.rawFree(self.data, @ctz(self.alignment), @returnAddress());
+                allocator.rawFree(
+                    self.data,
+                    std.mem.Alignment.fromByteUnits(self.alignment),
+                    @returnAddress(),
+                );
             }
         }
     };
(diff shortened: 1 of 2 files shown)


src/dvui.zig
```diff
@@ -2506,7 +2506,7 @@ pub const Window = struct {
 
         pub fn free(self: *const SavedData, allocator: std.mem.Allocator) void {
             if (self.data.len != 0) {
-                allocator.rawFree(self.data, @ctz(self.alignment), @returnAddress());
+                allocator.rawFree(self.data, @enumFromInt(@ctz(self.alignment)), @returnAddress());
```
nhanb, line 2509: Tbh memory alignment is way over my head, so please correct me if this "fix" is wrong...

src/dvui.zig
```diff
@@ -2506,7 +2506,11 @@ pub const Window = struct {
 
         pub fn free(self: *const SavedData, allocator: std.mem.Allocator) void {
             if (self.data.len != 0) {
-                allocator.rawFree(self.data, @ctz(self.alignment), @returnAddress());
+                allocator.rawFree(
+                    self.data,
+                    std.mem.Alignment.fromByteUnits(self.alignment),
```
nhanb, line 2511: Tbh memory alignment is way over my head, so please correct me if this "fix" is wrong...

david-vanderson:
Thanks!  This is great - I'm seeing at least one more issue on my Linux machine (`isAndroid()` went away?), so give me some time to test on all my machines.

david-vanderson:
Thank you!

PR #190, opened 2025-02-27
render newlines in text selection

sno2:
Both Chrome and Monaco (VS Code's editor) render newlines like this at the end of the line. We don't match their behavior completely because you cannot yet drag a selection to the right to include a newline without dragging to the next line. However, this can be added later.

On the other hand, Firefox does not render newlines at all even in <textarea> elements so this is slightly opinionated. However, rendering the newlines is useful when interacting with content so I think this is the way to go.

---

Here are some screenshots of various behaviors:

DVUI (this PR):
dvui paragraph with this pr
dvui textarea with this pr

Chrome:
chrome p tag and textarea tag

Firefox:
firefox p tag
firefox textarea tag

Monaco:
monaco newline handling


sno2:
Firefox still does show a space if the text is wrapping. When I was showing these examples, I used a \<br> tag to make them have the same shape, but if you do not have an explicit break, it shows the space before the line break in selections.

image

david-vanderson:
I want to make sure I understand the intent, so let me try to restate.

The intent is to render a newline character as a space, so you can see when you are selecting the newline character.

Is that correct?  If so, I'm thinking it makes sense in textLayout and textEntry (anywhere you can select it) but not in labels or buttons.  Does that make sense?

sno2:
> The intent is to render a newline character as a space, so you can see when you are selecting the newline character.

Yes, you are correct.

> If so, I'm thinking it makes sense in textLayout and textEntry (anywhere you can select it) but not in labels or buttons. Does that make sense?

I think you are right. I thought these would still be selectable, but it does not seem possible. I'll make sure those are not included.

david-vanderson:
Okay I think I was able to do it at the TextLayoutWidget level.  Please test.

For selection behavior, I'm seeing (on Chrome) that you can't select the newline just by dragging the mouse to the right of the line, you have to move the mouse down enough to be on the next line.  Can you double check that?

sno2:
Yes, that does work as well. I was trying to avoid the allocation but this is cleaner.

> For selection behavior, I'm seeing (on Chrome) that you can't select the newline just by dragging the mouse to the right of the line, you have to move the mouse down enough to be on the next line. Can you double check that?

Chrome only lets you include the newline on a single line selection if you did some method other than clicking once and dragging to create the selection. For example, you could do "Shift + RightArrow," double-click on a word and drag right, or triple-click to include the line with the newline.

david-vanderson:
Yeah - I tried to avoid it as well but at least it's rare.  If it becomes a problem we can do a trick with rendering the space by itself, but it would need additional machinery to make the selection background contiguous.

Okay good - can deal with the selection stuff in a separate issue if needed.

Thanks very much!

Issue #194, opened 2025-03-02
dvui.Window generic over backend


david-vanderson:
@alichraghi had the excellent suggestion to switch from the dvui module being combined with a backend at build time, to making dvui.Window accept the type of the backend.  This means we would have a single dvui module that is combined with a backend at compile time.

This should:
* reduce build complexity
* make having a backend outside of the dvui repo easier

david-vanderson:
Some followup thoughts:

We used to have a runtime-chosen backend that used the normal zig type-erased runtime interface.  That's one option.

Now we have a build-time chosen backend so the type is global.  It's working decently.

This option is comptime chosen backend so the type is generic.  The big question is what does this allow over the previous options?

The two advantages I listed above might be doable with the current approach.  Are there other things I'm missing?

@alichraghi pointed out that this would require moving all global dvui functions inside dvui.Window.  So we would trade the implicit global `current_window` for an explicit global - `dvui_win.button()`.

Are there any situations where a runtime-chosen backend is important?  The only thing I can think of is picking a Wayland backend or X11 backend at startup.  Unsure if this is ever needed.

alichraghi:
i scrolled through commits to see how it worked before and i think the old approach is pretty nice.

david-vanderson:
@alichraghi got me started on this and I've committed some improvements.

I think we've simplified the build complexity, and removed the forced choice of backend during build, allowing projects to host their own backends.

For now we'll keep the current_window situation, and leave the idea of runtime-chosen backend for the future.  Reopen if I missed something.

Thank you!

PR #195, opened 2025-03-02
revert to a more flexible way to set backend

alichraghi:
Related #194
Also deleted bitrotted mach backend impl since mach has changed *a lot* and updating them would require an almost full rewrite


src/dvui.zig
```diff
@@ -7212,7 +7215,7 @@ pub fn pngFromTexture(arena: std.mem.Allocator, texture: Texture, opts: pngFromT
     @memcpy(ret[split + extra ..], png_bytes[split..@as(usize, @intCast(len))]);
 
     if (wasm) {
-        backend.dvui_c_free(png_bytes);
+        dvui.BackendWeb.dvui_c_free(png_bytes);
```
alichraghi, line 7218: not good but it works until we figure out a better way..

david-vanderson:
I like the direction of splitting the dvui module from the backend module.  I don't understand how dvui can import all of the backend modules (like dx11 is only on windows)?

Maybe we can keep having dvui reference a single "backend" but instead of build.zig creating a composite module we require the using application to get both modules and combine them?

alichraghi:
they are lazily analyzed so it works as long as we don't reference them.

> Maybe we can keep having dvui reference a single "backend" but instead of build.zig creating a composite module we require the using application to get both modules and combine them?

good idea!

src/dvui.zig
david-vanderson: Yeah I punted here

david-vanderson:
I took this and mostly ran with it.  Sorry I didn't get it into the history.  Definitely an improvement.

Let me know if anything looks weird.  Thanks!

alichraghi:
LGTM!

PR #197, opened 2025-03-06
Add horizontal scroll support

Yinameah:
Hello,

First thank you so much for bringing dvui to the world. 
It's in such a cool sweet spot, I had a ton of fun hacking on it in the last few days. 

As an exercise, I added something I felt was missing because on my laptop's touchpad I use it quite a lot : horizontal scroll.

This commit implement only with sdl and breaks the other backends.
But I'm not very experimented neither with zig nor gui programming, especially in immediate mode, and before going further I wanted to :

a) know if you would be interested in such a PR.
b) have some feedback if this is going in the right direction. Maybe I'm doing something stupid

(but at least this works as I would expect it to with sdl, on my machine, in the scroll canvas demo)

### DIFF
--- a/src/Event.zig
+++ b/src/Event.zig
@@ -84,6 +84,7 @@ pub const Mouse = struct {
         press,
         release,
 
+        wheel_x,
         wheel_y,
 
         // motion Point is the change in position
@@ -111,6 +112,7 @@ pub const Mouse = struct {
     data: union {
         none: void,
         motion: Point,
+        wheel_x: f32,
         wheel_y: f32,
     } = .{ .none = {} },
 };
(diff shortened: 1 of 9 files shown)


david-vanderson:
Thank you!  Looks pretty good to me!  Give me some time to test it and fix the other backends.

Yinameah:
Oh ! the intent was not to just hand over the work to you.

I can try to implement the other backends as well. (but I'm not convince I'll work as fast as you)

Do as you feel is the best.

david-vanderson:
I think we can move forward with this, even if only implemented for a single backend.  We just need to update them so they don't break.  Give it a shot, or likely I'll get to it tomorrow.

Yinameah:
Ok, I already added the straightforward fix for other backends to compile, defaulting to vertical scrolling.

I'll have a look later is other backend's implementation is straightforward

Yinameah:
Turned out implementing it in raylib and the web backend was super straightforward.

Cannot really test the DX11 one easily, but at least it should work like before.

Let me know if there is anything you want me to do

david-vanderson:
Looks great - thank you!

PR #199, opened 2025-03-14
add build option to disable backends

dasimmet:
Hi David,

I'd like to bring my own wasm backend, but i need to disable raylib to be able to use the "dvui" module when i pass in the "wasm32-freestanding" target.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -12,10 +12,20 @@ pub const LinuxDisplayBackend = enum {
 pub fn build(b: *std.Build) !void {
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
+    const disable_backends = b.option(
+        bool,
+        "disable_backends",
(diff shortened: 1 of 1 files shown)


david-vanderson:
Thanks!  Trying to reproduce the problem I see that `zig build -Dtarget=wasm32-freestandaing` gives a panic from raylib's build.zig, which makes sense.

I think you are on the right track, but we can use this opportunity to give more fine grained control over which backends are built.  I've committed a potential solution.  I wasn't sure if you were trying to use the dvui module directly, or the dvui_web one.  Would that work for you?

dasimmet:
Cool, thanks yeah using `.backend = .none` now does the tick!
I tried the dvui_web one but that comes with the backend...so dvui directly and disabling the backends seems more universal and also prevents the download of SDL/raylib.

👍👍

PR #207, opened 2025-03-21
TextEntryWidget Deletion bugs

RedPhoenixQ:
There where two bugs when deleting whole words:

1. When deleting a word backwards, using `ctrl+delete`, `text_changed` was set by testing `self.cursor`, which  was reset a couple of lines above
2. When deleting the last word and setting `len` to 0, the null terminator wasn't added. When using a buffer, and possible other backing storages, this caused one frame to have the correct frame but the text to appear again on the next frame but with the cursor moved

The TextEntryWidget now works as expected in these scenarios!


david-vanderson:
Fantastic - thank you!

Issue #208, opened 2025-03-22
Data invalidation


RedPhoenixQ:
Hello!

I've been using the `dataGetSlice` API quite a bit to render out lists of items. These lists are editable and can have items removed or their data changed, in a SQLite database, from other parts of the program. To invalidate this data, I currently just set the data again with `dataSetSlice` with the fresh data. This doesn't work however if I need to invalidate the data while I'm iteration over the previous slice, as that is a use after free. This can happen when a context menu on one of the slice items removes the item or similar.

I'm proposing creating a trash for data, similar to `texture_trash`. The API could look something like:
```zig
dvui.dataInvalidate(id, "_some_key");
// OR
dvui.dataRemoveLater(id, "_some_key"); // More like textureDestroyLater
```
where `dataRemove` gets run on each trashed item during `Window.end()`.

I would be happy to submit a PR if this is of interest.

david-vanderson:
Interesting!  That sounds like a very reasonable request.  Could we add a trash like you suggest, but instead of adding a new function, have dataSet* put the existing data (if any) into the trash?  I think that means we delay the `sd.free(self.gpa);` around dvui.zig line 3752.

Does that make sense?  If so, a PR would be wonderful!

RedPhoenixQ:
> but instead of adding a new function, have dataSet* put the existing data (if any) into the trash?

That is a more elegant solution and prevent any accidental misuse. I submitted a PR with the implementation and tested that it successfully prevents segmentations faults in the scenario described.

RedPhoenixQ:
Closed by #209

PR #209, opened 2025-03-22
feat: Add datas_trash to free data at the end of the frame

RedPhoenixQ:
This PR addresses #208.

Calling any dataSet* function now waits with freeing that data until the end of the frame. This successfully prevents a segmentation fault in the following example:

```zig
const n1: []const u32 = &[_]u32{ 1, 2, 3, 4, 5, 6, 7, 8, 9 };
const n2: []const u32 = &[_]u32{ 100, 200, 300 };
const nums = dvui.dataGetSliceDefault(null, 0, "", []u32, n1);
for (0.., nums) |i, n| {
    try dvui.label(@src(), "{d}", .{n}, .{ .id_extra = i });
    if (n == 5) {
        dvui.dataSetSlice(null, 0, "", n2);
    }
}
```

### DIFF
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -3733,31 +3733,6 @@ pub const Window = struct {
         self.data_mutex.lock();
         defer self.data_mutex.unlock();
 
-        if (self.datas.getPtr(hash)) |sd| {
-            if (sd.data.len == (bytes.len * num_copies)) {
-                sd.used = true;
-                if (builtin.mode == .Debug) {
-                    sd.type_str = dt_type_str;
-                    sd.copy_slice = copy_slice;
-                }
-
-                // Someone might do dataSetSlice on the slice they got from
-                // dataGetSlice, and @memcpy would complain about aliasing
-                if (sd.data.ptr != bytes.ptr) {
-                    for (0..num_copies) |i| {
-                        @memcpy(sd.data[i * bytes.len ..][0..bytes.len], bytes);
-                    }
-                }
-                return;
-            } else {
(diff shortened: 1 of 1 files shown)


david-vanderson:
Thanks!

After sleeping on it, I'm worried that this solution will fail if the old slice and new slice have the same size, because then dvui internally overwrites one with the other.  Sorry to lead you astray.

I can't think of anything better than your dataRemoveLater(), that seems like a good targeted solution.  Mind reworking it?

david-vanderson:
I supposed we could do both - the "always remove it later if the size changes" could be an extra protection against anyone who mistakenly is in this situation.  The ideal would be detecting if they were and advising to use dataRemoveLater, but I can't think of a way to know.

What do you think?

david-vanderson:
Sorry you are getting my stream of conciousness today.  Maybe the best thing is to have dataSetSlice always put the existing entry in the trash even if they are the same length.

That means no conditional behavior and no extra api.

RedPhoenixQ:
> Sorry you are getting my stream of conciousness today. 

No problem.

> Maybe the best thing is to have dataSetSlice always put the existing entry in the trash even if they are the same length.
> 
> That means no conditional behavior and no extra api.

Makes sense. Then there is no risk of unexpected behaviour like times shifting for one frame or similar.

RedPhoenixQ:
I made use of the `fetchPut` to make the code more clear and slightly fewer hash lookups

david-vanderson:
Awesome thank you!

PR #210, opened 2025-03-22
TextEntryWidget: check for space in backing slice when adding null terminator

RedPhoenixQ:
When using internal buffer and deleting when there is no text, `text.len` was 0 which caused an index out of bounds error. This issues was partially surfaced due to the check for `self.len != 0` being removed in #207.

To fix the issue I have added a helper function to set the null terminator is there is space in the buffer and replaced all instances where the null terminator was set directly.


david-vanderson:
Good idea - thanks!

Issue #123, opened 2024-09-10
Text Entry with Suggested Values


VisenDev:
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist for example of desired widget behavior

david-vanderson:
Thanks for filing this and for the link!  I think we have the pieces needed, I just need to figure out the api.

VisenDev:
> I just need to figure out the api.

Perhaps just add a "suggested values" field to TextEntryWidget.InitOptions?
```zig
suggested_values: ?[]const []const u8 = null,
```

david-vanderson:
What I'd like to do is have the api be more like:
* make the textEntry
* make a separate "suggestion dropdown" widget that gets it's location off the textEntry
* the using code adds suggestions to the list
  * this way the suggestions don't have to be static, and
  * it can dynamically filter based on what's already in the textEntry
  * also the filtering is in the app code not dvui code (but we would offer a built-in one)
* detect if the user has selected an entry and populate the textEntry with that value

But there are still a lot of unknowns that can only be figured out by going through it:
* How to know when to show the suggestions dropdown
* How to make keyboard navigation work (pressing down while in the textEntry should move the focus to the first suggestion?)
* Can it be a separate widget, or does it need to wrap the textEntry?

I'm sure it'll end up really nice, but it's going to take some time.

RedPhoenixQ:
Added with #212, specifically `dvui.comboBox(...)`

PR #212, opened 2025-03-23
Suggetions widget

RedPhoenixQ:
Related issue: #123

This adds a `SuggestionsWidget` which takes a `TextEntryWidget` and augments it with a dropdown bellow the field with suggested values. See the `textEntrySuggestions` function in dvui.zig for an example of the usage of this widget.

{DA66B4A2-C4BC-4261-9DAF-1877C3656492}

## Notes on behavior and api
- This widget needs to be called between `TextEntryWidget`s `.install()` and `processEvents()`.
- The suggestion doesn't show unless the `TextEntryWidget` or one of the suggestions have focus
- Calling `.close()` on the `SuggestionsWidget` will hide the suggestions until `open()` is called
  - The common case will likely call `.open()` if the text has changed in the input
- If `"char_down"`  keybind is pressed, the suggestions open and the first item, if it exists, will be focused
- Currently the suggestions window is anchored to the bottom left corner of the `TextEntryWidget`
  - A wide suggestion will cause the suggestions window to extend to the right of the `TextEntryWidget`

### DIFF
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -4969,6 +4969,215 @@ pub fn dropdown(src: std.builtin.SourceLocation, entries: []const []const u8, ch
     return ret;
 }
 
+pub const SuggestionsWidget = struct {
+    pub var defaults: Options = .{
+        .corner_radius = Rect.all(5),
+        .padding = Rect.all(6),
+        .name = "Suggestions",
+    };
+
+    id: u32 = undefined,
+    options: Options = undefined,
+    text_entry: *TextEntryWidget = undefined,
+    drop: ?FloatingMenuWidget = null,
+    drop_first_frame: bool = false,
+    drop_mi: ?MenuItemWidget = null,
+    drop_mi_index: usize = 0,
+
+    pub fn init(src: std.builtin.SourceLocation, text_entry: *TextEntryWidget, opts: Options) SuggestionsWidget {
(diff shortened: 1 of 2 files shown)


RedPhoenixQ:
A more appropriate name for this widget might be `ComboBoxWidget`. That way we could make a function like `pub fn comboBox(src: ..., select_options: []const []const u8, opts: Options) !?usize` returning the index of the selected item, making it more like a combo box, a searchable dropdown.

It is also a bit strange for a widget to take another widget as a parameter, but I can't see another easy way to gain access to the `TextEntryWidget`.

david-vanderson:
I still need to implement restrictions so the suggestion window doesn't obscure the textEntry, but ran out of time today.  Then just some testing in low fps, no entries, and touch.

david-vanderson:
That was super useful both directly and in helping fix some long-standing weirdness with textLayout corner widgets and floating menus.

Thanks very much!

RedPhoenixQ:
Sorry, but this is funny. I should've said that I started working on a similar solution for menu placement. I added a `Direction` parameter to `placeOnScreen` as well. When to push and saw I was three commits behind. My bad for duplicated work.  Good changes though!

david-vanderson:
Hahaha great minds think alike!

PR #213, opened 2025-03-23
Add justify-content for FlexBoxWidget

RedPhoenixQ:
Previous behaviour was to center all rows based on the widest. The `start` option left aligns all the items, even after adding a new row. This imitates [`justify-content`](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content#try_it) from CSS Flexbox, were the naming is derived from.

The previous default behaviour of `center` has been kept, so no changes are needed for users.

An option for `end` could be added, but would require the row width logic to be reversed, starting from the right. Other options like `space-between` seem very difficult to implement in an immediate mode GUI, but may be possible by saving number/width of children between frames.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -51,6 +51,7 @@ var layout_border: Rect = Rect.all(0);
 var layout_padding: Rect = Rect.all(4);
 var layout_gravity_x: f32 = 0.5;
 var layout_gravity_y: f32 = 0.5;
+var layout_flex_content_justify: dvui.FlexBoxWidget.ContentPosition = .center;
 var layout_expand_horizontal: bool = false;
 var layout_expand_vertical: bool = false;
 var show_dialog: bool = false;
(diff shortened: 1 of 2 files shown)


david-vanderson:
This is a good idea.  Don't worry about space-between - if users want that they can add `dvui.spacer` things between the elements.

`end` should be doable like `.end => ret.x += (self.wd.contentRect().w - self.max_row_width_prev),` can you see if that works?

For a new feature like this, I like to add a demo of it to the demo window.  Can you add a row of radio buttons (or dropdown) above the flexbox (maybe in line with the label "Flexbox") in the "layout" section of src/Examples.zig?

RedPhoenixQ:
> `end` should be doable like `.end => ret.x += (self.wd.contentRect().w - self.max_row_width_prev),` can you see if that works?

I'll see. I though there would be more issues with the boxes starting from the left, but either way I can make it work.

> For a new feature like this, I like to add a demo of it to the demo window.

Sure! I'll add that too

RedPhoenixQ:
Ah, here is the issue with `end`. This is the result with only your change:
{23E93206-1B82-42AE-BDE8-D47787FE8B87}

If we are trying to match the CSS Flexbox behaviour, this is the expected result:
{58BFBFD4-6421-4FE2-A3DB-D6DB28436B15}

So we would need to start from the right and move left for each child. I'll see how hard that would be. 

It would also risk making the first frame worse as more items are likely to be outside the clipRect with the fallback width being 500.

david-vanderson:
Thank you!

PR #219, opened 2025-03-24
dx11backend cleanup

marler8997:
Bunch of cleanup, still lots to cleanup but it's a start.

This removes the globals from the dx11 backend.  Instead, now the backend returns an `HWND` value casted to a backend `Context`.  The window state is accessed with the `GetWindowLongPtr` method on the hwnd so it no longer needs to be global.

One notable API change, for the "ontop" example I replaced the API where you create a native window and then a DVUI window associated with it, and intead introduced the "attach" concept, where you attach DVUI to a native window.  The simpler `initWindow` will call the same attach function under the hood.

### DIFF
--- a/examples/dx11-ontop.zig
+++ b/examples/dx11-ontop.zig
@@ -1,90 +1,38 @@
 const std = @import("std");
 const dvui = @import("dvui");
 const Backend = dvui.backend;
-comptime {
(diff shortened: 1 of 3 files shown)


david-vanderson:
Awesome thank you!

PR #222, opened 2025-03-24
Multi window dx11

marler8997:
Addresses the first issue if a process wants to have multiple "native windows" that use DVUI (on MS Windows).  Namely, every new window needs a window class, but `RegisterClass` can only be called once for the same name.  If you call `initWindow` twice, then the second will fail when it calls `RegisterClass` with the same class name.

This PR contains two commits, the first allows the application to disable this call to `RegisterClass` to avoid subsequent calls to it.  Note that an app wouldn't disable the `RegisterClass` call if they are creating multiple windows with different options/classes. An approach that automatically manages this for the app requires storing/tracking a table of all the class names, however, I'd consider that overkill.  The first commit should suffice, but, I also included a second commit which I'd recommend over the first, namely, let the app call RegisterClass instead.  If the app forgets to do this, then I also created a nice panic with a message telling them they need to call RegisterClass.

### DIFF
--- a/examples/dx11-standalone.zig
+++ b/examples/dx11-standalone.zig
@@ -21,12 +21,20 @@ pub export fn main(
     _: ?[*:0]const u16,
     _: win32.SHOW_WINDOW_CMD,
 ) void {
+    return main2() catch |e| {
+        if (@errorReturnTrace()) |trace| {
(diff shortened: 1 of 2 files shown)


david-vanderson:
That looks like a good solution to me!

PR #223, opened 2025-03-25
windows: multiple windows work now

marler8997:
So...I guess my last commit actually did make it so DVUI could work with multiple windows, I just had a bug in dxui-standlone.zig where I was accidently re-using the same window_state for both windows...woops. After giving each window their own state DVUI now renders to both windows and we can gracefully quit the application with proper cleanup, hooray!

With that...I removed the comptime condition for the second window and made a button instead called "Spawn Another OS Window".  When clicked, it allocates a new window/state and creates a new backend instance, adds it to a list and it all seems to be working.

With this, we now need a way to know when to quit the application, for this I added an optional "close_handler" callback method the application can set to customize what a window close does.  Note that when we get WM_CLOSE, we can't simply destroy the window because the application would then have an invalid context pointer, that's why WM_CLOSE exists, it allows the application to get notified when the window wants to close and it can handle the teardown and call DestroyWindow when it's ready.

For dx11-standalone, I went with a simple scheme.  If you close the first window, it calls the default close handler which quits the application, otherwise, closing any other window simply removes that window from the list and destroys it.

### DIFF
--- a/examples/dx11-standalone.zig
+++ b/examples/dx11-standalone.zig
@@ -8,6 +8,16 @@ const window_icon_png = @embedFile("zig-favicon.png");
 var gpa_instance = std.heap.GeneralPurposeAllocator(.{}){};
 const gpa = gpa_instance.allocator();
 
+const ExtraWindow = struct {
+    state: *Backend.WindowState,
+    backend: Backend.Context,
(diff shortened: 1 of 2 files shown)


SoraTenshi (approved):
Neat!

PR #224, opened 2025-03-25
windows: replace close_handler with service result

marler8997:
Here's an idea to remove the close handler callback.  I've added a `received_close` field on `WindowState` that gets set to true once WM_CLOSE is received.  However, we probably don't want to be checking this field on all the windows on every frame, so, in addition I replaced the `isExitRequested` method with the `serviceMessageQueue` function. Instead of returning two states (false/true) it will return one of three states (queue_empty/quit/close_windows).  This way we know we only need to check if any windows have received WM_CLOSE when one of them has actuallly received it.

### DIFF
--- a/examples/dx11-ontop.zig
+++ b/examples/dx11-ontop.zig
@@ -35,30 +35,28 @@ pub export fn main(
         const win = backend.getWindow();
         log.info("dvui window also init.", .{});
 
-        main_loop: while (true) {
-            // This handles the main windows events
-            if (Backend.isExitRequested()) {
-                break :main_loop;
-            }
+        while (true) switch (Backend.serviceMessageQueue()) {
+            .queue_empty => {
+                // beginWait coordinates with waitTime below to run frames only when needed
+                const nstime = win.beginWait(backend.hasEvent());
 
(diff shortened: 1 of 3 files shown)


examples/dx11-ontop.zig
```diff
@@ -35,30 +35,28 @@ pub export fn main(
...
-            };
-
-            // marks end of dvui frame, don't call dvui functions after this
-            // - sends all dvui stuff to backend for rendering, must be called before renderPresent()
-            _ = win.end(.{}) catch continue;
-        }
+        while (true) switch (Backend.serviceMessageQueue()) {
+            .queue_empty => {
+                // beginWait coordinates with waitTime below to run frames only when needed
+                const nstime = win.beginWait(backend.hasEvent());
```
marler8997, line 41: Not related to this PR but this API doesn't work right when you have multiple windows...

david-vanderson: Yes totally right - the code would have to loop over the windows and do begin/end in series.  Then if we were using waitTime() you'd want the backend to wait for the min of those.

Does that make sense?

david-vanderson:
Looks good - thank you!

Issue #218, opened 2025-03-24
file upload in wasm


david-vanderson:
We already have file download shown in the plot demo.  Need to figure out how to upload a file from a browser and get the contents into dvui somehow.

RedPhoenixQ:
I could maybe make and attempt at this. The biggest issue is probably dealing with the async nature of picking and opening a file. Could probably be done with an exported callback function. We would have to block on the wasm side until the javascript promises have completed to match the blocking/synchronous behaviour of `dialogNativeFileOpen`. 

Btw I don't believe that `wasm_sleep` actually sleeps/blocks on the wasm side but I could be wrong.

Getting the data into dvui should be very similar to the current `wasm_add_noto_font` in web.js function, so not big hurdles there.

There is also the issue of the browser not having access to the filesystem, so returning the file path from `dialogNativeFileOpen` wouldn't be very useful. 

Here's some alternatives I though of on the spot:
- We could cache a [File](https://developer.mozilla.org/en-US/docs/Web/API/File) on the javascript side (keyed by path) and expose functions to get the file data on demand
  - The clearing of the cache needs to be figured out. Should it clear after one frame? 
- The file data can be stored in the arena keyed by the path to be removed next frame if unused
  - This would give access to both the path/filename and the data
  - Could bloat the arena excessively, especially with multifile selections
- Have different return type for wasm or a separate function entirely
  - Maybe the wasm version only returns the file data and does not give the file path
- The new [File System API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API) for browsers would be really interesting, but it is not well supported yet. Specifically [showOpenFilePicker](showOpenFilePicker) isn't supported in firefox or safari

There is also the [webkitdirectory](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory) attribute for directory selection, but it gives the list of files back, so its more of a convenience for multiple file select.

I think the first option is most realistic and would have good browser support.

david-vanderson:
That would be amazing!  I would love to see any code you come up with, even half-baked.  That would help me get my bearings about this stuff.

RedPhoenixQ:
I've done a little testing and I can't think of an easy way to wait without a callback. Using a callback here makes it so that we cannot get the result until the next frame at the earliest.

[...]

PR #225, opened 2025-03-25
Add wasm file upload

RedPhoenixQ:
Addresses #218.

This adds file uploads to wasm by a adding separate api that only works for wasm. This is needed to work around the inability to block while awaiting javascript promises.

The files are stored on the javascript side for 2 frames because the async resolve may happen after the check for the file, but before the end of the frame. To make the api simpler the whole file data is loaded eagerly on the javascript side. To access the file data, use `WasmFile.read_data(allocator)` to allocate a slice for the file data with any allocator. The `WasmFile` is only valid for one frame as the name is allocated in the arena.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -2771,9 +2771,11 @@ pub fn dialogs(demo_win_id: u32) !void {
         var hbox = try dvui.box(@src(), .horizontal, .{});
         defer hbox.deinit();
 
+        const single_file_id = hbox.widget().extendId(@src(), 0);
+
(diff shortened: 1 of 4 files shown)


david-vanderson:
This is fantastic!  Two nits and then we can merge - please:
* rename "read_data" to "readData" (slowly trying to adopt zig naming conventions)
* in the demo, use a widget id instead of a hardcoded number

david-vanderson:
Very cool - thank you so much!

Issue #158, opened 2024-10-22
tooltips


MichaelBelousov:
would be awesome if there was a built-in/recommended way to display a floating tooltip when hovering over something. I think it's a very common UX pattern.

david-vanderson:
Thanks for the issue!  I agree and will think about how to do it.

david-vanderson:
I think the best option here is "No-delay static position":
* tooltip is shown as soon as the mouse pointer is inside the target rectangle.
* tooltip is positioned relative to the target rectange (doesn't move with mouse).
* tooltip stays up as long as the mouse pointer is in the target rectangle.
* can move mouse into the tooltip and interact with widgets inside or even show nested tooltips (works similar to submenus).
* tooltip goes away when mouse pointer is no longer within target or tooltip or any child tooltip (similar to submenus but doesn't require a click to close)

I see this model in a lot of games so it has real-world experience.  However it is different from the traditional "delay after no mouse movement relative to mouse position".  Here are reasons I don't like the traditional:
* hard to know if a tooltip exists, especially with poor fine mouse control
* easy to accidentally nudge the mouse just after a tooltip appears, then hard to know how to get it back
* tooltip appears just next to mouse, almost certainly obscuring what you are trying to understand

That's my current thinking, but I don't have a use-case for this yet.

@MichaelBelousov Can you speak to whether this would work for you?

david-vanderson:
@MichaelBelousov I just merged tooltips based on @RedPhoenixQ initial implementation.  It's live in the online web demo - see the "Menus / Tabs" section, and there is also some text in the "Text Layout" section.

How does that look?

MichaelBelousov:
Sorry I should change my watching setting for the repo to "mentions" only so I don't ignore the emails haha. I'll take a look

MichaelBelousov:
LGTM! Thank you @RedPhoenixQ  and @david-vanderson ! I will definitely be using it in my application!

david-vanderson:
Awesome!  Reopen if you find any issues.  Thanks!

PR #226, opened 2025-03-27
TooltipWidget

RedPhoenixQ:
Addresses #158

Adds a tooltip widget that takes a triggering rect and shows a `FloatingWidget` with the users content. The tooltip waits for the mouse to stop moving to capture a starting horizontal location and from then onwards the tooltip is stationary. This means that the tooltip opens close to the users mouse, but still positioned around the triggering rect. This behavior is inspired by vscodes tooltips.

The user can move the mouse around the triggering rect and into the tooltip window, which can contain most other widgets. Widgets that extend outside the tooltip is currently not supported as the original tooltip would close when the mouse exits. Text, buttons, links and other useful short information should work fine.

An animation is also added by default to mimic the behavior of tooltips in vscode. They can be easily disabled by setting a duration of 0. The `AnimateWidget` is also being used as a container for styling purposes.

The waiting for mouse movements to stop does mean that a user that never stops moving the mouse won't see a popup. This is an unlikely scenario as the widget only requires a frame with no movement. It could be made to wait for a very small amount of movement to be more forgiving. It could also the the mouse position immediately.

There isn't currently an example added as I'm unsure of where it would fit. Will gladly add one with some pointers as to where. 

Usage:

```zig
var label = dvui.LabelWidget.initNoFmt(@src(), "Some hoverable text", .{});
try label.install();
try label.draw();
label.deinit();

var tooltip = try dvui.tooltip(@src(), .{ .rect = label.wd.borderRectScale().r }, .{});
defer tooltip.deinit();
if (try tooltip.shown()) {
    var text = try dvui.textLayout(@src(), .{ .break_lines = true }, .{ .background = false });
    defer text.deinit();
    try text.addText("Some tooltip text shown bellow", .{});
}
```

{2C4A6DCE-820B-4C31-96B4-16B559BCA49C}
(Cursor not visible, is roughly bellow the 'e' in Some)

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -1664,6 +1664,9 @@ pub fn layoutText() !void {
         const start = "\nNotice that the text in this box is wrapping around the stuff in the corners.\n\n";
         try tl.addText(start, .{ .font_style = .title_4 });
 
+        const col = dvui.Color.average(dvui.themeGet().color_text, dvui.themeGet().color_fill);
+        try tl.addTextTooltip(@src(), "Hover this for a tooltip.\n\n", "This is some tooltip", .{ .color_text = .{ .color = col }, .font = dvui.themeGet().font_body.lineHeightFactor(line_height_factor) });
+
         try tl.addText("Title ", .{ .font_style = .title });
         try tl.addText("Title-1 ", .{ .font_style = .title_1 });
         try tl.addText("Title-2 ", .{ .font_style = .title_2 });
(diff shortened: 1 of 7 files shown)


david-vanderson:
Awesome thank you!  Give me some time to review.

david-vanderson:
I think that is a solid tooltip.  We might have some more customization options in the future, but good for now.

Thanks so much!

Issue #215, opened 2025-03-24
sliderentry: cannot easily enter text_mode when using a mobile browser


barathrm:
It seems like there's currently now way to go into text_mode when using the sliderentry widget when running via wasm and in a mobile browser, as there's no way to press the enter key in that case (without doing it manually somehow).

What would be the preferred way to enable text input mode on a sliderentry on mobile?

- I intuitively long-pressed the slider to get into text input mode
- double-tap?

I'm not a web-developer so I may be wrong, but a quick search indicates that there's no "long press" event in javascript itself, so I assume we'd have to handle that logically manually somehow.

david-vanderson:
Thanks for the report!  Yes you are right there is no touch way to do it yet.

dvui doesn't (currently) have double tap or long tap - see https://david-vanderson.github.io/log/#2025-03-11 for my explanation.

I'm pretty sure in this case we don't need it.  sliderEntry is a very compact widget, so my finger completely obscures it.  What I've seen is that people touch down, drag down so they can see the sliderEntry, then drag left/right to set it.

Given that touch behavior, we can try making a single non-drag touch transition sliderEntry to the textInput mode.  Would you care to try implementing that?

To transition back, we need to catch when the sliderEntry loses focus.  I'm pretty sure this already works.

PR #230, opened 2025-03-28
sliderentry text mode touch support

barathrm:
Tested on android (firefox, vanadium), Firefox linux desktop and sdl-standalone.

----

##   sliderentry: fix not showing text mode when already focused
    
    Explicitly request a refresh after entering text mode via ctrl+click.
    
    Without this fix, entering text mode by ctrl+clicking when the
    sliderentry already has focus and in the absence of other events would
    result in the slider entry not visually transitioning to text input mode
    until the next event/refresh.

##  sliderentry: enter text_mode by touch-tapping
    
    Entering text mode via ctrl+click or the enter key is cumbersome on
    mobile devices, so this adds support for entering text mode by tapping
    on a sliderentry.
    
    Regular pointer (mouse/non-touch) events keep their current behavior
    of jumping the slider to the position of the pointer.
    
    Fixes https://github.com/david-vanderson/dvui/issues/215.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -824,7 +824,7 @@ pub fn basicWidgets(demo_win_id: u32) !void {
         try dvui.label(@src(), "Slider Entry", .{}, .{ .gravity_y = 0.5 });
         if (!slider_entry_vector) {
             _ = try dvui.sliderEntry(@src(), "val: {d:0.3}", .{ .value = &slider_entry_val, .min = (if (slider_entry_min) 0 else null), .max = (if (slider_entry_max) 1 else null), .interval = (if (slider_entry_interval) 0.1 else null) }, .{ .gravity_y = 0.5 });
-            try dvui.label(@src(), "(enter or ctrl-click)", .{}, .{ .gravity_y = 0.5 });
+            try dvui.label(@src(), "(enter, ctrl-click or touch-tap)", .{}, .{ .gravity_y = 0.5 });
         } else {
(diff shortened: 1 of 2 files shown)


david-vanderson:
Good catch on that refresh!

Your intuition about the touch-without-drag is good, but anyone without super good dexterity will have trouble maintaining the same pixel position.  The way we usually do it is by detecting whether a drag started or not.  Checkout how `fn labelClick()` detects if a drag has started and bails.  Does that make sense?

barathrm:
I didn't have any issues but you are right, makes sense :ok_hand: 

I was wondering why we call `dragEnd()` [in sliderentry](https://github.com/david-vanderson/dvui/blob/93072fd3f3bac7c0cfa5ae4d25bb404478afb3d8/src/dvui.zig#L6436) when we never called `dragStart()` :thinking: I guess that's in case a drag is going on in some parent widget which now won't receive it since sliderentry has `handled = true` it?

barathrm:
Pushed different approach using the drag functions :) Tested on the same backends/browsers.

Btw, should PR approval be required to be able to merge?

david-vanderson:
> I was wondering why we call `dragEnd()` [in sliderentry](https://github.com/david-vanderson/dvui/blob/93072fd3f3bac7c0cfa5ae4d25bb404478afb3d8/src/dvui.zig#L6436) when we never called `dragStart()` 🤔 I guess that's in case a drag is going on in some parent widget which now won't receive the mouse event since sliderentry has `handled = true` it?

Whoops - nope that was just a mistake on my part from when I split out dragEnd from releasing mouse capture.

> Btw, should PR approval be required to be able to merge?

dvui is still small enough that we don't have explicit rules for this.  Use your judgement, and don't worry too hard because we can always revert or fix up things after they are merged.

Thanks very much!

Issue #211, opened 2025-03-22
Questions about Event handling (hover detection in parent widget)


Yinameah:
I'm not sure if this is some kind of edge case bug, or if it's me that doesn't understand the logic correctly.

Here is a stripped down example of what I'm trying to achieve : 
(in my real code, this list of checkboxses is linked to a canvas and they share a hover state state per item, so I'm not trying to just style the checkbox hover behavior)

```zig
pub fn match_position_event_weirdness() !void {
    const Data = struct {
        var cb_val = [3]bool{ false, false, false };
        var cb_hover = [3]bool{ false, false, false };
    };

    const evts = dvui.events();
    for (0..3) |i| {
        var b = try dvui.box(@src(), .horizontal, .{
            .id_extra = i,
            .expand = .horizontal,
            .background = true,
            .color_fill = .{ .name = if (Data.cb_hover[i]) .fill_hover else .fill_window },
        });
        _ = try dvui.checkbox(@src(), &Data.cb_val[i], "cb label", .{ .id_extra = i });
        b.deinit();

        for (evts) |*e| {
            if (b.matchEvent(e)) {
                Data.cb_hover[i] = true;
            } else {
                Data.cb_hover[i] = false;
            }
        }
    }
}
```
My expectation with this code is that the `dvui.box` that surrounds the check-boxes will be _highlighted_ when the mouse is over the region. In practice however, it only works when being in the area that is part of the surround box, but not the checkbox area.

Took me a while to figure out what was going on, because the `.motion` mouse action event **do match** when the pointer is either on the cb or the surrounding box, but the `.position` one matches only when the pointer is in the surrounding box but **not** in the cb one.
And since the `.position` always comes at the end it kind of cancels the hover state.

I'm not even sure if the intended behavior regarding event matching for surrounding boxes when the cursor is in the inner region of child widget, but the fact that the 2 types of event gets a different treatment really puzzles me. Would be super grateful is somebody can shed some light on this.

david-vanderson:
Thanks for the explanation and reproduction code!  I'm not totally sure I understand all the issues here, so first, does this code do what you want?
```zig
pub fn match_position_event_weirdness() !void {
    const Data = struct {
        var cb_val = [3]bool{ false, false, false };
        var cb_hover = [3]bool{ false, false, false };
    };

    const evts = dvui.events();
    for (0..3) |i| {
        Data.cb_hover[i] = false;
[...]

PR #231, opened 2025-03-29
Prevent two tooltips chains at once

RedPhoenixQ:
This is a minor fix for tooltips. When a tooltip opens over another tooltip trigger area, both were shown at once. The second tooltip should not be shown before first one closes.

To fix this, a check is added to see if this is the first chain in the frame. Otherwise don't show the tooltip.

Before and after (with the mouse over the "Some text" tooltip in both cases)
{F03B6CD5-C807-4268-9627-12232541FB0F}
{862EDB1D-3619-4A1A-8E04-A20A1B8100E3}

### DIFF
--- a/src/widgets/FloatingTooltipWidget.zig
+++ b/src/widgets/FloatingTooltipWidget.zig
@@ -13,21 +13,11 @@ const FloatingTooltipWidget = @This();
 
 // maintain a chain of all the nested FloatingTooltipWidgets
 var tooltip_current: ?*FloatingTooltipWidget = null;
-// the last frame time a chain was open, only one allowed per frame
-var last_tooltip_set: i32 = 0;
 
 fn tooltipSet(tt: ?*FloatingTooltipWidget) ?*FloatingTooltipWidget {
-    if (tooltip_current == null) {
-        // Set the time of the current frame
-        last_tooltip_set = @truncate(dvui.frameTimeNS());
-    }
     const ret = tooltip_current;
     tooltip_current = tt;
     return ret;
-}
-
-fn tooltipCanOpen() bool {
-    return tooltip_current != null or last_tooltip_set != @as(i32, @truncate(dvui.frameTimeNS()));
 }
 
 pub var defaults: Options = .{
@@ -107,13 +97,10 @@ pub fn shown(self: *FloatingTooltipWidget) !bool {
         return true;
     }
 
-    if (!tooltipCanOpen()) {
-        return false;
-    }
-
     if (!self.showing) {
         // check if we should show
-        if (self.init_options.active_rect.contains(dvui.currentWindow().mouse_pt)) {
+        const cw = dvui.currentWindow();
+        if (cw.windowFor(cw.mouse_pt) == cw.subwindow_currentId and self.init_options.active_rect.contains(cw.mouse_pt)) {
             self.showing = true;
         }
     }
--- a/src/widgets/FloatingTooltipWidget.zig
+++ b/src/widgets/FloatingTooltipWidget.zig
@@ -13,11 +13,21 @@ const FloatingTooltipWidget = @This();
 
 // maintain a chain of all the nested FloatingTooltipWidgets
 var tooltip_current: ?*FloatingTooltipWidget = null;
+// the last frame time a chain was open, only one allowed per frame
+var last_tooltip_set: i32 = 0;
 
 fn tooltipSet(tt: ?*FloatingTooltipWidget) ?*FloatingTooltipWidget {
+    if (tooltip_current == null) {
+        // Set the time of the current frame
+        last_tooltip_set = @truncate(dvui.frameTimeNS());
+    }
     const ret = tooltip_current;
     tooltip_current = tt;
     return ret;
+}
+
+fn tooltipCanOpen() bool {
+    return tooltip_current != null or last_tooltip_set != @as(i32, @truncate(dvui.frameTimeNS()));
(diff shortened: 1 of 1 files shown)


david-vanderson:
Thanks! I saw that as part of your initial implementation, but still don't quite understand what is going on.  I think this plays into the the mouse .position question, but give me some time to reproduce it.

david-vanderson:
Ah right because we didn't use mouse .position to check, we directly used Window.mouse_pt.  Let me rework it to use the .position event, that should prevent this (because the .position event is only directed at the top subwindow).

david-vanderson:
See https://github.com/david-vanderson/dvui/issues/211

RedPhoenixQ:
I added a change that solves the issue in another more elegant way that probably applies to #211 too. Most, if not all, overlap of widgets is done via subwindows, so checking that the mouse pointer is in the current subwindow will effectively check the z-index of the "hovered" state.

If this is a viable solution, it could perhaps be a helper on `WidgetData` like `hovered()`, which checks if `mouse_pt` is in the current subwindow and within the widgets rect. This bypasses the event handling for simple hover states, minimising the need for individual state management. I believe there should one ever be one chain of widgets within one subwindow that would be considered hovered, which is similar to how CSS handles the hover state of child elements.

```zig 
pub fn hovered(self: *const WidgetData) bool {
    const cw = dvui.currentWindow();
    return cw.windowFor(cw.mouse_pt) == cw.subwindow_currentId and self.rect.contains(cw.mouse_pt);
}
```

david-vanderson:
That looks like a pretty good solution until I fix #211 - but don't we need to check the subwindow in deinit() as well?  That might be a bit tricky as we are testing two different subwindows - the original and the tooltip itself.

david-vanderson:
Even with this fix, if you mouse-down on a button (or slider), then drag and while dragging the mouse moves over a tooltip active_rect, then the tooltip will show even though the button (or slider) has mouse capture.

I haven't thought about this too much, but I think we don't want a tooltip in that scenario, and #211 should fix that.

RedPhoenixQ:
> Even with this fix, if you mouse-down on a button (or slider), then drag and while dragging the mouse moves over a tooltip active_rect, then the tooltip will show even though the button (or slider) has mouse capture.

That is a good point. Maybe a combination of `windowFor` and not marking `.position` events as handled unless they are actively being use like in the example you gave. I believe most hover states shouldn't "handle" the position event then, only mark them handled it they're part of something like a click-and-drag action.

> Maybe the logic for when to stop showing needs to be moved into install()?

It works currently, but might be a additional frame late. Since only one chain is open, there shouldn't be any interference but I agree it's not very robust

RedPhoenixQ:
Actually we couldn't use this check in the `deinit` as the trigger and tooltip are different subwindows. I don't think there is an easy way to get the subwindow of the trigger if we are hovering the tooltip. That means we need to check both rects unconditionally and rely on the triggering check to be in the correct subwindow.

RedPhoenixQ:
> That's a good point. Maybe not marking `.position` events as handled unless they are actively being use like in the example you gave. I believe most hover states shouldn't "handle" the position event then, only mark them handled it they're part of something like a click-and-drag action.

Nvm, doesn't work because of ordering. Windows like the debug window will always have the "move" cursor, even when hovering buttons which isn't correct.

Could mouse capture be used as a flag to see if position events are being used? Then this solution plus `dvui.captureMouseId() != null` might be enough to not trigger tooltips and other hover effects like cursor changes while some drag-like action is going on

RedPhoenixQ:
> Could mouse capture be used as a flag to see if position events are being used? Then this solution plus `dvui.captureMouseId() != null` might be enough to not trigger tooltips and other hover effects like cursor changes while some drag-like action is going on

This is effectively what `eventMatch` does... Sorry about the stream of thought

david-vanderson:
Not a problem - always useful to have someone else think through this stuff.  It is surprisingly subtle.

RedPhoenixQ:
Found this to get the correct events for the tooltip, but not sure when to do the processing yet. Will experiment some more

```zig 
dvui.eventMatchSimple(e, self.widget()) and dvui.eventMatch(e, .{  .id = self.id, .r = self.init_options.active_rect })
```

RedPhoenixQ:
Turns out handling the events in this case is difficult as the normal `matchEvent` logic cannot account for the current subwindow being the tooltip and states the `.position` events for the trigger is false. That means the event handling needs to be fully reimplemented within the tooltip widget, which I don't like for maintainability/consistency. 

I added the check for mouse capture. It's a pretty good 80% solution.

The last 20% (probably achieved through handling the events) would be:
- mark `.position` events as handled for other widgets
- not respecting the handled state of the position events
  - This causes issues in the order the widgets handle events. Could a widget with a hovered state have a tooltip? In that case, which widgets handles the event first? Do they mark it as handled, and in that case who ignores the handled state?

If this requires another handled marker on the `.position` event, wouldn't that be very similar to this current solution (check mouse capture, subwindow is active and mouse in rect)? As the handled state is ignored either way

david-vanderson:
I agree with all that.  Sit tight and let me fix .position handling in all the widgets for #211 and then we'll see where we are.

david-vanderson:
Okay now that I've committed fixes for #211 can you double check me that this is fixed?

RedPhoenixQ:
Works beautifully. No overlapping tooltips and the tooltips stays hidden while mouse is captured by another widget.

PR #236, opened 2025-03-31
Changed MouseCapture API for the simpler case

Yinameah:
Make dvui.MouseCapture() the default way to interact with the api, making clear that MouseCaptureCustomZone is not something you want normally.


david-vanderson:
Yes this is good.  Please change "mouseCaptureCustomZone" to just "mouseCaptureCustom" and commit.  Thanks!

PR #237, opened 2025-03-31
Make Tooltip interactivity optional

RedPhoenixQ:
From using the new tooltips, I've realised that I often just want to show contextual information to the user. Having the tooltips always stay open on hover makes this difficult to use in a list were the tooltip covers the next list item. I believe most tooltips are NOT interactive and have set that as the default.

In short most tooltips shouldn't hinder the access to other elements unless is has a reason to.


david-vanderson:
Looks good!

Issue #235, opened 2025-03-31
wasm: make it easier to integrate other wasm imports / libraries


barathrm:
Apologies if there's already a ticket or plans for this, couldn't find a good match.

I'm currently using [zjb](https://github.com/scottredig/zig-javascript-bridge/) in the same wasm file as the one I'm using dvui in. Since zjb requires setting some zjb imports and the wasm instance, I currently need to copy and patch the web.js file provided by dvui and then add the imports and set the zjb instance manually like so:

```diff
@@ -1,3 +1,4 @@
+var zjb = new Zjb();
 
 async function dvui_sleep(ms) {
     await new Promise(r => setTimeout(r, ms));
@@ -173,7 +174,10 @@ function dvui(canvasId, wasmFile) {
     const utf8encoder = new TextEncoder();
 
     const imports = {
+        zjb: zjb.imports,
         env: {
+        memory: new WebAssembly.Memory({initial: 1}),
+        __stack_pointer: 0,
         wasm_about_webgl2: () => {
             if (webgl2) {
                 return 1;
@@ -544,6 +548,7 @@ function dvui(canvasId, wasmFile) {
     .then((response) => response.arrayBuffer())
     .then((bytes) => WebAssembly.instantiate(bytes, imports))
     .then(result => {
+        zjb.setInstance(result.instance);
 
         wasmResult = result;
 
```

It would be nice if the dvui javascript backend code could be created similar to `new Zjb()`, to then let me get the imports it needs similar to `zjb.imports` etc. Then the responsibility of instantiating the wasm file would fall to the user and I could have my own minimal inline js to combine zjb and dvui and potentially other things. 

Example for how zjb is used: https://github.com/scottredig/zig-javascript-bridge/blob/main/simple/static/index.html

I imagine something similar as the Zjb example but concatenating the zjb imports with those of dvui and then instead of calling `dvui()` onLoad, I'd call `dvui.run()` or something after hooking everything up.

I hacked together a working-ish proof of concept, but it turned quite ugly quickly as I haven't really touched javascript in ages, but I guess there's no real reason this shouldn't work. 

For more inspiration, [zig-js](https://github.com/mitchellh/zig-js?tab=readme-ov-file#usage) seems to do something similar.

RFC/opinions?

david-vanderson:
This is a good idea, and I would like to do it, but I don't really understand what a solution would look like.  Can you post your proof of concept code?

barathrm:
[...]

Issue #165, opened 2024-11-22
higher-level app shim


david-vanderson:
@MichaelBelousov has been building an app on top of dvui that functions with multiple backends, including web.  One of the pain points is that the web stuff was not written in a way that makes building on top of it work well.  In particular merging dvui changes is annoying.

This issue is about a possible way to make this better by adding a higher-level "app" api that inverts control of the mainloop.  The app code would expose a few functions (init/deinit/frame) and pass them to some dvui "app runner" interface.

The advantage is more separation between app code and dvui code.  The disadvantage is less control over the mainloop.

To be clear this would only be an addition - an extra way to structure apps where this tradeoff makes sense.

david-vanderson:
https://wiki.libsdl.org/SDL3/README/main-functions#main-callbacks-in-sdl3

Likely a good idea to follow SDL3's lead here, with similar functions.

david-vanderson:
@marler8997 showed problems with resizing windows (at least on Windows and Mac) - so this also needs to integrate with backends so that the "render" callback can be run inside the resize callback (WM_PAINT on Windows)

nhanb:
Janky window resizing has been bothering me too, and I can confirm that using SDL3's main callbacks, resizing on Windows and Linux has been buttery smooth:

https://github.com/user-attachments/assets/2c3d6c4b-b1df-423c-be95-00b5b5652f1c

Here's the source code: https://github.com/nhanb/sdl3-zig-probe/blob/d8034c0eff0550178647b2972b982fde52990463/src/main.zig#L35 - I simply needed to set `SDL_HINT_MAIN_CALLBACK_RATE = "waitevent"` to get the same "render only on interaction" behavior as dvui currently has.

david-vanderson:
Excellent - thanks for experimenting!

RedPhoenixQ:
I've thought a bit about this and looked tested some things. Many of the different backends have different main functions, like the dx11 backend has an exported main function while the web backend sets restrictions on the parameters to the lifetime functions.

I've done a test were the sdl backend provides a main function where the user app can be imported into. The user code provides the functions which the main function imports and uses. With some cross-backen api for the parameters and return values of these functions the user code wouldn't have to change at all between backends, at the cost of less control of the main loop.

For the web and dx11 backends, dvui would define all the required exports (behind a compile option for dx11? or the library exported main gets ignored?) so the user wouldn't need to know/care about what to export.

```zig
// src/backends/sdl.zig

[...]

PR #239, opened 2025-04-01
Web refactor for added modularity

RedPhoenixQ:
Addresses #235

This cleans up the javascript file for the web backend. Moreover all extern functions are now under the namespace "dvui". See the new `dvui` function for an example of how to instantiate the wasm module with the new refactor. Most of the code is now part of the `Dvui` class and most variables and functions have JSDoc strings added for editor integration.

Some remaining questions are:

- Should the `dvui` function remain in web.js? Perhaps this is something that should be a part of `index.html` to be copied by users that want to instantiate it themselves?
- Could/should the `wasm_` prefix be removed from the extern functions? This is purely a cosmetic change, but i believe `wasm.sleep` reads better than `wasm.wasm_sleep` and it fits with the web refactor

### DIFF
--- a/src/backends/web.js
+++ b/src/backends/web.js
@@ -985,10 +985,44 @@ class Dvui {
         );
     }
 
+    init() {
+        let app_init_return = 0;
+        let str = utf8encoder.encode(navigator.platform);
+        if (str.length > 0) {
+            const ptr = this.instance.exports.gpa_u8(
+                str.length,
+            );
+            var dest = new Uint8Array(
+                this.instance.exports.memory.buffer,
+                ptr,
+                str.length,
+            );
+            dest.set(str);
+            app_init_return = this.instance.exports.app_init(
+                ptr,
+                str.length,
+            );
+            this.instance.exports.gpa_free(ptr, str.length);
+        } else {
+            app_init_return = this.instance.exports.app_init(
+                0,
+                0,
+            );
+        }
+
+        if (app_init_return != 0) {
+            console.log(
+                "ERROR: app_init returned " + app_init_return,
+            );
+            return;
+        }
+    }
+
     run() {
+        this.init();
+
         let renderRequested = false;
         let renderTimeoutId = 0;
-        let app_initialized = false;
 
         const render = () => {
             renderRequested = false;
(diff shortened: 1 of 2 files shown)


david-vanderson:
Thank you very much!

I will look into this, but I'm not super up to date with modern javascript.  @MichaelBelousov can you take a look and weigh in here?

MichaelBelousov:
@david-vanderson will do

src/backends/web.js
```diff
@@ -29,670 +40,1035 @@ async function dvui_open_file_picker(accept, multiple) {
...
-            const cached = filesCache.get(id);
-            if (!cached) return 0;
-            return cached.files.length;
-        },
-        wasm_clipboardTextSet: (ptr, len) => {
-            if (len == 0) {
-                return;
-            }
+        if (!canvas) {
+            alert("Could find canvas element.");
```
MichaelBelousov, line 863:
```suggestion
            alert("Could not find canvas element.");
```

src/backends/web.js
```diff
@@ -29,670 +40,1035 @@ async function dvui_open_file_picker(accept, multiple) {
...
+            this.instance.exports.gpa_free(ptr, str.length);
+        } else {
+            app_init_return = this.instance.exports.app_init(
+                0,
+                0,
+            );
+        }
+
+        if (app_init_return != 0) {
+            console.log(
```
MichaelBelousov, line 1012: minor nitpick
```suggestion
            console.error(
```
maybe even consider throwing/returning an error code for people to handle
but maybe you didn't write this part

MichaelBelousov (approved):
LGTM for the most part.
definitely looks like an improvement.

imo:
- keep `dvui` in web.js

fyi @RedPhoenixQ lmk if you are using dvui in both a web and native context since I have some thoughts on doing that related to #165

src/backends/web.zig
```diff
@@ -27,42 +27,42 @@ const EventTemp = struct {
...
-    pub extern fn wasm_renderTarget(u32) void;
-    pub extern fn wasm_textureDestroy(u32) void;
-    pub extern fn wasm_renderGeometry(texture: u32, index_ptr: [*]const u8, index_len: usize, vertex_ptr: [*]const u8, vertex_len: usize, sizeof_vertex: u8, offset_pos: u8, offset_col: u8, offset_uv: u8, clip: u8, x: i32, y: i32, w: i32, h: i32) void;
-
-    pub extern fn wasm_cursor(name: [*]const u8, name_len: u32) void;
-    pub extern fn wasm_text_input(x: f32, y: f32, w: f32, h: f32) void;
-    pub extern fn wasm_open_url(ptr: [*]const u8, len: usize) void;
-    pub extern fn wasm_download_data(name_ptr: [*]const u8, name_len: usize, data_ptr: [*]const u8, data_len: usize) void;
-    pub extern fn wasm_clipboardTextSet(ptr: [*]const u8, len: usize) void;
+    pub extern "dvui" fn wasm_about_webgl2() u8;
```
MichaelBelousov, line 30: TIL you can choose the wasm import namespace using extern dll name specifiers (or whatever they're called)

src/backends/web.js
MichaelBelousov: I assume this also is just code you restructured without editing

src/backends/web.js
RedPhoenixQ: This remained from the previous code which has just been moved. I'm debating if the user should call this `init` function or if it's only going to be called from `run`. In either case it is probably better to throw an error so that is can be handled or shows up in the console

MichaelBelousov: I would vote you throw the error, but probably want to start doing that elsewhere here too so you can leave it for now if you like... when I migrate my project's SDK to this new version, I will give it more thought

src/backends/web.zig
david-vanderson: Me too!

RedPhoenixQ:
> fyi @RedPhoenixQ lmk if you are using dvui in both a web and native context since I have some thoughts on doing that related to #165

I'm currently making a windows only application, but I'd be interested in discussing an abstraction of backends. The SDL3 approach seems the be very similar to the implicit setup of `examples/web-test.zig`. I my head you'd make these extern function for web like it is now, and then have a main function for other platforms that looks something like:

```zig
pub fn main() !void {
    try dvui.runApp(myAppState, myInitFn, myUpdateFn, myDeinitFn);
}
```

We can continue the discussion in #165 to keep this PR clean

david-vanderson:
Looks good to me - go ahead an commit when you are ready!

PR #243, opened 2025-04-04
Update autodocs look and feel

Yinameah:


david-vanderson:
Very excellent.  I've pushed this to https://david-vanderson.github.io/docs/ as well.  Thank you!

PR #245, opened 2025-04-04
App runner with dvui provided main functions

RedPhoenixQ:
This related to #165.

This adds a proof of concept with `sdl-app` and `raylib-app` examples which reuses the same application code for both backends. The backend defined a main function for use as the applications root and the user code gets imported as `app` with predefined exports.

This can be extended with more optional lifetime functions like `pub fn configure_window(win: *dvui.Window) void {...}` which would only be called if `@hasDecl(App, "configure_window")` is true.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -86,8 +86,9 @@ pub fn build(b: *std.Build) !void {
 
         const dvui_sdl = addDvuiModule(b, target, optimize, "dvui_sdl", true);
         linkBackend(dvui_sdl, sdl_mod);
-        addExample(b, target, optimize, "sdl-standalone", dvui_sdl);
-        addExample(b, target, optimize, "sdl-ontop", dvui_sdl);
+        addExample(b, target, optimize, "sdl-standalone", dvui_sdl, null, null);
+        addExample(b, target, optimize, "sdl-ontop", dvui_sdl, null, null);
+        addExample(b, target, optimize, "app", dvui_sdl, sdl_mod, .sdl);
     }
 
     // Raylib
@@ -147,8 +148,8 @@ pub fn build(b: *std.Build) !void {
 
         const dvui_raylib = addDvuiModule(b, target, optimize, "dvui_raylib", false);
         linkBackend(dvui_raylib, raylib_mod);
-        addExample(b, target, optimize, "raylib-standalone", dvui_raylib);
-        addExample(b, target, optimize, "raylib-ontop", dvui_raylib);
+        addExample(b, target, optimize, "raylib-standalone", dvui_raylib, null, null);
+        addExample(b, target, optimize, "raylib-ontop", dvui_raylib, null, null);
     }
 
     // Dx11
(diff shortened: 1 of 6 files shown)


david-vanderson (approved):
Mostly looks great!  I have 3 things I'm unsure about:
* Instead of passing the kind of backend via build.zig, can we share the Backend enum between build.zig and the rest of the code, then each backend can have `pub const kind: Backend = `?
* Do the two different web examples clobber each other's index.html (and possibly other files)?  Maybe we make each web example have it's own subdir under zig-out?
* I'm struggling with how the dvui.App functions should work.  I think you are on the right track, but maybe the first should be `startFn` and returns the InitOptions (maybe renamed), then `initFn` would be the one that takes `*Window`, so then we say `deinitFn` is run before the backend deinits the Window and Backend?

No obligation though - if you want to go ahead and commit and I'll address these later.

Thank you!

RedPhoenixQ:
Thanks for the feedback. Will work on it tomorrow. 

> * Instead of passing the kind of backend via build.zig, can we share the Backend enum between build.zig and the rest of the code, then each backend can have `pub const kind: Backend = `?

Isn't this effectively what adding it via build.zig does? Share the enum that is. I preferred this version as it makes build.zig the source of truth. 

> * Do the two different web examples clobber each other's index.html (and possibly other files)?  Maybe we make each web example have it's own subdir under zig-out?

They do completely overwrite each other. Separate subdirs is a great solution, will add that.

> * I'm struggling with how the dvui.App functions should work.  I think you are on the right track, but maybe the first should be `startFn` and returns the InitOptions (maybe renamed), then `initFn` would be the one that takes `*Window`, so then we say `deinitFn` is run before the backend deinits the Window and Backend?

That naming seems good, better connection between `initFn` and `deinitFn`. Maybe `InitOptions` should be `WindowInitOptions` for clarity. 

Also I'm thinking the new `initFn` and `deintiFn` could be optional. If you don't have anything to change on the window, you don't need to provide a function for it. Only mandatory functions should be generating the `WindowInitOptions` and `frameFn`.

david-vanderson:
> Isn't this effectively what adding it via build.zig does? Share the enum that is. I preferred this version as it makes build.zig the source of truth.

It's almost the same, but doing it in build.zig means it's possible to screw it up.  I think we want the source of truth to come from the backend itself, in case anyone is doing something tricky with build logic.

david-vanderson:
> That naming seems good, better connection between `initFn` and `deinitFn`. Maybe `InitOptions` should be `WindowInitOptions` for clarity.

Or maybe `StartOptions` in case we ever put stuff in there that's unrelated to Window stuff.  Can't really think of anything right now though.

> 
> Also I'm thinking the new `initFn` and `deintiFn` could be optional. If you don't have anything to change on the window, you don't need to provide a function for it. Only mandatory functions should be generating the `WindowInitOptions` and `frameFn`.

Good idea!

RedPhoenixQ:
> It's almost the same, but doing it in build.zig means it's possible to screw it up. I think we want the source of truth to come from the backend itself, in case anyone is doing something tricky with build logic.

Good point about build logic. I removed the backend enum and used @hasDecl to determine the backend in the example. I believe that shows how you could determine the backend as well as being able to differentiate between many custom backends if needed. It could be made a helper function or const in Backend.zig with the known backends

david-vanderson (approved):
Excellent!  I like the union for the start options too.  Thank you!

PR #246, opened 2025-04-05
Dx11 cursors

RedPhoenixQ:
Dx11 backend did not set the cursors previously.

The avoid some quirks with the window resizing cursors the registered class cursor is modified as that is used as the default cursor when not resizing.


david-vanderson:
I think you know better than me in this area, so go ahead and commit!

RedPhoenixQ:
I just tested with the example until it felt correct. Any potential issues, if any, can be fixed in the future

PR #247, opened 2025-04-05
Animation easing

RedPhoenixQ:
This adds animation easing, which is a very cheap and simple improvement to the feel of an application. This could be done on the user side before by having `start_val` and `end_val` be 0 and 1 and doing this easing yourself. Having access to premade functions makes the barrier to nicer animation much lower.

`Animation` now has a `easing` field for an `EasingFn`. A linear function is passed as a default so the old behaviour stays the default.

Currently only the spinner is making use if the new easing functions, but it is a significant visual improvement imo.

chrome-capture-2025-4-5

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -94,7 +94,7 @@ const AnimatingDialog = struct {
         const winHeight = win.data().rect.h;
 
         if (dvui.animationGet(win.data().id, "rect_percent")) |a| {
-            win.data().rect.h *= a.lerp();
+            win.data().rect.h *= a.value();
 
             // mucking with the window size can screw up the windows auto sizing, so force it
             win.autoSize();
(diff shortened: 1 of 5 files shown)


david-vanderson:
This is great!  Can you add a clamp to the last line of `lerp()`?
`return (a.start_val * (1.0 - t)) + (a.end_val * t);`

So we ensure that even with floating point inaccuracies the return value is between start_val and end_val.

Good to commit - Thank you very much!

RedPhoenixQ:
There are reasons to not do that. The [elastic](https://easings.net/#easeInOutElastic) easing overshoots to get a swaying effect.

We could return `end_val` directly once the animation has finished to guarantee that the final value is what was requested. How does that sound

david-vanderson:
Very good point, then disregard that.

RedPhoenixQ:
Added some clamping for pre and post animation. Seemed unreasonable that you could have a start delay and not receive the `start_val` precisely.

Added an elastic easing to the dialog drop example. Looks pretty good!

The other examples uses `dvui.animate` which only takes a duration parameter. I could add easing option to it as a part of this PR

david-vanderson:
Good idea about the start and end - I agree.  Yeah I think adding an easing option to `dvui.animate` is good.  Thanks!

david-vanderson (approved):
Looks great.  If you have a chance, a doc comment for `lerp()` that warns about values outside start_val/end_val would be good.

Thank you!

david-vanderson:
Especially appreciate the demo additions!

david-vanderson:
In fact, now that I think about it, `lerp()` is a bad name now.  How about `val()`?

RedPhoenixQ:
`value()` sounds good. Didn't want to change the public api too much. 

Will add some doc comments for `Animation.value` as well as the individual easing functions to warn about them overextending past 0 or 1.

RedPhoenixQ:
I believe that is all. Feel free to merge if this looks good to you!

david-vanderson:
Looks fantastic - thank you!

PR #249, opened 2025-04-06
Update docs and file split

### LINKED ISSUE #244: Improve autodocs
I played a bit (ok, no, I was totally absorbed most of my afternoon) with autodocs. This is super super super cool. Zig is incredible.

Anyway, I customized a bit the `index.html` file and update the build system to inject the logos at built time so we can change it easily. It's a bit of an overkill tbh, but building docs is so absurdly fast I think it's cool like that.

With #243 the docs looks like this : 

Image

Image

Yinameah:
So, I started to add some docs around.

But first, I wanted to fix the fact that the widgets display in the Navlist leads to a "declaration not found" page.

image

I tried some `build.zig` approaches, but that was adding way too much complexity, so I instead created `import_widgets.zig` where all wigets are imported, and then change the import syntax in `dvui.zig`. This is a bit more clubersome when adding more widgets, but has the benefit to display a nice recap page of all widgets in the doc.

Since I was at this, I moved the widgets (DropdownWidget, SuggestionWidget, TabsWidget, PlotWidget) that were in the main `dvui.zig` file into their respective file in `widgets` folder and also extracted `Window` struct out of the main file. This is almost 2'000 lines that were making the navigation sometimes confusing. 

I had to mark a few extra stuff as `pub` for it to work. It's a straightforward change and I'm confident I didn't break anything, but since it's a big diff, I don't want to merge it without notice, since it might be annoying for merge of ongoing work.

@david-vanderson @RedPhoenixQ If you have ongoing work that might be tricky to merge after that, let me know, I can wait a bit an rebase on main when your changes are merge.

Cheers.


RedPhoenixQ:
> If you have ongoing work that might be tricky to merge after that, let me know,

Not currently. All good for me

Yinameah:
Ah, and also in the new widget files, I adopted the convention to have basically : 

- fields
- types
- functions
- and imports at the end

This match what the doc does. I read an argument recently about having a lot of imports at the top is a lot of screen space when opening a file for very little useful information. I find the argument very compelling. Let me know if you have a problem with that.

david-vanderson (approved):
Looks great - thanks very much!

RedPhoenixQ (approved):
Looks good! Tested the sdl-standalone example and everything indeed seems to work

Issue #248, opened 2025-04-06
Image render backend for docs (and test ?)


Yinameah:
I just discovered that if you do : 

```zig
/// ![basicWidgets image](basicWidgets.png)
pub fn basicWidgets(demo_win_id: u32) !void {
    {
        ...
```

well... It just work™

![Image](https://github.com/user-attachments/assets/1a43c740-1d50-41d7-aa3d-7d0b5313d232)

This is super cool, but I don't really feel like doing this manually because it's a bit cumbersome, but most importantly it will very easily gets out of sync.

This got me thinking : How hard could it be to have a simple mock backed that would render a single frame into an image ? If we had this, we could `build.zig` our way into making the examples functions automatically generate their doc comment with an image. How cool would that be ?

Also, I'm reflecting about testing, and I recently read https://kristoff.it/blog/dead-simple-snapshot-testing/ (and other stuffs about snapshot testing) and it immediately made me think that it would be super cool to have a gui library tested against images that way. Clearly not straight forward, but having an image backend would be the first step.

Do you think this is a good idea ? Feasible ?

RedPhoenixQ:
@Yinameah Where are the images located? Are they in the output `zig-out/docs` directory or somewhere in `src` relative to the source file?

Some questions and initial thoughts:

Most widget would also need a separate "preview" function to render its image. I don't know where a good place would be for such functions. Maybe a separate file with the name of the image, minus the extension?

How will the "preview generator" files be located and rendered? 
 1. Maybe, if an preview image = one source file, a file name like `X.png.zig` to indicate that it renders the `X.png`?
 2. Maybe just one big `preview_images.zig` with a function for each image and then a executable file used by build.zig that calls all fn decls in that file?

Snapshot testing definitely seems like the most reasonable way to test the current codebase. Would be interesting. 

The "static render backend" would also need to run of at least 2 frames so widgets can get their sizes correctly.

david-vanderson:
Thanks!  I also did not know you could do that in docs.

That is an interesting idea.  I think it's plausible, and could be useful beyond dvui docs.  But there are a lot of details to work out so we might practically be a ways off.

We have the basic building blocks:
* renderTarget means we can produce an image of what we rendered, even without a separate backend
* CacheWidget does a (currently undocumented and bad api) thing where it waits until a frame where none of the child widgets request another frame
  * this solves the "how to wait until everything is sized correctly"
  * this doesn't solve "how to capture a frame in the middle of an animation"

[...]

Issue #171, opened 2024-12-22
Add ability to tag widgets for tests


MichaelBelousov:
I have a large UI where I would like to be able to do something like:

```zig
test "click user 3" {
  var app = app.init();
  defer app.deinit();
  app.frame();
  
  // imaginary API, probably can use existing event API but might be nice to have a helper
  // here that just assumes a center insertion
  dvui.testing.click(.UserWidget, .{ .id_extra = 2 });
  
  app.frame();
  
  std.testing.expectEqual(app.selected_user_index, 2);
}
```

Then my app would have code like:

```zig
var selected_user_index: ?usize = null;
fn frame() !void {
  var hbox = try dvui.box(.horizontal);
  defer hbox.deinit();
  for (users.items, 0..) |user, i| {
    var user_box = dvui.box(.vertical, .{ .id_extra = i });
    defer user_box.deinit();
    try dvui.label("user {}", .{i}, .{ .id_extra = i, .test_tag = .UserWidget });
  }
}
```

And the zig enum literal passed to `test_tag` could be used to have an efficient name that can be used to refer to the widget, with dvui erroring if you duplicate the same `test_tag,id_extra` pair.

As a separate issue, it would be nice to add a "null backend" to ignore drawing widgets, for
test scenarios, I will add such an issue later, right now I just don't initialize my backend which seems to work ok.

fyi, this is lightly based on the common pattern in webdev of adding `data-testid` attributes to DOM elements to query them in tests

david-vanderson:
This is a good idea, and having a null testing backend is also a great idea.  Could in the future be used to test if widgets are in/out of the clipping region for example.

Will put some thought into how to refer to widgets for testing purposes, how to insert events to match them, and the flow of tests.  Need some way to specify longer logical tests like:
* click button "Show Dialog"
* maybe wait for dialog to finish animating
* test if dialog button "ok" is visible (and maybe has focus)
* click "ok"
* maybe wait for dialog to animate closed
* test if dialog is no longer shown

MichaelBelousov:
I would avoid complexity but it might be worth considering if the API gets complex enough to try to follow zig's `std.testing`:

```zig
test "show dialog" {
  var app = App.init();
  defer app.deinit();
  
  try dvui.testing.actions.click(.ShowDialogButton, null);
  try dvui.testing.runUntilNotAnimating(.ShowDialogButton, App.frame, .{app});
  try dvui.testing.expectVisible(.DialogOk, null);
[...]

PR #251, opened 2025-04-08
Programmatic interactions and testing (with snapshots)

RedPhoenixQ:
**Edit**: Expanded the scope of the PR to include testing,  see [my comment bellow]( https://github.com/david-vanderson/dvui/pull/251#issuecomment-2794902928)

This adds a proof of concept headless runner that can interact with the application and capture frames.

The frame capturing can be used for #248 and the programmatic interactions are similar to #171.

The widget tagging cannot easily be done with enum literals like proposed as that would force some level of comptime knowledge on the Window. A string id is used instead.

Also implements a headless option for the SDL backend which sets the appropriate flags as mentioned in #248.

Here is a sample of two captured frames before and after programmatically pressing the "File" menu item.

before interaction

after interaction


Yinameah:
So cool.

I'll have a look and try asap.
On business trip currently so not so much time for the next days. But will check

david-vanderson:
Wow - Awesome!  I'll dig into this in the next few days.

RedPhoenixQ:
I added a `testing` module which uses the Runner to perform tests as well as handle snapshot testing.

I've some design questions to think about:

- The testing api mostly calls directly to the Runner.It would be a nicer API if the runner and testing where combined. I have two ideas here:
   - The Runner could implement the `expectX...` methods with a comment that they're only intended for tests.
     - The `testing` module could remain only for `init` and `deinit` of a testing window, by ultimately just return a `Runner` OR another init/deinit pair could be added for creating the testing window, e.g `Runner.initTest` and `Runner.deinitTest`
   - The `Runner` gets moved to the `testing` module and `@import("builtin").is_test` asserts that the runner is only used for testing
     - This would remove the usefulness for outside test, if there was any
     - This allows for all the overhead in `Window` to be removed completely with comptime fields and branches as it's guaranteed 
there won't be a `Runner` during normal operations
       - It would also be nice to remove the `test_id` field from Options in this case, if possible
- The `testing` module currently gets passed and allocator, which will almost always be `std.testing.allocator`. Perhaps it should just always use `std.testing.allocator` directly without being passed anything.
- I'm not sure the test in the examples should stay. Perhaps they should still exits with a custom folder for snapshots to show examples of how dvui's testing works?

david-vanderson:
This is going in a good direction, and I have some ideas I'll work on tomorrow:

The test_id thing is useful in general, not just in a testing scenario, so I'm thinking dvui.Window should have a general "tag" capability that can associate a known name (plus id_extra?) with a widget (id, visible, rect on screen, other stuff?)
* can be a way for one widget to lookup the id of another widget (to move focus for example) without having to store it on the app side
* can enable UI scripting that's sort of like running a test, but just automates part of the UI - imagine a game tutorial where during the "scripting" phase it highlights some widgets and draws arrows towards them

Extending the snapshot to be able to take pictures of part of the screen, especially a widget rect.

Testing how long running tests take, maybe we can reuse the backend OS window part, and just reinit dvui.Window?

As I'm trying to write fake scripting/test code, I'm thinking instead of saying "run a frame from Window.begin to Window.end", we might want "run Window.end and Window.begin".  Then even in a normal dvui.App.frameFn we could write something like:

```zig
if (scripting) {
    if (micros_from_last_window_end == 0) {
        // not settled
        myFrame();
        return;
    }
    win.tagClick("foo");
    myFrame();
    win.endBegin();
    win.tagClick("bar");
    myFrame();
}
```

Not sure if this will survive reality, but wanted to get some of my thoughts out.

david-vanderson:
Good comments.  Yes a lot of the "run a frame" kind of things would be very weird if they were called in the middle of a frame.  I think we can protect against that by checking that the parent widget is still the base Window.

From my explorations, I think the most useful place for testing code to happen is just after Window.begin - that way all of the dvui functions can be called, and events can be added.

Sorry I'm going slowly on this, but hang with me and I'll get there.

RedPhoenixQ:
> Good comments. Yes a lot of the "run a frame" kind of things would be very weird if they were called in the middle of a frame. I think we can protect against that by checking that the parent widget is still the base Window.
>
> From my explorations, I think the most useful place for testing code to happen is just after Window.begin - that way all of the dvui functions can be called, and events can be added.

Thinking about it I quite like that for a couple of reasons:
 - The testing `init` function could do the first `win.begin()` call, and `deinit()` do the last one
 - You wouldn't have to pass your `frameFn` anywhere, just call it yourself, interleaving `win.endbegin()`
 - All the API would be available on the Window
 - The `dvui.testing.expectX` functions can call `dvui.currentWindow`, which mean they could look very similar to the `std.testing` api with no additional parameters

My only concern is the "run until settled" case, which I think should have a nice api. Maybe `Window.settled()` to call in a while loop predicate? Or `endbegin()` could return true if not settled so it can be used in a while loop? 

> Sorry I'm going slowly on this, but hang with me and I'll get there.

No problem at all. I'm happy to do the work with your pointers if you want, otherwise take your time. I don't actively have a use case for this, it's just an interesting problem.

RedPhoenixQ:
I also created a quit no-op dummy backend on [this branch](https://github.com/RedPhoenixQ/dvui/tree/dummy-backend) that can be merged later if needed. It does no rendering and it useful for for quickly testing dvui logic (event handling and visibility and so on)

Yinameah:
Currently, this branch fails for me to `zig build sdl-picture` with root source file struct 'dvui' has no member named 'Runner'. Actually, I don't see the definition for Runner.

Did you guys forgot to commit a file or am I missing something obvious ?

david-vanderson:
@RedPhoenixQ had everything working and then I broke it while reworking it.  I will get to it, but yeah things are going to be broken off and on here until it gets merged to main.

Yinameah:
Ok, all good. Wanted to test and see if I can help / feedback. I'll wait a bit.

RedPhoenixQ:
Forgot to fix the picture example. There are new tests with in the app.zig example though

RedPhoenixQ:
@Yinameah, @david-vanderson This should be more finished now, the temporary `sdl-picture` example has been removed in favour of `test-sdl-app`.

david-vanderson:
I think we got this to a pretty good place, so we can do followup work in main.

Excellent work!  Thank you!

PR #252, opened 2025-04-10
App interface panic handlers (for web)

RedPhoenixQ:
On the web backend, no panic  information was provided as stderr isn't present and all default handling relies on it. This adds a panic handler to the `App` interface which handles this. I wasn't able to generate stack traces within zig, but the error message shown in 

Additionally, web.js now properly stops when encountering a panic instead of continuing when new events come in.


david-vanderson:
Thank  you!

Issue #253, opened 2025-04-12
SDL3 backend is broken


weskoerber:
It looks like commit 69de3187943eb6dd75bf353a711b43337f25b000 broke the SDL3 backend due to an unused return value. `SDL_EnableScreensaver` had a `void` return type in [SDL2](https://wiki.libsdl.org/SDL2/SDL_EnableScreenSaver), but in [SDL3](https://wiki.libsdl.org/SDL3/SDL_EnableScreenSaver) it returns `bool`.

<details><sumary>Click to show failed build output</summary>

```
❯ zig build -Dbackend=sdl -Dsdl3
install
└─ compile-sdl-standalone
   └─ install sdl-standalone
      └─ zig build-exe sdl-standalone Debug native 1 errors
examples/sdl-standalone.zig:49:36: error: value of type 'bool' ignored
    Backend.c.SDL_EnableScreenSaver();
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
examples/sdl-standalone.zig:49:36: note: all non-void values must be used
examples/sdl-standalone.zig:49:36: note: to discard the value, assign it to '_'
error: the following command failed with 1 compilation errors:
[...]

PR #254, opened 2025-04-12
fix: sdl backend broken due to unused return values

weskoerber:
closes: #253

I looked around at other call of SDL functions and it looked like in the vast majority of cases, the return values are discarded. Let me know if this is appropriate!


RedPhoenixQ:
Good catch! That's my bad for not testing sdl3. Thank you

david-vanderson:
Excellent - thank you!

Issue #221, opened 2025-03-24
(feature request) gradient  backgrounds


MichaelBelousov:
would it be possible to support gradient backgrounds or have a demo showing how to do it with existing APIs?

david-vanderson:
I've pushed an initial commit of this (also in the web demo).  See the "Styling" part of the demo.

@MichaelBelousov Do you need more complicated gradients?  If so, can you describe and/or post a picture?  Thanks!

MichaelBelousov:
No I don't, that looks ok to me for now! I would probably add the word "gradient" somewhere to the web demo though since it's hard to tell what that dropdown does unless you know what you're looking for.

david-vanderson:
Whoops - thanks!

Issue #220, opened 2025-03-24
(feature request) simple box shadows


MichaelBelousov:
could DVUI, for common styling purposes, support the ability to render a "box shadow" behind a box?
I imagine you could do it now with low level features but could their be a more standard API if there is a desire?

RedPhoenixQ:
Hello. I don't know if there will be a standard API for this, but here is an example which should get the results you want. You do need to use the widgets directly to control the order of what is drawn.

```zig 
var box = dvui.BoxWidget.init(@src(), .horizontal, false, .{ .background = true, .min_size_content = .{ .w = 40, .h = 30 } });
try box.install();

// The drop shadow with offset
try box.data().borderRectScale().r
    .offset(.{ .x = 10, .y = 10 })
    .fill(box.data().options.corner_radiusGet(), dvui.Color.black);

try box.drawBackground(); // draw box after shadow
box.deinit();
```

Hope this helps

david-vanderson:
Good response!

@MichaelBelousov is a relatively "hard" shadow like this okay?  I've been thinking about ways to do more diffuse stuff, but maybe it's not necessary right now for your purposes?

david-vanderson:
It think I have a plan for dealing with this and #221 at the same time:
* break rendering functions into a triangulation step and a `renderTriangles()` step
* user would call something like `Rect.fillTriangles()` that returns/fills a vertex and index list
* then the user is free to manipulate the vertices in any way
* but we would start providing easy functions like `applyGradient()` that go over the vertex list and change the colors
* or `applyTexture()` to set uv coords to handle #204 
* then call the `renderTriangles()` function, which mostly just calls the backend `drawClippedTriangles()` but handles deferred rendering and clipping

Luckily for all 3 issues we only need to handle the rounded rect case for now.

david-vanderson:
Ok good - that means we probably also need to add an option to fillRect to control how large the antialias fringe is.  Hopefully that (maybe combined with rendering it multiple times at different offsets) will be enough.

david-vanderson:
I've committed the infrastructure support for this.  There is a new part in the styling part of the demo (and online demo) where you can play around with box shadows.

@MichaelBelousov please play around with it and recommend some default values for those things.  The next step would be to add some kind of dvui.boxShadow() function to make it easier to use.

MichaelBelousov:
[...]

PR #255, opened 2025-04-12
Triangle abstraction

### LINKED ISSUE #204: draw texture with rounded corners
Motivation: to show an avatar (might be a person's picture) with rounded corners (circle outline is a special case).

Instead of a separate function for this, I'm going to try adding a generic function that takes a path and an origin (and scale?) and draws a texture in the triangles produced by that path.  Each triangle vertex would get a texture uv based on how far from the origin that vertex was.

Maybe instead of an origin/scale, it should take a rect (in physical pixels) that defines where the texture would be drawn on the screen (possibly rescaled) and the vertexes should be inside that rect.

RedPhoenixQ:
Closes #204

This effectively implements the base for https://github.com/david-vanderson/dvui/issues/220#issuecomment-2758640352
This PR could also be expanded to include gradients by creating a gradient triangle function for #221 and "feathered" box shadow triangles for #220.

The layout example for images have been extended to show these new features.

{69F0EB95-508B-4F8B-AF8B-38836264E992}

There is some issue with rounded corners on images combined with shrinking. I've compromised by ratio shrinking not being able to be used without the `Option.expand` being `.ratio` as well.  The previous implementation could also be left as it works for non-rounded corners.

The ratio shrink with rounded corners could be solved by rendering the image to a texture matching the size of the rect leaving some padding, and then rendering that texture with the rounded corner to fix this. There is no good place to inject this extra textures without the `shrink` needing to be passed down to the lower level functions. There is also an issue of caching this temporary texture if possible, otherwise creating multiple textures every frame would be too much.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -53,6 +53,8 @@ var layout_border: Rect = Rect.all(0);
 var layout_padding: Rect = Rect.all(4);
 var layout_gravity_x: f32 = 0.5;
 var layout_gravity_y: f32 = 0.5;
+var layout_rotation: f32 = 0;
+var layout_corner_radius: Rect = Rect.all(5);
 var layout_flex_content_justify: dvui.FlexBoxWidget.ContentPosition = .center;
 var layout_expand: dvui.Options.Expand = .none;
 var show_dialog: bool = false;
@@ -1362,6 +1364,7 @@ pub fn layout() !void {
         const Static = struct {
             var img: bool = false;
             var shrink: bool = false;
+            var background: bool = false;
             var shrinkE: dvui.Options.Expand = .none;
             var size: Size = .{ .w = 16, .h = 16 };
         };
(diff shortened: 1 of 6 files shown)


david-vanderson:
I'm trying to understand the issue here.  Why do we need to use uv to do the ratio shrinking?

My mental model of this is:
* get the rect for where the texture is going to be
* make rounded filled path triangles inside that rect
* map the uv coords (will all be inside the texture)

What am I missing?

RedPhoenixQ:
The problem is the rounded corner applies only to the full rect, so expanding the rect also stretches the corner rounding the image is supposed to be contained within. 

{79CE2EF1-E40A-4ADA-9BA8-D3DC369E2AD9}

We could leave it as before with a note in the docs that vertical/horizontal shrinking might break with corner radius. I appears to work ok with with shrink set to .both and .ratio on commit dfd285992feee92cfda2ca0c09cf93dfe929a883

{45A297BC-D558-4CE9-A22F-17CDF8AAA045}

david-vanderson:
Something about this is just not going into my brain.  I don't get "expanding the rect also stretches the corner rounding".  I'll need to checkout the branch and play around with it.

RedPhoenixQ:
> I wonder if we have different expectations about how expanding/shrinking should interact with rounded corners?
> 
> In this implementation, the expanding/shrinking happens first. Then you have a final rectangle on the screen, and then round the corners of that. So the rounded corners always "eat into" the texture. Does that help at all?

Yeah we did, I come mostly from web so expected the image to be "clipped" first and then rounded, but this makes sense.

> I took your code and ran with it. Sorry I couldn't figure out a way to easily keep the attribution.

Don't worry about the attribution, the mention in the commit is enough. Nice to have this effectively merged to extend with a function for gradient triangles and so on.

david-vanderson:
Awesome thank you!

PR #257, opened 2025-04-15
Add check and test build steps

RedPhoenixQ:
Adds a check and test build step that together, currently, semantically checks all declarations in the code base. While adding these steps I also found some bugs in code paths that weren't referenced, see b71d95145b6d463094ac21c1f8e6a00fc1858338.

This is motivated by me creating compilation errors in compile configurations I did not actively test or for functions that weren't referenced. It could also be integrated into CI to verify that everything compiles and run the tests in CI too, which I would be happy to add. Don't know how opening OS windows works in that environment, so maybe not `dvui.testing` tests.

Adding `std.testing.refAllDecls(@This())` to every file does prevent some patterns like `pub const something = @compileError("...")` as that would always fail the test compilation. This has been worked around by checking `builtin.is_test` or by reworking the api, like for `dvui.App` which now has `dvui.App.get()` which returns an optional to prevent a compile error for dereferencing a compile time `null`.

As a part of this the sdl3 backend was separated into its own `enums.Backend` variant, which allows for both of them to be compiled at the same time when not explicitly picking a backend. This would have prevented #253.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -25,7 +25,7 @@ pub fn build(b: *std.Build) !void {
 
     const test_step = b.step("test", "Test the dvui codebase");
 
-    // SDL
+    // SDL2
     if (back_to_build == null or back_to_build == .sdl) {
         const sdl_mod = b.addModule("sdl", .{
             .root_source_file = b.path("src/backends/sdl.zig"),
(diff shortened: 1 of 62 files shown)


david-vanderson:
This looks great, go ahead and merge.  I'll try later to add some error messages to help people upgrade, but no reason to hold for that.

PR #258, opened 2025-04-17
dx11: implement renderTarget, textureCreateTarget and textureRead

RedPhoenixQ:
These functions are all needed for `CacheWidget` to be able to create a cached image. This improves the fps of the demo grid by alot, enabling 60fps compared to around 30fps before on my machine.

There is a lot of copying data in the implementation, but dx11 seems to require making a copy of the texture to a specific readable texture.


david-vanderson:
Looks fantasic - merge when ready.  Great work!

PR #259, opened 2025-04-17
Dummy backend (for testing)

RedPhoenixQ:
This PR adds a dummy, or no-op, backend that does no rendering or OS interaction. It is intended to be used for tests were OS windows and rendering isn't needed, which is almost all test except `dvui.testing.snapshot`. If you are not using snapshots this backend will be much faster as it does not need to do any rendering but still tests all dvui logic like focus and visibility. 

The name could be changed to `NoOpBackend` or `TestingBackend` if dummy is a poor choice of word. If more testing functionality was going to be added, like verifying that triangles sent to the backend are correct, could also be added and in that case `TestingBackend` would be more fitting.

Additionally this changes the cache widget to prevent endless calls to `dvui.refresh` which stops the window from settling. If it cannot get a renderTarget on the second attempt, it doesn't call refresh and makes a new attempt during the next frame, stopping the endless loop.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -48,17 +48,16 @@ pub fn build(b: *std.Build) !void {
         }
     }
 
-    // Dummy, does no rendeing and is intended for tests
-    if (back_to_build == null or back_to_build == .dummy) {
-        const dummy_mod = b.addModule("dummy", .{
-            .root_source_file = b.path("src/backends/dummy.zig"),
+    // Testing
+    if (back_to_build == null or back_to_build == .testing) {
+        const testing_mod = b.addModule("testing", .{
+            .root_source_file = b.path("src/backends/testing.zig"),
             .target = target,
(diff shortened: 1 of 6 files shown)


david-vanderson:
That looks great!  Good catch on the CacheWidget bug - did you use the refresh logging to find it, or something else?

For naming - let's call it "TestingBackend".  That seems most descriptive and I could see us adding extra testing checks and stuff later.  Merge when ready - thank you!

RedPhoenixQ:
> Good catch on the CacheWidget bug - did you use the refresh logging to find it, or something else?

I did use refresh logging on the test cases that failed due to `error.settle` and saw only CacheWidget in there. It was a pretty good debug experience.

I'll rename to `TestingBackend` and `enums.Backend.testing` and merge, thank you!

PR #260, opened 2025-04-17
Backend logging scopes

RedPhoenixQ:
Previously only the Dx11 backend had it's own `std.log.scope` which this PR implements for all backends. There was also some printing that happened through `std.debug.print` which has now also been changed to use the log api.


david-vanderson:
Good idea - merge when ready!

PR #261, opened 2025-04-18
Documentation image generation

### LINKED ISSUE #248: Image render backend for docs (and test ?)

RedPhoenixQ:
Closes #248

This is the wip PR for documentation image generation. Thanks to the many changes and the addition of `dvui.testing` this is now quite trivial. As of now an image of the Example demo window is created and linked to through a top level doc comment. 

It would be nice if there was a nicer api to create new image functions. I believe we could use the zig tests to run all tests and use the test names as file names. To make that work a custom test runner, which would replace the current `generate_images.zig` main function, would get the output dir and allow the tests to save files there. I haven't figured out how to do multiple images from the same test, perhaps the save function takes a `sub_name` param that will be appended to the test name.


RedPhoenixQ:
Now all tests that end with ".png" are ran with the doc generation. When it's not running with `image_gen_test_runner.zig`, it will ignore the image generation function and treated as a normal test. This way the dvui codebase can increase its test coverage while generating doc images

RedPhoenixQ:
This is the current state of the generated documentation images

{F9E849E4-1DA4-4650-87CB-AE0EF958EB5E}

{E031DFBB-D5D0-48B3-8523-9996689E23B6}

david-vanderson:
This is very cool!

It seems like we might be overloading zig's tests, but I can't think of another way of doing it.  I have a feeling that there's a bit too much "magic" in terms of overloading the test name and whether we are doing testing or doc image generation, but I haven't gotten a chance to really engage with the code yet.

RedPhoenixQ:
> It seems like we might be overloading zig's tests, but I can't think of another way of doing it. I have a feeling that there's a bit too much "magic" in terms of overloading the test name and whether we are doing testing or doc image generation, but I haven't gotten a chance to really engage with the code yet.

That might be fair. There is nothing preventing the current solution from being moved to a new specific image gen file, like the first iteration in e0088d32f274b98766b3f0ac839650c65c77b04b. I preferred this as the tests still run during normal tests, encouraging wider test coverage. Additionally it makes the "discovery" easier as you don't have to list all image gen functions somewhere to be ran

david-vanderson:
This is good enough to merge and I can deal with my concerns in main.  Well done!

Yinameah:
@RedPhoenixQ : I have a question about theses lines : https://github.com/david-vanderson/dvui/blob/312366ef2d61e41bed8dc7d79e54e857ef2b819f/build.zig?plain=1#L287-L293

Isn't this a typo ? If you pass `-Dbackend=testing`, this results in the message saying "only sdl2 or testing backend to be used".

Also, If I change the L289 for 
```zig
        if (mod == null) break :blk .{ b.modules.get("dvui_testing"), false };
```
(which I suppose was the original intent), and run `zig build docs -Dbackend=testing` not only the images aren't generated, but the tests don't run at all.

Which, to be honest is not a bad thing to have (i.e gen docs without running tests), because as cool as this image generation thingy is, it makes the doc build quite slow and it's valuable to have a fast path to update the doc when working on it, without the need to re-gen the images.

I'm happy to provide a fix but I don't want to mess with it without understanding the original intent.

Yinameah:
Ok, all good. 

Fully agree that "no gen image" option is an important thing to have. Personally I would find "-Dno-doc-images" more explicit and understandable.

And if there is no image generation, no backend needs to be passed as doc generation is fully static, as it should be. Which leaves us with only sdl2 available for "doc build with backend"

I'll leave it up to you to do what you see best fit then. Thanks for the clarification.

RedPhoenixQ:
I believe the docs still have to link a backend to resolve the `@import("backend")` so something has to be linked, probably should be the testing backend for compatibility. 

There isn't really anything preventing other backends from being used. I'll look into a nicer solution with an explicit `-Dno-doc-images` option

david-vanderson:
This is good, but let's try having a `zig build docs` and a separate `zig build docs-images` command.  Can that work?

Yinameah:
@RedPhoenixQ I think your right, the @import("backend") results in a semantic error otherwise.

Having separate `zig build docs` and `zig build docs-image` can definitely work and on the bright side, it makes docs building without images more a "standard" zig build docs. `zig build docs docs-images` would perform the "full build".

@RedPhoenixQ : do you want to take care of this ?

RedPhoenixQ:
Made a PR with `-Dno-doc-images` solution. Can change to the `zig build docs docs-images` version. Didn't realise you could specify multiple steps at once

PR #267, opened 2025-04-20
Allow choosing the target for the web backend

RedPhoenixQ:
@MichaelBelousov raised the need to be able to specify the target for the web backend. With this change the web backend gets built with the user specified target, which was already used for tests. This does mean users need to specify wasm32 as their target to use the web backend in the browser, which they had to do anyway.

The web examples still force the use of the wasm32 target, but that does not affect users of dvui.


david-vanderson:
Excellent!

PR #270, opened 2025-04-21
Fix failing tests

Yinameah:
While playing with the new stuff for doc and tests, I noticed a couple of typo in `Rect.zig` that were failing tests.

I also made the `saveDocImage` output something a bit more explicit in case of duplicate image name.

@david-vanderson let me know if you would prefer this kind of small stuff to be pushed directly to main instead of going through a PR. 
I prefer the PR way, but don't want to create annoying noise either.

### DIFF
--- a/docs/readme-docs.md
+++ b/docs/readme-docs.md
@@ -13,6 +13,19 @@ I also add success temporarily adding in the `<head>` section of `docs/index.htm
 ` <script type="text/javascript" src="http://livejs.com/live.js"></script>`  
 For a full auto reload experience. Cool enough to be mentioned.
 
+# About Images
+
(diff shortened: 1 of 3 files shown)


Yinameah:
Also add a small explanation about images in docs.

RedPhoenixQ:
Embarrassing that I added failing tests. Thank you for fixing!

The readme-doc looks great too!

david-vanderson:
Looks good - make that change and commit!

Yinameah:
Ah ! True. Fully agree with @RedPhoenixQ comments about triggering a stack trace and checking of filename, not on `src.fn_name`. Good thing I did a PR ;-)

Fixed.

david-vanderson:
Thanks!

PR #272, opened 2025-04-22
More doc improvements

RedPhoenixQ:
- adds references with \`\` to create links when using `zig build docs`
- adds links to the each backends docs
- move `Backend` doc comments  from `VTableTypes` to the methods to make them visible
- make comments in `Window` into doc comments (they are not perfect but making the comment information accessible in editor and generated docs is nicer)


david-vanderson:
Amazing love it!

PR #276, opened 2025-04-29
options: max_size_content is type MaxSize

david-vanderson:
When setting .max_size_content for Options, we don't want the default behavior of Size where unspecified means 0.

So this attempts to solve it by using a new type, MaxSize.  It doesn't have defaults so the user must specify both the width/height or use a decl literal.

The idea is to prevent confusion when trying to do .max_size_content = .{} - does that mean zeros or large numbers?

### DIFF
--- a/src/Options.zig
+++ b/src/Options.zig
@@ -45,6 +45,25 @@ pub const FontStyle = enum {
     title_4,
 };
 
+pub const MaxSize = struct {
+    w: f32,
+    h: f32,
+
+    pub const zero: MaxSize = .{ .w = 0, .h = 0 };
+
+    pub fn width(w: f32) MaxSize {
+        return .{ .w = w, .h = 1_000_000 };
+    }
+
+    pub fn height(h: f32) MaxSize {
+        return .{ .w = 1_000_000, .h = h };
+    }
+
+    pub fn size(s: Size) MaxSize {
+        return .{ .w = s.w, .h = s.h };
+    }
+};
+
 // used to adjust widget id when @src() is not enough (like in a loop)
 id_extra: ?usize = null,
 
@@ -110,7 +129,7 @@ min_size_content: ?Size = null,
 /// Use when a child textLayout or scrollArea is making the parent too big.
 ///
 /// padding/border/margin will be added to this.
-max_size_content: ?Size = null,
+max_size_content: ?MaxSize = null,
 
 // whether to fill the background
 background: ?bool = null,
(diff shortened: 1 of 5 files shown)


RedPhoenixQ:
Zig 0.14.0 mentions [faulty default field values](https://ziglang.org/documentation/0.14.0/#Faulty-Default-Field-Values) and recommend associated constants. Something line `pub const zero: Size = .{ .w = 0, .h = 0 }` to where `.zero` would replace `.{}`. That way it would avoid duplicate types.

It might be too large of a change for all `.{}` size to become `.zero`, but it would be more explicit and idiomatic.

david-vanderson:
Yes that's the kind of thing I'm going for here.  `MaxSize` has decl literals for specifying stuff like that.

A common case is to only care about either width or height but not both.  I think for normal `Size` it makes sense to say `.min_size_content = .{.w = 100}` where the `.h` is unspecified, meaning zero.  `MaxSize` implies the opposite - an unspecified field should be large (`1_000_000`).

If the code changes look reasonable, then we'll move forward and try this out.  I just wanted a double-check because this kind of "use different types to differentiate intent" is a new strategy for me.

src/Options.zig
```diff
@@ -45,6 +45,25 @@ pub const FontStyle = enum {
...
+
+    pub fn width(w: f32) MaxSize {
+        return .{ .w = w, .h = 1_000_000 };
+    }
+
+    pub fn height(h: f32) MaxSize {
+        return .{ .w = 1_000_000, .h = h };
+    }
+
+    pub fn size(s: Size) MaxSize {
```
RedPhoenixQ, line 62: Should this be named `all` like `Rect.all` for consistency?

RedPhoenixQ:
> Yes that's the kind of thing I'm going for here. `MaxSize` has decl literals for specifying stuff like that.
> 
> A common case is to only care about either width or height but not both. I think for normal `Size` it makes sense to say `.min_size_content = .{.w = 100}` where the `.h` is unspecified, meaning zero. `MaxSize` implies the opposite - an unspecified field should be large (`1_000_000`).
> 
> If the code changes look reasonable, then we'll move forward and try this out. I just wanted a double-check because this kind of "use different types to differentiate intent" is a new strategy for me.

That's fair, just wanted to raise the option. Types declaring the api is very reasonable and might make it easier to document too.

src/Options.zig
david-vanderson: I don't understand - `all` takes a scalar and applies it to all fields.  This is converting from Size to MaxSize.  What am I not getting?

RedPhoenixQ: I'm very sorry, I misread. Ignore this

phatchman:
This does create an asymmetry between setting the min and max widths e.g:
```
        .fixed_width => default_header_options.override(.{
            .min_size_content = .{ .w = fixed_width_w },
            .max_size_content = .width(fixed_width_w),
        }),
```
You could add width() etc to Size and remove the default initialization to 0, but I suspect that would touch a lot of code?

david-vanderson:
It's a good point.  I wasn't sure how much that asymmetry would stick out.  Clearly it does.  How about adding width() and height() to Size but keeping the default initialization of 0?  We can try that for a while and see how it feels.

PR #285, opened 2025-05-02
Suggestion filtering

### LINKED ISSUE #280: add filtering example to suggestions widget demo
add filtering example to suggestions widget demo

RedPhoenixQ:
Closes #280

This adds filtering to the "Suggest" example as well as changing how the `comboBox` helper works to allow the user to filter the entries based on the text content.


david-vanderson:
Great work!

Issue #282, opened 2025-05-02
Type safety for pixel units


david-vanderson:
#281 was the last in a long series of easy-to-make mistakes.

Normally we have 3 different kinds of Rect with different units:
* physical pixels (mouse events, and drawing use this)
* natural pixels (window positions/sizes use this)
* logical pixels (widget positioning uses this)

This issue is to investigate if we could have better type safety here, by adding new types:
* RectPhysical
* RectNatural
* RectLogical (maybe this stays as just Rect)

david-vanderson:
This also affects Point and Size.

RedPhoenixQ:
I suggest adding simple wrapper types like:

```zig
const PhysicalRect = struct { rect: dvui.Rect };
```

That way all the helper function on Rect and other types don't need to be duplicated or worked around. Simple to create from the base type as well. This type would only be used for argument or return types for clarity right?

david-vanderson:
That's a good idea.  We could also add some conversion functions to the wrapper types.  I'm pretty sure these would primarily be used as argument and return types.  And it means we can introduce them slowly.

My memory is that the largest source of these issues comes from the floating widgets (floatingWindow, floatingMenu, etc.) - because those need natural pixels, but are often based off of some widget's physical pixel rect.  So that's where I'm going to try it first.

RedPhoenixQ:
Incorrect reference on [d332a48](https://github.com/david-vanderson/dvui/commit/d332a48b7752c30acd51ace85d5ee5b12fac189c),  should be #284

david-vanderson:
Excellent thank you!

PR #287, opened 2025-05-03
Add Physical and Natural variants of Rect, Size and Point

RedPhoenixQ:
Closes: #282

This touches many lines, most being types that were previously commented as being physical pixels. There is also many lines changed linked to Paths as their `Point` type should be `Point.Physical`.

Many of the helper functions on `Rect` and `Point` have been proxied for their physical versions. This limits the amount of conversion needed in user code or the rest of the library. This isn't as elegant as I would have liked, but thanks to the proxying there is minimal logic duplication.

Clip rects is always an `Rect.Physical`. Checking `clippedBy` on rects other than `Rect.Physical` isn't  relevant, so the method has been moved to `Rect.Physical`.

The `fill` and `stroke` has been moved to `Rect.Physical` as they require physical pixels.

More conversions between `Rect.Physical` and `Size.Physical` could be added. These conversions are uncommon so has not been included, they can be achieved using:

```zig
const pys_size = dvui.Size.Physical.fromSize(pys_rect.toRect().toSize())
```


david-vanderson:
Looking good!  I will go through this carefully but it'll take me a few days.

One thing I noticed was `inline` on some functions.  Do you know if that makes a difference?  Usually we wouldn't want to force it like that and leave the decision up to the compiler.

RedPhoenixQ:
I started with inline only on the to/from functions as this transform should be essentially free. Later when adding the proxies my thought was that with the inlined transform, it should be equivalent to calling the original function. Adding inline _should_ make the call site just a function call to the original function, nothing extra added.

I haven't verified this. I'll look at the assembly with/without inline and see what changes, if anything.

RedPhoenixQ:
I started without creating a function returning a type believing they wouldn't work as nice with the LSP, but I was wrong and this is much nicer.

david-vanderson:
Very cool!  I did not realize `test` didn't have to be at the top level.  Does this mean we are duplicating those tests for each units variant?

RedPhoenixQ:
Yes, but the image generating test only run once. The other tests are very small and I believe it's worth it for the better docs

RedPhoenixQ:
There is an known issue with the `.png` test for rect and easings that haven't been ported to use the new doc image functions. Will fix and adress that is a followup PR.

PR #289, opened 2025-05-04
Datagrid first cut

phatchman:
## This comment is out of date.

I don't intend for this to be merged yet, rather it is for review of the approach and code so far.

Something has happened recently with the smoothness of the virtual scrolling which I need to investigate further.

I've come to the view that the GridWidget probably needs to do all of the layout itself, rather than relying on boxes. I've not really looked into how difficult that would be yet, but it would bring several advantages:
1) Smoother scrolling where header and body have to scroll together.
2) More control over virtual scrolling
3) Maybe easier to keep header and body sizes in sync as can layout the header and body at the same time.
4) Likely easier to implement draggable columns.

There are probably a few instances where the widgets are a bit too intimate with each other, such as setting the "invisible_height" in the body widget. It depends how much you want to de-couple the widgets.

A copy of TODO.MD is below:
### Issues
- [ ] Virtual scrolling is sometimes smooth and sometimes flickers by 1 row. 
    - Likely some floating point rounding issue.
    - Is less apparent when vsync = false
    - Possibly resolved by not forcing the grid to scroll on full rows.
- [ ] Virtual scrolling with large row heights doesn't work well because of the way it snaps to the next visible row.
- [ ] scrolling header and body simultaneously is not a smooth as I'd like.
- [ ] Columns flash when sorting. Only the separator is shown and appears that label is not rendered or rendered with no label text.
- [ ] Is there a better name than window size for the scroller? It's not really a window size. It's a number of extra rows to render above and below the visible rows.
- [ ] Checkbox doesn't expand to the full height of the header.
- [ ] Remove the need to pass the same (or sometimes different) styling to the header vs the body.
- [ ] Make column headers respect column width "ownership" so that .expand can be used on the body columns. 
- [ ] Grid header widget assumes vertical scroll bar width is 10 and that it will always be displayed. 
- [ ] Example needs to be added to Example.zig, rather than a stand-alone.

### Future
* Some better visual indication that columns are sortable.
* Make the GridWidget do the layout calculations, rather than relying on hbox / vbox layouts. 
    - Then each column have a style of fixed, expanding, size_to_content etc.
    - The grid Widget can then take into account the available space and size each column accordingy.
    - Potentially it would allows layout out data by row rather than just by col.
    - Would likely fix the issue of the header and body not being 100% in sync while scrolling.
* Resize columns via dragging
* Filtering headers

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -1,0 +1,23 @@
+### Todos
+- [ ] Get rid of the padding for the sortable header. It isn't const the size of the symbol is dependent on the font size?
+- [ ] Handle padding when header column width is smaller than the heading button width. Currently doesn't show the separator.
+- [ ] Draws over vertical scroll bar when horizontal scrolling. Need to change the clip or reserve space for the scrollbar.
+- [ ] Need some defalt padding, so the first column isn;t hard against the edge of the grid.
+- [ ] Fix checkbox padding or size? It's not shown the speparator.
+
+### Issues
+- [ ] Is there a better name than window size for the scroller? It's not really a window size. It's a number of extra rows to render above and below the visible rows.
+        - Windows size > 1 doesn't seem to make any difference really.
+- [ ] Grid header widget assumes vertical scroll bar width is 10 and that it will always be displayed. 
+- [ ] Example needs to be added to Example.zig, rather than a stand-alone.
+- [ ] Issue with grid headers moving while virtual scrolling. Header is set to scroll_info.vieport.y, but this is not always at the top of the viewport (due to floating point precision?)
+- [ ] Gravity is applied incorrectly when virtual scrolling? i..e things will only center when scrolling stops.
+        - It doesn't happen with non-virtual scrolling, so something is not expanding correctly?
+        - Actually, I don't think it is gravity, it is some cells being drawn too wide while scrolling and then corrected when the scrolling stops.
+        - Also note in the demo there is an "Expand" column for Description. That's not really compatible with virtual scrolling.
+        - Need to investigate further.
+
+### Future
+* Some better visual indication that columns are sortable.
+* Resize columns via dragging
+* Filtering headers
--- a/build.zig
+++ b/build.zig
@@ -149,6 +149,7 @@ pub fn build(b: *std.Build) !void {
 
         linkBackend(dvui_sdl, sdl_mod);
         addExample("sdl2-standalone", b.path("examples/sdl-standalone.zig"), dvui_sdl, true, dvui_opts);
+        addExample("sdl2-datagrid", b.path("examples/data_grid_example.zig"), dvui_sdl, true, dvui_opts);
         addExample("sdl2-ontop", b.path("examples/sdl-ontop.zig"), dvui_sdl, true, dvui_opts);
         addExample("sdl2-app", b.path("examples/app.zig"), dvui_sdl, test_dvui_and_app, dvui_opts);
     }
--- a/examples/data_grid_example.zig
+++ b/examples/data_grid_example.zig
@@ -1,0 +1,588 @@
+const std = @import("std");
+const builtin = @import("builtin");
+const dvui = @import("dvui");
+const Backend = dvui.backend;
+comptime {
+    std.debug.assert(@hasDecl(Backend, "SDLBackend"));
+}
+
+const window_icon_png = @embedFile("zig-favicon.png");
+
+var gpa_instance = std.heap.GeneralPurposeAllocator(.{}){};
+const gpa = gpa_instance.allocator();
+
+const vsync = true;
+var scale_val: f32 = 1.0;
+
+var g_backend: ?Backend = null;
+var g_win: ?*dvui.Window = null;
+var filter_grid = false;
+var frame_count: usize = 0;
+/// This example shows how to use the dvui for a normal application:
+/// - dvui renders the whole application
+/// - render frames only when needed
+///
+pub fn main() !void {
+    if (@import("builtin").os.tag == .windows) { // optional
+        // on windows graphical apps have no console, so output goes to nowhere - attach it manually. related: https://github.com/ziglang/zig/issues/4196
+        _ = winapi.AttachConsole(0xFFFFFFFF);
+    }
+    std.log.info("SDL version: {}", .{Backend.getSDLVersion()});
+
+    defer if (gpa_instance.deinit() != .ok) @panic("Memory leak on exit!");
+
+    // init SDL backend (creates and owns OS window)
+    var backend = try Backend.initWindow(.{
+        .allocator = gpa,
+        .size = .{ .w = 800.0, .h = 600.0 },
+        .min_size = .{ .w = 250.0, .h = 350.0 },
+        .vsync = vsync,
+        .title = "DVUI SDL Standalone Example",
+        .icon = window_icon_png, // can also call setIconFromFileContent()
+    });
+    g_backend = backend;
+    defer backend.deinit();
+
+    _ = Backend.c.SDL_EnableScreenSaver();
+
+    // init dvui Window (maps onto a single OS window)
+    var win = try dvui.Window.init(@src(), gpa, backend.backend(), .{});
+    defer win.deinit();
+    filtered_cars = try filterCars(gpa, cars[0..], filterLongModels);
+    defer gpa.free(filtered_cars);
+
+    main_loop: while (true) {
+
+        // beginWait coordinates with waitTime below to run frames only when needed
+        const nstime = win.beginWait(backend.hasEvent());
+
+        // marks the beginning of a frame for dvui, can call dvui functions after this
+        try win.begin(nstime);
+
+        // send all SDL events to dvui for processing
+        const quit = try backend.addAllEvents(&win);
+        if (quit) break :main_loop;
+
+        // if dvui widgets might not cover the whole window, then need to clear
+        // the previous frame's render
+        _ = Backend.c.SDL_SetRenderDrawColor(backend.renderer, 0, 0, 0, 255);
+        _ = Backend.c.SDL_RenderClear(backend.renderer);
+
+        // The demos we pass in here show up under "Platform-specific demos"
+        try gui_frame();
+
+        // marks end of dvui frame, don't call dvui functions after this
+        // - sends all dvui stuff to backend for rendering, must be called before renderPresent()
+        const end_micros = try win.end(.{});
+
+        // cursor management
+        backend.setCursor(win.cursorRequested());
+        backend.textInputRect(win.textInputRequested());
+
+        // render frame to OS
+        backend.renderPresent();
+
+        // waitTime and beginWait combine to achieve variable framerates
+        const wait_event_micros = win.waitTime(end_micros, null);
+        backend.waitEventTimeout(wait_event_micros);
+    }
+}
+
+const num_cars = 500_000;
+pub const testing = false;
+pub var scroll_info: dvui.ScrollInfo = .{ .horizontal = .auto, .vertical = .given };
+var virtual_scrolling = true;
+var horizontal_scrolling = false;
+var sortable = true;
+var header_height: f32 = 0;
+var row_height: f32 = 0;
+var selectable = false;
+const ColumnSizing = enum {
+    equal_width,
+    proportional,
+    col_info,
+    col_width,
+    expand,
+};
+var column_sizing: ColumnSizing = .equal_width;
+const prev_height: f32 = 0;
+var resize_rows: bool = false;
+
+const fixed_width_w: f32 = 50;
+pub fn headerCheckboxOptions() dvui.Options {
+    if (header_height == 0) {
+        return .{
+            .min_size_content = .{ .w = 40 },
+            .max_size_content = .width(40),
+        };
+    } else {
+        return .{
+            .min_size_content = .{ .w = 40, .h = header_height },
+            .max_size_content = .width(40),
+        };
+    }
+}
+
+//pub fn rowCheckboxOptions() dvui.Options {
+//    return headerCheckboxOptions();
+//}
+
+const col_widths_default: [7]f32 = .{ 50, 20, 30, 40, 50, 60, 70 };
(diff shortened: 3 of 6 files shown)


phatchman:
I'll likely be tied up for at least the next few days. My apologies in advance if I don't respond for a while.

phatchman:
Thanks for taking the time to review. Yeah, the test data and descriptions are all from chat GPT. I found them amusing enough to leave them as-is, even the ones that don't really make sense. You can still tell what it was getting at.

OK. Seems like we are in good agreement on the direction. My plan was to get of the full row scrolling also. It looked good at first, but as you say, once the row height becomes larger, it doesn't really work.

I did have an experiment where I externalized the layout and basically just provided a default implementation that did equal spacing. Then the min/max columns sizes are always passed in when the column is created. If someone wants "expand" then the "column sizer" can work that out what that actually means. I think this totally gets rid of the need to remember column sizes between header and body as they will always be pre-calculated. 

And then if I do add a "size to content" layout, it can work more nicely with virtual scrolling as the layout would provide the max size for the entire dataset field, rather than just sizing it to what is currently rendered. In the current implementation, the columns dynamically resize as you scroll, which is not ideal.

I just need to think what that would mean when I allow the user to resize the columns via drag-handles. If the column size is calculated outside the grid, the dragging would need to be done externally as well.

phatchman:
OK. Below looks fairly promising to me. It feels more "dvui" than keeping column widths as internal state.

For the below, I bypassed all of the code in GridWidget that remembers column sizes and instead sizes are supplied each frame by the user.  I've included 2 sample layout methods that can assist with this:
1) Equally spaced.
2) "Proportional", which keeps the same width proportion when the window is resized. (i.e. a ratio expand, but only for width).
* If you supply a content_w as an init_option it will size to that value, otherwise it will size to parent.
* reserved_w is space that is under control of the user, rather than the layout. e.g. reserving space for the fixed-width checkbox columns. 

The changes I plan to make are: 
1) Pass an actual column width (using Size), to the beginCol() functions, rather than using min/max content size options. The layouts will return Size, rather than dvui.Option.
2) When I make changes for GridWidget to perform the layout, widths will probably only be supplied for headers, rather than header and body. (height would still be allowed on the body).
3) Have some way for things like checkbox columns to tell the user what their "preferred width" (maybe this is just documentation), so users know how much space to reserve for any fixed-width columns that are not under control of the layout.
4) It's not clear to me how to make this approach work with "draggable" column sizes, because the grid widget itself will likely need to handle all of the events related to dragging. But I'll cross that bridge later. 
5) Make columns have a default width if not supplied.

```
const ColumnLayoutProportional = struct {
    const InitOpts = struct {
        // Space to reserve for fixed width columns
        reserved_w: ?f32 = null,
        // Size to specified width. Otherwise sizes to width of parent grid.
        content_w: ?f32 = null,
    };
    initial_w: []const f32,
    initial_total_w: f32,
    available_w: f32,
    num_cols: usize,
    current_header_col: usize,
    current_body_col: usize,

    pub fn init(grid: *dvui.GridWidget, init_opts: ColumnLayoutProportional.InitOpts, comptime initial_widths: []const f32) ColumnLayoutProportional {
        const scroll_bar_w: f32 = 10; // Don't necessarily know if SB is showing?
        const reserved_w: f32 = init_opts.reserved_w orelse 0;
        const available: f32 = init_opts.content_w orelse grid.data().contentRect().w;
        const initial_total_w = sumSlice(f32, initial_widths);
        return .{
            .initial_w = initial_widths,
            .initial_total_w = initial_total_w,
            .available_w = available - reserved_w - scroll_bar_w,
            .num_cols = initial_widths.len,
            .current_header_col = 0,
            .current_body_col = 0,
        };
    }

    pub fn nextHeaderColOption(self: *ColumnLayoutProportional, opts: dvui.Options) dvui.Options {
        // TODO: Sanitize the options or warn that opts are being ignored.
        const result = self.colOption(opts, self.current_header_col);
        self.current_header_col += 1;
        return result;
    }

    pub fn nextBodyColOption(self: *ColumnLayoutProportional, opts: dvui.Options) dvui.Options {
        const result = self.colOption(opts, self.current_body_col);
        self.current_body_col += 1;
        return result;
    }

    fn colOption(self: *ColumnLayoutProportional, opts: dvui.Options, col_num: usize) dvui.Options {
        if (col_num < self.initial_w.len) {
            const ratio: f32 = self.initial_w[col_num] / self.initial_total_w;
            const w: f32 = ratio * self.available_w;
            return opts.override(.{
                .min_size_content = .{ .w = w },
                .max_size_content = .width(w),
            });
        } else {
            // TODO: Probable return an error?
            return .{};
        }
    }
};

const ColumnLayoutEqualWidth = struct {
    const InitOpts = struct {
        // Space to reserve for fixed width columns
        reserved_w: ?f32,
        // Size to specified width. Otherwise sizes to width of parent grid.
        content_w: ?f32,
    };
    col_width: f32,

    pub fn init(grid: *dvui.GridWidget, init_opts: ColumnLayoutEqualWidth.InitOpts, num_cols: usize) ColumnLayoutEqualWidth {
        const num_cols_f: f32 = @floatFromInt(num_cols);
        const scroll_bar_w: f32 = 10; // Don't necessarily know if SB is showing?
        const reserved_w: f32 = init_opts.reserved_w orelse 0;
        const content_w: f32 = init_opts.content_w orelse grid.data().contentRect().w;
        return .{
            .col_width = (content_w - reserved_w - scroll_bar_w) / num_cols_f,
        };
    }

    pub fn nextHeaderColOption(self: *ColumnLayoutEqualWidth, opts: dvui.Options) dvui.Options {
        // TODO: Sanitize the options or warn that opts are being ignored.
        return opts.override(.{
            .min_size_content = .{ .w = self.col_width },
            .max_size_content = .width(self.col_width),
        });
    }

    pub fn nextBodyColOption(self: *ColumnLayoutEqualWidth, opts: dvui.Options) dvui.Options {
        return self.nextHeaderColOption(opts);
    }
};
```

david-vanderson:
I think you are on the right track, it all sounds reasonable to me.

For column widths, could try to mirror what horizontal box does - if the column is not .expand, then give it it's min_size.w (which would come directly from the user, or calculated from what was there last frame).  If there is any extra space, then allocate among the .expand columns.  But for now I would just give everything min_size.w (of the column header).

Can do the same for height of rows, and we'll need a (optional?) row header.  Then the individual grid cells don't get any say on their size.

For dragging (in the future), grid could maintain an internal store of "overridden" column widths, or we have a struct that holds all the column/row sizes that you pass a pointer to the grid?

Designing this stuff is tough, and I appreciate the work!

phatchman:
Happy to spend the time to iterate on this. Would rather get it "right". 

I'm wondering if you had a trick I could use to prevent drawing of any row(s) under the header?

Currently I'm supplying the header last, so that you can't see any rows that have scrolled underneath it, but that's not a great interface as the user would typically expect to supply the header before the body. 

Currently the grid layout works like this:
1) Laid out column by column, meaning the user must supply the header and body for col1, then header and body for col2 etc.
2) Each column is a vbox with the x and y position of the header and body rows within the vbox are calculated by the grid. 
3) The header is positioned based on scroll position so that it is always the first visible row.
4) The y position of the rows are offset by the header height, so that the first row is not hidden behind the header.


```
                    try grid.colBegin(@src(), layout.nextHeaderColOption(.{}).max_size_content.?.w); // TODO: Take either an option or a width
                    defer grid.colEnd();
                    try dvui.gridColumnFromSlice(@src(), grid, Car, cars[0..], "make", "{s}", .{});
                    if (try dvui.gridHeadingSortable(@src(), grid, "Make", &sort_dir, .{})) {
                        sort("Make", sort_dir);
                    }
```

david-vanderson:
Sounds great!

phatchman:
OK. Not expecting you to look at this, but if you are interested in the shape of the new code, see [this diff](https://github.com/phatchman/dvui-dev/compare/main..grid_widget_layout_refactoring)
Be warned, the code is still in a pretty rough state. 

But the below shows how much simpler this new implementation is:
src/widgets/GridWidget.zig | 537 ++++++++++++++-------------------------------
 1 file changed, 164 insertions(+), 373 deletions(-)

david-vanderson:
That is looking better!  One suggestion: have `colBegin()` return a pointer to the BoxWidget, then the user code can directly interact with it and call deinit().  Same idea with `bodyCellBegin()`.  Once you `.init()` a widget, you can get its height (to calculate next_row_y.

phatchman:
Yep, that makes sense. I've implemented that and it works fine.  One thing I've still not properly internalised yet is that you can get the dimensions before you have laid everything else out. But yes, it makes a good improvement.

Another improvement is with col_info.  The Layout structs can just produce a slice of f32 now, rather than behaving like iterators. In fact they can be converted to just a single function that takes in a slice of f32 requested widths, and then populates that slice with actual column widths. (I might change it to populate a different, output slice. still working it out)

I'm using positive numbers to request a fixed width and negative numbers to request a calculated width when the slice is passed to the "Layout" function. It's mildly hacky, but it's simple. That also means that that the "proportional" and "equal width" layouts can be made the same. Equal width can be implemented via proportional when all of the proportions set to the same number.

[diff](https://github.com/phatchman/dvui-dev/compare/main..grid_widget_layout_refactoring)

phatchman:
Hi @david-vanderson,

I would like your thoughts on the resizing of row heights.

What I used to do was store the maximum height of all the cells and use that to set the minimum height of all cells for the next frame. This has the advantage that all cells are the same height, but the disadvantage is that the height could never shrink, because the minimum cell height is always set to the max height of the previous frame.

What I'm currently doing is remembering the max height of the last frame and using that to increment next_row_y (which determine the top-left of the next row's hbox). I set the height to of the cell's hbox rect to 0 so that it will only take up as much height as it needs. The still keeps the rows aligned as they are laid out using next_row_y, and lets  rows shrink in size, but has the disadvantage that each cell is now a different height.

If someone were to put a border on the cells, in the old method, all of the borders would line up. In the new method they wouldn't. It is fairly common to have things like banded rows etc, so the current solution isn't really practical.

Unless you have a nicer solution, I'm thinking of just providing a resized() function that would lay out the next frame with all cells having a min_height of 0, to reset the max row heights. It would be up to the user to work out that either the data or column sized had changed and then they would call resized() on the grid. Maybe I could handle the case where the actual grid widget size has changed, but anything to do with data or column sizes is up to the user.
The other nice thing about making this the user's responsibility is that for anyone just using labels in cells, they wouldn't even need to call resized() as the labels are going to truncate, rather than wrap.

[GridWidget.zig](https://github.com/phatchman/dvui-dev/blob/18f87cb7fd520d0025325c48dc4f14c2d75b5c10/src/widgets/GridWidget.zig)

Currently the grid makes all rows the same height. Which is just a consequence of laying out the data by column. I'd need to remember the height for each row to allow variable row heights. But I've a plan to change it so that if the user provides a row height for a cell, we'll use that height instead of the calculated "max row height", which will let the user have full control over the heights for each row.

So, I think all that is left to do for "phase 1" is this row height issue, fixing some bugs with virtual scrolling and then cleaning up the example. Then I'll submit for review.

phatchman:
> > Unless you have a nicer solution, I'm thinking of just providing a resized() function
> 
> I don't have a better solution, so this sounds good to me!

OK that actually turned out well. It's now implemented as a "resize_rows" init option which, by a nice accident of the implementation, works symmetrically. i.e. you need to pass it to make the rows resize in either direction. (except for the first frame where resizing is assumed). 

It occurs to me that almost every issue I've raised while building this grid, has been solved by pushing state management back the user. Hopefully I'll learn that lesson one day! :smile:

david-vanderson:
> It occurs to me that almost every issue I've raised while building this grid, has been solved by pushing state management back the user. Hopefully I'll learn that lesson one day! 😄

I've also had more difficulty learning that than expected!  Where to draw that line has constantly surprised me, making it very difficult to predict.

phatchman:
I've pushed a new cut of the code just now.  It should now just be reworking the demo and fixing some minor display issues and whatever cleanups are marked in the code.

Can you please take a look at the changes I made with the options processing, especially how I've enabled banding in dvui.columnFromSlice and let me know if you are OK with it? I could not think of a good way to do this without adding some optional callbacks to allow per-row options to be supplied for each cell.
I'm also thinking of cases like using red text to represent negative numbers, which would also need this per-cell formatting.

Otherwise, each time the user needs some slightly different formatting they would end up creating their own "column" function. Which is fine if you don't like the callback approach.

## Todos
- [ ] Get rid of the padding for the sortable header. It isn't const as the size of the symbol is dependent on the font size?
- [ ] Handle padding when header column width is smaller than the heading button width. Currently doesn't show the separator.
- [ ] Draws over vertical scroll bar when horizontal scrolling. Need to change the clip or reserve space for the scrollbar.
- [ ] Need some default padding, so the first column isn;t hard against the edge of the grid.
- [ ] Fix checkbox padding or size? It's not shown the separator.

### Issues
- [ ] Is there a better name than window size for the scroller? It's not really a window size. It's a number of extra rows to render above and below the visible rows.
        - Windows size > 1 doesn't seem to make any difference really.
        - Will remove option to configure and leave it set to 1.
- [ ] Grid header widget assumes vertical scroll bar width is 10 and that it will always be displayed. 
- [ ] Example needs to be added to Example.zig, rather than a stand-alone.
- [ ] Issue with grid headers moving while virtual scrolling. Header is set to scroll_info.vieport.y, but this is not always at the top of the viewport (due to floating point precision?)
- [ ] Gravity is applied incorrectly when virtual scrolling? i..e things will only center when scrolling stops.
        - It doesn't happen with non-virtual scrolling, so something is not expanding correctly?
        - Actually, I don't think it is gravity, it is some cells being drawn too wide while scrolling and then corrected when the scrolling stops.
        - Also note in the demo there is an "Expand" column for Description. That's not really compatible with virtual scrolling.
        - Need to investigate further.

phatchman:
[Oops!](https://imgur.com/a/O3Zwz9G)

phatchman:
We have [examples](https://imgur.com/a/1EOhQV8). If there is anything else you think you'd like as an example, let me know. 

Quite a few things left to polish, but I think we're close!

PR #291, opened 2025-05-05
Fix check step not performing comptime checks

RedPhoenixQ:
I appears that depending on Step.Compile only parses the zig files and doesn't do semantic analysis and comptime evaluation. This became apparent in #287 where the ci did not catch issues in backends not having updated their types.

To force this, the check step now has to install the compiled tests and thus also run llvm. This is much slower but the binaries can at least be reused for running the tests later.

The check step also needs check all backends and their dvui import, otherwise changes to the Backend api wouldn't be caught.


david-vanderson:
Okay I agree with this, but we need to swap them - "zig build check" is now a common enough thing that people do on editor save that we can't break that speed.

david-vanderson:
Thank you!  Merge when ready.

RedPhoenixQ:
All examples already compiled together on the default install step `zig build`, so I removed the step `compile` in favour of that

PR #292, opened 2025-05-05
Doc image generation changes

RedPhoenixQ:
There were some painpoints about the existing solution. Running `zig build test -Dbackend=sdl2 -Ddoc-image-dir=zig-out/docs -Dtest-filter=".png"` will still run the steps for other example files, which makes it take longer. It is also very inconvenient to type out all of these arguments when they will be the same every time you want to generate doc images.

Now running `zig build docs -Dgenerate-images` will run only the dvui modules test with the `.png` filter. It will also use the `docs/image_gen_test_runner.zig` which is needed for:
- Providing the output directory
  - It can get a cached directory by the build system, which allows for testing that no two tests generate an image with the same name.
- Ignoring error logs
  - Running the `Example-debbuging.png` test without this causes the test step to fail, failing the full build step.
- Making the doc testing comptime known, allowing for smaller/quicker test when not generating images.

It also didn't make much sense to me to have a `saveImage` function in `dvui.testing` and a runtime known directory. The `capturePng` function allows for the same functionality for users of dvui and a comptime known `saveDocImage` function is more explicit and more efficient.

Additionally, the `saveDocImage` function now always runs one frame. Before it didn't actually test the frame and could cause tests to have different outputs when not generating images because the frame count would change. This became apparent in the Rect tests where the new types didn't fail the tests when not generating images.

There also seems to be an issue with `Picture` and passing a rect to it. On my computer the `Example-demo.png` looks very wrong. I haven't been able to figure out why.
Examples-demo

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -380,7 +380,7 @@ pub fn build(b: *std.Build) !void {
                 const image_tests = b.addTest(.{
                     .name = "generate-images",
                     .root_module = dvui,
-                    .filters = &.{".png"},
+                    .filters = &.{"DOCIMG"},
                     .test_runner = .{ .mode = .simple, .path = b.path("docs/image_gen_test_runner.zig") },
                 });
                 docs_step.dependOn(&b.addRunArtifact(image_tests).step);
--- a/build.zig
+++ b/build.zig
@@ -24,8 +24,8 @@ pub fn build(b: *std.Build) !void {
     const use_lld = b.option(bool, "use-lld", "The value of the use_lld executable option");
     const test_filters = b.option([]const []const u8, "test-filter", "Skip tests that do not match any filter") orelse &[0][]const u8{};
 
+    // NOTE: Currently unused, but is not removed because it's imported everywhere for future use cases
     const build_options = b.addOptions();
-    build_options.addOption(?[]const u8, "doc_image_dir", b.option([]const u8, "doc-image-dir", "Directory for documentation images"));
 
     const dvui_opts = DvuiModuleOptions{
         .b = b,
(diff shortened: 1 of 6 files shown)


david-vanderson:
I will try to replicate that problem with the sdl2 backend.

I agree that this needs some work, but I want a solution that:
* dependencies can use this to generate their own doc images, so no "internal use only" and should be useable with any backend (web can wait)
  * this should also be useable from outside test blocks (like a downstream program that accepts a flag where it loads, waits for the initial screen to settle, saves a doc image, and exits)
* distinguishing tests between behavior and doc images shouldn't be matching "png" on the test name - it should either be a much more obvious string match (like "DOCIMG") or some code inside the test that differentiates it
* as much as possible doesn't run tests multiple times

I was trying to go in that direction, but I agree it wasn't looking great.  I like putting the doc images as an option on "zig build docs".

Does this make sense?

david-vanderson:
> I can understand the dependencies generating doc images to some degree. Is it common for a dvui user to generate docs?

Not yet, but it will come.  I'm not super sure of where the line dividing dvui-only stuff and stuff from an application using dvui should be - we will have to keep feeling that out.

> This is why I have been referring to `capturePng`.

This is somewhat reasonable, but we should strive for providing the same api that we use internally.  I think saveImage with a runtime-known path passed to it for where to save the image is the api people (and us) want.  That's why I put it as an init_opt into testing - other people can use it easily, but internally we can provide a build-time known string.

> Makes it easier to figure out which test generated which image when debugging.

I'm not convinced it's easier than if the filename is passed to saveImage - either way the person is searching the codebase for a string right?  Also it makes it so you can produce multiple images per test without embedding format stuff into the test name.  I'm also thinking of having some of the tests that reference decl names produce doc images as well?

RedPhoenixQ:
Ok I understand the no exclusive api part of dvui. I'll revert back an go the other way, making all test use `saveImage`. I'm going to update `-Dgenerate-images` to set the doc_image_dir option directly to get the best of both worlds.

> Also it makes it so you can produce multiple images per test without embedding format stuff into the test name. I'm also thinking of having some of the tests that reference decl names produce doc images as well?

How about not including formatting in the `saveImage` function at all? You would pass a string directly or if you can, comptime concat the filename directly.

david-vanderson:
Thanks for going through this with me!

> How about not including formatting in the `saveImage` function at all? You would pass a string directly or if you can, comptime concat the filename directly.

Yes that sounds good.

RedPhoenixQ:
With these changes, the only reason for `docs/image_gen_test_runner.zig` to exist is for the debugging Example to not fail the tests because of it's error log. Maybe this could be done nicer, or integrate the very short file in the build script like for the deprecated modules?

RedPhoenixQ:
> I'm less sure how useful images for the parts of the demo window are

That is true. I think it just got added because the demo window was on the first page, then we I generalized the image generation, the examples was the most obvious place to start. They had a clear correct image.

I don't think the test runner hurts anyone so it might as well stay for now. Doesn't affect normal test or generating images without error logs.

> also representative images for widgets

We have the same idea here, like `ContextWidget` wasn't immediately obvious to me. An image would have quickly clarified that.

david-vanderson:
Great!  Merge when ready.

RedPhoenixQ:
Just wanted to note that the issue with the demo window is fixed. The texture was offset twice when there was a render target present, fixed in da153d52ef0a9689881b77e3fdfc86f97d701d51

david-vanderson:
Good catch, sorry about that.  I think you are right that the render target offset should only happen just before a call to `drawClippedTriangles`.  I'm not too worried about the performance impact, drawing to a target is rare, and we can eliminate the for loop if the offset is zero.  I just spied another double-offset bug though - pathStrokeRaw with a length of 1.

I'll get on it.

Issue #293, opened 2025-05-05
bug: focusing text entry that isn't fully inside browser viewport can push browser view around


MichaelBelousov:
When I select a text entry that is partially offscreen when running DVUI in a browser canvas,
my browser view gets pushed to view the full fake input element, which is partially offscreen.
This leads to an ugly view of white off-page content in my web application:

![Image](https://github.com/user-attachments/assets/c127eaa3-9897-4c89-a1f4-bd695fc964a8)

Should dvui be doing that? It occurs because the hidden input element in the web backend is being focused off screen, and it is attached to the document top level.

My initial thoughts are that the hidden input element should be attached to a container div, not the document directly. Then that container div can hide overflow.

PR #294, opened 2025-05-05
web: fix overflow on inputs at the edge of the window

RedPhoenixQ:
Closes: #293

This limits the input width to the window. If the input is focused but completely of screen the width or height will be limited to 0 making the element hidden and not causing overflow.

Adding `overflow: hidden;` doesn't quite work unless the window has a fixed size with `100vw` or similar, but that's fragile and inconsistent between browsers so it's not desirable.


david-vanderson:
looks good - let's push and check with @MichaelBelousov that this solves it for him

PR #296, opened 2025-05-08
removed ColorOrName from the Codebase

nat3Github:


nat3Github:
- removes ColorOrName
- ive ran the examples and there are no changes as far as i can tell.
- added fn Option.fromAny()  which parses a Tuple that is structured like Option, only that you can use ColorFromTheme as theme for a color option
- override the fromAny(DefautlDef) with (var defaults) and then with opts
- added a fromComptimeHex for use cases where we have the color at comptime (which i think is mostly the case)

david-vanderson:
When you load the examples and change the theme, everything looks okay?  You should see a bunch of the old theme colors since they got baked into widget defaults.

nat3Github:
will check

nat3Github:
yes everything looks good aka the theme changes. i dont know how it looked before but i think its supposed to work like that.

david-vanderson:
Oh I see you are running some nasty comptime introspection in Options.fromAny().  That is clever but way more complicated that the original scheme.  Now we would have tuples that kind of look like Options, but they aren't really.

Unfortunately this comes down to a judgement call.  I think we should try a simpler solution of adding some decl literals to the ColorOrName to make it easier to type.

nat3Github:
you are right, the tuples are a placeholder for something like ColorOrName you can't get rid of it completely the complexity is at a different place.

nat3Github:
i think that's a fair call, will leave a pull request for the other idea .fromTheme and .fromComptimeHex

david-vanderson:
Thank you!

PR #297, opened 2025-05-08
add utility fn to Color and ColorOrName

nat3Github:
in Color:
- fromTheme(...)
- fromComptimeHex(...)

in Options.ColorOrName
- fromColor(...)
- fromTheme(...)
- fromHex(...)
- fromComptimeHex(...)

### DIFF
--- a/src/Color.zig
+++ b/src/Color.zig
@@ -12,6 +12,7 @@ a: u8 = 0xff,
 
 pub const white = Color{ .r = 0xff, .g = 0xff, .b = 0xff };
 pub const black = Color{ .r = 0x00, .g = 0x00, .b = 0x00 };
+pub const magenta = Color{ .r = 0xFD, .g = 0x3D, .b = 0xB5 };
 
 /// Convert normal color to premultiplied alpha.
 pub fn alphaMultiply(self: @This()) @This() {
@@ -111,10 +112,6 @@ pub fn transparent(x: Color, y: f32) Color {
 pub fn format(self: *const Color, comptime _: []const u8, _: std.fmt.FormatOptions, writer: anytype) !void {
     try std.fmt.format(writer, "Color{{ {x} {x} {x} {x} }}", .{ self.r, self.g, self.b, self.a });
 }
-
-pub const white = Color{ .r = 0xff, .g = 0xff, .b = 0xff };
-pub const black = Color{ .r = 0x00, .g = 0x00, .b = 0x00 };
-pub const magenta = Color{ .r = 0xFD, .g = 0x3D, .b = 0xB5 };
 
 /// Average two colors component-wise
 pub fn average(self: Color, other: Color) Color {
@@ -274,62 +271,6 @@ test tryFromHex {
     try std.testing.expectEqual(Color{ .r = 0xa1, .g = 0xa2, .b = 0xa3, .a = 0xa4 }, Color.tryFromHex("#A1A2A3A4"));
     try std.testing.expectEqual(FromHexError.InvalidCharacter, Color.tryFromHex("XXX"));
     try std.testing.expectEqual(FromHexError.InvalidHexStringLength, Color.tryFromHex("#12"));
-}
-
-/// Comptime Converts slice of HexString to Color
-pub fn fromComptimeHex(comptime rgb_hex: []const u8, alpha: u8) @This() {
-    const m = struct {
-        inline fn hexToRgb(hex: []const u8) ![4]u8 {
-            var xrgba: [4]u8 = .{ 0, 0, 0, 255 };
-            if (hex.len == 6) {
-                for (xrgba[0..3], 0..) |_, i| {
-                    const start = i * 2;
-                    const slice = hex[start .. start + 2];
-                    const value = try std.fmt.parseInt(u8, slice, 16);
-                    xrgba[i] = value;
-                }
-                return xrgba;
-            }
-            if (hex.len == 7 and hex[0] == '#') {
-                const hex1 = hex[1..];
-                for (xrgba[0..3], 0..) |_, i| {
-                    const start = i * 2;
-                    const slice = hex1[start .. start + 2];
-                    const value = try std.fmt.parseInt(u8, slice, 16);
-                    xrgba[i] = value;
-                }
-                return xrgba;
-            }
-            return error.FailedToParseHexColor;
-        }
-    };
-    const rgba = comptime m.hexToRgb(rgb_hex) catch {
-        @compileError("failed to convert rgba from hex code");
-    };
-    return @This(){
-        .r = rgba[0],
-        .g = rgba[1],
-        .b = rgba[2],
-        .a = alpha,
-    };
-}
-
-/// Get a Color from the active Theme
-///
-/// Only valid between `Window.begin`and `Window.end`.
-pub fn fromTheme(theme_color: ColorsFromTheme) @This() {
-    return switch (theme_color) {
-        .accent => dvui.themeGet().color_accent,
-        .text => dvui.themeGet().color_text,
-        .text_press => dvui.themeGet().color_text_press,
-        .fill => dvui.themeGet().color_fill,
-        .fill_hover => dvui.themeGet().color_fill_hover,
-        .fill_press => dvui.themeGet().color_fill_press,
-        .border => dvui.themeGet().color_border,
-        .err => dvui.themeGet().color_err,
-        .fill_window => dvui.themeGet().color_fill_window,
-        .fill_control => dvui.themeGet().color_fill_control,
-    };
 }
 
 /// Comptime Converts slice of HexString to Color
--- a/src/Color.zig
+++ b/src/Color.zig
@@ -273,44 +273,6 @@ test tryFromHex {
     try std.testing.expectEqual(FromHexError.InvalidHexStringLength, Color.tryFromHex("#12"));
 }
 
-/// Comptime Converts slice of HexString to Color
-pub fn fromComptimeHex(comptime rgb_hex: []const u8, alpha: u8) @This() {
-    const m = struct {
-        inline fn hexToRgb(hex: []const u8) ![4]u8 {
-            var xrgba: [4]u8 = .{ 0, 0, 0, 255 };
-            if (hex.len == 6) {
-                for (xrgba[0..3], 0..) |_, i| {
-                    const start = i * 2;
-                    const slice = hex[start .. start + 2];
-                    const value = try std.fmt.parseInt(u8, slice, 16);
-                    xrgba[i] = value;
-                }
-                return xrgba;
-            }
-            if (hex.len == 7 and hex[0] == '#') {
-                const hex1 = hex[1..];
-                for (xrgba[0..3], 0..) |_, i| {
-                    const start = i * 2;
-                    const slice = hex1[start .. start + 2];
-                    const value = try std.fmt.parseInt(u8, slice, 16);
-                    xrgba[i] = value;
-                }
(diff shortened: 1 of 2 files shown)


RedPhoenixQ:
The normal `fromHex` is already comptime compatible, so I don't think there is a need for `fromComptimeHex`.

I do like `fromTheme` on `Color`. It wasn't super obvious before when you had to get the theme and then then color.

Having `fromHex` available on `Options.ColorOrName` is nice, but is that really a big difference from:

```zig
Options{ .color_text = .{ .color = .fromHex("#113355".*) } };
```

nat3Github:
> The normal `fromHex` is already comptime compatible, so I don't think there is a need for `fromComptimeHex`.
> 
> I do like `fromTheme` on `Color`. It wasn't super obvious before when you had to get the theme and then then color.
> 
> Having `fromHex` available on `Options.ColorOrName` is nice, but is that really a big difference from:
> 
> ```zig
> Options{ 
>.color_text = .{ .color = .fromHex("#113355".*) }
>};
> ```
this btw wont work because fromHex returns an error so more like 
```zig
 Options{ 
.color_text = .{ color = try .fromHex("#113355".*) },
 };
```
instead of
```zig
 Options{ 
.color_text = .fromComptimeHex("#113355"),
 };
```

this of course is purely a matter of taste. but when reading a large fn with lot of color changes i find the second one less straining for eyes and brain and also for my hands because "{" is a lot harder to type than "." writing a gui means defining a lot of things and the second one is shorter and arguably clearer. that saves time and brain capacity when quickly changing things around.
we could also add some colors for debugging purposes as pub decls for quick changes like
```zig
 Options{ 
.color_fill = .Blue,
 };
```
or even add the ColorsFromThem enum tags as pub decls:

```zig
 Options{ 
.color_fill = .accent,
 };
```
i am against shortforming anything just for "developer ergonomics" but in this case its easier to read and easier to write and i think thats worth something.

nat3Github:
>```zig
>Options{ 
>.color_fill = .Blue,
>.color_text = .Transparent,
>};
>Options{ 
>.color_fill = .accent,
>};
```
i can also add this changes to the PR but first waiting for more feedback on this subject!

**Comment by RedPhoenixQ** [comment]
When I tested if `fromHex` was comptime I took some time update it. I got this version:
```zig
/// Converts hex color string to `Color`
///
/// Supports the following formats:
/// - "#RGB"
/// - "#RGBA"
/// - "#RRGGBB"
/// - "#RRGGBBAA"
/// - "RGB"
/// - "RGBA"
/// - "RRGGBB"
/// - "RRGGBBAA"
pub fn fromHex(hex_color: []const u8) !Color {
    const hex = if (hex_color[0] == '#') hex_color[1..] else hex_color;

    const num = try std.fmt.parseUnsigned(u32, hex, 16);

    const is_nibble_size, const has_alpha = switch (hex.len) {
        3 => .{ true, false },
        4 => .{ true, true },
        6 => .{ false, false },
        8 => .{ false, true },
        else => return error.InvalidHexStringLength,
    };
    const mask: u32 = if (is_nibble_size) 0xF else 0xFF;
    const step: u5 = if (is_nibble_size) 4 else 8;
    const offset: u5 = @intFromBool(has_alpha);
    return .{
        .r = @intCast((num >> step * (2 + offset)) & mask),
        .g = @intCast((num >> step * (1 + offset)) & mask),
        .b = @intCast((num >> step * (0 + offset)) & mask),
        .a = if (has_alpha) @intCast(num & mask) else 0xFF,
    };
}
```

It has all the formats and is comptime compatible, but you're right in that it would have to be `try comptime .fromHex("#FFF")` to guarantee that its comptime. Just `try .fromHex("#FFF")` _should_ also be comptime known an not executed but I'm not 100% on that.

I'm gonna make a separate PR with this `fromHex` function, so your solution will be upgraded with other formats later.

> we could also add some colors for debugging purposes as pub decls for quick changes like

I would be in favour of more color decls, but that doesn't have to be a part of this PR.

> or even add the ColorsFromThem enum tags as pub decls:
> 
> ```zig
>  Options{ 
> .color_fill = .accent,
>  };
> ```
> 
> i am against shortforming anything just for "developer ergonomics" but in this case its easier to read and easier to write and i think that is the reason why the ability to do this found its way into the zig compiler.

These would have to be functions as they look up the current theme, so `.color_fill = .accent(),`. That's not bad but i prefer `.fromThem(.accent)` for clarity and simplicity.

david-vanderson:
I'm liking this - let's go with Color.fromTheme(), ColorOrName.fromTheme(), and ColorOrName.fromColor() for now.

Let's see what we can do with fromHex() - maybe we can throw a compileError if `@inComptime()`?

If it's at runtime, I'm tempted to have it not throw and error and instead log a message and give some kind of error color (like magenta).  I'd like to remove as many `try` as we reasonably can.  What do you guys think?

RedPhoenixQ:
> ColorOrName.fromRGB() and .fromHSLuv()?

I think the common name for the "#FFFFFF" format is hex, so that should probably stay `fromHex`. I don't mind also having `fromHSLuv`.

> If it's at runtime, I'm tempted to have it not throw and error and instead log a message and give some kind of error color (like magenta). I'd like to remove as many `try` as we reasonably can. What do you guys think?

I think it's important for have a failing version so you can tell if a input field is invalid in an application. We could have a wrapping function that just return a default. `fromHex` could be the infallible one and all the logic extracted into a `tryFromHex` function. That is an api inspired by Rust where a fallible version is available under a `try_` prefix.

> Let's see what we can do with fromHex() - maybe we can throw a compileError if `@inComptime()`?

That should work and would avoid a separate version. Even better if we make `fromHex` infallible.

david-vanderson:
> I think the common name for the "#FFFFFF" format is hex, so that should probably stay `fromHex`. I don't mind also having `fromHSLuv`.

I agree - I meant that if we have ColorOrName.fromColor() that is a bit ambiguous, so we should probably have all 3 (fromHex, fromHSLuv, and fromRGB).  fromRGB would accept a dvui.Color.

But given your response, maybe that's not a good idea, and we should have fromHex, fromHSLuv, and fromColor?

RedPhoenixQ:
> But given your response, maybe that's not a good idea, and we should have fromHex, fromHSLuv, and fromColor?

I prefer that as the type system will make is clear very quickly. Imo I would expect fromRGB to take 3 u8s as the color paramaters or similar

nat3Github:
@david-vanderson 
> If it's at runtime, I'm tempted to have it not throw and error and instead log a message and give some kind of error color (like magenta). I'd like to remove as many `try` as we reasonably can. What do you guys think?

i like this a lot! one thing less that can error.

@RedPhoenixQ 
> These would have to be functions as they look up the current theme, so .color_fill = .accent(),. That's not bad but i prefer .fromThem(.accent) for clarity and simplicity.

i dont think so, you could normally also put .{ .name = .theme_col_name} so we duplicate the fields as pub decls.


how do you like this:
```zig
// Either specify the color directly or name a color from the Theme
pub const ColorOrName = union(enum) {
    color: Color,
    name: ColorsFromTheme,

    pub const theme = struct {
        pub const accent = .{ .name = .accent };
        pub const err = .{ .name = .err };
        pub const text = .{ .name = .text };
        pub const text_press = .{ .name = .text_press };
        pub const fill = .{ .name = .fill };
        pub const fill_window = .{ .name = .fill_window };
        pub const fill_control = .{ .name = .fill_control };
        pub const fill_hover = .{ .name = .fill_hover };
        pub const fill_press = .{ .name = .fill_press };
        pub const border = .{ .name = .border };
    };
    pub const debug_color = struct {
        pub const Red = fromHex("#FF0000");
        pub const Blue = fromHex("#0000FF");
        pub const Green = fromHex("#00FF00");
    };

    pub fn fromColor(col: Color) @This() {
        return .{ .color = col };
    }
    pub fn fromHex(hex: Color.HexString) @This() {
        if (@inComptime()) return comptime .fromColor(Color.fromHex(hex) catch unreachable);
        return .fromColor(Color.fromHex(hex) catch unreachable);
    }
};
```

syntax is like (its just namespaces)
```zig
 Options{ 
.color_fill = .debug_color.Blue,
 };
 Options{ 
.color_fill = .theme.accent,
 };
```

david-vanderson:
That is very interesting.  I'm torn - one the one hand it is pretty cool, but on the other it's different from the rest.  Let me sleep on it.

I think we don't need debug_color - using fromColor() and then adding a bunch of well-known names into Color should work for that.  I need to figure out what set of names we should use.

RedPhoenixQ:
> i dont think so, you could normally also put .{ .name = .theme_col_name} so we duplicate the fields as pub decls.

Oh I misunderstood how you meant it to work.

I'm not against moving the decls to the main scope like:

```zig
pub const ColorOrName = union(enum) {
    color: Color,
    name: ColorsFromTheme,

    pub const accent = .{ .name = .accent };
    pub const err = .{ .name = .err };
    pub const text = .{ .name = .text };
    pub const text_press = .{ .name = .text_press };
    pub const fill = .{ .name = .fill };
    pub const fill_window = .{ .name = .fill_window };
    pub const fill_control = .{ .name = .fill_control };
    pub const fill_hover = .{ .name = .fill_hover };
    pub const fill_press = .{ .name = .fill_press };
    pub const border = .{ .name = .border };
  
    ...
};
```

```zig
Options{ 
    .color_fill = .fromColor(.white),
};
Options{ 
    .color_fill = .fromColor(.red),
};
Options{ 
    .color_fill = .fromColor(.magenta),
};
Options{ 
    .color_fill = .fromHex("#AABBCC".*),
};
Options{ 
    .color_fill = .accent,
};
Options{ 
    .color_fill = .fill_window,
};
``` 

Might as well go all the way right!


Perhaps `ColorOrName.fromHex` should always be comptime? That way we could guarantee that the easiest solution for a hex color has no runtime impact. If you need a dynamic hex color, you can use `.fromColor(.fromHex(...))` directly as it's probably coming from some dynamic input anyway. That way all examples above would be comptime values!

```zig
    pub fn fromHex(comptime hex: Color.HexString) @This() {
        return .fromColor(comptime Color.fromHex(hex) catch unreachable);
    }
```

RedPhoenixQ:
#298 was just merged which has some `fromHex` improvements. There is a minor api change where the argument for `fromHex` changed from `Color.HexString` to `[] const u8`. If you want to rebase/merge these changes against main that's fine, otherwise we'll make that minor change when merging this PR.

nat3Github:
>Might as well go all the way right!

yes!
> Perhaps `ColorOrName.fromHex` should always be comptime? That way we could guarantee that the easiest solution for a hex color has no runtime impact. If you need a dynamic hex color, you can use `.fromColor(.fromHex(...))` directly as it's probably coming from some dynamic input anyway. That way all examples above would be comptime values!

i think thats a good idea. runtime conversion is a very niche use case imo and not preferable anyway. 

>https://github.com/david-vanderson/dvui/pull/298 was just merged which has some fromHex improvements. There is a >minor api change where the argument for fromHex changed from Color.HexString to [] const u8. If you want to >rebase/merge these changes against main that's fine, otherwise we'll make that minor change when merging this PR.

as you like! you guys can tell me what changes you like in the PR, ColorOrName.fromColor(), comptime ColorFromHex(), pub decls of ColorsFromTheme, etc.

david-vanderson:
Okay!  I'm sold on both the decls and the comptime ColorOrName.fromHex.  @nat3Github please rebase with those changes.  Thanks!

nat3Github:
i think thats it! :) will go sleep.

david-vanderson:
Thank you!

PR #298, opened 2025-05-08
`Color.fromHex` improvements

RedPhoenixQ:
Relevant for #297

Makes the `Color.fromHex` accept more formats of hex colors and makes the function infallible. A `tryFromHex` function has been added for when you want to know if the string is invalid.

Magenta was used a the fallback color to indicate the error and `@inComptime()` used to give a compile error for invalid strings when possible, as suggested [here](https://github.com/david-vanderson/dvui/pull/297#issuecomment-2864320259).

### DIFF
--- a/src/Color.zig
+++ b/src/Color.zig
@@ -172,10 +172,18 @@ pub fn alphaAverage(self: Color, other: Color) Color {
 
 pub const HexString = [7]u8;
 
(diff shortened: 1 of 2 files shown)


david-vanderson:
Looks great!  Merge when ready.

PR #300, opened 2025-05-09
Named Colors

RedPhoenixQ:
This adds more named colors, finishing up #297.

As suggested by @kcbanner on the zig discord the [web basic colors](https://en.wikipedia.org/wiki/Web_colors#Basic_colors) was was used. In addition I added some aliases from the extended colors that are more commonly used, like "magenta" instead of "fuchsia".

These are mostly just intended for debugging or styling during development. Most application will want to use the theme colors and if needed create their own theme.

### DIFF
--- a/src/Color.zig
+++ b/src/Color.zig
@@ -10,9 +10,31 @@ g: u8 = 0xff,
 b: u8 = 0xff,
 a: u8 = 0xff,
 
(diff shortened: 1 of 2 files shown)


david-vanderson:
Perfect - merge when ready.  Thanks!

RedPhoenixQ:
CI failure seems like a build cache issue. The windows build works and the tests pass locally so I will merge

PR #301, opened 2025-05-11
add svg2tvg as replacement for tvg-sdk

nat3Github:
we had a lot of problems with tvg-sdk. the converter written in c# dotnet was delivering subpar results when we tried to convert popular svg icons to tvg. rendering wasnt robust either.

svg2tvg implements a new renderer and parser in zig on top of the old tvg-sdk. its available here: https://github.com/nat3Github/zig-lib-svg2tvg under MIT
- this pull request integrates the new renderer into dvui.
- the old tvg-sdk src files are deleted from the src tree

please let me know if everything is working as expected!


marler8997:
I started an svg2tvg converter, here's a link in case it's helpful: https://github.com/marler8997/svg2tvg

RedPhoenixQ:
The build failure is due to an assumption in z2d that usize can be coerced from u63 (positive values of i64), which is not true for wasm32. I've submitted https://github.com/vancluever/z2d/pull/117 to be up streamed before this can be merged.

nat3Github:
> https://github.com/marler8997/svg2tvg
@marler8997 i wrote this in a misleading way. svg2tvg renders tvg and converts svg to tvg hence the name. so you're also welcome to checkout the repo if you like. im not sure if my converter handles all path parsing commands but i converted feathericons, lucide and heroicons and fixed many path parsing bugs along the way. but nice to have something to compare to maybe ill give it a spin on the icon libraries...

david-vanderson:
Do you have a clever way to compare before/after?  Or should I run two separate apps (before, after) and compare?

nat3Github:
unfortunately not, as a heads up, in the icon browser i noticed one icon (at the bottom of the list) that did not look right. if that blocks this merge its a ok for me. i will not be able to work on this till i finished my bachelor thesis (10 days) but i can look into it after that.

david-vanderson:
Which icon?  Okay let's pause on this, because it's definitely going to generate questions for you, and I don't want to interrupt you doing that.

I looked side-by-side at the entypo icons in the example, and they all look the same to me, so I guess the next step for me is to start looking at the code.

nat3Github:
that sounds good, will come back to the icon thing i mentioned, could well be that it's a conversion error since it was converted with the old converter or a render issue since i'm using the same logic for filled paths than the old renderer...

either way if you are ok with that i'll add an option for icon stroke thickness and color since it's available in the renderer anyway

david-vanderson:
> either way if you are ok with that i'll add an option for icon stroke thickness and color since it's available in the renderer anyway

Color makes sense.  I'm less sure that overriding the stroke thickness will be useful, but no harm in having the option.  Thanks!

nat3Github:
stroke thickness only makes sense for icons that are made out of strokes like https://lucide.dev/icons/?focus= 
on entypo it probably has no effect.
but since stroked icons are the defacto standard fethericons, lucide, and heroicons all would benefit from it!
(there is a a slider on the page to see the effect, we can have this in dvui for more customization ability)

nat3Github:
api changes: everything icon related uses an additional IconRenderOptions.
IconRenderOptions allow customizing the tvg icons:
- overwrite stroke color
- overwrite fill color
- overwrite stroke width
- disabling fill

RedPhoenixQ:
Zig 0.14.1 just released and the CI appears to be trying to load it and it's failing. I thought it would go with the version specified in build.zig.zon. Will fix today

RedPhoenixQ:
CI fixed on main in 38be270b446d85576cb7f2fb734ecb93538c49ef. Rebase/merge main here again to run the CI

RedPhoenixQ:
CI seems to pass, just some caching issue. There where a couple of those issues fixed in zig 0.14.1, so maybe CI would be more reliable on that version

david-vanderson:
This is looking really good.  I have 2 requests:
- can we swap the order of IconRenderOptions and Options so that Options is always the last parameter?
- In IconRenderOptions, let's remove the "override", and can we use a transparent color to indicate disable_fill?  So that would look something like:

```zig
pub const IconRenderOptions = struct {
    /// if null uses original fill colors, use .transparent to disable fill
    fill_color: ?Color = .white,

    /// if null uses original stroke width
    stroke_width: ?f32 = null,

    /// if null uses original stroke colors
    stroke_color: ?Color = .white,
};
```

nat3Github:
i think thats it

david-vanderson:
I think `buttonIcon` got missed for the options switch?  Otherwise looks good.  Fix that and merge!  (Let me know if you can't)

nat3Github:
why does the CI still fail is this an issue on my part?

RedPhoenixQ:
No, it's a FileNotFound caching error. Windows passes so there shouldn't be any issues for this PR

david-vanderson:
Got ahead and merge anyway.  When I write this in my monthly update post, do you want to be credited as "nat3Github" or a different name?

nat3Github:
> Got ahead and merge anyway. When I write this in my monthly update post, do you want to be credited as "nat3Github" or a different name?

done! very happy to see this finally land :). sure you can use my github username or just nat3 as you like.

PR #302, opened 2025-05-15
Hash based snapshots

RedPhoenixQ:
This changes the previously added `dvui.testing.snapshot` feature from using images generated with a "real" backend render to a hash of all reasonable widget state from `WidgetData` and `Options`. The function is limited to the testing backend as other backend get state like window scaling from the OS, which is not fully reproducible.

Each hash is still stored in individual files because it's a simple solution. To see a visual difference between snapshots, `-Dsnapshot-images` was added where you pass a suffix for the images. This lets you create an "file.zig-test.X-after.png", go back to a previous commit and create "file.zig-test.X-before.png". An environment variable could be added to specify this, removing the need to recompile to generate the images, but it made the code messier so it was not added. Having to go back to a previous commit will like also force a recompile.

The current implementation uses a global variable `dvui.testing.widget_hasher` that contains the hasher and `WidgetData` checks if it's not null and populates is. The assumption is that if no other code interacts with the variable, it is comptime known to be null and the hashing code will be omitted. This has not been verified.


david-vanderson:
Excellent - thank you!

PR #306, opened 2025-05-18
Color picker

### LINKED ISSUE #78: Color Picker Widget
Add support for a specialized color picker widget similar to the following.

Screenshot 2024-07-29 at 11 58 15 AM

.
The current implementation of a basic color picker in `src/Example.zig` looks like this

```zig
// Let's wrap the sliderEntry widget so we have 3 that represent a Color
pub fn rgbSliders(src: std.builtin.SourceLocation, color: *dvui.Color, opts: Options) !void {
    var hbox = try dvui.box(src, .horizontal, opts);
    defer hbox.deinit();

    var red: f32 = @floatFromInt(color.r);
    var green: f32 = @floatFromInt(color.g);
    var blue: f32 = @floatFromInt(color.b);

    _ = try dvui.sliderEntry(@src(), "R: {d:0.0}", .{ .value = &red, .min = 0, .max = 255, .interval = 1 }, .{ .gravity_y = 0.5 });
    _ = try dvui.sliderEntry(@src(), "G: {d:0.0}", .{ .value = &green, .min = 0, .max = 255, .interval = 1 }, .{ .gravity_y = 0.5 });
    _ = try dvui.sliderEntry(@src(), "B: {d:0.0}", .{ .value = &blue, .min = 0, .max = 255, .interval = 1 }, .{ .gravity_y = 0.5 });

    color.r = @intFromFloat(red);
    color.g = @intFromFloat(green);
    color.b = @intFromFloat(blue);
}
```

RedPhoenixQ:
Closes #78

ColorPickerWidget

This adds a HSV based color picker, mostly inspired by [photoshops color picker (see fig 16b)](https://en.wikipedia.org/wiki/HSL_and_HSV#Use_in_end-user_software). It uses two small 4 and 7 pixel textured and the builtin texture interpolation to make the colored rectangles. They are generated on demand with most of the pixel data being comptime known. 

Additionally this also adds `textEntryColor` which makes use of `Color.tryFromHex` to create a hex color input field. This was not needed for the color picker, but allows for the creation of a fairly complete color picker.

Examples-styling

Perhaps there should be a minimal version of the color picker in the image above available as `dvui.colorPicker`? I'd be happy to add it before merging this

### DIFF
--- a/src/Color.zig
+++ b/src/Color.zig
@@ -387,6 +387,7 @@ pub const FromHexError = std.fmt.ParseIntError || error{
 /// - `RRGGBB`
 /// - `RRGGBBAA`
 pub fn tryFromHex(hex_color: []const u8) FromHexError!Color {
+    if (hex_color.len == 0) return error.InvalidHexStringLength;
     const hex = if (hex_color[0] == '#') hex_color[1..] else hex_color;
 
     const is_nibble_size, const has_alpha = switch (hex.len) {
--- a/src/Color.zig
+++ b/src/Color.zig
@@ -64,6 +64,124 @@ pub fn brightness(self: @This()) f32 {
     return 0.2126 * r + 0.7152 * g + 0.0722 * b;
 }
 
+pub fn toRGBA(self: @This()) [4]u8 {
+    return .{ self.r, self.g, self.b, self.a };
+}
+
+/// https://en.wikipedia.org/wiki/HSL_and_HSV#HSV_to_RGB
+pub const HSV = struct {
+    /// Hue 0-360 (degrees)
+    h: f32 = 0.0,
+
+    /// Saturation 0-1 (%)
+    s: f32 = 1.0,
+
+    /// Value 0-1 (%)
+    v: f32 = 1.0,
+
+    /// Alpha 0-1 (%)
+    a: f32 = 1.0,
+
+    pub fn fromColor(color: Color) HSV {
+        const r: f32 = @as(f32, @floatFromInt(color.r)) / 255.0;
+        const g: f32 = @as(f32, @floatFromInt(color.g)) / 255.0;
+        const b: f32 = @as(f32, @floatFromInt(color.b)) / 255.0;
+        const a: f32 = @as(f32, @floatFromInt(color.a)) / 255.0;
+
+        const max = @max(r, g, b);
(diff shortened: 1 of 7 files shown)


david-vanderson:
This looks fantastic!

One tiny note: you can use `dvui.boxEqual` instead of the lower-level api, but the fact that you didn't use it suggests it's hard to find.  Do you think we should change `dvui.box` to accept a BoxOptions?

> Perhaps there should be a minimal version of the color picker in the image above available as `dvui.colorPicker`? I'd be happy to add it before merging this

That would be lovely.  Otherwise it looks great and you can merge when you are ready.

Thank you!

RedPhoenixQ:
> One tiny note: you can use `dvui.boxEqual` instead of the lower-level api, but the fact that you didn't use it suggests it's hard to find. Do you think we should change `dvui.box` to accept a BoxOptions?

I think that would be a very large change for little win. I just didn't know about it, so maybe adding references to each other in the docs would be enough for discoverability.

david-vanderson:
> I think that would be a very large change for little win. I just didn't know about it, so maybe adding references to each other in the docs would be enough for discoverability.

Good idea I will do it.

david-vanderson:
I added some docs to box, boxEqual, and flexbox.  Does that help?  I'm still unsure of how people see the docs - are you seeing it via `zig build docs`?  Or some language server thing?

RedPhoenixQ:
> I added some docs to box, boxEqual, and flexbox. Does that help? I'm still unsure of how people see the docs - are you seeing it via `zig build docs`? Or some language server thing?

That's perfect. Makes it obvious that these are related and serve roughly the same purpose. I use zls in vscode via the zig extension mostly, rarely read the docs while coding

RedPhoenixQ:
`dvui.colorPicker` can be both horizontal and vertical. It justs moved the sliders bellow in vertical mode. It, and the new doc image, looks like:

ColorPickerWidget

RedPhoenixQ:
Right, the problem was slightly different than I stated. The text does update when the value changes but here is the issue:

1. Clear out the text input (make it empty)
2. You can no longer access the currently selected hex color
3. Press the reset button
4. The current color is now present again

RedPhoenixQ:
That sounds fair. I'll remove the button and replace it with that for now.

PR #309, opened 2025-05-20
Path refactor

RedPhoenixQ:
This is only motivated by me noticing inconsistencies in functions that should be very similar. The triangle builder gives some more guarantees for indices needing to be in sets of 3 as well as you needing to specify the exact amount of vertexes and indices needed. 

0511df413820dd2951444da26636e8ba5933e0ad touches a lot of api surface area and I would be okay with reverting it.

This also includes changing `Color.transparent` to be a decl for a transparent color, and the previous function taking the name `opacity`, inspired by CSS.

### DIFF
--- a/src/Color.zig
+++ b/src/Color.zig
@@ -36,6 +36,8 @@ pub const magenta = fuchsia;
 pub const darl_cyan = teal;
 pub const dark_magenta = purple;
 
+pub const transparent = Color{ .r = 0, .g = 0, .b = 0, .a = 0 };
+
 /// Convert normal color to premultiplied alpha.
 pub fn alphaMultiply(self: @This()) @This() {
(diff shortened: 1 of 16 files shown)


david-vanderson:
This looks good.  Let me do some testing tomorrow.

I'll also audit where we convert colors to premultiplied alpha and add that to the doc comments.  Although now that I write that, maybe we should add that to the type system - like Color.PMAColor - how does that sound?

RedPhoenixQ:
> like Color.PMAColor - how does that sound?

I like that. I can get on that tomorrow if you'd like

I've also been close to making a Color.RGBAFloat type where all the fields are 0-1, but the usage sites aren't consistent enough to be able to easily use a common abstraction

PR #310, opened 2025-05-20
Dx11 handle interpolation

RedPhoenixQ:
The new color picker relied on linear interpolation of its textures which was not respected in dx11. The interpolation is not set on or stored on the textures so we maintain a hashmap of texture pointers to interpolation values ourselves.


david-vanderson:
Great!  Merge when ready.

PR #311, opened 2025-05-20
Add types for premultiply alpha colors and pixels

RedPhoenixQ:
This clarifies which APIs expect colors premultiplied by alpha (PMA).


david-vanderson:
Looking pretty good, just a few nits:
- `castPremultiplied` -> let's go with just `cast`
- `Color.multiply` -> kill it, as only Color.PMA should have a `multiply` function
- for casting from/to Color/Color.PMA, let's have only a `cast` function on each one that converts the other

Those sound okay, or do you disagree?

RedPhoenixQ:
> * for casting from/to Color/Color.PMA, let's have only a `cast` function on each one that converts the other

I chose to add `PMA.cast` and `PMA.castToColor` to almost mirror `fromColor` and `toColor` which is already used by `HSV`. If `Color.fromHSLuv` was moved to `HSLuv.toColor` that would unify most of the color transformations.

david-vanderson:
> I chose to add `PMA.cast` and `PMA.castToColor` to almost mirror `fromColor` and `toColor` which is already used by `HSV`. If `Color.fromHSLuv` was moved to `HSLuv.toColor` that would unify most of the color transformations.

Good point, but I'm not crazy about the exiting fromColor/toColor, so maybe this is a chance to fix it.  I'm liking the ".from" convention, where we expect the user to know the type they want, and look for a function in that type to convert from what they have.

But this is a weak preference for now.  So I'll leave the choice to you.  Merge when ready.

Thank you!

RedPhoenixQ:
My main motivation for the `fromColor`/`toColor` is that `Color` is always the intermediate, for now at least, and it keeps the `to` and `from` transformations close together in the code. I think that will make it easier to keep them up to date.

I'm not firm on to the `X.toColor` and would be up for changing it to `Color.fromX` if you want. I'll keep it like it is for now though.

PR #313, opened 2025-05-27
Sdl3 live resize

david-vanderson:
Use SDL3's main callback functionality to enable rendering during OS window resizing events on Windows and MacOS.

The really tricky part is making it work while keeping the variable framerate.  SDL does not expect you to call SDL_WaitEventTimeout when using main callbacks.  The trick is to track if we are getting called nested during a waitEventTimeout, and make sure we don't call waitEventTimeout or any SDL event functions in a nested context.

Edit: During resizing, we run as fast as possible.  By "making it work" I mean having variable framerate during normal operation (not during resizing).

To test the tricky parts, run `zig build sdl3-app`, open demo section Animations > Clock and resize the OS window.

@RedPhoenixQ Can you please verify that this works for you on Windows?

### DIFF
--- a/src/App.zig
+++ b/src/App.zig
@@ -36,7 +36,7 @@ fn nop_main() !void {}
 /// ```
 /// pub const main = dvui.App.main;
 /// ```
-pub const main: fn () anyerror!void = if (@hasDecl(dvui.backend, "main")) dvui.backend.main else nop_main;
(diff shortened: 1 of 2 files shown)


RedPhoenixQ:
FYI I'm currently away so might not be able to test until the weekend.

phatchman:
Works for me on windows. Clock runs at full speed while resizing and window renders correctly while resizing.

david-vanderson:
Awesome thank you @phatchman!

Issue #268, opened 2025-04-21
dvui grid widget


phatchman:
## Proposal for dvui.GridWidget

Create a dvui grid widget with the following features:
1) Display of a fixed header row with a scrollable data area beneath it.
2) Auto-scale column widths to fit data / or allow setting of min and max widths
3) Ability to sort data by clicking on column headers in ascending or descending order.
4) Horizontal / Vertical scrolling
5) Display sort direction in column heading
6) Mouse selection
7) Keyboard navigation / selection
8) Single select and multi-select support
9) Only render visible set of elements, while providing virtual scrolling over full set of elements
10) Support for slice and iterator population of the grid.
11) Data columns (column data is bound to a field of a struct)
12) Checkbox columns
13) Custom layout columns.
14) Auto-fit content, auto-fit window, equal-width, and any other fit modes.
15) (optional) Draggable / resizable columns
16) (optional) Drag and drop between multiple grids.

I have created a proof-of concept for this component at https://github.com/phatchman/dvui-dev/tree/datagrid_proto.  If you are able, I'd appreciate any feeback on the approach taken in the PoC as I plan to polish this further and develop it into a PR. 

I split the grid into 3 widgets to make the layout of the header and body easier. I had tried some magic with setting the parent when adding a header or a body column, but this way seems much cleaner.
The downside of this approach is that the user can the GridHeader and GridBody outside of the GridWidget. I'm not sure if there is a to detect / warn about this?

My aim is to allow the user to create any type of "gridHeader" or "gridColumn" function to customize the look of the grid / columns. ~~Ideally the parent vbox for each column would be moved out of the "gridColumn" function and I'll see if I can work out a way to do that.~~ It does this now in the BeginHeaderCol/SendHeaderCol, BeginBodyCol/EndBodyCol functions.

Cheers,

Paul

phatchman:
Awesome. 

For the sorting, I can think of a couple of options:

1) I could put the state management fully back on the user:
* dvui.gridHeading fn returns true if it has been clicked on.
* user needs to keep track of which column was last sorted on and manage asc / desc themselves.
* user needs to supply asc / desc to the header column as an init option so it can draw the sorting arrow.

2) I could return a SortDirection from gridHeading that is none, asc or desc.
* User then sorts data based on which header return asc or desc. This open up the potential for multi-column sorting in the future. (e.g. with shift-click or similar)
* Widget remembers sorting direction and whether it is being sorted on. 
[...]

PR #314, opened 2025-05-27
GridWidget implementation

phatchman:
# Grid Widget

This seems to be in a pretty reasonable state now. Please review for merge.

Partially implements #268

## Discussion / Questions

Things I'd like your input on below:
1) Does columnFromSlice provide enough value vs the complexity of the generic function and the need for styling callbacks? It is saving the user from writing the following code for each column:
```
{
    for (data_set[0..], 0..) | *item, num | {
            var cell = try grid.bodyCell(@src(), num, .{ });
            defer cell.deinit();
            try dvui.labelNoFmt(@src(), item.name,  .{});
    }
}
```
2) Should VirtualScroller live in the GridWidget namespace or be exported through the dvui namespace? Or anything else exported for that matter. Only GridWidget is currently exported.
3) Unsure about the naming of the "add column" function in GridWidget. Currently named column() vs addCol/colAdd? Which is different to addTab(), but similar to dvui.label() etc.
4) Decided that supporting margins and padding on columns as well as cells wasn't required as user can just set them on each cell. This makes layout simpler as don't have to place each row in the column's content rect. And top padding to a column would apply before the header and bottom padding after the bottom row, which doesn't seem useful as they can just pad the top and bottom of the grid instead. Thoughts?
5) The row highlighting on hover in the virtual scrolling example feels a bit awkward? Is it OK or have any suggestions? The code makes perfect sense once you understand how dvui works, but for a beginner it may seem difficult to understand? At least there is an example they can copy.
6) That issue came up again with gravity_x = 0.5 not working for boxes unless .expand is used on the inner widget. See the virtual scrolling demo and placement of the tick box. I couldn't use the .expand as it would distort the image, so I had to center it "manually" instead.

Please feel free to be as picky/thorough as you need when reviewing. 

## Completed
- [X] Display of a fixed header row with a scrollable data area beneath it.
- [X] Provide various column layouts (e.g. fit to window, proportional layout, fixed_width, user calculated)
- [X] Ability to sort data by clicking on header rows
- [X] Display sort direction
- [X] Ability to select rows via checkbox
- [X] Horizontal / Vertical scrolling
- [X] Virtual scrolling (only render visible elements)
- [X] "Databound" columns - gridColumnFromSlice
- [X] User customizable cell and column styling and layouts
- [X] Ability to highlight hovered rows (userland)

## TODO
- [ ] Get rid of the padding for the sortable header. It isn't const the size of the symbol is dependent on the font size?
- [ ] Handle padding when header column width is smaller than the heading button width. Currently doesn't show the separator.
- [ ] Need some default padding, so the first column isn't hard against the edge of the grid.
- [ ] Support single select vs multiple select on the checkbox columns?
- [ ] Fix checkbox separator location when placed in a too big or too small column
- [ ] Change mouse pointer and behavior when hovering over grid. i.e. stop the "move window" that happens when clicking on a demo grid.
- [ ] Fix alignment of checkbox headers vs checkbox rows.

## Issues
- [ ] Grid header widget assumes vertical scroll bar width is 10 and that it will always be displayed. 
- [ ] Headers "wobble" with large virtual row heights. Almost certainly fp precision issue
        - Header is set to y position of scroll_info.virtual_size.y, but this doesn't always correspond to the top-left of the window.
        - Doubt this is solvable unless draw headers independently of the scroll canvas. i.e. relative to the outer grid's box?
- [ ] For 1 million rows our scroll height calc is out by 0.5 lines. Maybe work out some small padding to add to compensate for large numbers.
- [ ] Variable row heights demo seems to have an extra pixel between rows?

## Remaining Features
* Row headings?
* Some better visual indication that columns are sortable.
* Resize columns via dragging 
* Filtering headers - May not be worth providing anything generic here. Likely just an example of string filtering.
* Drag / Drop? - Guessing this can be implemented in user-land. Likely just needs an example.
* Keyboard navigation / selection - Likely userland and example only as well.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -17,6 +17,7 @@ const ButtonWidget = dvui.ButtonWidget;
 const FloatingWindowWidget = dvui.FloatingWindowWidget;
 const LabelWidget = dvui.LabelWidget;
 const TextLayoutWidget = dvui.TextLayoutWidget;
+const GridWidget = dvui.GridWidget;
 
 const enums = dvui.enums;
 
@@ -316,6 +317,7 @@ pub const demoKind = enum {
     animations,
     struct_ui,
     debugging,
+    grid,
 
     pub fn name(self: demoKind) []const u8 {
         return switch (self) {
@@ -335,6 +337,7 @@ pub const demoKind = enum {
             .animations => "Animations",
             .struct_ui => "Struct UI\n(Experimental)",
             .debugging => "Debugging",
+            .grid => "Grid",
         };
     }
 
@@ -356,6 +359,7 @@ pub const demoKind = enum {
             .animations => .{ .scale = 0.45, .offset = .{} },
             .struct_ui => .{ .scale = 0.45, .offset = .{} },
             .debugging => .{ .scale = 0.45, .offset = .{} },
+            .grid => .{ .scale = 0.45, .offset = .{} },
         };
     }
 };
@@ -468,6 +472,7 @@ pub fn demo() !void {
                     .animations => try animations(),
                     .struct_ui => try structUI(),
                     .debugging => try debuggingErrors(),
+                    .grid => try grids(0),
                 }
             }
 
@@ -520,6 +525,7 @@ pub fn demo() !void {
             .animations => animations(),
             .struct_ui => structUI(),
             .debugging => debuggingErrors(),
+            .grid => grids(scroll.si.viewport.h - hbox.data().rect.h - 10),
         };
     }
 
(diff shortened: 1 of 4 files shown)


phatchman:
> This is awesome!
Thanks. This version is so much nicer than the "alpha", both in code and behavior. Appreciate all your help to guide it in this direction.
> 
1. columnFromSlice()
> My guess is it's a bit overkill....
Yeah. My fear is someone will take a look at that function and think that creating a column is much more difficult than it really is. Writing the examples, I never really felt compelled to reach for it (though I am mostly using generated data there). 
2. Namespace
> Unsure, but by default I would put it in the GridWidget namespace unless/until it becomes awkward.
I prefer more namespacing, over less and it gives the user more options. So, I'll leave as is and maybe add the column-style functions there from dvui if it makes sense.
3. Column function naming
> Sorry about that, dvui naming is not super consistent.  I think "column()" is fine.  (addTab() should probably be just tab(), I will look at that)
If you are looking at the TabWidget.zig, I think this line shouldn't have .debug = true? 
`self.scroll = ScrollAreaWidget.init(src, scroll_opts, self.options.override(.{ .debug = true }));`
5. Finding hovered row to highlight
> It's not bad, maybe we could do better like having a function that gives you the RectScale for a particular row?  Then you can run a normal event loop on that.  I'm unsure, will have try it.
What about adding a function to the grid that returns the mouse position relative to the "body" area. mousePosBody() or something? Then they would do mousePosBody().y / grid.row_height to work out the highlighted row or alternatively .x and some simple math for the hovered column? I think GridWidget is going to end up with its own processEvents() anyway.
6. Centering widgets
>This is a real problem that I need to solve.  dvui needs a widget like box but that has a simpler layout so that a single child can be .expanded in both directions.  I will add that functionality to BoxWidget, and add "dvui.boxBasic()" for it.  Is there a better name?
Can only suggest boxSimple? But I can't give you any reason to prefer that over boxBasic. Seems fine to me.

### Data by Row
> My only real question at this point is whether we need the ability to provide data by row as well as by column. I'm thinking of things like a database viewer, where we know the columns, but the data comes by row. I think we can make that happen by reusing your concept of providing the columns up front, and then passing a row and col to bodyCell, where the individual body cells are not wrapped inside a column vbox, but get their rect directly passed from the grid.

Now let me try and convince you why every other grid widget ever created has got this wrong that laying out by column is the one and only true way to make a dvui grid widget. (Joking of course. Want to hear your counter-points).

For an immediate mode grid, you are going to need some backing storage for the data. Let's say it is a large database query, with some sort of scrolling cursor that allows the user to page through the data one chunk at a time. They can't just stream that data from the database query result to the grid row by row as they're going to need to render the same data again next frame. Either the database API or user is going to need to store that chunk of results in memory. Because they are going to have the data in-memory anyway, I don't think requiring them to layout by column is going to be an issue. 

The other case I was thinking about is a large-file csv reader. Here we'd probably need to run a parsing pass over the whole file to determine the number of columns, number of lines and the file-offsets of each line-break. Then presumably we'd load 'n' lines-worth of the file into a buffer, parse whatever lines are visible into some array of slices (one slice per column) and use those slices to populate the grid. I think that unless there is a scenario where the data needs to be re-generated each frame, (and can only be generated as rows) you are always going to have some data-structure to iterate over. And even then, we are only talking about needing the visible rows, so maybe 50 rows max?

Having said that, it does make using an actual iterator more painful as you need to reset or create a new iterator for each column, but iterators are tricky with virtual scrolling anyway, so they aren't necessarily a nice fit for providing the data.

I did spend some time thinking about how to implement data by row, and came up with the same solution as you by adding a bodyCell function that takes a col_num. If we did this, I'd probably get rid of the vboxes altogether. They are really only being used to store the column widths, for styling the column and for the "expand" option used when no column width is provided. All of this is pretty easily replicated within the widget and forcing all styling to be on the cell rather than the column.
> 
> Even so, I think this code is good enough to be merged and worked on in main. Please merge when you are ready - let me know if you can't.
> 
Cool! Will do.

phatchman:
Re: row based data

The only other thing I can think of is some datastructure that is so expensive to iterate over, that you only want to do it once? But in that case you'd probably want to implement caching of visible rows anyway.

david-vanderson:
Oh - when I write this up in my monthly update post, do you want to be credited as "phatchman" or another name?

phatchman:
You can use my actual name. Paul Hatchman. 

There was a time when unix login names had a max of 8 characters :)

david-vanderson:
Hahaha I remember that craziness - thanks!

phatchman:
Hmm. The below doesn't seem right to me, but I'm not sure what else I would do to stop the mouse events going up to the floating window and enabling the "drag" cursor and click-drag functionality.

Is there any reason the I should let the mouse events propagate outside the grid widget?
Probably key events be left alone because they could be used for shortcuts etc.

```
pub fn processEventsAfter(self: *GridWidget) void {
    const evts = dvui.events();
    for (evts) |*e| {
        if (!dvui.eventMatchSimple(e, self.data()))
            continue;

        switch (e.evt) {
            .mouse => {
                e.handle(@src(), self.data());
            },
            else => {},
        }
    }
}
```

phatchman:
OK that makes sense. Hmmm.  Maybe the real issue is I don't like that the floating window defaults to dragging when you anywhere click on it, rather than just when you click on the window header?

With the current functionality, it just seems strange that when you hover over a label in a grid, you get the drag cursor, but if you hover over a text area or other widget, you get a either a pointer or I-beam and no drag interaction. 

It also feels weird when moving over the header as it goes from Pointer to Drag just at the time you think you could drag the column separator, but instead you drag the window.

david-vanderson:
> OK that makes sense. Hmmm. Maybe the real issue is I don't like that the floating window defaults to dragging when you anywhere click on it, rather than just when you click on the window header?

This is a real thing, I also am a bit weirded out by it sometimes.  We copied that behavior from dear imgui.  Maybe we should revisit it.  Want to file an issue for it?

> See below where I lay out 4 boxes in a grid, next to each other:

Sorry - that looks like a bug.  The border is being drawn inset from where it should be.  I'll look at that as soon as I can - thanks for the report!

david-vanderson:
> Sorry - that looks like a bug. The border is being drawn inset from where it should be. I'll look at that as soon as I can - thanks for the report!

Just pushed a fix for it, definitely just a bug.  Does that fix it?

phatchman:
> styling/banding - we might need to move some of that into grid, so that when you ask for a cell, it comes already banded.

Just noting if we do this, no widgets in the cell would either not draw background or need access to the cell's fill color so they can draw the same background color as the cell. Need to find example of where widgets might need to draw their own background.

Issue #318, opened 2025-05-30
Change Floating Windows so that "click to drag" is not the default


phatchman:
## Background
When clicking within a FloatingWindowWidget you will get one of two different behaviors:
1) If the widget under the cursor _does not_ handle the mouse click event, the floating window will capture the mouse and drag the window as the mouse moves.
2) If the widget under the cursor _does_ handles the mouse click event, the click will be processed by that widget and the window does not drag.
This is also indicated by the drag cursor changing to another cursor type depending on whether the widget handle mouse motion events.

This causes 3 issues:
1) The user experience for floating windows is different to that of non-floating windows.  Clicking on an empty box will cause a window drag in the floating window, but not in a main window.
2) For widgets like the GridWidget, which are made up of many widgets, it makes the user experience feel like interacting with separate parts, rather than an integrated widget.
3) The user experience is different to the operation of a normal Linux, Mac or Windows desktop or web browser experience with pop-up and dialog-style windows.

The current functionality is similar to Dear ImGUI, so there must be value in retaining the current functionality.

## Proposal
1) Provide an init option to choose between full window click to drag and header only click to drag
2) If click to drag is set to header only, only show the drag cursor and allow dragging with the user clicks in the header.

PR #319, opened 2025-05-30
Floating window limit dragging proto

phatchman:
This is prototype code to demonstrate the concept for #318

The issue I see is that the window heading is created separately to the floating window, or might not be created at all. So either we:
1) Remove the dragging code out of FloatingWindow and into the window header. But this means there is no way to move floating windows without headers and each different style of window heading would need to implement it. Not good.
2) Allow the user to specify the area of the floating window that allows dragging.
3) Have the header widget register itself with the floating window and set the drag rect. (Which is just option 2 with the header widget passing the rect instead of the user). 

I feel like option 2 returning the rect of the overlay widget is the best of those options? It will still break all users of dvui.windowHeader(), but in a way that people can just put a _ = in front of it if they want to keep existing behavior.

So I propose that:
1) dvui.windowHeader either return the overlay widget or some other way to access the window header rect.
2) FloatingWindowWidget take an init option of "drag_rect" or similar which optionally limits dragging functionality to within that rect.

While this solution is going to be fairly simple to implement, it does feel a little inelegant? Though I'm not sure what other solution we'd come up with if designing this from the beginning assuming the desire is to have floating windows draggable. It is quite flexible as someone could add a drag handle or similar to the window using this.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -375,7 +375,7 @@ pub fn demo() !void {
 
     const width = 600;
 
-    var float = try dvui.floatingWindow(@src(), .{ .open_flag = &show_demo_window }, .{ .min_size_content = .{ .w = width, .h = 400 }, .max_size_content = .width(width), .tag = demo_window_tag });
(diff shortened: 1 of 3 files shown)


david-vanderson:
In FloatingWindowWidget.deinit() we start with `const rs = self.wd.rectScale();` - I think it needs a separate rect that can be set anytime before deinit() is called.

Yeah - I'm pretty sure I know what this should look it, give me a few days to have the time.

PR #320, opened 2025-05-30
PanedWidget: Make the handle size configurable

foxnne:
Please let me know if I did this wrong, I was behind several commits and pulled from upstream but its been a while since I've done this. If I need to redo it so it doesn't show all these changes somehow please let me know!


david-vanderson:
I'm not good enough with git to know what the right fix here is, I think it would involve force pushing this branch?  In any case, I cherry-picked your changes into main.  Reopen if anything looks off.

Thank you!

Issue #192, opened 2025-02-28
error handling strategy


david-vanderson:
Error handling in dvui lacks a coherent strategy, but here's a proposal: "make it to the next frame".

Currently most dvui functions can return various errors, and there's no guidance on how to handle them.  I personally have been putting "try" in front of everything and the whole app goes down if there is an error.

What could you do if a dvui function returns an error?  We can imagine:
1. something is horribly wrong (like out of memory) and likely the app will crash, or at the least many things will start erroring
1. something temporary went wrong (like a backend dropped a texture)
1. something is consistently wrong with this part of the code (like misusing the dvui api)

I'm proposing moving to a strategy where if dvui encounters an error, it will log it and continue as best it can, skipping whatever needed to make it to the next frame.

This has the benefit of reducing the use of "try" all over application code, and shifting the burden of dvui error handling from the application to dvui.  It also can turn temporary bugs from crashers into "this part didn't render", which is a clear win for robustness.

Any feedback about this would be helpful!

rohlem:
> It also can turn temporary bugs from crashers into "this part didn't render", which is a clear win for robustness.

My only concern with this strategy would be that a clear "oh, it crashed, and here is the error trace" might become "huh, this doesn't render, I don't know what's wrong with it/ the code".

I'm not sure what structure you have in mind - would it be feasible/helpful if the package user could configure dvui's error handling strategy?
It could be a single flag of crash vs swallow errors, or as fine-grained as feasibly useful.

david-vanderson:
> My only concern with this strategy would be that a clear "oh, it crashed, and here is the error trace" might become "huh, this doesn't render, I don't know what's wrong with it/ the code".
> 
> I'm not sure what code structure you have in mind - would it be feasible/helpful if the package user could configure dvui's error handling strategy? It could be a single flag of crash vs swallow errors, or as fine-grained as feasibly useful.

Good thinking - this is a valid concern.  I'm not sure exactly how this will work, but here are some thoughts.

I'm imagining in the place where dvui catches an error, it always logs to stderr, and then (optionally?) stores the error (or maybe a limited number of errors).  The app can check every frame if there is a stored error, and either crash or do nothing, or maybe in the future show a user-friendly error-reporting thing?

I think dumping something to stderr is enough for developers during app development.  We can outline in red the widget the error was in.  If that's not enough we could have a switch to crash on all errors.

I'm not sure if overloading the return value fits well here - we want something where the code doesn't have to change between development and production.

[...]

PR #322, opened 2025-06-02
Add shrinking arena allocator to prevent capacity leak

RedPhoenixQ:
I mentioned this before in https://github.com/david-vanderson/dvui/issues/192#issuecomment-2925202681.

Before when using the `std.heap.ArenaAllocator` the capacity would never shrinking as the `.retain_capacity` option was passed to the arena. We could not easily use `.retain_with_limit` before as the reasonable limit would change per application or view in the application.

To address this a wrapper around the `std.heap.ArenaAllocator` was created that keeps track of the peak memory usage before the reset and retains only that much, with some padding. This allows the arena to shrinking over time but still adapt and be as fast as before on "static" views.

I first tried to calculate the memory used based on the capacity and `end_index` of the `std.heap.ArenaAllocator`. This was unreliable as memory could have been freed before the check, like in the plot example, causing the needed capacity to be freed and reallocated every frame.

### DIFF
--- a/src/Window.zig
+++ b/src/Window.zig
@@ -102,7 +102,7 @@ capture: ?dvui.CaptureMouse = null,
 captured_last_frame: bool = false,
 
 gpa: std.mem.Allocator,
-_arena: std.heap.ArenaAllocator,
+_arena: dvui.ShrinkingArenaAllocator,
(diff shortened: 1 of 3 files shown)


david-vanderson:
This rocks - thanks!

PR #325, opened 2025-06-02
Use stb_truetype when cpu == wasm64

knightpp:
When I build for wasm64 I get error

```
~/.cache/zig/p/freetype-2.13.0-AAAAAI6NqAA4EbABl7qfk5qBFJQb2xjUIcP6SY7EBFvJ/include/freetype/config/ftstdlib.h:88:10: error: 'string.h' file not found
```

After this fix my build passes.


david-vanderson:
Thank you!

PR #326, opened 2025-06-02
PanedWidget: New option `handle_interactive` 

foxnne:
When true enables the handle to grow/shrink based on mouse position, rather than just toggling visibility. `false` by default, so everything should remain as is currently.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -1830,7 +1830,7 @@ pub fn layout() !void {
     }
     try dvui.label(@src(), "Collapsible Pane with Draggable Sash", .{}, .{});
     {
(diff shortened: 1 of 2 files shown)


david-vanderson:
@foxnne Sorry about that.  That widget had a bunch of units errors, they should be fixed now.  I moved the dynamic stuff into it's own optional struct.  How does that look?  Any advice on what those default values should be (for .handle_size_max and .distance_max)?

foxnne:
@david-vanderson No worries at all! I figured we would need some changes :) Looks good! I just pulled and tried this out in my project. For my stuff, it seemed to look best with `handle_size_max = 10` and `distance_max = 60`, but that may be too much movement for smaller windows like the demo window, though it is off by default.

Anyway, I think what you have here is great! Thanks!

david-vanderson:
Okay we'll compromise - I like the handle size, but let's make the distance small by default so it stays relatively local.

Thanks very much!

PR #330, opened 2025-06-03
Add TrackingAutoHashMap with support for usage tracking

RedPhoenixQ:
This removes the need for the structs contained within to maintain a `used` field. The tracking can be customized to only apply on insertions like for tags and min_size where the used flag indicates multiple insertions in a single frame.

The `TrackingAutoHashMap` is an unmanaged hashmap, which I deemed to be most flexible but I can easily make it a managed hashmap like before. What is your thoughts on unmanaged hashmaps and arraylists?

### DIFF
--- a/src/WidgetData.zig
+++ b/src/WidgetData.zig
@@ -292,7 +292,7 @@ pub fn minSizeSetAndRefresh(self: *WidgetData) void {
 
     var cw = dvui.currentWindow();
 
-    const existing_min_size = cw.min_sizes.fetchPut(self.id, .{ .size = self.min_size }) catch |err| blk: {
+    const existing_min_size = cw.min_sizes.fetchPut(cw.gpa, self.id, self.min_size) catch |err| blk: {
         // returning an error here means that all widgets deinit can return
         // it, which is very annoying because you can't "defer try
         // widget.deinit()".  Also if we are having memory issues then we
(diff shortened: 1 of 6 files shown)


david-vanderson:
Looks good!  I'm generally following the zig core team here - they are recommending unmanaged by default, so moving in that direction is great!

david-vanderson:
I'm seeing stuff like this when running `zig build test` now:

```
[gpa] (err): memory address 0x111202400 leaked: 
/Users/dvanderson/code/dvui/src/Window.zig:1218:82: 0x1052f8593 in dataSetAdvanced__anon_30487 (dvui_sdl3)
    var sd = SavedData{ .alignment = alignment, .data = self.gpa.allocWithOptions(u8, bytes.len * num_copies, alignment, null) catch |err| switch (err) {
```

Are you?

RedPhoenixQ:
Yes, weird because CI passed.

RedPhoenixQ:
It's only "DOCIMG demo" and examples/app.zig "open example window" tests that leaks

RedPhoenixQ:
I made a mistake while copying code for removing unused data, accidentally removing the call to free on SavedData. Fixed in 0ad53e471b3d1766647087eab3dcbe1156572086. Still weird that CI passed with this issue...

david-vanderson:
That is weird, but thanks for fixing it!

david-vanderson:
I'm working on fixing tags to allow non-static tag names, and I want to write:
```zig
    if (cw.tags.getPtr(name)) |old_data| {
        if (old_data.used) {
```

How do I get the KV thing (so I can test .used) without putting a new value (using fetchPut)?

RedPhoenixQ:
@david-vanderson I think the easiest way is to reach down to the internal `map` to get the value from there, as you also don't want it to track this get, otherwise used would always be true

Could add a doc comment about untracked access being done by using the internal map directly

PR #332, opened 2025-06-03
Add Path namespace and add improved Builder experience

RedPhoenixQ:
This change makes the API easier to discover as well as minimizing the pollution of the dvui namespace, which is quite large. Keeping all related path functions within a struct also allows for better documentation as the relationship between parts are clearer.

I've though about how dvui makes almost everything available under `dvui.xxx` but some amount of namespacing for related operations would be nice for API discoverability and code navigation. This moves all functions that were already pseudo namespaced with `dvui.pathXxx` to a real namespace of `dvui.Path.xxx`, with the added benefit of `Path` being able to be a struct passed to the related functions. I have a branch doing similar namespacing for `dvui.dataXxx`. What are you thoughs on this kind of namespacing @david-vanderson?

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -1504,11 +1504,11 @@ pub fn styling() !void {
             defer drawBox.deinit();
             const rs = drawBox.data().contentRectScale();
 
-            var path: dvui.PathArrayList = .init(dvui.currentWindow().arena());
+            var path: dvui.Path.Builder = .init(dvui.currentWindow().arena());
             defer path.deinit();
-            try dvui.pathAddRect(&path, rs.r, dvui.Rect.Physical.all(5));
+            try path.addRect(rs.r, dvui.Rect.Physical.all(5));
 
-            var triangles = try dvui.pathFillConvexTriangles(path.items, .{ .center = rs.r.center() });
+            var triangles = try path.build().fillConvexTriangles(.{ .center = rs.r.center() });
             defer triangles.deinit(dvui.currentWindow().arena());
 
             const ca0 = backbox_color;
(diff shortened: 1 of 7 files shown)


david-vanderson:
This looks good - well done!

I'm not totally sure about the naming.  The Path case makes more sense to me, because you can instantiate it.  The data functions don't have a similar thing (closest might be WidgetId), so let's leave them for now.

I agree it's kind of a lot of stuff all jammed under dvui.* so we'll slowly chip away where it makes sense.  But I'm also worried about the api churning too fast.

RedPhoenixQ:
> But I'm also worried about the api churning too fast.

This is very valid. I feel a little bad every time I change some public API.

I'll hold of on more namespacing changes, unless there is other good reasons for them. Might work on moving helper functions like `dvui.box` to `BoxWidget.zig` and reexporting to make the `dvui.zig` file a bit cleaner.

david-vanderson:
> This is very valid. I feel a little bad every time I change some public API.

Me too.  I think it's okay as long as we do it slowly enough for users to adapt.

> I'll hold of on more namespacing changes, unless there is other good reasons for them. Might work on moving helper functions like `dvui.box` to `BoxWidget.zig` and reexporting to make the `dvui.zig` file a bit cleaner.

Sounds good.  In that particular case, I don't know if it's more or less useful to have all those higher-level functions next to each other in the same file, or spread around.

RedPhoenixQ:
> In that particular case, I don't know if it's more or less useful to have all those higher-level functions next to each other in the same file, or spread around.

On second thought, you might be right. Looked at the actual line number and even `dvui.textEntryX` is around 200-300 lines, which in my mind was one of the larger easily grouped clusters. Maybe I'll try it out but I see less value in it now.

david-vanderson:
Cool.  Merge when ready (this and the tracking hashmap) - I wasn't sure if you were waiting on something or not.

PR #334, opened 2025-06-04
Make all render functions to use `renderTriangles` 

RedPhoenixQ:
This updates `renderText` and `debugFontAtlases` to use `renderTriangles` and `renderTexture` respectively. This now means all calls to `Backend.drawClippedTriangles` are done via `renderTriangles`, ensuring the safeties applied there is applied for all dvui rendering. Along with this, manual triangle creation is now limited to only `Path.fillConvexTriangles`, `Path.strokeTriangles` and `renderText`. 

Making `debugFontAtlas` use `renderTexture` also means it doesn't have to handle deferred rendering, so one `RenderCommand` variant can be removed.

### DIFF
--- a/src/Window.zig
+++ b/src/Window.zig
@@ -1167,9 +1167,6 @@ pub fn renderCommands(self: *Self, queue: std.ArrayList(dvui.RenderCommand)) !vo
             .text => |t| {
                 try dvui.renderText(t);
             },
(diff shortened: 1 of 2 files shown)


david-vanderson:
Great work!  Merge when ready.

Thank you!

PR #338, opened 2025-06-06
Dx11 dpi scaling issues

RedPhoenixQ:
There were some odd behaviours for the dx11 scaling were the mouse position was *slightly* off towards the bottom right of the window. Turns out `win32.GetWindowRect` includes the title bar and other things in the window rect, messing up the natural scale factor.

To fix this issue, the dpi scaling is manually applied to the pixel window size. Also turns out dx11 gives the mouse position in out Physical coordinates, so we have to unapply the Natural to Physical scaling. I left a todo about changing the parameter type for mouse events at comptime.

### DIFF
--- a/src/backends/dx11.zig
+++ b/src/backends/dx11.zig
@@ -1035,8 +1035,7 @@ pub fn refresh(self: Context) void {
     _ = self;
 }
 
-fn addEvent(self: Context, window: *dvui.Window, key_event: KeyEvent) !bool {
-    _ = self;
+fn addEvent(_: Context, window: *dvui.Window, key_event: KeyEvent) !bool {
     const event = key_event.target;
     const action = key_event.action;
     switch (event) {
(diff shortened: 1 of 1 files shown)


src/backends/dx11.zig
```diff
@@ -1052,7 +1050,12 @@ fn addEvent(self: Context, window: *dvui.Window, key_event: KeyEvent) !bool {
             return window.addEventMouseButton(ev, if (action == .up) .release else .press);
         },
         .mouse_event => |ev| {
-            return window.addEventMouseMotion(.{ .x = @floatFromInt(ev.x), .y = @floatFromInt(ev.y) });
+            const state = stateFromHwnd(hwndFromContext(self));
+            // Coordinates are already Point.Physical, so the scale from Natural has to be unapplied
+            const point = dvui.Point.Physical{ .x = @floatFromInt(ev.x), .y = @floatFromInt(ev.y) };
+            // TODO: Maybe addEventMouseMotion should take a comptime arg to choose between
+            //       Physical and Natural points as input
+            return window.addEventMouseMotion(point.scale(1 / (state.dvui_window.natural_scale / state.dvui_window.content_scale), dvui.Point.Natural));
```
SoraTenshi, line 1058: I guess this is something for the higher level api?

SoraTenshi (approved):

src/backends/dx11.zig
RedPhoenixQ: Not sure which higher level api you mean. I was just thinking of making the signature:

```zig
fn addEventMouseMotion(win: *Window, comptime Point: anytype, pt: Point)  std.mem.Allocator.Error!bool {
    ...
    const newpt = switch (Point) {
        dvui.Point.Natural => pt.scale(self.natural_scale / self.content_scale, Point.Physical),
        dvui.Point.Physical => pt,
        else => unreachable,
    };
    ...
}

david-vanderson: I've pushed a less fancy addition.  Not totally sure, but generally I'm trying to avoid "anytype" in user-facing apis.  Not a strong preference, so we'll see how it goes.  Thanks!

david-vanderson:
Merge when ready.  Thank you!

PR #340, opened 2025-06-06
IconWidget: Initial color set to white 

foxnne:
...to prevent final fill color being multiplied with text color.

When using the IconWidget, it seems that the color is defaulted to be the theme's text color, which is fine except when setting the fill color of the icon, the final color seems to be multiplied with the text color, meaning you can't set the icon fill to the color you want. 

I believe we just need to set that `color_mod` to white to prevent that behavior, but please let me know if this is wrong, we may need to also set the fill color to .text by default?

### DIFF
--- a/src/widgets/IconWidget.zig
+++ b/src/widgets/IconWidget.zig
@@ -49,7 +49,7 @@ pub fn matchEvent(self: *IconWidget, e: *dvui.Event) bool {
 
 pub fn draw(self: *IconWidget) !void {
     const rs = self.wd.parent.screenRectScale(self.wd.contentRect());
-    try dvui.renderIcon(self.name, self.tvg_bytes, rs, .{ .rotation = self.wd.options.rotationGet(), .colormod = self.wd.options.color(.text) }, self.icon_opts);
+    try dvui.renderIcon(self.name, self.tvg_bytes, rs, .{ .rotation = self.wd.options.rotationGet(), .colormod = .white }, self.icon_opts);
 }
 
 pub fn deinit(self: *IconWidget) void {


RedPhoenixQ:
I think this removed the behaviour of defaulting to text color. I think the better option is to not set `.color_mod` at all and instead overriding the `self.icon_opts` colors to `.text` if they are null (both stroke and fill I think?). This way the user can still choose the color manually, without `.color_mod` interfering. The only downside is that `IconWidget` wouldn't be able to render icons with their predefined colors.

foxnne:
Hmm, seems like this will need a bit more thought as I'm sure we don't want to give up `IconWidget` being able to render icons with predefined colors. I'll investigate a bit more.

foxnne:
@david-vanderson do you have advice on this? I just have a local patch in my project so i can set specific colors, but I'm not sure what the right answer is, as setting it to white I don't think is the correct fix. Not setting the color_mod makes them all white as well. I also see this may be tied into icons that have their own colors?

david-vanderson:
> @david-vanderson do you have advice on this? I just have a local patch in my project so i can set specific colors, but I'm not sure what the right answer is, as setting it to white I don't think is the correct fix. Not setting the color_mod makes them all white as well. I also see this may be tied into icons that have their own colors?

This was more tricky than expected.  Here's the commit message:

> The problem is that we now have 2 ways to color an icon:
> * set the color when rasterizing
> * apply a color mod in the shader
> 
> This solution is a compromise.
> 
> If the code is rasterizing with default values (meaning a white icon),
> then we will always pass a colormod, so the icon will have the same
> color as text.  This is the normal case.
> 
> If the code rasterizes with non-default values, then we only pass a
> colormod if the code explicity passed a .color_text value.  This allows
> full control over both rasterization and colormod.
> 
> The only catch is that if someone gives explicit rasterization
> parameters AND they happen to be the defaults, then they might be
> surprised that the text color is multiplied in.
> 

Does that sound reasonable?

foxnne:
Sounds perfectly reasonable to me, and thank you so much!

Issue #317, opened 2025-05-30
Grid Widget planned Functionality


phatchman:
Now that the initial implementation of GridWidget merged, this issue will replace #268 to track planned enhancements to the grid.

## Planned Enhancements
- [ ] Allow columns to be resized by clicking and dragging the header separator
- [ ] Add an example with Filtering Headers.
- [ ] Add an example with keyboard selection.
- [ ] Look at ways to simplify gridColumnFromSlice into potentially a handful of smaller column functions. Including looking at better ways to providing styling options. e.g. for banding.

## Potential Enhancements
- [ ] Add an example with drag and drop
- [ ] Ability to provide data by row, instead of by column. This is expected to be a niche use-case as most applications will need to cache the data backing the visible rows, making it relatively easy to provide data by column instead of by row.
- [ ] Provide support for banding within the grid, rather than requiring the user to band themselves.
   - +ve is that removes the need to implement callbacks in column functions just to allow simple formatting like banding.
   - -ve is that banding is just one case of lots of different types of formatting. It is probably by far the most common case?

phatchman:
Hi @david-vanderson,

If you have some time, I've just started implementation of column resizing and wanted your quick thoughts on the GrabHandleWidget. 

I can't decide if this should really be a widget or a function more like dvui.labelClick()? It seems a little too complicated for a dvui function, though I could make use of an "internal" struct, similar to what you did for the recent change to dialogs.

Alternatively, I could try and develop this further and make a generic "grabbable" widget to be re-used by paned and possibly scrollbar? At first glance it seems that each of those widgets needs a slightly different version "processEvents()" and that seems to be really the only code that would be saved by making it a generic widget. 

So, I'm leaning towards changing it to use the function approach, but not sure. If I want to provide a custom drawing function to make the grab handles look better than just a separator in the future, maybe it's better to stay as a Widget? But it doesn't really feel like a widget since it only interacts with "width", rather than directly with other widgets. i.e. without the grid interpreting the calculated width, it won't "drag" anything.

Any other thoughts appreciated. (The code for horizontal orientation can be ignored for now).

The theory is that the "grab handles" will interact with the grid solely through the col_widths, init_opt slice. I didn't need to make any code changes to GridWidget to implement this resizing, so that was very nice! That shows off one of the clear strengths of DVUI's approach. 

https://github.com/phatchman/dvui-dev/compare/main..draggable2

[...]

PR #342, opened 2025-06-07
GridWidget: Added draggable column resizing and example

phatchman:
Adds the GridWidget.HeaderResizeWidget to allow column resizing via mouse dragging.
Added resizing ability to the layouts and data example.

Partially implements #317

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -3903,7 +3903,7 @@ fn gridStyling() !void {
                 grid.colSortSet(local.sort_dir);
             }
 
-            if (try dvui.gridHeadingSortable(@src(), grid, "Celcius", &local.sort_dir, .{}, .{})) {
+            if (try dvui.gridHeadingSortable(@src(), grid, "Celcius", &local.sort_dir, .fixed, .{}, .{})) {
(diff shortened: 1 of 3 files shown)


phatchman:
One thing I noticed while testing this is that something has changed recently (within the last week?) where the sorting chevrons used to totally obscure the button text as the column shrunk too small but now, they are drawn on top of the button text. i.e. merged with it.

~~I don't think that is a big deal as it can probably be resolved by the chevrons drawing a background. Just noting it in case it was unintentional. The current behavior is probably the correct one, I think?~~

Edit: No drawing the background doesn't help :(. It draws the background, but still merges the two lots of text. dvui.buttonLabelAndIcon() is what creates the two. I wonder if this is the same thing as reported in PR #340?  
I've created #343 for this.

P.S. I'm still amazed that this was possible without changing a single line of code in the GridWidget.

phatchman:
Wondering how well this works with touch?

Edit: Just tested this. It works "OK", but it's not obvious the the handles are draggable. It's not obvious on non-touch either. Perhaps use a double bar? Need to research what other UIs do.

Added to #317

david-vanderson:

PR #350, opened 2025-06-08
set self to undefined in deinits

phatchman:
The zig std library commonly sets self to undefined on deinit() to help detect use after deinit() errors. While I think use after deinit() is fairly unlikely the way DVUI is commonly used, it is trivial to implement and gives some extra protection.

This PR sets self.* = undefined in all deinit()s where self is a non-const pointer.
This operation becomes a noop in release fast and release small builds, but fills the struct with all 0xAA in debug and release safe builds.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -3559,7 +3559,7 @@ pub fn debuggingErrors() !void {
             try dvui.label(@src(), " - fix by passing .id_extra = <loop index>", .{}, .{ .id_extra = i });
         }
 
-        if (try dvui.labelClick(@src(), "See https://github.com/david-vanderson/dvui/blob/master/readme-implementation.md#widget-ids", .{}, .{ .gravity_y = 0.5, .color_text = .{ .color = .{ .r = 0x35, .g = 0x84, .b = 0xe4 } } })) {
+        if (try dvui.labelClick(@src(), "See https://github.com/david-vanderson/dvui/blob/master/readme-implementation.md#widget-ids", .{}, .{ .color_text = .{ .color = .{ .r = 0x35, .g = 0x84, .b = 0xe4 } } })) {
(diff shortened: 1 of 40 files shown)


phatchman:
Hmm. There is a difference in functionality with this PR applied in the demo. 

* Without this applied, the GridWidget displays at the correct height to prevent the floating window from scrolling. 
* With this applied, the GridWidget ignores the space taken up by the "Back to Demos" button causing scrolling.

Let me look into this further. Either I broke something, or this picked up an actual issue.

phatchman:
Ha. It is a real use after deinit() issue. 

hbox is deinitted on line 506, but used by me on line 530 in Examples.zig. It's sort of easy to miss that the deinit) didn't have a defer, as per the usual pattern,  in my defense!
```
        try dvui.label(@src(), "{s}", .{demo_active.name()}, .{ .font_style = .title_2, .gravity_y = 0.5 });
        hbox.deinit();
```
```
            .grid => grids(scroll.si.viewport.h - hbox.data().rect.h - 10),
```
And if I print out hbox.data().rect it shows
```
Rect{ -0.00000000000030316488 -0.00000000000030316488 -0.00000000000030316488 -0.00000000000030316488 }
```
Nice!

RedPhoenixQ:
I also have a branch in progress requiring all deinit calls to be via defer for error handling. It's about retrying frames on OOM errors. Setting to undefined will help a lot in mandating that

david-vanderson:

phatchman:
@RedPhoenixQ Can you please just double check the last commit in this PR vs your error handling changes? There were a couple of small conflicts.

RedPhoenixQ:
Looks good! Everything compiles and tests pass, sdl3 works without issue

PR #354, opened 2025-06-09
labels: new labelEx and init_opts

david-vanderson:
This leaves `label` alone adding a new `labelEx` that takes the extra parameter.

init_opts is added to `labelNoFmt` and LabelWidget.

### DIFF
--- a/examples/dx11-standalone.zig
+++ b/examples/dx11-standalone.zig
@@ -169,7 +169,7 @@ fn gui_frame() !void {
     }
 
     {
-        try dvui.labelNoFmt(@src(), "These are drawn directly by the backend, not going through DVUI.", .{ .margin = .{ .x = 4 } });
+        try dvui.labelNoFmt(@src(), "These are drawn directly by the backend, not going through DVUI.", .{}, .{ .margin = .{ .x = 4 } });
 
         var box = try dvui.box(@src(), .horizontal, .{ .expand = .horizontal, .min_size_content = .{ .h = 40 }, .background = true, .margin = .{ .x = 8, .w = 8 } });
         defer box.deinit();
--- a/examples/raylib-ontop.zig
+++ b/examples/raylib-ontop.zig
@@ -110,7 +110,7 @@ fn colorPicker(result: *dvui.Color) !void {
         var hbox = try dvui.box(@src(), .horizontal, .{});
         defer hbox.deinit();
 
-        try dvui.labelNoFmt(@src(), &color_hex, .{
+        try dvui.labelNoFmt(@src(), &color_hex, .{}, .{
             .color_text = .{ .color = result.* },
             .gravity_y = 0.5,
         });
--- a/examples/raylib-standalone.zig
+++ b/examples/raylib-standalone.zig
@@ -161,7 +161,7 @@ fn dvui_frame() !void {
             }
         }
 
-        try dvui.labelNoFmt(@src(), "Below is drawn directly by the backend, not going through DVUI.", .{ .margin = .{ .x = 4 } });
+        try dvui.labelNoFmt(@src(), "Below is drawn directly by the backend, not going through DVUI.", .{}, .{ .margin = .{ .x = 4 } });
 
         var box = try dvui.box(@src(), .vertical, .{ .expand = .horizontal, .min_size_content = .{ .h = 40 }, .background = true, .margin = .{ .x = 8, .w = 8 } });
         defer box.deinit();
--- a/examples/sdl-standalone.zig
+++ b/examples/sdl-standalone.zig
@@ -179,7 +179,7 @@ fn gui_frame() !void {
             }
         }
 
-        try dvui.labelNoFmt(@src(), "Below is drawn directly by the backend, not going through DVUI.", .{ .margin = .{ .x = 4 } });
+        try dvui.labelNoFmt(@src(), "Below is drawn directly by the backend, not going through DVUI.", .{}, .{ .margin = .{ .x = 4 } });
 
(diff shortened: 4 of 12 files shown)


david-vanderson:
@RedPhoenixQ and @phatchman I would appreciate any feedback here.

First I tried adding the extra parameter to `label()` but then there were so many lines that looked like:

`try dvui.label(@src(), "My Label", .{}, .{}, .{});`

And I already think having 2 `.{}` parameters is pushing it.  So this is a compromise - we have one new function in the rare case that people need it.

Does it work?  Any other ideas?

RedPhoenixQ:
> `try dvui.label(@src(), "My Label", .{}, .{}, .{});`

I don't think that is a huge problem because at least one of them is going to have format args, otherwise you should use `labelNoFmt`. `try dvui.label(@src(), "fps: {d}", .{dvui.FPS()}, .{}, .{ .gravity_y = 0.5 });` isn't too bad IMO

phatchman:
I don't think it is too bad, but that would be my absolute limit in terms of default empty parameters. LabelEx seems like a good pattern we can follow with other functions.

As an alternative, I had a similar issue with the grid options and am playing with a solution to that now. For each grid cell, callers usually won't need to supply any options. If they do supply options, then only one of the cell options or the normal "widget" options. Then in rare cases, they would need to supply both types of options.

So, what I'm playing with is combining the CellOptions and "widget" options into a "GridOptions" struct. Most users will just pass .none to set both options as empty, otherwise they set one or both struct members as they need.

Above is a little is a little simplified because I'm actually looking at taking an anytype, so that the user can pass their own "GridOptions" struct to return different options based on the current row/col. Very early stages, but looking promising so far.

phatchman:
Hmm. I just tried this by changing the dvui.buttonLabelAndIcon as below, but it doesn't display the text correctly. It's [squished](https://imgur.com/a/FAfNNaK).
```
        var outer_hbox = try box(src, .horizontal, .{ .expand = .horizontal });
        defer outer_hbox.deinit();
        try icon(@src(), label_str, tvg_bytes, .{}, opts.strip().override(.{ .gravity_x = 1.0 }));
        try labelEx(@src(), "{s}", .{label_str}, .{ .align_x = 0.5 }, .{});
```

I tried various combinations of .expand, but it didn't help.

The API is good. I wouldn't mind if it was an extra param to label either. The only downside is I had to change from labelNoFmt to labelEx and specify the format param, but I think that is understandable if need the extra layout options.

david-vanderson:
> Hmm. I just tried this by changing the dvui.buttonLabelAndIcon as below, but it doesn't display the text correctly. It's [squished](https://imgur.com/a/FAfNNaK).

Looks like with your code you are getting the default padding of a label, but the grid column header is constrained to be a certain height, so the label is pushed down too far and you only see the top bit of the text.  We'd want code more like:

```zig
        try icon(@src(), label_str, tvg_bytes, .{}, opts.strip().override(.{ .gravity_x = 1.0 }));
        try labelEx(@src(), "{s}", .{ label_str }, .{ .align_x = 0.5 }, opts.strip().override(.{ .expand = .both }));
```

Does that look better for you?

david-vanderson:
@foxnne, @Yinameah, @marler8997, @MichaelBelousov any opinions here?  We need to add a (likely rarely used) parameter to `dvui.label()`:
* do we just add it (maybe too many `.{}` params) or
* add a separate `dvui.labelEx()` function

foxnne:
>     * add a separate `dvui.labelEx()` function

I think this would be my vote. In a perfect world, there is only one function for a label and it covers everything you need, but I think in reality many of the widgets will end up exposing more settings/features over time. 

Dear Imgui went the route of having a standard function and an `exampleEx` function with more control/settings/features and I think that works pretty well and allows you to offer as much configuration as you like while maintaining that the "default" way of doing things isn't cluttered.

MichaelBelousov:
I vote `labelEx` in the current situation. 

Fwiw I think it would actually be better to have `labelFmt` be the name of the default label function. In that case, I think the 3 default arguments is ok since `Fmt` kind of makes it more readable imo.

david-vanderson:
> Fwiw I think it would actually be better to have `labelFmt` be the name of the default label function. In that case, I think the 3 default arguments is ok since `Fmt` kind of makes it more readable imo.

Do you mean we would have `dvui.label()` not do formatting, and `dvui.labelFmt()` do formatting with the extra format args parameter?

MichaelBelousov:
> Do you mean we would have `dvui.label()` not do formatting, and `dvui.labelFmt()` do formatting with the extra format args parameter?

Yes except you could also keep `label` as a deprecation notice and keep the name `labelNoFmt`

You could also add an optional fmt_args field to the int_opts of label but I personally still prefer adding "format" to the name somehow. It's a slight preference though, all of the options I think have their very slight pros/cons.

phatchman:
Ahh... OK.  I didn't think about that. Passing stripped options is not the same as passing empty options.
> Does that look better for you?
Yes.  Thanks!

Did you want me to add the change to the buttonLabelAndIcon to this pr?

Yinameah:
I think the extra function has my preference as well. It convey intent much better I find. 
(if the name itself is not enough, the doc comment will do) 


`. {}, .{}` is already not very nice. Why is there two set of parameters? In which one goes `.gravity` again?
Plus looking doc requires two jump to definition (at least on my setup) instead of one.

david-vanderson:
> Did you want me to add the change to the buttonLabelAndIcon to this pr?

Yes - thanks!

david-vanderson:
Thank you everybody for commenting, it was very useful.  Nothing is perfect, and I agree with @MichaelBelousov:

> It's a slight preference though, all of the options I think have their very slight pros/cons.

I'm not sure `labelEx` is the right decision, but we're going to live with it for a while and see.

PR #356, opened 2025-06-09
Introduces CellStyle which can supply both CellOptions and Options to column functions

phatchman:
All dvui column functions now take "cell_style", which is an anytype.

This type must supply the following functions:
1) cellOptions(col: usize, row: usize) CellOptions
2) options(col: usize, row: usize) Options

Some default "CellStyle" types are supplied:
    - CellStyle - supplies the same options to all rows and cols
    - CellStyle.Banded - implements row banding
    - CellStyle.HoveredRow - highlights the row currently hovered by the mouse.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -4247,7 +4247,7 @@ fn gridLayouts() !void {
                     };
                 }
             }
-            _ = try dvui.gridColumnCheckbox(@src(), grid, Car, all_cars[0..], "selected", banded.overrideOptions(.{ .gravity_y = 0 }));
+            _ = try dvui.gridColumnCheckbox(@src(), grid, Car, all_cars[0..], "selected", banded.optionsOverride(.{ .gravity_y = 0 }));
         }
         // Make
         {
@@ -4390,13 +4390,13 @@ fn gridVirtualScrolling() !void {
     defer grid.deinit();
 
     // Each column has slightly different border requirements. Create separate options for each.
-    // Using the override avoids having to duplicate the event processing from .init for each column.
+    // Using the override avoids having to re-run the event processing from .init for each column.
     const highlight_hovered_1: GridWidget.GridOptionsHighlightHovered = .init(grid, &local.scroll_info, .{
         .border = .{ .x = 1, .w = 1, .h = 1 },
         .background = true,
         .color_fill_hover = .fill_hover,
     }, .{});
-    const highlight_hovered_2 = highlight_hovered_1.overrideCellOptions(.{ .border = .{ .w = 1, .h = 1 } });
+    const highlight_hovered_2 = highlight_hovered_1.cellOptionsOverride(.{ .border = .{ .w = 1, .h = 1 } });
     const scroller: dvui.GridWidget.VirtualScroller = .init(grid, .{ .total_rows = num_rows, .scroll_info = &local.scroll_info });
     const first = scroller.startRow();
     const last = scroller.endRow(); // Note that endRow is exclusive, meaning it can be used as a slice end index.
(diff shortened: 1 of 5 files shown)


phatchman:
This needs some more cooking, but looks like a promising direction. GridOptions probably isn't the right name, but its the best I have so far.

I like this because:
1)  most of the cells in a grid will have the same options. So a single GridOption type can be constructed and passed to all column functions
2) They are "inheritable". I could have made it more generic, but the Condition text colour options is an example of using the banded options as a base and applying the text options on top of it.
3) It simplifies the options handling in the column functions because they don't have to deal with callbacks.
4) Banding needs two lots of Cell Options one for normal and one for alt rows. This model caters well for different styling needing different types of options.
5) Allows the same options to be easily shared amoung dvui-provided and user-provided column functions. E.g. the custom description column uses exactly the same banded grid options as the other columns.

I don't like this because:
1) I'm a little concerned about the amount of copying of options that might happen. It's probably not a big deal
2) It's a little more difficult to do your own styling since you have to write a struct with two methods, rather than just writing a single callback function. But I think this is minimal.
3) The .init for the GridOptions. Maybe it is better to have the members set explicitly? Or as functions returning the underlying struct?

I want to see how far I can push this. It might be possible to have a GridOption that performs the hover highlighting, including handling the mouse position etc. It would be nice to encapsulate that sort of logic so that it can be re-used easily.

Edited to add: Just pushed a commit that implements the above. Seems to work well.
Edited2: I think it's mostly there now. Need to either document "anyopts" or put some validation to check that anyopts supports the GridOptions "interface".

david-vanderson:
Good progress!

phatchman:
I'm happy with this if you are. I'm still unsure about those init() functions as the banded one takes 3 separate lots of options, and it's not clear what is for what just looking at the calling code.  Without the init(), they could just set .alt_cell_opts and leave the others as defaults. But I also don't like some to have init() and others not. It's pretty minor either way I suppose.
```
    const banded = GridWidget.CellStyle.Banded.init(
        .rows,
        .{},
        .{ .color_fill = .{ .name = .fill_press }, .background = true },
        .{},
    );
```

I kinda like the organisation of this now with GridWidget.CellStyle being the default styling struct and then all of the variants coming under that namespace i.e. GridWidget.CellStyle.Banded, GridWidget.CellStyle.HoveredRow. 

I did play around with some alternatives to anytype as well as doing type checking on the passed in struct. But everything just seemed to make it more complicated for little benefit. I think the error message you get if you don't have the required functions is probably good enough by itself:
```
src\dvui.zig:4467:56: error: no field or member function named 'options' in 'widgets.GridWidget.CellStyle'
    const heading_opts = heading_defaults.override(opts.options(g.col_num, 0));
                                                   ~~~~^~~~~~~~
```
I believe there is also some opportunity to package up grid functionality in a similar way. e.g. being able to encapsulate the finding the hovered row so people could use it for multiple things, like selecting a row when clicked-on. But that's all a bit fuzzy in my mind and needs to stew a bit more. It will become more important for more advanced selection logic, e.g. with shift and ctrl-click to extend ranges, spacebar to select/unselect the current row, arrow key, pgup, pgdn navigation etc. It would be nice to be able to write that logic once (in composable pieces) and provide it as part of dvui.

Let me know if happy to merge as-is?

phatchman:
Thinking some more about the .init(). 

The alternative is to have the user set the members of the struct they want to provide e.g. .alt_cell_opts = . {... } and remove init() from everything. I can then just add a processEvents() to the row hover one. That also makes it more transparent that it is processing events, so you'd know to only call it once per grid per frame. Whereas if the event processing is buried in init() that is much less obvious.

I'm pretty sure I'll make that change as I have a general preference for passing things using names over using function parameter order. 

Edit: above implemented. The side-benefit of this is that the user can directly set highlighted_row if they have their own row calculation logic.

david-vanderson:
Merge when ready.  Great!

Issue #335, opened 2025-06-05
explore memory reduction ideas


david-vanderson:
Sparked by #322 this is a placeholder issue so I don't forget to explore this at some point.

Ideas:
* how much memory are we using vs. how much is reasonable?
* should make a memory benchmark

It seems like we should be able to do better, because most widgets (all?) follow a stack, so we should be able to pop them all off the arena:
* Each widget could have a ?Allocator field, and if not null, then it frees itself from that allocator in deinit()
  * should set the memory to 0xaa in debug mode to catch use-after-free (I know we have a few that will need to be fixed)

Maybe Window should have 2 (3?) arenas, often we can't immediately free what we've allocated because:
  * a function allocates some temporary work space then needs to allocate for the return value (so can no longer pop the temp space)
  * a widget allocates space for itself, but then in the middle has to allocate space for drawing stuff
  * unsure if we could use an arena for both deferred drawing stuff and random scratch space

RedPhoenixQ:
I've thought about this and tried some things. Having a "scratch space" arena and a "alive for whole frame" arena is definitely a good idea.

I have a branch reworking errors across dvui, I'll make a PR for it to discuss there.

`ShrinkingArenaAllocator` allows us to easily detect if we are not freeing the memory we allocate, perhaps the "scratch space" arena should check that it has 0 allocations at the end of the frame?

> * Each widget could have a ?Allocator field, and if not null, then it frees itself from that allocator in deinit()

Should this be on `WidgetData` perhaps, so that all widget get it? And `WidgetData` gets a `deinit` function taking in the widgets self ptr? If there is a dedicated arena for the "widget stack", then we could just check if the widgets `self` ptr is within the arena and free it like that, without having to store the allocator.

> * hould set the memory to 0xaa in debug mode to catch use-after-free (I know we have a few that will need to be fixed)

Wouldn't setting the ptr to `undefined` be caught in debug builds?

RedPhoenixQ:
Btw, my initial test didn't lower memory very much for the process. The arena, even when not freeing, was barely 1MB in size in the sdl3-app example with only +-5% arena memory usage for the split arena.

david-vanderson:
> perhaps the "scratch space" arena should check that it has 0 allocations at the end of the frame?

Unsure - I don't know how much user code uses that arena.  Probably a good idea for dvui debugging at least.

[...]

PR #359, opened 2025-06-09
Split arenas and add widget stack

RedPhoenixQ:
Relates to #335

Dvui applications should mostly be able to work on one large stack of data, reusing it as it goes since all widgets are immediate. Some reasons the arena used as this stack could not free its memory was because items with "frame lifetime" got allocated along side it.

Widgets enforce this stacking behaviour with `parentReset` already, which we can leverage for a separate widget stack were functions needing to return a pointer to a widget can use. `widgetAlloc` will allocate a widget on this "fixed" stack, falling back to the arena. All widgets should now call `widgetFree` in its `deinit` to remove itself from that stack, if it was allocated there. `widgetAlloc` keeps track of the total space used so the stack can be reallocated if it overflows.

We also want to be able to ignore or panic on OOM errors, which in turn means we should aim to reduce the number of points were allocation are performed. `ShrinkingArenaAllocator` have been changed to not free like the previous behaviour for this purpose. We can think about maybe only shrinking at some time/frame interval to not leak memory but still limit reallocations.

This also changes the `ShrinkingArenaAllocator` to check for alignment padding and still free items in those cases. It is unlikely that a value of < 8 bytes gets allocated and not freed correctly, so this change should be safe.

All of these changes reduces the total amount of allocations performed by the arenas/widget stack during a frame as well as reducing the peak total arenas/widget stack memory use of the app example from ~1800KB to ~950KB.

A followup to this will begin handling errors, probably by logging returning early or perhaps `catch @panic("OOM")` if needed.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -1019,8 +1019,8 @@ pub fn textEntryWidgets(demo_win_id: dvui.WidgetId) !void {
         }
     };
 
-    var font_entries: [][]const u8 = try dvui.currentWindow().arena().alloc([]const u8, dvui.currentWindow().font_bytes.count() + 1);
-    defer dvui.currentWindow().arena().free(font_entries);
+    var font_entries: [][]const u8 = try dvui.currentWindow().lifo().alloc([]const u8, dvui.currentWindow().font_bytes.count() + 1);
+    defer dvui.currentWindow().lifo().free(font_entries);
     {
         font_entries[0] = "Theme Body";
         var it = dvui.currentWindow().font_bytes.keyIterator();
@@ -1139,7 +1139,7 @@ pub fn textEntryWidgets(demo_win_id: dvui.WidgetId) !void {
                 .{},
                 .{ .expand = .ratio, .gravity_x = 1.0 },
             )) {
-                new_filename = try dvui.dialogNativeFileOpen(dvui.currentWindow().long_term_arena(), .{ .title = "Pick Font File" });
+                new_filename = try dvui.dialogNativeFileOpen(dvui.currentWindow().arena(), .{ .title = "Pick Font File" });
             }
 
             try dvui.label(@src(), "File", .{}, .{ .gravity_y = 0.5 });
@@ -1164,22 +1164,22 @@ pub fn textEntryWidgets(demo_win_id: dvui.WidgetId) !void {
                     try dvui.toast(@src(), .{ .subwindow_id = demo_win_id, .message = "Add a Name" });
                     name_error.* = true;
                 } else if (dvui.currentWindow().font_bytes.contains(name)) {
-                    const msg = try std.fmt.allocPrint(dvui.currentWindow().arena(), "Already have font named \"{s}\"", .{name});
-                    defer dvui.currentWindow().arena().free(msg);
+                    const msg = try std.fmt.allocPrint(dvui.currentWindow().lifo(), "Already have font named \"{s}\"", .{name});
+                    defer dvui.currentWindow().lifo().free(msg);
                     try dvui.toast(@src(), .{ .subwindow_id = demo_win_id, .message = msg });
                     name_error.* = true;
                 } else {
                     var bytes: ?[]u8 = null;
                     if (!std.fs.path.isAbsolute(filename)) {
                         file_error.* = true;
-                        const msg = try std.fmt.allocPrint(dvui.currentWindow().arena(), "Could not open \"{s}\"", .{filename});
-                        defer dvui.currentWindow().arena().free(msg);
+                        const msg = try std.fmt.allocPrint(dvui.currentWindow().lifo(), "Could not open \"{s}\"", .{filename});
+                        defer dvui.currentWindow().lifo().free(msg);
                         try dvui.dialog(@src(), .{}, .{ .title = "File Error", .message = msg });
                     } else {
                         const file = std.fs.openFileAbsolute(filename, .{}) catch blk: {
                             file_error.* = true;
-                            const msg = try std.fmt.allocPrint(dvui.currentWindow().arena(), "Could not open \"{s}\"", .{filename});
-                            defer dvui.currentWindow().arena().free(msg);
+                            const msg = try std.fmt.allocPrint(dvui.currentWindow().lifo(), "Could not open \"{s}\"", .{filename});
+                            defer dvui.currentWindow().lifo().free(msg);
                             try dvui.dialog(@src(), .{}, .{ .title = "File Error", .message = msg });
                             break :blk null;
                         };
@@ -1193,15 +1193,15 @@ pub fn textEntryWidgets(demo_win_id: dvui.WidgetId) !void {
                             error.OutOfMemory => @panic("OOM"),
                             error.fontError => {
                                 dvui.currentWindow().gpa.free(b);
-                                const msg = try std.fmt.allocPrint(dvui.currentWindow().arena(), "\"{s}\" is not a valid font", .{filename});
-                                defer dvui.currentWindow().arena().free(msg);
+                                const msg = try std.fmt.allocPrint(dvui.currentWindow().lifo(), "\"{s}\" is not a valid font", .{filename});
+                                defer dvui.currentWindow().lifo().free(msg);
                                 try dvui.dialog(@src(), .{}, .{ .title = "Bad Font", .message = msg });
                                 break :blk;
                             },
                         };
 
-                        const msg = try std.fmt.allocPrint(dvui.currentWindow().arena(), "Added font named \"{s}\"", .{name});
-                        defer dvui.currentWindow().arena().free(msg);
+                        const msg = try std.fmt.allocPrint(dvui.currentWindow().lifo(), "Added font named \"{s}\"", .{name});
+                        defer dvui.currentWindow().lifo().free(msg);
                         try dvui.toast(@src(), .{ .subwindow_id = demo_win_id, .message = msg });
                     }
                 }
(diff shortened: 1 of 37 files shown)


david-vanderson:
I like this direction.  Here are some immediate thoughts/comments:
* Did you find out that doing an alloc/free on an arena doesn't necessarily work due to alignment padding?  What's going on there?
* I don't get how widget_stack having a fixedbuffer allocator is better than ShrinkingArena.  What am I missing?
* Do arenas have multiple non contiguous memory allocations?  If so that would break the `widgetFree` logic right?
* Anything allocated by a widget that has the same lifetime as the widget could go on the same stack right?
* "long_term_arena" should probably be named something that indicates the lifetime, like "frame_arena".  Or maybe we keep "arena" meaning "for the frame" and use some other word to mean "should be freed", like "stack_arena", or "widget_arena".  Unsure here.
* The widgetFree in ScrollContainerWidget is commented out?

RedPhoenixQ:
> Did you find out that doing an alloc/free on an arena doesn't necessarily work due to alignment padding? What's going on there?

Yes. The Arena and Fix buffer allocator both just check that their end index is exactly equal to the end of the ptr + the size of the allocation. The only scenario in which this would cause problems is if you allocate something < 8 bytes after an allocation that ended on an unaligned index, and then freed the previous item before the < 8 bytes one. Optionally we could always align the end index to 8 bytes and change free to always align the end to 8 as well.

> I don't get how widget_stack having a fixedbuffer allocator is better than ShrinkingArena. What am I missing?

FixedBufferAllocator has ownPtr method which is useful in determining if the ptr should be freed at all. A widget allocation would always be a point where we'd have to panic, so having a stack that is guaranteed to fit the max depth makes that even more safe. There is also an issue of widgets containing other widgets which would trigger widgetFree and free only part of the allocation, which would then prevent the full widget from being freed. Using a fixed buffer allows us to easily override a simpler free logic that accounts for this.

> Do arenas have multiple non contiguous memory allocations? If so that would break the widgetFree logic right?

Yes they do. As mentioned above, this is one of the reasons why we use a fixed buffer allocator.

> Anything allocated by a widget that has the same lifetime as the widget could go on the same stack right?

Yes, but as mentioned this is more to guarantee widget ptrs. Other allocations a widget does, like Labels printing or TextArea copying/selecting, would bloat the widget stack and there are often reasonable fallbacks that can be used if the arena allocator would fail. Everything else should be a field in the widget.

> "long_term_arena" should probably be named something that indicates the lifetime, like "frame_arena". Or maybe we keep "arena" meaning "for the frame" and use some other word to mean "should be freed", like "stack_arena", or "widget_arena". Unsure here.

I agree the names are hard. I had the arena function renamed to "temp" for a while, would "temp" or "temp_alloc" be better? That might better convey that the allocations should be freed normally as any other allocator. The "long_term_arena" is probably better named "frame_arena" or "arena" if the other one is renamed "temp" or "stack" arena.

> The widgetFree in ScrollContainerWidget is commented out?

This is some testing left behind when investigating the widgets in widgets I mentioned. Scroll container is the field of ScrollArea if I remember correctly.

david-vanderson:
> The only scenario in which this would cause problems is if you allocate something < 8 bytes after an allocation that ended on an unaligned index, and then freed the previous item before the < 8 bytes one. Optionally we could always align the end index to 8 bytes and change free to always align the end to 8 as well.

I'm still not getting it sorry.  This example has the stack getting freed out of order right?  Doesn't that always cause a problem?

> There is also an issue of widgets containing other widgets which would trigger widgetFree and free only part of the allocation, which would then prevent the full widget from being freed. Using a fixed buffer allows us to easily override a simpler free logic that accounts for this.

Does this mean we have to be very careful about which widgets get freed?  (Like only free a "containing" widget, not the "contained" in a compound widget?)

The fixed buffer thing seems overkill for this, and I'm worried that we won't normally run situations that trigger the edge cases.  I get the desire to ensure that widget allocation won't fail, but we can't guarantee that anyway.

I'm also concerned that it's tying widget allocation into dvui internals too much.  We could imagine a program wanting to allocate their widgets from a freelist.

Are we getting anything from using a fixed buffer (vs arena) beyond being able to check the pointer for freeing?

RedPhoenixQ:
> This example has the stack getting freed out of order right? Doesn't that always cause a problem?

Yeah you're absolutely right. I was just explaining why it's sane to do, as out of order frees is required to cause problems.

> Does this mean we have to be very careful about which widgets get freed? (Like only free a "containing" widget, not the "contained" in a compound widget?)

No, because the special handling in widgetFree that always frees, even if a "contained" widget is freed before. This would not be as easily done with an ArenaAllocator as it would have to rely on specific offsets in its buffer list. 

> I'm also concerned that it's tying widget allocation into dvui internals too much. We could imagine a program wanting to allocate their widgets from a freelist.

The widgetAlloc and widgetFree does not have to be used by anyone depending on dvui. The helper functions, like dvui.box, already have to allocate widgets within dvui, this just changes how those internal allocations are done. If you want to allocate your own widgets you couldn't use the helper functions anyway. widgetFree called on any other ptr, including from the normal stack, is a no-op so it's safe for us to add in the dvui provided widgets.

david-vanderson:
> Yeah you're absolutely right. I was just explaining why it's sane to do, as out of order frees is required to cause problems.

Argh - something about this is not going into my head.  Maybe I am assuming that all widget stack allocations are freed in the correct last-in-first-out order?  Is that not assumed here?  Can you give a step-by-step example of allocations/frees that show how to get into the bad situation that this alignment stuff fixes?

It's possible we are imagining slightly different goals.  In my head the primary benefit of splitting into multiple arenas is to lower peak memory usage, primarily by ensuring that the things we allocate in stack fashion actually work that way.  Is there something else you are trying to do here?

I was imagining broadly 2 changes - a separate arena specifically for last-in-first-out allocations, plus some debugging stuff to catch when we violate last-in-first-out.

The out of order frees stuff makes me think you are approaching it differently.  Are you trying to ensure freeing works in the face of out of order frees?

RedPhoenixQ:
I'm sorry. Last in first out is definitely assumed. I just mentioned the error case where it's not upheld. The widget stack and temporary arena both assume normal fifo behaviour.


As for the alignment example:
- you allocate a 30 byte widget, then end is at 30
- allocate another 30 byte widget aligned to 8, end index is at 62 (30 + 2 bytes alignment padding + 30 for the new widget)
- free the top widget, where end index is = ptr + 30, end index is 32
- free first widget, end index is at 32 and not 30, so the free fails

david-vanderson:
Ahhh thank you.  That really seems like a bug in zig's arena allocator right?  I don't know the right fix (either store extra info with each allocation, or like you said force alignment to the max for each).  Should we file a PR for it?

RedPhoenixQ:
> Should we file a PR for it?

I don't think so. Zigs arenas don't guarantee freeing will work past the last allocation, which is fine. We can support it ourselves in ShrinkingArena. 

Maybe always aligning to 8 bytes is a more performant/less error prone at the cost of slightly more memory usage.

The ShrinkingArena could store debug info similar to the DebugAllocator, saving the ret_addr to give stack traces of memory that wasn't freed correctly. That would be quite a bit of code though. We can still log that fifo was violated without a trace with only the current information we have.

david-vanderson:
> I don't think so. Zigs arenas don't guarantee freeing will work past the last allocation, which is fine.

This is different right?  You alignment example says that even when always freeing the last allocation, sometimes zig's arena allocator will fail to free it?

I'm thinking we could have logic in `ShrinkingArena.free` that logs an error if the free didn't succeed (meaning it wasn't the last allocation).  I think that would be enough debug information to catch when we screw up?

RedPhoenixQ:
> You alignment example says that even when always freeing the last allocation, sometimes zig's arena allocator will fail to free it?

The very last allocation always succeeds, but 2 in a row is not guaranteed.

david-vanderson:
> The very last allocation always succeeds, but 2 in a row is not guaranteed.

Just to be clear, the last allocation always frees successfully.  But if you make multiple allocations of different alignments, then free them in reverse order, you are only guaranteed that the last will free successfully.  Is that correct?

RedPhoenixQ:
> Just to be clear, the last allocation always frees successfully. But if you make multiple allocations of different alignments, then free them in reverse order, you are only guaranteed that the last will free successfully. Is that correct?

Yes, spot on. Sorry again about the confusion

RedPhoenixQ:
There is now a FIFO version of the `std.mem.Allocator` for `ShrinkingArenaAllocator` that logs an message and stack trace to where a free failed, giving you a good starting point for figuring out what allocation/free is incorrectly ordered. (if there is a lot of  failed allocations, printing the stack trace can cause significant performance issues, especially if the framerate is uncapped. Maybe the stack trace should be conditionally by a compilation flag as a debugging tool?)

The allocators have been renamed to `arena` for the frame lifetime arena, keeping the previous behaviour with the major advantage that existing code using dvui doesn't have to change or implement perfect FIFO memory management. 

The FIFO allocator has been named `temp` as anything `..._arena` isn't quite correct as we expect you to free the memory, which isn't true for a normal arena. Other options I considered was:
- `temp_alloc` (makes it clear this is an allocator)
- `fifo_alloc` (specifies the behaviour of the allocator without needing to read docs)
  - optionally just `fifo` 
- `allocator` (this could be either very clear, as in use this for most allocations, or unclear because it is confused with the common allocator interface and it doesn't make much sense for a `Window` to be an allocator)

david-vanderson:
Excellent!  I wouldn't worry about too much debug printing, it'll be turned off in Release modes right?

> The FIFO allocator has been named `temp` as anything `..._arena` isn't quite correct as we expect you to free the memory, which isn't true for a normal arena. Other options I considered was:

I like `lifo` the best, or `filo`, but unfortunately `fifo` is not correct, the first thing freed (out) is the last thing added (in), right?

RedPhoenixQ:
Oh my bad, got fifo/lifo mixed up. Will rename and merge this

RedPhoenixQ:
> debug printing, it'll be turned off in Release modes right?

It's currently a `dvui.log.debug`, so it would be turned of if the log level for the program or the `.dvui` scope was set higher than debug.

david-vanderson:
Thanks very much for going through this all with me!

Issue #358, opened 2025-06-09
Feature: option to load png and svg icons (xdg)


vlappa:
I list applications in my dvui app and I would like to also display the icons of those applications. Only applications with a specific key in the xdg *.desktop files will be listed. In those *.desktop files is also a key for the name of the icon of the particular application, which then should be loaded in my dvui app. 

There is a xdg specification for it: 

https://wiki.archlinux.org/title/Desktop_entries#Icons

On Linux most of them seems to be png or svg (or xpm, but deprecated).

RedPhoenixQ:
For raster images, like pngs, you can use `dvui.image` or look at its implementation. See [stb_image](https://github.com/david-vanderson/dvui/blob/05bdd090df98e8e91b852d85216c21c6d47c3cdc/src/stb/stb_image.h#L23C1-L33C36) for all supported formats.

As for svgs, the parts to do this already exits in the `svg2tvg` dependency they're just not exposed. A function like `dvui.iconFromSvg` giving you the bytes to pass to `dvui.icon` should perhaps be added?

david-vanderson:
Thanks for the issue!  I agree with @RedPhoenixQ.  I don't know how useful having something at the dvui level that tries to combine the two (like trying to use a png image as a scalable icon) would be.

@vlappa Can you check some of the applications and see whether they mostly reference svgs or pngs?  I think using pngs for icons is older, and maybe most of them are using svgs now?

vlappa:
> [@vlappa](https://github.com/vlappa) Can you check some of the applications and see whether they mostly reference svgs or pngs? I think using pngs for icons is older, and maybe most of them are using svgs now?

Most of them offer both it seems. Some only png I think.

RedPhoenixQ:
@vlappa #360 should now allow for most svgs to be converted to tvg. Not every svg will work because the underlying library doesn't support 100% of svgs features, but most should.

vlappa:
Thanks!

nat3Github:
> [@vlappa](https://github.com/vlappa) [#360](https://github.com/david-vanderson/dvui/pull/360) should now allow for most svgs to be converted to tvg. Not every svg will work because the underlying library doesn't support 100% of svgs features, but most should.

yes! and if some svg has conversion errors your welcome to drop an issue or pr at [svg2tvg](https://github.com/nat3Github/zig-lib-svg2tvg) and i'll look into it in my free time!

david-vanderson:
[...]

PR #360, opened 2025-06-09
Expose svg2tvg.tvg_from_svg in dvui

RedPhoenixQ:
Relates to #358

This allows applications to make use of this functionality to import svgs without extra dependencies.

Current width/height handling in zig-lib-svg2tvg doesn't allow for the zig logo to be loaded, so this PR depends on  https://github.com/nat3Github/zig-lib-svg2tvg/pull/1

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -39,8 +39,8 @@
         },
         // for rendering tvg icons
         .svg2tvg = .{
-            .url = "git+https://github.com/nat3Github/zig-lib-svg2tvg#628e32e87e7a2f57493a4af1b70870877e461358",
-            .hash = "svg2tvg-0.0.0-HpMZuTN2BgAWaLEMONTjaMrPki2_9x6KJZtO4W7NNHh2",
+            .url = "git+https://github.com/nat3Github/zig-lib-svg2tvg#0ff7a68c3d3999d6273d9058ef206a58d49147d6",
+            .hash = "svg2tvg-0.0.0-HpMZuVl2BgAZMTmi-VfBOxDxauBRRSvx936sSzlU099q",
         },
     },
 }


nat3Github:
this is a cool idea! ...but it will somewhat require svg2tvg to correctly handle the svg format in its entirety, it currently does not! and it will be hard to get it to this level. one edge case for example are the different line-join attributes that are currently unsupported. for simple icon stuff maybe its acceptable

david-vanderson:
> but it will somewhat require svg2tvg to correctly handle the svg format in its entirety,

This is not required.  The intent behind this is to dynamically load an svg icon at runtime, so we can support the same kinds of icons we would otherwise.

RedPhoenixQ:
Web build is broken because of debug.print at https://github.com/nat3Github/zig-lib-svg2tvg/blob/bc20ee70a70443ae557667c2f390cad381667e96/src/conversion.zig#L932. Will have to fix that before merging

nat3Github:
> Web build is broken because of debug.print at https://github.com/nat3Github/zig-lib-svg2tvg/blob/bc20ee70a70443ae557667c2f390cad381667e96/src/conversion.zig#L932. Will have to fix that before merging

the debug.print is not important we could either hide it behind a flag or remove it.

RedPhoenixQ:
Replacing `std.debug.print` with `std.log.debug` would be enough if you don't want to make larger changes now.

nat3Github:
> Replacing `std.debug.print` with `std.log.debug` would be enough if you don't want to make larger changes now.

done its in main, you can now update the svg2tvg dependency.

RedPhoenixQ:
Thank you, will do!

david-vanderson:
Looking pretty good!  For me the svg icon renders as black - is that just me?  If not, do we know why?

RedPhoenixQ:
> For me the svg icon renders as black - is that just me? If not, do we know why?

I does for me too but it should right? Because it defaults to the text color like icons. #340 might change that.

david-vanderson:
Yep, sorry, you are right.  Merge when ready.  Thanks!

PR #365, opened 2025-06-11
Lifo free account for the alignment of memory pointer

### LINKED ISSUE #364: lifo errors when using for Path.Builder
@vlappa hit this when running a labelClick in a loop like:

```zig
    for (0..10) |i| {
        _ = try dvui.labelClick(@src(), "{d}", .{i}, .{ .id_extra = i, .gravity_x = 0.0, .expand = .horizontal, .padding = dvui.Rect.all(1.7), .background = true, .color_fill = .blue });
    }
```

I was able to reproduce it by putting that loop into sdl-standalone.zig.  I pushed a commit to use `arena()` for Path Builder until we figure this out.

I'm not sure, but my suspicion is that ArrayList is doing a resize and that is causing a new allocation maybe?

@RedPhoenixQ any ideas?

RedPhoenixQ:
Fixes #364

This was an oversight as the pointer of the memory doesn't necessarily have to be aligned to 8, like in `[]const u8` buffers were alignment is 1.

A note about lifo errors, the trace back given does not indicate that the code there as a fault. It indicates that somewhere between its allocation and free call, there is faulty handling.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -1514,7 +1514,7 @@ pub fn styling() !void {
             defer drawBox.deinit();
             const rs = drawBox.data().contentRectScale();
 
-            var path: dvui.Path.Builder = .init(dvui.currentWindow().arena());
+            var path: dvui.Path.Builder = .init(dvui.currentWindow().lifo());
             defer path.deinit();
             try path.addRect(rs.r, dvui.Rect.Physical.all(5));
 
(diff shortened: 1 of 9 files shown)


david-vanderson:
Ah thank you!

I got confused with the LabelWidget, and I think now I see that since we never call widgetAlloc on it, the widgetFree does nothing?

RedPhoenixQ:
> I think now I see that since we never call widgetAlloc on it, the widgetFree does nothing?

They use a separate stack so it doesn't matter in this case. *If* they were on the same stack, widgetAlloc would have been called first, so deferring it at the to of `deinit` is still the correct order.

Maybe all the `sefl.* = undefined` for widgets should be moved into `widgetFree`? Less lines of code and potential to miss one or the other?

david-vanderson:
> Maybe all the `sefl.* = undefined` for widgets should be moved into `widgetFree`? Less lines of code and potential to miss one or the other?

Unsure, let's keep it the way it is for now.

PR #369, opened 2025-06-12
Create stack allocator for lifo allocations

RedPhoenixQ:
This is an modified arena allocator with support for full lifo frees. It's not thoroughly tested but the app example works without issue.

The new allocator does not handle resizing very well because we need to keep the size the same to allow for free to know what to free. This means this allocator currently isn't very useful for use with dynamic arrays and should only be used to create a static allocation and then free it in lifo order. So unfortunatly this will not easily be able to replace the ShrinkingArena for general allocations yet.

This allocator successfully catches the error in 8bf8717b1a3b8a8ec5fa6d5c1193e8efd36bbeb5 by causing a segfault on the use after free.

There is however some issue when running tests. When I launch sdl3-app there is no issue but during tests there is something wrong. I'll figure it out tomorrow, just wanted to share.

<details>
<summary>Example output when frees are incorrectly ordered</summary>

The test looks like:
```zig
test "out of order frees" {
    var instance = StackAllocator.init(std.testing.allocator);
    defer instance.deinit();
    const alloc = instance.allocator();
    log.warn("IGNORE ERRORS PRINTED FOR THIS TEST, THEY ARE EXPECTED", .{});

    {
        const a1 = try alloc.alloc(usize, 1);
        a1[0] = 0xdeadbeaf;
        defer alloc.free(a1);

        try std.testing.expectEqual(@sizeOf(usize) * 1, instance.current().?.data.end_index);
        const a2 = try alloc.alloc(usize, 4);
        try std.testing.expectEqual(@sizeOf(usize) * 5, instance.current().?.data.end_index);
        const a3 = try alloc.alloc(usize, 5);
        try std.testing.expectEqual(@sizeOf(usize) * 10, instance.current().?.data.end_index);
        alloc.free(a2);
        try std.testing.expectEqual(@sizeOf(usize) * 1, instance.current().?.data.end_index);
        alloc.free(a3);

        // The first value is untouched even though the earlier values failed
        try std.testing.expectEqual(0xdeadbeaf, a1[0]);
        try std.testing.expectEqual(@sizeOf(usize) * 1, instance.current().?.data.end_index);
    }
    try std.testing.expectEqual(0, instance.current().?.data.end_index);

    {
        const a1 = try alloc.alloc(usize, 1);
        a1[0] = 0xdeadbeaf;
        defer alloc.free(a1);

        const a2 = try alloc.alloc(usize, 4);
        try std.testing.expectEqual(@sizeOf(usize) * 5, instance.current().?.data.end_index);
        // Force a second buffer
        _ = try alloc.alloc(usize, 100);
        try std.testing.expectEqual(2, instance.buffer_list.len());
        try std.testing.expectEqual(@sizeOf(usize) * 100, instance.current().?.data.end_index);
        alloc.free(a2);

        // We have reset back to the first buffer
        try std.testing.expect(instance.current() != instance.buffer_list.first);

        // The first value is untouched even though the earlier values failed
        try std.testing.expectEqual(0xdeadbeaf, a1[0]);
        try std.testing.expectEqual(@sizeOf(usize) * 1, instance.current().?.data.end_index);
    }
}
```

```bash
[stack_allocator] (warn): IGNORE ERRORS PRINTED FOR THIS TEST, THEY ARE EXPECTED
[stack_allocator] (warn): The following item as freed when not at the top of the stack: 
...\dvui\src\stack_allocator.zig:498:19: 0x7ff77aafa2dd in test.out of order frees (test.exe.obj)
        alloc.free(a2);
                  ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:214:25: 0x7ff77ab170ac in mainTerminal (test.exe.obj)
        if (test_fn.func()) |_| {
                        ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:62:28: 0x7ff77ab12538 in main (test.exe.obj)
        return mainTerminal();
                           ^



These items should be freed in the following order:
Item 0:
...\dvui\src\stack_allocator.zig:496:35: 0x7ff77aafa18e in test.out of order frees (test.exe.obj)
        const a3 = try alloc.alloc(usize, 5);
                                  ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:214:25: 0x7ff77ab170ac in mainTerminal (test.exe.obj)
        if (test_fn.func()) |_| {
                        ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:62:28: 0x7ff77ab12538 in main (test.exe.obj)
        return mainTerminal();
                           ^


Item 1:
...\dvui\src\stack_allocator.zig:494:35: 0x7ff77aafa02e in test.out of order frees (test.exe.obj)
        const a2 = try alloc.alloc(usize, 4);
                                  ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:214:25: 0x7ff77ab170ac in mainTerminal (test.exe.obj)
        if (test_fn.func()) |_| {
                        ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:62:28: 0x7ff77ab12538 in main (test.exe.obj)
        return mainTerminal();
                           ^


[stack_allocator] (warn): An item lower in the stack has already been freed!

...\dvui\src\stack_allocator.zig:500:19: 0x7ff77aafa3a6 in test.out of order frees (test.exe.obj)
        alloc.free(a3);
                  ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:214:25: 0x7ff77ab170ac in mainTerminal (test.exe.obj)
        if (test_fn.func()) |_| {
                        ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:62:28: 0x7ff77ab12538 in main (test.exe.obj)
        return mainTerminal();
                           ^

[stack_allocator] (warn): An item lower in the stack has already been freed!

...\dvui\src\stack_allocator.zig:519:19: 0x7ff77aafa99c in test.out of order frees (test.exe.obj)
        alloc.free(a2);
                  ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:214:25: 0x7ff77ab170ac in mainTerminal (test.exe.obj)
        if (test_fn.func()) |_| {
                        ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:62:28: 0x7ff77ab12538 in main (test.exe.obj)
        return mainTerminal();
                           ^

[stack_allocator] (warn): The following item as freed when not at the top of the stack:
...\dvui\src\stack_allocator.zig:519:19: 0x7ff77aafa99c in test.out of order frees (test.exe.obj)
        alloc.free(a2);
                  ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:214:25: 0x7ff77ab170ac in mainTerminal (test.exe.obj)
        if (test_fn.func()) |_| {
                        ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:62:28: 0x7ff77ab12538 in main (test.exe.obj)
        return mainTerminal();
                           ^



These items should be freed in the following order:
Item 0:
...\dvui\src\stack_allocator.zig:516:28: 0x7ff77aafa7f2 in test.out of order frees (test.exe.obj)
        _ = try alloc.alloc(usize, 100);
                           ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:214:25: 0x7ff77ab170ac in mainTerminal (test.exe.obj)
        if (test_fn.func()) |_| {
                        ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:62:28: 0x7ff77ab12538 in main (test.exe.obj)
        return mainTerminal();
                           ^


Item 1:
...\dvui\src\stack_allocator.zig:513:35: 0x7ff77aafa692 in test.out of order frees (test.exe.obj)
        const a2 = try alloc.alloc(usize, 4);
                                  ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:214:25: 0x7ff77ab170ac in mainTerminal (test.exe.obj)
        if (test_fn.func()) |_| {
                        ^
...\zig\x86_64-windows-0.14.1\lib\compiler\test_runner.zig:62:28: 0x7ff77ab12538 in main (test.exe.obj)
        return mainTerminal();
  
```

</details>

### DIFF
--- a/src/Window.zig
+++ b/src/Window.zig
@@ -106,7 +106,7 @@ gpa: std.mem.Allocator,
 _arena: dvui.ShrinkingArenaAllocator,
 _lifo_arena: dvui.ShrinkingArenaAllocator,
 /// Used to allocate widgets with a fixed location
-_widget_stack: std.heap.FixedBufferAllocator,
+_widget_stack: dvui.StackAllocator,
 /// The peak amount of bytes used by the widget stack since
 /// the creation of `Window`. Used to resize the stack ifs needed.
 peak_widget_stack: usize = 0,
@@ -190,7 +190,7 @@ pub fn init(
         .gpa = gpa,
         ._arena = init_opts.arena orelse .init(gpa),
         ._lifo_arena = .init(gpa),
-        ._widget_stack = .init(try gpa.alloc(u8, init_opts.default_widget_stack_capacity)),
+        ._widget_stack = .init(gpa),
         .subwindows = .init(gpa),
         .tab_index_prev = .init(gpa),
         .tab_index = .init(gpa),
@@ -417,7 +417,7 @@ pub fn deinit(self: *Self) void {
     self.keybinds.deinit();
     self._arena.deinit();
     self._lifo_arena.deinit();
-    self.gpa.free(self._widget_stack.buffer);
+    self._widget_stack.deinit();
 
     {
         var it = self.font_bytes.valueIterator();
@@ -454,7 +454,7 @@ pub fn deinit(self: *Self) void {
 /// For allocations that should live for the entire frame, see
 /// `Window.arena`
 pub fn lifo(self: *Self) std.mem.Allocator {
-    return self._lifo_arena.allocatorLIFO();
+    return self._lifo_arena.allocator();
 }
 
 /// A general allocator for using during a frame. All allocations
@@ -1706,31 +1706,32 @@ pub fn end(self: *Self, opts: endOptions) !?u32 {
 
     // self._arena.debug_log();
     _ = self._arena.reset(.retain_capacity);
-    if (self._lifo_arena.current_usage != 0 and !self._lifo_arena.has_expanded()) {
-        log.warn("Arena was not empty at the end of the frame, {d} byte left. Did you forget to free memory somewhere?", .{self._lifo_arena.current_usage});
-        // const buf: [*]u8 = @ptrCast(self._lifo_arena.arena.state.buffer_list.first.?);
-        // std.log.debug("Arena content {s}", .{buf[@sizeOf(usize)..self._lifo_arena.current_usage]});
-    }
+    // if (self._lifo_arena.current_usage != 0 and !self._lifo_arena.has_expanded()) {
+    //     log.warn("Arena was not empty at the end of the frame, {d} byte left. Did you forget to free memory somewhere?", .{self._lifo_arena.current_usage});
+    //     // const buf: [*]u8 = @ptrCast(self._lifo_arena.arena.state.buffer_list.first.?);
+    //     // std.log.debug("Arena content {s}", .{buf[@sizeOf(usize)..self._lifo_arena.current_usage]});
+    // }
     // self._lifo_arena.debug_log();
     _ = self._lifo_arena.reset(.retain_capacity);
 
     // Widget stack
-    if (self._widget_stack.end_index != 0) {
-        log.warn("Widget stack was not empty at the end of the frame. Did you forget to call deinti?", .{});
-        self._widget_stack.reset();
-    }
-    if (self._widget_stack.buffer.len < self.peak_widget_stack) {
-        log.warn("Widget stack overflowed, consider increasing the default widget stack size", .{});
-        const new_len = self.peak_widget_stack + 0x400;
-        if (self.gpa.resize(self._widget_stack.buffer, new_len)) {
-            self._widget_stack.buffer.len = new_len;
-        } else {
-            // do realloc ourselves as we don't need to copy any memory from the old allocation
-            const new_buf = try self.gpa.alloc(u8, new_len);
-            self.gpa.free(self._widget_stack.buffer);
-            self._widget_stack.buffer = new_buf;
-        }
-    }
+    // if (self._widget_stack.end_index != 0) {
+    //     log.warn("Widget stack was not empty at the end of the frame. Did you forget to call deinti?", .{});
+    //     self._widget_stack.reset();
+    // }
+    _ = self._widget_stack.reset(.retain_capacity);
+    // if (self._widget_stack.buffer.len < self.peak_widget_stack) {
+    //     log.warn("Widget stack overflowed, consider increasing the default widget stack size", .{});
+    //     const new_len = self.peak_widget_stack + 0x400;
+    //     if (self.gpa.resize(self._widget_stack.buffer, new_len)) {
+    //         self._widget_stack.buffer.len = new_len;
+    //     } else {
+    //         // do realloc ourselves as we don't need to copy any memory from the old allocation
+    //         const new_buf = try self.gpa.alloc(u8, new_len);
+    //         self.gpa.free(self._widget_stack.buffer);
+    //         self._widget_stack.buffer = new_buf;
+    //     }
+    // }
 
     try self.initEvents();
 
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -92,6 +92,7 @@ pub const enums = @import("enums.zig");
 pub const easing = @import("easing.zig");
 pub const testing = @import("testing.zig");
 pub const ShrinkingArenaAllocator = @import("shrinking_arena_allocator.zig");
+pub const StackAllocator = @import("stack_allocator.zig").StackAllocator;
 pub const TrackingAutoHashMap = @import("tracking_hash_map.zig").TrackingAutoHashMap;
 
 pub const wasm = (builtin.target.cpu.arch == .wasm32 or builtin.target.cpu.arch == .wasm64);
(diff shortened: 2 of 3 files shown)


david-vanderson:
This is technically impressive.  However I'm not sure of my abilities to properly review such complex code in a reasonable amount of time.  I'm worried that any bug will take a lot of effort to track down, and unsure if that is worth the memory savings.  I don't know what we should do here.

Is there a way we could reduce the complexity of the implementation by giving up something?

Maybe we could give up the extra debugging info, and only report the stack trace of the free that didn't work?  Maybe add a logging mode so that if a bad free is reproducible you can rerun with a log of all the allocs/frees?

Could we make the alignment stuff easier by always 8-byte aligning everything?  Or having 4 separate underlying arena allocators, one for each alignment?

@RedPhoenixQ From all the effort you've put into this, we've learned a great deal about the complexities of implementing a "simple" lifo-enforcing stack allocator.  Are you also concerned by the complexity/benefit tradeoff here?  Is the memory savings still expected to be around 1MB?

RedPhoenixQ:
This was an fun exercise and I'm glad to have made something work. If it's not merged here I'll pull it out into a separate zig project, no problem. 

> I'm worried that any bug will take a lot of effort to track down,

That is very fair and I full respect you not wanting dvui to take on the maintenance of something like this. I had some ideas about allocators that I though would work and made an attempt at it. It is completely okay to not accept this PR (in that case maybe look into logging widget stack allocations like you suggested?). 

> Is the memory savings still expected to be around 1MB?

Yes, for the example app. The impact is greater for larger programs as there is more memory reuse. If I turn of caching for the examples, idle memory (when not generating frames) goes from ~47MB to ~44.5MB and at peak when moving the mouse from 54.5MB to ~51MB (with sdl3-app on my windows 11 machine).

This allocator is more about the debugging capabilities on top of reusing memory. As I mentioned above, not wanting to take on maintenance for this is completely fine if another solution would suffice for debugging.

> Could we make the alignment stuff easier by always 8-byte aligning everything? Or having 4 separate underlying arena allocators, one for each alignment?

Yes but that would limit us to 8 byte aligned types, which might be okay. The current solution relies on the fact that it's a stack, so the padding from the alignment must have been added from the item directly above. We save the alignment of the last freed item and use that to know how much padding must have been added.

> Maybe we could give up the extra debugging info, and only report the stack trace of the free that didn't work? Maybe add a logging mode so that if a bad free is reproducible you can rerun with a log of all the allocs/frees?

Then we would give up on the ability to see the correct order of the allocations in between. This is a valid solution, but then I'd rather not have that be an allocator and just add it to the widget allocation functions along with the logging mode as that is simpler.

RedPhoenixQ:
Also, other note. This is a very separated part of dvui, so there is little risk of merge conflicts or similar. This does not have to happen now if we want to explore other alternatives in the mean time.

david-vanderson:
> Also, other note. This is a very separated part of dvui, so there is little risk of merge conflicts or similar. This does not have to happen now if we want to explore other alternatives in the mean time.

Very good point.  Also spinning it into a separate library is a good idea as well.

This is super useful information.  What I'd like to do is spend some time exploring some alternatives that might not pan out, while keeping this code to compare against.  Might take me a while - are you okay with me keeping this PR open for a while?

RedPhoenixQ:
> Might take me a while - are you okay with me keeping this PR open for a while?

Absolutely! Take your time, this is not a high priority issue. I'm also happy to do some improvement to the current `widgetAlloc` and `widgetFree` as exploration.

nat3Github:
this is quite impressive and cool i actually thought about lifo arenas recently!
as for optimizing memory usage this is still a pretty coarse method as it requires a certain order of alloc and frees.
naturally in a program you will have some mem thats used temporarily while other mem lives longer. that means to use a stack allocator would require manual decision or reordering of memory operations and you would want some sort of separation between temp mem and storage mem.

another, much simpler idea that takes this further would be to use a second arena thats purely for temp mem. by resetting said arena often one could also achieve more optimal memory usage in the program at the cost of more complex memory management. what do you think? imo the strength of an arena is program simplicity and reduced reallocation, not size.

david-vanderson:
You are exactly right - what this direction of exploration is looking into is using multiple arenas.  In this case, we special case a particular arena for stack-based allocation.

@RedPhoenixQ Has done an amazing job going deep into this and learning valuable information.  It'll take some time for us to understand how to do this with the right tradeoffs.

RedPhoenixQ:
> another, much simpler idea that takes this further would be to use a second arena thats purely for temp mem. by resetting said arena often one could also achieve more optimal memory usage in the program at the cost of more complex memory management. what do you think?

This is what we are currently doing with `arena` (the frame long lifetime) and `lifo` (temporary memory). They are both using `ShrinkingArenaAllocator` doing effectively what you said with resetting more often.

Because of the nature of immediate mode rendering, there will be a lot of temporary allocations following a stack order. That's why this is reasonable for dvui specifically.

RedPhoenixQ:
Converting to draft for to clarify that this is not ready to be merged. Could maybe be closed as well if it clutters the PRs

david-vanderson:
Going to close this until/unless we have to wrestle with this again.

PR #370, opened 2025-06-12
GridWidget: Introducing Navigation and SelectionInfo.

phatchman:
Edit: Data Adapters are dead. RIP. We found better ways to support this.

This PR is really a concept demonstration right now, though I like the direction. Some of the naming is a bit meh.

I'm putting it out there early, because of the DataAdapter concept, which at first glance doesn't feel like a good fit for an IM gui.

Why do we need a DataAdapter in an immediate mode GUI? While I was skeptical of the concept, it does brings a few nice improvements:
1) It simplifies the "gridColumnFromSlice" very considerably. The data adapters convert the underlying data structure to a per-row value that can be read or written. All of the nasty trying to cater for slices of values vs structs vs pointers to structs gets reduced to `data_adapter.value(row_num)`. 

2) Means that things like the checkbox columns can work with a variety of types data, including bitsets. Setting the selection state just becomes `data_adapter.setValue(row_num, is_selected);`

3) However, the most important part for me, is that it allows me to package up various behaviors i.e. Actions. Currently I have Shift-Select and Single Select actions. There is no real other way to do this without the user providing some way to read and write a boolean flag per row so the action can set what is selected/unselected.

While shift-select is not very complicated to implement, my first implementation of it was different to the other GUIs I later tested.  I think that if we want dvui apps to always have consistent behaviors for things like grid selection, then dvui needs to provide that logic. 
(Selection will also get slightly more complicated when it's combined with keyboard navigation and selection. e.g. hitting space to select and then shift-selecting with the mouse)

The SelectionInfo struct is used to externalize the selection state, so that the Actions can be used independently of the widgets being selected.

Open to any thoughts or suggestions. 

The plan is to improve the selection further with:
* Keyboard selection
* Click row to select
* Make Hovered Row/Col/Cell an action that can be shared?


phatchman:
I did some investigation into how Dear IMGUI handles the multi-selection.  

Below is their example code for creating a list of text labels that can be selected. A selectable is basically a label you can select. 

```
static vector<MyItem> items;                  // Your items
static ImGuiSelectionBasicStorage selection;  // Your selection
selection.UserData = (void*)&items;           // Setup adapter so selection.ApplyRequests() function can convert indexes to identifiers.
selection.AdapterIndexToStorageId = [](ImGuiSelectionBasicStorage* self, int idx) { return ((vector<MyItem>*)self->UserData))[idx].ID; };

ImGuiMultiSelectIO* ms_io = ImGui::BeginMultiSelect(ImGuiMultiSelectFlags_None, selection.Size, items.Size);
selection.ApplyRequests(ms_io);
for (int idx = 0; idx < items.Size; idx++)
{
    bool item_is_selected = selection.Contains(items[idx].ID);
    ImGui::SetNextItemSelectionUserData(idx);
    ImGui::Selectable(label, item_is_selected);
}
ms_io = ImGui::EndMultiSelect();
selection.ApplyRequests(ms_io);
```

1) They create a selection object that uses a "Basic Storage" adapter to read from the `vector<MyItems>`. In this case it reads the ID field, which will be used as a key when determining what data is selected.
2) The SetNextItemSelectionUserData() basically tags the "Selectable" with user data containing the "row number" of the Selectable, which is then used by the selection adapter to set which indexes (and thus which ID's) are selected.
4) The ApplyRequests(ms_io) processes the current selection changes (e.g. clicks, shift-selects, drag-selects) and applies them to the selection object.
5) This does not actually update the vector. The user code would need to query the selections object to find what is selected or not selected. e.g. by using selection.contains().
    - IM GUI have a philosophy that the "user owns the data", so they won't update anything on your behalf unless you specifically ask.
6) They do however provide an adapter that can update the underlying data structure named ImGuiSelectionExternalStorage.
    - In the example ImGuiSelectionBasicStorage would be replaced by ImGuiSelectionExternalStorage.
    
My takeaways from this:
1) Selection is not just a grid-specific thing, but a more generalized concept. However, I think that's too big a task to tackle up front. I'm happy if I can just get something working for the grid for now. 
2) Having a DataAdapter with an IM mode GUI is not so weird. I'm mindful of pushing some sort of "framework" onto people. 
    - Whatever I do needs to be light-weight and opt-in and easily discoverable.
3) Selectable things are the only things that need to set values in user data.  
    - I'd previously rejected providing a list of selected / unselected items as in a large virtual scrolling list, this would not be scalable. It is worth re-considering this vs the complexity of an adapter. But that selection.Contains() really doesn't sit well with me. I want the user to control the type of selection data structure, not query one I've built for them. It is just duplicating state that the user needs to resynchronize every time a selection changes.
    - Make separate data adapters for selection vs those for display. Selection can work only with bools and is the only adapter to need a setValue()
    - If we end up deciding that the "display" data adapters aren't useful, then the selection side won't be impacted.
4) Think more about the row number vs ID concept. Is there a  case where it's difficult to map a row number back to the underlying data?
    - Initial thoughts are the user will know how to maintain that mapping better than me?
    - Think about impacts of sorting/filtering - i.e. tracking selections as row nrs change.
    - There must be a reason IMGUI went with the Id approach.
 5) Should there be a way to access the WidgetId or the WidgetData of the selection or display widgets?
    - first impression is that would be niche case and user would create the widgets themselves rather than relying on dvui-supplied functions
 6) Data adapters for text entry columns? They would also need setValue()

phatchman:
@david-vanderson Do you think I ever need to worry about the possibility for 2 or more widgets being clicked on or typed into in the same frame?

For example, with my column of checkboxes or text entries, it would be much nicer to return the "row number" or id of the selected checkbox or which textEntry was modified that frame. If I have to potentially return multiple ids, the interface gets more complicated.

I really can't think of a way in normal usage, but before I commit to that path, I wanted to check I'm not missing something. (Outside of someone programmatically sending multiple click events per frame, which I don't care about).

vlappa:
> Do you think I ever need to worry about the possibility for 2 or more widgets being clicked on or typed into in the same frame?

FYI: I saw this the other day: 

https://chaos.social/@blinry/114688346729960000

phatchman:
> FYI: I saw this the other day:
> 
> https://chaos.social/@blinry/114688346729960000

Hmm.. makes me think about touch as well.

david-vanderson:
> @david-vanderson Do you think I ever need to worry about the possibility for 2 or more widgets being clicked on or typed into in the same frame?

This can technically happen, but almost never.  If the framerate is low enough, then you can get a handful of events each frame, which could type into one textEntry, tab (or click) into another, and type some more.

If you think of the DataAdapter stuff as a higher-level convenience api, then I would have it only support one changing per frame.  If someone needs to support multiple changes per frame then they can use the lower level api and do each textEntry themselves.  Does that make sense?

david-vanderson:
I haven't looked into supporting multiple independent mice or touches like that at all.

phatchman:
> If you think of the DataAdapter stuff as a higher-level convenience api, then I would have it only support one changing per frame. If someone needs to support multiple changes per frame then they can use the lower level api and do each textEntry themselves. Does that make sense?

Don't want to drag you too far into this as I'm still trying to find a good balance. But it's basically the difference between a "text entry column" returning the single row number of the modified text entry, or have a callback to let the user know when something was updated (yuck) or extending the updateable data adapters to keep a list of changed rows (a little better), or have the user pass in a bitset or similar so I can mark the modified rows (overkill if it is almost alway going to be 1 change). 

And the main reason I'm contemplating updatable text entry columns is because there is typically behavior associated with navigating between text entries e.g. enter either moves 1 row down or tabs to the next entry, scrolling the viewport if required,  then wrapping to the first row if enter is pressed on the last row.

I think I'll just return a single row for now. Thinking about how to handle validation will probably give more insight. (Though trying to also "generically" handle validation might be a bridge too far.

david-vanderson:
> I think I'll just return a single row for now. Thinking about how to handle validation will probably give more insight. (Though trying to also "generically" handle validation might be a bridge too far.

I agree with this - almost certainly something later will make it clearer.

phatchman:
Below gives an idea of what the current Data Adapater API looks like. I'm relatively happy with this, but I think still needs to prove its usefulness  in real-word scenarios. So I want to come up with some more scenarios to test this with.

I've chosen to keep the layout by column, rather than row, even though that option is now available. e.g. It's much easier to just lay out all of the checkboxes at once, than have to be called once per row.

Regarding the API, there are 3 classes of DataAdapters:
1) Read Only - just supplies a value through value()
2) Read Only with Conversion - supplies a value() after being run through a conversion function. 
    - e.g. below we convert between an enum and an icon using an enum array. The enumToString converter is also used in the example. 
    - The conversion function is just a function that takes a value and returns a value. While I use functions from DataAdapater in the example, the user can easily supply their own. e.g fn convert(value: usize) usize { return value * 10; }
3) Updatable - provides value() and setValue(), but no conversion. 

In each Class, there are currently three types:
1) Values from a slice
2) Values from a slice of struct
3) Values from a bitset. (I expect a bitset will be good for managing selection, especially with virtual scrolling).

The below gives a flavour of the code. (It is cut-down from a bigger example, so might not be syntactically correct). "Data" is the name of the struct's type.

```
            // Headers
            _ = dvui.gridHeadingCheckbox(@src(), grid, 0, &selection, .{});
            dvui.gridHeading(@src(), grid, "Value", 1, .fixed, .{});
            dvui.gridHeading(@src(), grid, "Parity", 2, .fixed, .{});
            dvui.gridHeading(@src(), grid, "Icon", 3, .fixed, .{});
            dvui.gridHeading(@src(), grid, "Entry", 4, .fixed, .{});

            // Body
            const adapter = DataAdapter.SliceOfStructUpdatable(Data, "selected"){ .slice = data };
            selection_changed = dvui.gridColumnCheckbox(@src(), grid, 0, 
               .{ .selection_info = &local.selection_info }, adapter, CellStyleTabIndex{});

            dvui.gridColumnLabel(@src(), grid, 1, "{d}", DataAdapter.SliceOfStruct(Data, "value"){ .slice = data }, .{});
            dvui.gridColumnLabel(
                @src(),
                grid,
                2,
                "{s}",
                DataAdapter.SliceOfStructConvert(Data, "parity", DataAdapter.enumToString){ .slice = data },
                .{},
            );
            const icon_map: std.EnumArray(Data.Parity, []const u8) = .init(.{
                .even = dvui.entypo.aircraft_take_off,
                .odd = dvui.entypo.aircraft_landing,
            });

            dvui.gridColumnIcon(
                @src(),
                grid,
                3,
                .{},
                DataAdapter.SliceOfStructConvert(
                    Data,
                    "parity",
                    DataAdapter.enumArrayLookup(icon_map),
                ){ .slice = data },
                CellStyle{ .opts = .{ .gravity_x = 0.5 } },
            );

            if (dvui.gridColumnTextEntry(
                @src(),
                grid,
                4,
                .{},
                DataAdapter.SliceOfStructUpdatable(
                    Data,
                    "text",
                ){ .slice = data },
                CellStyleTabIndex{},
            )) |row_num| {
                std.debug.print("text changed: {}:{s}\n", .{ row_num, data[row_num].text });
            }
```

The really nice thing about this so far is that the dvui column functions no longer need to care about what is supplying the data. e.g. gridColumnLabel() which replaces gridColumnFromSlice() is infinitely simpler and loses almost all of the meta programming.

```
pub fn gridColumnLabel(
    src: std.builtin.SourceLocation,
    g: *GridWidget,
    col_num: usize,
    comptime fmt: []const u8,
    data_adapter: anytype, // GridWidget.DataAdpater
    cell_style: anytype, // GridWidget.CellStyle
) void {
    const opts = if (@TypeOf(cell_style) == @TypeOf(.{})) GridWidget.CellStyle.none else cell_style;
    GridWidget.DataAdapter.requiresReadable(data_adapter);

    const label_defaults: Options = .{
        .expand = .horizontal,
    };
    for (0..data_adapter.len()) |row_num| {
        var cell = g.bodyCell(
            src,
            col_num,
            row_num,
            opts.cellOptions(col_num, row_num),
        );
        defer cell.deinit();

        label(
            @src(),
            fmt,
            .{data_adapter.value(row_num)},
            label_defaults.override(opts.options(col_num, row_num)),
        );
    }
}
```

phatchman:
Some use-cases I've come up with to try/simulate below. Anyone think of other common uses for data population/update?

1) Data Entry Grid
  - Backed by a database (in-memory e.g. sqlite, and server based e.g. mysql)
  - Editable fields.
  - Keyboard navigation between rows / cols
  - Insert / Delete rows.
  - User needs some way to validate input / display errors.
    - Maybe that's not a grid thing. Maybe that is more generic validation like what is done on the number text entry field?
  - Probably biased to populating out by row, rather than column if not using backing cache.
  
 2) Display some more complicated data structure than an array
   - E.g. walk a tree and display
   - Or anything that returns an iterator, like a directory listing.
   - Probably also biased to populating by row if walking the data structure is expensive.
     - But not expensive enough to use a cache? i.e. would not call into OS to get a directory listing each frame.
     - Is there any benefit in dvui providing a "cache" or will that be too application-specific?

3) Query from "server"
  - Some sort of request / response interface
  - Data paging (e.g. rows 50-100)
  - Possibly async result return (though not sure if that impacts)
  - Virtual scrolling over this large dataset.
  - Might not be different enough from (1) to need a separate case?
  
4) Data entry grid with update to "server"
  - As above, retrieve results from server, but with editable fields
  - Send updates to server
  - Possibly receive validation response that needs to be displayed
  - Possibly threaded (if that matters)

vlappa:
I did not dive into the grid in any way, but when I read your last message I'm thinking: are this really questions for a GUI widget? Shouldn't data and gui be separated?  Isn't this getting way to complex? Is this Zig style or C++ style? I'm fan of a 'keep it simple stupid'  / 'don't try to be smart' approach, but maybe a grid is that complex.

Feel free to ignore this comment. ;) 

A more practical 'concern' is how to deal with keybindings. Is it able to simply set custom keybindings and adjust the 'jumps' a keybinding makes? Dvui is handling up/down keys for scroll area I believe. Is that something dvui should do or should it leave that up to the user, which keybindings they want to set and use?

phatchman:
> I did not dive into the grid in any way, but when I read your last message I'm thinking: are this really questions for a GUI widget? Shouldn't data and gui be separated? Isn't this getting way to complex? Is this Zig style or C++ style? I'm fan of a 'keep it simple stupid' / 'don't try to be smart' approach, but maybe a grid is that complex.
> 
> Feel free to ignore this comment. ;)
> 
> A more practical 'concern' is how to deal with keybindings. Is it able to simply set custom keybindings and adjust the 'jumps' a keybinding makes? Dvui is handling up/down keys for scroll area I believe. Is that something dvui should do or should it leave that up to the user, which keybindings they want to set and use?

No, it is a very valid comment. I'm prepared for this to just be a bad experiment. It makes some things much easier, like implementing multiple selection in a generic way. (i.e. the thing that handles the selecting needs to update what has been selected somewhere. So, using a data adapter for that is a nice solution).

Dear IMGUI, also implements data adapters for similar reasons. Leaving the display / update of data 100% up to the user is fine, until you want to have consistent behavior across different applications (like shift-select for multiple selecting). And I think the key thing that makes the grid different it operates over multiple rows. So you either have the user calling label() 100 times in a loop, and they have to handle all events themselves etc, or you provide a function to create the column and handle events, but the function needs to be supplied data to be able to do that.

Re: keybindings. I'm not 100% sure I follow what you mean? One of the demo apps I have takes over control of the keys in the scroll area and handles them before the scroll area can. I don't really have much issue with the current functionality. We could add an option not to handle those keys, but I think most people would want pgup/pgdn and arrows handled in the scroll area by default? Especially for accessibility.

re: Controlling the jumps, I've also done that in another app using the scroll_info directly. But yeah it might be worth to have some configuration there.

Edit: Oh, I should Add I probably gave the wrong impression. I'm not looking to create some pattern to solve all of those use-cases. I'm just trying to validate that what we have would support those style of interfaces. I suspect from the data side, they really all come down to the one case from the GUI's point of view, which is to cache data in an Array or MultiArrayList and update the grid from that array.

Edit2: And I guess I'd say the goal of the adapters is to achieve exactly what you said. Separate the provision of the data from the GUI. The data adapter, only has one method named value() to provide a value to the GUI. And setValue if it is updatable. So the idea is anyone should be able to write an adapter to almost anything they want as long as it can provide values.

vlappa:
> Re: keybindings. I'm not 100% sure I follow what you mean? One of the demo apps I have takes over control of the keys in the scroll area and handles them before the scroll area can. I don't really have much issue with the current functionality. We could add an option not to handle those keys, but I think most people would want pgup/pgdn and arrows handled in the scroll area by default? Especially for accessibility.
> 
> re: Controlling the jumps, I've also done that in another app using the scroll_info directly. But yeah it might be worth to have some configuration there.

Ah is that "another app using the scroll_info directly" available to study that code? I need something for a list browser, but left it for a later moment to dive into. 

For keybindings: maybe a option to pass a custom data structure with the keybindings, where dvui just passes such data structure with the default keybindings? It should be easily possible to not use the default up/down pgup/pgdown keys one woud think. Scroll area is quite a complex thing, I mean it wasn't just copying a dvui function and customize it with your own bindings, but it's behavior was defined by multiple source files in dvui if I recall correctly.  Anyway, like I said, I've to dive into this myself deeper. 

> Separate the provision of the data from the GUI.  

Sounds good.

david-vanderson:
Excellent ideas and use-case enumeration!  Here are some random thoughts about that stuff:

> But not expensive enough to use a cache? i.e. would not call into OS to get a directory listing each frame.

Most OS stuff caches the filesystem info in ram - I expect people to try this (directory listing each frame) and we should see how that pans out before requiring a caching layer.

Best guess is that we want to stay out of caching altogether if we can, and make that the apps responsibility.

For editable fields and validation, I think it's reasonable to write a few demos and see where to draw the line between "we just give you an empty grid cell", and "we give you lots of machinery by default".

One use-case I can think of is a grid showing x,y pairs that are being plotted next to the grid.  This requires:
* some way to enter values for a new row
  * do we require the user to say "new row" before being able to enter values
  * is there a way to support empty "template" rows that become "real" when values are entered?
* simple validation
* maybe using different widgets (like sliderEntry) instead of label/textLayout/textEntry

For keyboard nav, I'm unsure whether it should:
* work like normal scrollContainer
* up/down moves by a whole cell
* whatever the default is, a custom cell widget should (maybe?) be able to override it

This brings up the issue of cell focus and selection:
* can you tab/keyboard nav through cells that weren't rendered?
* does the grid maintain a cursor saying which cell has focus?
  * this feels like the kind of thing we'd do by default, but also allow passing in a pointer to a value so user code could change it (and scroll to the new focused cell)
* who maintains selection info?
  * how do we even represent this?  Could be a series of single cells, no cells, a whole row/col, a contiguous rectangle of cells, multiple rectangles of cells?
  * if we maintain this info, then how does the app use it (like say you click "clear selected cells", do we run a whole frame so each cell can say "I was selected")?

All of this is pure speculation until we start writing demo code.  I could be way off.

phatchman:
> 
> Ah is that "another app using the scroll_info directly" available to study that code? I need something for a list browser, but left it for a later moment to dive into.

Sure, the code for that app is just local and in a pretty messy state, but I'm planning to include something like that in either the demo or just a stand-alone example. So will ping you when I've got something. I don't think it ended up being as difficult as it first appeared.
> 
> For keybindings: maybe a option to pass a custom data structure with the keybindings, where dvui just passes such data structure with the default keybindings? It should be easily possible to not use the default up/down pgup/pgdown keys one woud think. Scroll area is quite a complex thing, I mean it wasn't just copying a dvui function and customize it with your own bindings, but it's behavior was defined by multiple source files in dvui if I recall correctly. Anyway, like I said, I've to dive into this myself deeper.

I'm not sure if it is exactly the same thing, but generally dvui lacks the ability to assign keys to things, like keyboard shortcuts / accelerators for buttons, menu items etc. You have to process events yourself for this and it's not always straight forward. e.g. if I'm handling ctrl-a for something, I need to handle the control key press before any other widget can swallow it, but the ctrl-a might be handled by a TextArea so I need to checl if the ctrl-a event was handled after the TextArea has deinit-ed.

I'm sure there is probably a better solution with existing dvui, but in my ADGUI app, the shortcuts are handled in a particularly hacky way. They check if a text-area is receiving input and modify the key handling based on that. I think I also had to block navigation keys going to the scroll area when a dialog window was showing.

I can go back at take a look that code and write up an issue for the difficulties I had with the shortcuts. I wasn't able to make much use of the "keybinds" for cut-paste etc, either. I can't remember why.

david-vanderson:
> I can go back at take a look that code and write up an issue for the difficulties I had with the shortcuts. I wasn't able to make much use of the "keybinds" for cut-paste etc, either. I can't remember why.

If you get a chance that would be useful - I don't know what kinds of problems people are having.  Thanks!

phatchman:
Thanks very much for your thoughts. I'll digest and incorporate in the list above and see what comes out!

> > But not expensive enough to use a cache? i.e. would not call into OS to get a directory listing each frame.
> 
> Most OS stuff caches the filesystem info in ram - I expect people to try this (directory listing each frame) and we should see how that pans out before requiring a caching layer.

Well, you are 100% right on this. For 5000 Windows directory entries all rendered to the grid, I cannot tell the difference between repopulating using the Dir.Iterator each frame and caching the data in an array. Neither in scrolling speed, CPU usage or anything. In release fast mode, it just works, in Debug mode it's a little jittery, but that is more to do with 10k labels.  If windows can do this easily, I'm sure mac and linux would perform even better.

Right now, it is just displays name and kind, not any stats. I expect adding stats might change things significantly on Windows. Something to try out. -- Yeah it does, but only on 1000's or dir entries.

I'm now convinced your case of using something like sqlite and streaming directly from the "db" each frame will be something users would want to support. It's prob going to be cached inside sqlite anyway? So, might actually be the preferred way to do it?

All of this is bad news for my attempt to push a column-centric view on the world. Yep, I suppose there is a reason every other grid ever created works by row and not col. Who could have guessed?

Edit: It's quite funny watching file sizes and modification timestamps update in real-time, but only when you move the mouse or press a key!

vlappa:
> Sure, the code for that app is just local and in a pretty messy state, but I'm planning to include something like that in either the demo or just a stand-alone example. So will ping you when I've got something. I don't think it ended up being as difficult as it first appeared.

Cool. I'm hanging out in #dvui on IRC (Libera.Chat) regularly, feel free to ping me there as well. Then I can show you also what I try to accomplish.

phatchman:
EDIT: Given the progress I've made on the alternative approach, it's not really worth looking at the code below and most of the issue mentioned below have been solved another way. So you can safely skip this one.  Sadly, I think I've already killed my data adapter idea. RIP.

I've put together a very simple directory viewer [diff](https://github.com/phatchman/dvui-dev/compare/upstream_main...directory_browser ) using cached and non-cached data. This demonstrates the data adapter for selection and multi-selection. If you are inclined to look, at the code [sdl-dir-grid.zig](https://github.com/phatchman/dvui-dev/blob/directory_browser/examples/sdl-dir-grid.zig) is the application.

I think I need to add:
  - filtering - Done
  - sorting (though not sure how for non-cached)
  - make clicking on a row do something.
  - There needs to be co-ordination between the "select all" header and the selection of individual cells. Already have an idea for this. - Done

What I've found, so far.
   - providing functions for columns of labels etc isn't that useful here. The new grid api is so minimal, that you don't save much and also have much more flexibility if you don't use them.
   - providing a function for a column of checkboxes is useful. It can provide functionality like saving the last 2 selected rows.
   - SelectionInfo seems a good abstraction so far. Users can ignore the data adapter bits and just use selection_info directly. (though the field naming needs an overhaul).
   - Not convinced Actions are the right abstraction for handling the shift-click.
 
To remove the DataAdapter we need:
1) A way for the user to create their own checkboxes, but also be able to track which ones were selected.
2) Provide the user with knowledge of the shift key state during the selection, so they can manually apply the results in selection_info to their own dataset. 
    - they need to know whether to apply 1 selection or a selection range. It doesn't have to directly be shift key state.
    - And ignores that multiple single selections can theoretically happen in 1 frame. Data adapter doesn't have that issue.
    - Edit: this  isn't an issue with the "event" approach.

phatchman:
This is a very different idea I had for doing selection. I'm hoping it would be a simpler solution to the data adapter one.

What if in the widget options there is a "selection_id" style tag that can be put on the widget?
- For any widgets that can be "selected" they push their selection_id onto some sort of selection queue, or as a selection event? ~~(via bubbling? or whatever the replacement for that will be)~~
- It would be then simple to write something that listens to those selection events and looks at the last one for single select or the last two for multi-select. (across multiple frames)
- There needs to be something else associated with the selection id. the rect I suppose, so we can tell which widget the event belongs to.

Then the user can just create their own checkboxes, give it the selection_id of the row number.
run some function that spits out the selections info they need for that frame, based on whether shift key is pressed, or whatever other criteria they need?

* No need for data adapter as there is no other functionality required than creating a checkbox with an id.
* Flexibility because it uses id over row_number. Up to user to choose what they want to use.
* Works across anything that support selecting one or more things from a group of things. e.g. the tree widget
* Not 100% clear what all the selectables are? buttons (covers checkboxes and other clickables). Anything else?  
* Maybe user can make other things selectable, like using boxes and processing click events? Would we include that as well? They'd need to push a selection event somehow.
* It wouldn't need to dynamically allocate, limiting to 5 selection events per frame would be more than enough. Maybe even just 3.

If that sounds promising I can try and prototype something along those lines. Or maybe it sparks someone to have a better idea :)

Edit: here is the [prototype](https://github.com/phatchman/dvui-dev/compare/upstream_main...new_selection_idea). I only implemented it for checkboxes, but I think that is the right level. A button doesn't know if it is "selected" or not. It has to be something like a checkbox that has a selection state.
* Should it be an "Event" event? And added to the event queue instead?
* Pretty simple code. Works for single select as well, you just iterate though the selection events which have the selection_id and whether it should be selected or not.
* Does selection_id belong in options? Maybe it is a separate thing that gets passed optionally to selectables?

Edit2: We now have working single and multi-select with this new method. It feels like a much more dvui-integrated approach than the previous data adapter version. And this would work for anything selectable, not just grids.

Edit3: In middle of adding filtering, which shows that selection needs to run off id's rather than grid row numbers. - Done.

Edit4: This is the [integrated event version](https://github.com/david-vanderson/dvui/compare/main...phatchman:dvui-dev:selections_as_events). It seems to work fairly well. You should be able to build whatever selection strategy you want out of this. I like this approach a lot.
P.S. Put one of the checkboxes half visible at the top/bottom of the grid window. Then click on it and you can see the nice benefit of the scroll-to-focus changes.

phatchman:
> 
> Ah is that "another app using the scroll_info directly" available to study that code? I need something for a list browser, but left it for a later moment to dive into.
> 
@vlappa Actually you can take a look at the virtual scrolling grid example if you want to see something now. It is relatively simple because it can assume all rows are the same height. The only weird thing it does is that it uses the "fixed viewport" stuff that was added recently for the scroll container. You would nomrally use the scroll_info viewport x and y, rather than the saved viewport that I'm using in this example.

phatchman:
I've now added a demonstration of how users can add their own selection events with the "row selection" option. It's in the same [integrated version](https://github.com/david-vanderson/dvui/compare/main...phatchman:dvui-dev:selections_as_events) branch. This demo now covers all the selection scenarios I can think of outside:
1) The interaction with sorting. 
  - I don't see any issues there. Now the user has control over what id's to pass, they can use whatever makes sense for them to identify the start and end of the selection range.
2) Interaction with virtual scrolling. 
  - Again, I don't see any issues. By using id's provided by the user, rather than row numbers, it avoids all of the issues I hadn't solved with DataAdapters using row number directly. 
  - Really this is just a different type of "filtering"

I think this is developed enough to make a PR consisting of:
1) The Selection event and Options change
2) The change to checkbox and anything else we think needs to raise selection events. 
    - Honestly, I think it is only checkbox right now.
    - IMGUI has "selectable boxes", I think they are called "selectable", not sure if we need these?
    - I imagine the TreeWidget could link into this as well if it wants to support multi-select?
3) Adding Single and MultiSelect structs to a selection namespace or similar. - Done.
4) Changes to the "Layouts Data" example to enable multi-selection and row selection with highlighting.
5) Just added keyboard selection via the "select_all" keybinding. Simplified how the checkbox header works so now user controls what the select-all state is, not the header widget. 

Things I'm not sure about:
1) Any issues with me adding these as events to the event queue? They won't be there are start of frame, but I don't think that matters? 
2) Whether selection_id belongs in "Options" or just as a param to checkbox and other widgets. I like it in options, but no widget actually uses it.
3) Do we care that users would raise their own events?
~~4) There is a bug with row select that when you click on the checkbox, it selects on mouse down (via the row selection), but then unselects on mouse up (via the checkbox / button). Needs some tweak here.~~
    ~~- Behaviour of other UIs I've tried is is basically, checkboxes select on mouse-up, unless being row-selected, then they select on mouse down, and don't react to mouse up.~~
    ~~- For Windows explorer in row-select mode, clicking on the checkbox does a single-select, rather than a row-select. i.e. similar to making the shift-select rect not include the checkbox cols.~~
    - I have implemented a solution for this now. Grid returns the cell of the clicked point. User can decide which columns / rows count as row selection based on the cell. In the example, I exclude col 0, (i.e. the checkbox column) from counting as row selection.
    
Let me know your thoughts or if you have a better idea for this. But something along these lines seems good to me. I'm not in a rush to merge it. It can wait until I've got a couple of the other examples put together and it's shaken out more.

Ramble: I mean you wouldn't think there was that much to shift-select right? But so many little corner cases. Like what if you select something that then gets filtered. Oh, and the user wants to hold shift on the first select, so that we never ever set a second selection point until they let go of shift. And they row-select to start and checkbox select to end. You want to select-all while filtering? Sure! And lots of other minor things. But I think it is all sorted now! At least I can't find any more bugs after trying to torture it for 20 mins. At least it's not a text-entry widget! I've looked on in awe at the code for the event handling in that widget!!! (He says and then finds another bug. So. Select then Select-All, Then Shift-Select, what should happen?? I guess I'll have to find out. Yeah, this is too complicated to leave up to users to re-implement. OK Turns out Mac and Windows handle that in the same non-way I do. So I'll call this done.

P.S. Am I doing the check for select all correctly in "SelectAllKeyboard"? i.e. I check if the mouse position is in the widget's rect to see if I should handle the key. I don't think there is any other way to check? I don't want to react to select-all unless the mouse is within the grid.

david-vanderson:
Very cool!

> * Any issues with me adding these as events to the event queue? They won't be there are start of frame, but I don't think that matters?
> * Do we care that users would raise their own events?

I honestly don't know right now.  #410 is working towards removing those extra kinds of events.  But maybe they should still be unified into custom events that get added to the queue?  This is an area of active investigation, so any ideas/attempts/feedback is useful.

> 
> * Whether selection_id belongs in "Options" or just as a param to checkbox and other widgets. I like it in options, but no widget actually uses it.

My advice is to circle back to this after trying more usage code, hopefully it'll be more clear.

phatchman:
> > * Any issues with me adding these as events to the event queue? They won't be there are start of frame, but I don't think that matters?
> > * Do we care that users would raise their own events?
> 
> I honestly don't know right now. #410 is working towards removing those extra kinds of events. But maybe they should still be unified into custom events that get added to the queue? This is an area of active investigation, so any ideas/attempts/feedback is useful.
> 
As I understand the aim of #410 is to get rid of bubbling? Even though these are "artificial" events, they can't bubble because there is no widget to process them. So, I think this is a new category of "event" compared to what we have today.

I like the queue solutions, because dvui is just supplying what was selected, in the order it was selected. Anything can listen to that and implement a selection strategy. e.g. the current single-select doesn't handle the case of multi selections per frame. But the user could instead listen to the raw selection events and unselect everything you receive except for the last one. 

The alternative would be to re-introduce a selection_info and keep the queue inside that rather than adding to the dvui event queue. It just means the selection code processes both the event and selection queues, and a "selection_info" gets passed to everything that handles selection. And I'd need to supply an equivalent of  eventMatch/eventMatchSimple for it. Yeah, I can see this approach working.

I'll add something like that plotting application you mentioned, and a "database" driven one. This will cover paging of data, virtual scrolling, filtering, cell to cell navigation and standard data-entry behavior, add row, edit row, delete row etc. That should give us a pretty good sample to look at.

david-vanderson:
> The alternative would be to re-introduce a selection_info and keep the queue inside that rather than adding to the dvui event queue. It just means the selection code processes both the event and selection queues, and a "selection_info" gets passed to everything that handles selection. And I'd need to supply an equivalent of eventMatch/eventMatchSimple for it. Yeah, I can see this approach working.

My gut says treating selection like an event seems weird, but I am realizing that I don't understand the selection needs enough, so I would proceed with your best guess.

phatchman:
> My gut says treating selection like an event seems weird, but I am realizing that I don't understand the selection needs enough, so I would proceed with your best guess.

I think your gut is right. All I need is a queue and a "selection_id". I can wrap that in a separate Selection stuct and pass that wherever I need as an InitOption (i.e. to checkbox). Solves the polluting Options with the selection id as well. 
The thing I do lose is the auto-reset to 0 events each frame. Oh well.

I think I was just trying to make my code look pettier by only dealing with 1 queue.

Edit: That was almost [too easy](https://github.com/david-vanderson/dvui/compare/main...phatchman:dvui-dev:selections_as_events). Moved it all into the SelectionInfo struct. Kept the "SelectionEvent", but not as a real event. 

Two minor downsides:
1) Checkbox gets a new parameter, breaking the API for everyone named "SelectionOptions". If I find it too annoying passing the extra .{}, I'll create a new checkbox function for it. 
`_ = dvui.checkbox(@src(), &is_set, null, .{ .selection_id = dir_num, .selection_info = &selection_info }, .{ .gravity_x = 0.5 });`
2) User needs to call "reset()" each frame to dump the selection event queue. 

Big benefit of this version is that if you aren't interacting with selection it doesn't impact your application in any way. (except for that extra checkbox param!)

Thanks. It really helps when you point out the "smells".

phatchman:
~~@david-vanderson I not sure how to do what I need with the keyboard handling for the grid. Well not how to do it nicely anyway.~~
~~Let's say a textbox in the grid has focus and the user wants to take over the up and down arrow keys to navigate up and down the column. My expectation in "data heavy" grids, you would want use tab and shift-tab to navigate left and right and up/down or ctrl-up/down to move within a column.~~

~~In the grid's processEvent(), I currently check if the mouse pointer was in the grid's rect for that frame and only then handle the up and down arrow keys. But I suspect there is some better way to do this?~~

Actually, what I *really* think I want to do is check if they key would be captured by the currently focused widget and that the currently focused widget is in the grid's rect. I think that is what the .focus_id to eventMatch is for, right?

Got it working using the .focus_id stuff. It's a bit clunky in places, but is a good starting point. We now have a basic cursor that the user can use to focus a widget. I don't even think this needs to live in the grid, which would give users the ability to implement their own navigation strategy and key bindings easily.

david-vanderson:
> Got it working using the .focus_id stuff. It's a bit clunky in places, but is a good starting point. We now have a basic cursor that the user can use to focus a widget. I don't even think this needs to live in the grid, which would give users the ability to implement their own navigation strategy and key bindings easily.

Excellent!

phatchman:
@david-vanderson can you please explain when I should use `focusedWidgetId()` vs `lastFocusedIdInFrame`?

P.S. This navigation stuff seems to have just fallen into place very easily. Just worked out the last bit, which was how to update the cursor if the user clicks in a widget. There is a very handy mouse focus event for that.

I assume it's fine to give all of the "entry" widgets in the grid a tab index of 0? When they want keyboard navigation, I'm taking control of the tab key. (To be clear the user would set it to 0, not the grid).

P.S. I've pushed the latest code containing navigation and selection to this PR if you are curious to try it. The plotting app is still pretty basic, but gives a nice demonstration of the instant updates you get with dvui. Can't believe it turned out to be only 361 lines of code for both features!

phatchman:
> For keybindings: maybe a option to pass a custom data structure with the keybindings, where dvui just passes such data structure with the default keybindings?

I implemented this for the navigation and was fairly easy. User can pass in their own keybindings or use defaults.
```
  pub const NavigationKeys = struct {
        up: dvui.enums.Keybind,
        down: dvui.enums.Keybind,
        left: dvui.enums.Keybind,
        right: dvui.enums.Keybind,

        pub const none: NavigationKeys = .{
            .up = .{},
            .down = .{},
            .left = .{},
            .right = .{},
        };

        pub fn defaults() NavigationKeys {
            const cw = dvui.currentWindow();
            return .{
                .up = cw.keybinds.get("char_up") orelse unreachable,
                .down = cw.keybinds.get("char_down") orelse unreachable,
                .left = cw.keybinds.get("prev_widget") orelse unreachable,
                .right = cw.keybinds.get("next_widget") orelse unreachable,
            };
        }
    };
```

david-vanderson:
> @david-vanderson can you please explain when I should use `focusedWidgetId()` vs `lastFocusedIdInFrame`?

`focusedWidgetId()` is for when you want to check if a particular widget is focused.  "Should I draw a focus ring"?

`lastFocusedIdInFrame()` is for checking if any widget between 2 calls was focused.  Usually this is to check if any of a widget's children have focus.  This replaced keyboard event bubbling, but can also be used to do stuff like "enable keyboard shortcuts only if the focus is in this part of the app".

Does that make sense?

> 
> P.S. This navigation stuff seems to have just fallen into place very easily. Just worked out the last bit, which was how to update the cursor if the user clicks in a widget. There is a very handy mouse focus event for that.
> 
> I assume it's fine to give all of the "entry" widgets in the grid a tab index of 0? When they want keyboard navigation, I'm taking control of the tab key. (To be clear the user would set it to 0, not the grid).

I don't understand - a tab index of 0 removes it from the keyboard nav.  So you are manually using tab to change the focus?
> 
> P.S. I've pushed the latest code containing navigation and selection to this PR if you are curious to try it. The plotting app is still pretty basic, but gives a nice demonstration of the instant updates you get with dvui. Can't believe it turned out to be only 361 lines of code for both features!

Will do - thank you!

phatchman:
> Does that make sense?

Yeah. Last focused in frame kept returning 0 for me. I'll play with it some more.

> I don't understand - a tab index of 0 removes it from the keyboard nav. So you are manually using tab to change the focus?

I just give the user a cursor and it is up to them to focus the right widget. So they can still put / omit the tab index if they want and get the normal tab handling. And the user can configure which keys are used for the nav. By default tab / shft tab / up / down.

But in keyboard nav mode my app sets tab index to 0 for all the cells. Just thinking if someone has 1000's of widgets in the grid, do we want to add them all as tab indexes? They can use the nav cursor instead.

Edit2: i guess that brings up the question of how you tab into / out of the grid. I'll do some research on how people handle that.

Edit: leaving a note for myself. Keyboard  select all doesn't respect if text widget should handle the ctrl-a. Fix this. And add ability to skip cols, and ability for tab to wrap to next row.

phatchman:
> but can also be used to do stuff like "enable keyboard shortcuts only if the focus is in this part of the app".

This reminds me. It seems like from my experiments that ctrl-a / select all is supposed to work if the mouse is inside the grid, regardless of whether anything in the grid has focus. That seemed like a new kind of interaction, i don't think we have anywhere else?

So that's why the processEvents() for it is weird.

david-vanderson:
> > but can also be used to do stuff like "enable keyboard shortcuts only if the focus is in this part of the app".
> 
> This reminds me. It seems like from my experiments that ctrl-a / select all is supposed to work if the mouse is inside the grid, regardless of whether anything in the grid has focus. That seemed like a new kind of interaction, i don't think we have anywhere else?
> 
> So that's why the processEvents() for it is weird.

So even if keyboard focus is somewhere outside the grid, if the mouse is hovered over the grid ctrl-a should be handled by the grid?  Wow that is weird.

phatchman:
> So even if keyboard focus is somewhere outside the grid, if the mouse is hovered over the grid ctrl-a should be handled by the grid? Wow that is weird.

Hmm. I went back and tested this again and after testing some other applications, it does behave more sanely than I thought. We can just adopt the normal if the grid or child has focus and receives a non-handled select-all key approach.

I did find one other interesting [feature](https://imgur.com/a/OapdSKW) in Windows Explorer though. If you click on the folder in the left pane and press ctrl-a, it does do a select all on the grid. But that is something users could implement if they needed with existing functionality.

david-vanderson:
> > So even if keyboard focus is somewhere outside the grid, if the mouse is hovered over the grid ctrl-a should be handled by the grid? Wow that is weird.
> 
> Hmm. I went back and tested this again and after testing some other applications, it does behave more sanely than I thought. We can just adopt the normal if the grid or child has focus and receives a non-handled select-all key approach.
> 
> I did find one other interesting [feature](https://imgur.com/a/OapdSKW) in Windows Explorer though. If you click on the folder in the left pane and press ctrl-a, it does do a select all on the grid. But that is something users could implement if they needed with existing functionality.

Oh good - yeah that sounds good.

phatchman:
>lastFocusedIdInFrame() but can also be used to do stuff like "enable keyboard shortcuts only if the focus is in this part of the app".

So, if I have this right, I should look at last focused in frame before the grid is rendered and compare it to last focused in frame after all the cells have been laid out. If focused widget has changed, I know the focus is somewhere inside the grid and should react to the key. I've pushed a change to selection.SelectAllKeyboard that implements this method.

Actually, I think this is now might even cover the windows explorer scenario I showed. Where you call kb_selection.reset() and kb_selection.processEvents(), determines which widgets get included in the focus change test. So, if you reset the keyboard select all before laying out the side-panel and the grid, and call processEvents() after creating all the side-panel and grid content, any widget within the side-panel would also count towards handling select-all. Nice! - Yeah it does, just tried it.

That lastFocusedInFrame()) seems really flexible!

phatchman:
OK, I have come across an issue. 

Edit: The TLDR is that I wanted to intercept keys (like up / down arrow) before a text widget does. But that means I have to run processEvents before creating the grid body, which means I can't check if the focused widget changed.

I now think doing that is a bad idea and a better approach is to be able to either mask certain keys from being processed by text entry widget, or make it possible to configure text entries keys and set some of them to null, so the keys can be processed by the grid instead, when the user wants that.

---------------- Below is obsolete ----------

For the navigation struct I want to handle the keys *before* the text entry widget does in case we have been assigned a key that the text entry also wants to process (like up arrow). So, I need to run processEvents() before the cell is created, but I won't know if focus has changed until after it is created.

I could make the user do the processing in two steps; process the events first and then make the user make a second call to apply the result afterwards if the focused widget has changed. But it seems a bit messy. 

If you have any better ideas, would be appreciated. Or maybe we just don't support overriding keys that a widget already processes. I can find other defaults for the navigation keys, but keeping using up/down arrow would be nice.

edit: Splitting into two functions (a processEventsBefore and processEventsAfter()) doesn't work. I need to know the cursor position before the cells are created (well before they are deinited anyway), so that the correct widget can be given focus. 
So, I need to process those keypress events before I know if I should be doing so.

i.e. the below doesn't work because the cursor is updated too late.

```
    pub fn processEvent(self: *GridKeyboard, e: *Event, grid: *GridWidget) void {
        self.last_focused_in_frame = dvui.lastFocusedIdInFrame(null);

        defer self.enforceCursorLimits();

        switch (e.evt) {
            .key => |*ke| {
                // Check if focus is in a child of our widget.
                // TODO: This doesn't work yet because processEvents() needs to run before the body.
                if (true or self.last_focused_in_frame != dvui.lastFocusedIdInFrame(null)) {
                    if (ke.action == .down) {
                        if (ke.matchKeyBind(self.navigation_keys.up)) {
                            e.handle(@src(), grid.data());
                            self.cursor_temp.row_num = if (self.cursor.row_num > 0) self.cursor.row_num - 1 else 0;
                        } else if (ke.matchKeyBind(self.navigation_keys.down)) {
                            e.handle(@src(), grid.data());
                            self.cursor_temp.row_num += 1;
                        } else if (ke.matchKeyBind(self.navigation_keys.left)) {
                            e.handle(@src(), grid.data());
                            self.cursor_temp.col_num = if (self.cursor.col_num > 0) self.cursor.col_num - 1 else 0;
                        } else if (ke.matchKeyBind(self.navigation_keys.right)) {
                            e.handle(@src(), grid.data());
                            self.cursor_temp.col_num += 1;
                        }
                    }
                }
            },
            .mouse => |*me| {
                if (me.action == .focus) {
                    const focused_cell = grid.pointToColRow(me.p);
                    if (focused_cell) |cell| {
                        std.debug.print("updating mouse cursor", .{});
                        self.cursor.col_num = cell.col_num;
                        self.cursor.row_num = cell.row_num;
                        self.cursor = self.cursor_temp;
                    }
                }
            },
            else => {},
        }
    }

    pub fn processEventsAfter(self: *GridKeyboard) void {
        if (self.last_focused_in_frame != dvui.lastFocusedIdInFrame((null))) {
            self.cursor = self.cursor_temp;
        }
    }
```

phatchman:
The further issue is that text entries use up a lot of keys. So, if we go with the "no override" approach, you don't have many keys left to play with.

So maybe what I really should be looking at is a way to either re-define or limit what keys are used in a text entry and continue to process events based on the last focused in frame approach.  It doesn't address the problem that I need to know which widget to focus before it is de-inited, but maybe it gets me closer. - Edit. I have solved this now with an is_focused bool the user can check before setting their widget focus.

I'll play with this key remapping / overriding idea and see how it pans out. Unless you have something that seems clearly better.

(Add a textbox outside the grid in the plot demo and you'll see how broken the existing event handling is.) - Should be fixed now.

Edit: Yeah, I got something working properly when using different keys to textbox when processing events after the textentry. So, I either need a way to block textbox from processing some keys or a way to configure which keys it will process. Have pushed that change now. You have to use ALT W/A/S/D to navigate until I work out the next bit.

Q: Should TextEntryWidget handle the tab / next_widget key? I took it out and everything still seems to work?

Q2: char_up is currently:
```
try self.keybinds.putNoClobber(self.gpa, "char_up",          .{ .key = .up,   .shift = false });
```
which means it also matches alt up and ctrl up, because they are not explicitly set to false. In the short-term I can just set control or alt to false for those. And if textWidget doesn't need to handle tab, it can be tab/shift-tab, ctrl/cmd-up, ctrl/cmd-down as default keys until I figure something else out.

I checked what other apps do on ctrl-up and is varies. Notepad++ scrolls up/down a line as does vstudio. Regular notepad moves the cursor up and down, except if it is on the last line, then it does nothing??? Microsoft word jumps to the start/end of a paragraph. React text entry widget moves up and down 1 line. 

And control-shift is similar. On notepad++ it moves the selection up and down 1 line, i.e. the selected text actually moves. In react text widget, it moves the selection up/down a whole line. In notepad it does the same as react. In vscode it works exactly the same as shift up/down as best I can tell.  Microsoft word selects to the top/bottom of the paragraph.

So, I'm thinking I can steal ctrl up/down from text entry, at least for now? It will also mean ctrl-shift up/down won't work in the text area. I've pushed those changes if you want to see what it feels like with the keybinds changed. So the grid now uses "next_widget", "prev_widget", "grid_cell_up" and "grid_cell_down", otherwise known as tab/shift-tab/ctrl-up and ctrl-dn (or cmd on mac)

Now we have some reasonable behavior when tabbing in/out of the grid. Only one cell has a tab index, so every time the grid gets focus, it goes to that cell. Ideally I guess it would go to the first cell on tab and the last cell on shift-tab, which I think is do-able, but a job for another day. There seems to be an extra widget taking a tab-index somewhere that I've not looked into. So one lot of tab/shift tab goes no-where. Could be a bug with my code though.

 I think this is worth playing with the tabbing to see what you think. I've made wrapping the cursor and tabbing in/out options.

david-vanderson:
> That lastFocusedInFrame()) seems really flexible!

Yeah I've been surprised too!

> Edit: The TLDR is that I wanted to intercept keys (like up / down arrow) before a text widget does. But that means I have to run processEvents before creating the grid body, which means I can't check if the focused widget changed.

One thing to try is remembering if anything in the grid had focus from last frame?

It's tempting to special case textEntry, which might be correct.  But we could imagine someone putting multiple widgets into a grid cell, so it's worth putting in some extra investigation to see if we can make it generic over what ends up in the grid cell.  The "correct" solution is processing the events before the grid cells, but maybe we have to figure something else out.

> Q: Should TextEntryWidget handle the tab / next_widget key? I took it out and everything still seems to work?

The reason it did that was to be able to, in the same frame, handle text going to one text entry, tab to the next, and put text into the next.  This makes sense when trying to quickly fill out a form in a low fps environment.

I still need to dig into this more, this is just some quick reactions, sorry!

phatchman:
I think remembering the last focus should work. I'll try that.

david-vanderson:
>Is the model "the grid cell has focus", or "the first element in a grid cell has focus"?  Maybe considering multiple widgets in a cell is wrong.  That would mean special casing how the grid uses each kind of cell though.

I think it either has to be each cell gets focus OR each widget gets focus.
If cells, you can use the cell keyboard nav.  And by cell I mean a single widget in the cell gets focus.
If widget, you give each widget an increasing tab id(which will happen by default). But you only get tab navigation when you do this.

phatchman:
> I still need to dig into this more, this is just some quick reactions, sorry!

Breaking the problem down, there are 3 scenarios when running processEvents() *before* the grid is created.
1) Something has focus before the grid is created. 
    - dvui.lastFocusedIdInFrame(null) will return the id of the focused widget. (This alone is enough to tell that a grid "cell" doesn't have focus)
    - dvui.focusedWidgetId() = dvui.lastFocusedIdInFrame(null) from *this* frame.
    
 2) Something inside the grid has focus
    - dvui.lastFocusedIdInFrame(null) will return .zero as nothing has focus yet
    - dvui.focusedWidgetId() = dvui.lastFocusedIdInFrame(null) from *last* frame captured immediately after grid.deinit().
    
 3) Something after the grid has focus
     -  dvui.lastFocusedIdInFrame(null) should return .zero as nothing has focus yet.
     - dvui.focusedWidgetId() != dvui.lastFocusedIdInFrame(null) from *last* frame captured immediately after grid.deinit().

i.e. for 3 all we need to know is that the focus is with some widget we hadn't seen before grid.deinit() in the last frame.

I've implemented the above in the processEvent() and it seems to work fine. (self.last_focused_widget is the 
lastFocusedIdInFrame() from the previous frame).

```
self.is_focused = self.last_focused_widget == dvui.focusedWidgetId() and dvui.lastFocusedIdInFrame(null) == .zero;
```
Then I use self.is_focused in processEvent() to work out if I need to process the navigation keys or let them through to the text area.

I think it can probably be simplified further because I probably don't have to remember the exact widget id, only that it wasn't .zero when the grid was deinit()-ed last frame. I'll try that next time. 

* Need to make sure I test this with scrolling. e.g. what should pg-up and pg-dn do. Just scroll the grid or move the cursor as well?  - Undecided, probably want to scroll the cursor as well. Excel scrolls the cursor, but want more examples. 
* Think about other keys like ctrl-end/home? - Implemented.

I've pushed these changes. Seems to work fine, except for the mystery tab index I still need to track down. It's just the tab in / tab out behavior I need to think about, whether it returns to the cursor or the start. Thinking about tabbing in to a grid you've scrolled part way down for example. And you've set the navigation keys to arrow keys or something. - Here I'm thinking just tab back into whatever the last cell that had focus was. Otherwise clicking out of the grid and tabbing back in would take you to the first cell, rather than back to the last cell you were on.

P.S. `dvui.currentWindow().debug_widget_id = dvui.focusedWidgetId() orelse .zero;` was my useful for tracking down where the extra tab index was coming from. Don't know how to fix it yet, but that's another issue.
I'm going to play around with having the debug window up while developing this to see if it helps.

phatchman:
I now get the same behavior as this minimal app:
```
    for (0..3) |i| {
        var text = dvui.textEntry(@src(), .{}, .{ .id_extra = i });
        text.deinit();
    }
```
The focus tabs from first to last, but then the tab focus is set to "null" after tabbing out of the last textbox. Pressing tab again, takes you back to the first textbox. So, on the 3rd tab, there is no widget with focus.

phatchman:
> > Is the model "the grid cell has focus", or "the first element in a grid cell has focus"?  Maybe considering multiple widgets in a cell is wrong.  That would mean special casing how the grid uses each kind of cell though.
> 
> I think it either has to be each cell gets focus OR each widget gets focus. If cells, you can use the cell keyboard nav. And by cell I mean a single widget in the cell gets focus. If widget, you give each widget an increasing tab id(which will happen by default). But you only get tab navigation when you do this.

~~There is another way to do this, but the user will need to implement the intra-cell focus. The user checks if the curent cell has focus. If it does they run their own event loop for tab/shift-tab and keep track of which widget within the cell has focus themselves. This would only be necessary if they are trying to combine keyboard nav with multiple focusable widget per cell. Otherwise they'd just use the existing tabbing between widget.~~ I'll test out this idea in the plotting app. Having a drop-down tied to a text-entry or something.

~~My future idea to make this easier would be to allow multiple body cells per header cell or potentially per row (row 1.5, 1.6?). Then they could put each widget in its own cell and still use cell navigation. Not sure how feasible this would be. It breaks my brain to think about it.~~

Have a much simpler idea, will see if it works. The dimensions of the cursor don't need to match the dimensions of the grid! Each widget can be in it's own virtual cell.

phatchman:
Random thought:

For scroll key events, these are propagated upward if the scroll container is already at its min or max and can't process the event. Does it make sense to do this for other keys? If a text widget is already at end and gets a ctrl-end , what would happen if it left that event unhandled and let the parent widget handle it? What do other ui kits do?

Thinking about case where grid is scrolled to end and someone presses down arrow. This should prob scroll any parent scroll container.

david-vanderson:
I think I'm following.  Sounds like you've got a good understanding of the focus issues.

One thing for tabbing into a grid, the grid could only set a tab index on the first visible grid cell each frame, so tabbing in never scrolls it?  Or set the tab index only on the last focused cell?

Extra tab index problem - dvui has a state where nothing is focused, so you are seeing that.  Is it messing you up?  I **think** that is okay behavior, but haven't done a thorough investigation.

> For scroll key events, these are propagated upward if the scroll container is already at its min or max and can't process the event. Does it make sense to do this for other keys? If a text widget is already at end and gets a ctrl-end , what would happen if it left that event unhandled and let the parent widget handle it? What do other ui kits do?

I'm not sure about this.  My guess is that keyboard events should not propagate to containing scroll areas, but mouse events should.  That seems to be at least what browsers do?

phatchman:
> One thing for tabbing into a grid, the grid could only set a tab index on the first visible grid cell each frame, so tabbing in never scrolls it? Or set the tab index only on the last focused cell?

Currently it remembers focused cell.
> 
> Extra tab index problem - dvui has a state where nothing is focused, so you are seeing that. Is it messing you up? I **think** that is okay behavior, but haven't done a thorough investigation.

No, it's not messing anything up and is consistent with the handful of apps I tried. So, all good.

> > For scroll key events, these are propagated upward if the scroll container is already at its min or max and can't process the event. Does it make sense to do this for other keys? If a text widget is already at end and gets a ctrl-end , what would happen if it left that event unhandled and let the parent widget handle it? What do other ui kits do?
> 
> I'm not sure about this. My guess is that keyboard events should not propagate to containing scroll areas, but mouse events should. That seems to be at least what browsers do?

Agree, I doubt there is anything to do here, but worth checking. I'll do some testing and see what I find.

phatchman:
One minor issue I'm having with this focus stuff is getting access to the underlying widget data. e.g. for drop-down:
```
pub fn dropdown(src: std.builtin.SourceLocation, entries: []const []const u8, choice: *usize, opts: Options) bool {
    var dd = dvui.DropdownWidget.init(src, .{ .selected_index = choice.*, .label = entries[choice.*] }, opts);
    dd.install();

    var ret = false;
    if (dd.dropped()) {
        for (entries, 0..) |e, i| {
            if (dd.addChoiceLabel(e)) {
                choice.* = i;
                ret = true;
            }
        }
    }

    dd.deinit();
    return ret;
}
```
It's not a big deal as I can just easily write my own "dropdown()" to give me the widget data, but just leaving here in case it sparks an idea.

Previously I set a tag on the widget I knew had focus and that worked when there might only be 1 widget in the cell that could have focus. But now I need to know the widget id's of both widgets (dropdown and textEntryNumber) so I can tell which one of those was focused with the mouse click. 

I have an idea to try and work-around this, but being able to access to the underling wd for all the dvui functions would be a nice to have.

phatchman:
OK. Here's a question for you. [This](https://imgur.com/a/wuq1sbr) is the current plot grid.

The dropdown grabs focus on hover. I don't seem to get an event telling me that the dropdown now has focus. And just knowing the focused id doesn't help me. I need to know the screen rect of the focused widget so I can find out which cell it lives in. 

I tried it with "Debug Focus", (which I now realise does the same thing as my setting the debug widget ot the focused widget) and it doesn't show any notification that the focus changed either.

I'm a bit at a loss on where to go with this one.

Also, I'm not sure changing the focused cell on widget hover is going to be a good interaction. 

I get widgets changing their appearance when hovered, but actually taking focus? It honestly doesn't feel good to me. I'm typing away in my text area and I move my mouse out of the way, but just happen to very briefly hover a drop-down. Now I'm typing into a drop-down instead.

Then with a drop-down in a data entry grid, where you are basically tab, type, tab, type, tab arrow etc. I can see it being particularly annoying if moving the mouse changes which cell becomes focused.

david-vanderson:
> Also, I'm not sure changing the focused cell on widget hover is going to be a good interaction.

I agree, this needs to change. Dropdown reused the MenuItemWidget stuff, which grabs focus on hover, because I when I wrote that I didn't understand the problems. https://github.com/david-vanderson/dvui/issues/441

Can you move forward assuming that will be fixed?

phatchman:
Fffffffinally... I've got some sort of solution to the multiple widgets within a cell. You would not believe the number of promising dead ends I've been chasing.

You can use the grid navigation cursor as a virtual cursor, so that it navigates over widgets, rather than cells. All that cursor does is give the user a way to know what to focus.

That is fine for keyboard navigation, BUT.. The big issue is when focus changes externally, like the user clicking into a text box. I need to convert that back into a row and col somehow. IF it is one widget per cell, I can just ask the grid for the row and col for that point, but with multiple widgets, that doesn't work.

So, my 5-line solution is to allow the user to provide an optional callback for that mapping function. I know, I know. But once the code is cooked, I'll leave it as your challenge to find a way to get rid of it. :)

This still doesn't work with the focus-stealing dropdowns. But at least we can have 2 text boxes in a cell now,

The code is pretty raw, but have pushed the latest.

phatchman:
> 
> Can you move forward assuming that will be fixed?

Sorry, I keep hitting edit on your messages, instead of reply.. I think I've managed to restore your original comment.

Yes. I'm just using 2 text boxes for now. The drop-down just makes it look prettier.

david-vanderson:
> Fffffffinally... I've got some sort of solution to the multiple widgets within a cell. You would not believe the number of promising dead ends I've been chasing.

Oh man sorry for that.  I really appreciate you looking into all the different ways to try this, thank you!

> So, my 5-line solution is to allow the user to provide an optional callback for that mapping function. I know, I know. But once the code is cooked, I'll leave it as your challenge to find a way to get rid of it. :)

Hahaha no problem.  Looking forward to merging this - grid stuff is really hard!

phatchman:
> Oh man sorry for that. I really appreciate you looking into all the different ways to try this, thank you!
> 

Despite some frustrations like that, I'm enjoying trying to break this down and come up with a nice solution. I really don't mind throwing away code that's not the right fit. And I'm trying to see how far I can push the GridWidget without having a processEvents() and implementing it all externally as pluggable behavior. So far it is working out, though I can see situations where it wouldn't.

After doing some experimenting and playing with excel and some web pages, I think this is a sensible set of keys. They make sense on Windows at least! It is mostly a subset of the text entry kb controls, so hopefully it makes sense for Mac as well.)

1) Arrow Up / Down - Move the focus up/down one row. [grid_cell_up, grid_cell_down] (Note sure if these stay as Window keybinds, I'm thinking probably not)

2) Tab / Shift - Move left/right one col. [next_widget, prev_widget]
    - At start/end of col optionally wrap to next/prev row 
    - If tab on last cell, tab out of grid.
    - If shift-tab on first cell, tab out of grid.
    - If tab into grid, focus last cell/widget that had focus.
 
3) PgUp / PgDn - Move the cursor up / down a defined number of rows. [.page_up, .page_down]
    - If cursor is off-page, scroll so it is at top/bottom of scroll area
    - I scroll both the cursor and the scroll container, so the cursor stays roughly in the same position on screen.
    - It needs a little tweaking for when the row height is not an exact multiple of the viewport height, but I think is good enough for v1.0
    
4) Home / End - Goto first / last col on row. (line_start/line_end) 
    - I don't set this by default as text area would usually want to process this instead. But if using "click to edit" or non-editable grids it could be useful.
    - If cursor is off-page, scroll so it is at top of scroll area

5) Ctrl-Home / Ctrl-End - Goto first / last cell (text_start/text_end)
    - If cursor is off-page, scroll so it is at bottom of scroll area.

6) Tab into grid:
    - Restore focus to last cell that had focus
    - If focused widget is not on screen, scroll to bring it to top/bottom of scroll area.
    
7) If click into grid, set cursor to cell of focused widget. 

8) User scrolls focused widget off screen and then hits a key
    - Perform any nav if needed
    - Scroll to bring it to top/bottom of scroll area

Those recent "scroll to focused" changes came in really handy as I didn't need to do anything except make sure the right widget is focused.

 So, I think I'd call this cell navigation functionally complete. I'll clean this up and then flesh out the rest of the demo app so we can add / remove data points.
 
 I'm not going to add specific support for skipping columns with "tab". If user treats it as a "virtual" column cursor, rather than a physical one, they can implement the skipping themselves. It's just the opposite of having 2 widgets in a single cell.
 
 Above functionality is pushed.
 
 I've also removed all of the dvui.gridColumn functions (I don't think they add any value anymore) and updated the examples to the use the latest APIs. 
 
The most interesting one to look at is the gridLayouts() example as that was the example making use of the column functions. I don't think we'll miss them.
 
 I also quite like the new `Join` CellStyle, which lets you combine different styles together and is used in the VirtualScrolling example. Not sure Join is the best name, but best I can come up with so far. Combine?
 ```
    const borders: CellStyle.Borders = .{
        .num_cols = 2,
        .num_rows = num_rows,
        .external = .{ .x = 1, .w = 1, .h = 1 },
        .internal = .{ .w = 1, .h = 1 },
    };

    const cell_style: CellStyle.Join(CellStyle.HoveredRow, CellStyle.Borders) = 
.{ .style1 = highlight_hovered, .style2 = borders };

 ```

vlappa:
These keys are they configurable or hardcoded?

phatchman:
> These keys are they configurable or hardcoded?

All configurable, with defaults.

```
 pub fn defaults() NavigationKeys {
            const cw = dvui.currentWindow();
            return .{
                .up = cw.keybinds.get("grid_cell_up") orelse unreachable,
                .down = cw.keybinds.get("grid_cell_down") orelse unreachable,
                .left = cw.keybinds.get("prev_widget") orelse unreachable,
                .right = cw.keybinds.get("next_widget") orelse unreachable,
                .first = cw.keybinds.get("text_start") orelse unreachable,
                .last = cw.keybinds.get("text_end") orelse unreachable,
                .col_first = .{},
                .col_last = .{},
                .scroll_up = .{ .key = .page_up },
                .scroll_down = .{ .key = .page_down },
            };
        }
```

phatchman:
The interface is [coming together](https://imgur.com/a/lWUpb7j). @david-vanderson was there anything else you were thinking to add?

I'll prob add some sort of sliders and/or whatever those up/down arrow things are called into each cell to give another way to set the number.

This demo might be showing a scrollbar issue? Delete the grid down to a single row. Then add some new rows. You can briefly see the vertical scroll bar flash up as they are added. I've not done any investigation into this yet, but seems similar to the issue from a couple of days ago?

Edit: note to self. Explore whether the kb navigation should store widget data, so it can be created each frame, rather than being "global".

david-vanderson:
> The interface is [coming together](https://imgur.com/a/lWUpb7j). @david-vanderson was there anything else you were thinking to add?

Nope, you've already put tons more functionality into this than I was thinking!

> 
> I'll prob add some sort of sliders and/or whatever those up/down arrow things are called into each cell to give another way to set the number.
> 
> This demo might be showing a scrollbar issue? Delete the grid down to a single row. Then add some new rows. You can briefly see the vertical scroll bar flash up as they are added. I've not done any investigation into this yet, but seems similar to the issue from a couple of days ago?

This shouldn't hold up merging this, I can investigate after.

> 
> Edit: note to self. Explore whether the kb navigation should store widget data, so it can be created each frame, rather than being "global".

This work is excellent!  Merge when ready.  Thank you very much!

phatchman:
This is starting to look [cool](https://imgur.com/a/YndDXWJ) and a little fun to play with.

And it shows how flexible the navigation can be. I've set it up so when you click on the slider, the texbox keeps the focus. But it still allows you to arrow down to all of the sliders or all of the text boxes.

I reckon this one can prob go in the Demo when we're done.

RedPhoenixQ:
@phatchman FYI if you want to push changes [but not run ci](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/skipping-workflow-runs), you can add `[no ci]` to the commit message

phatchman:
OK only one question for you today. So you can have a rest day!  :)

Is there a way to have tab labels wrap in the TabWidget? I'm adding the directory app (using fake data) and plotting app to the examples, and the tabs go off the screen unless you resize the window. 

I tried wrapping the tab labels in a flex box, but that doesn't help because the tabs are in a scroll area.

It's in the pub fn grids(height: f32) in Examples.zig where I'm trying to do this.

P.S. We have some shiny new examples to play with now!! The plot one is particularly nice and really easy to control via kb. 
I'll start polishing this up now and making it into some sort of sensible commit.

david-vanderson:
> Is there a way to have tab labels wrap in the TabWidget? I'm adding the directory app (using fake data) and plotting app to the examples, and the tabs go off the screen unless you resize the window.

The individual labels can be multiline (or whatever you want).  But it sounds like when there are too many tabs, you want them to wrap instead of scrolling?  We don't have that yet sorry.  Your intuition to try a flexbox makes sense, but it'll have to be done inside TabWidget.  https://github.com/david-vanderson/dvui/issues/446

Super exciting!

phatchman:
> But it sounds like when there are too many tabs, you want them to wrap instead of scrolling? 

Yeah, that's exactly right. 

Currently, without using the horizontal scroll wheel, it's hard to select the off-screen tabs. Sometimes only a very small part of the off-screen tab is visible and you need to click in that small area and then repeat for each further off-screen tab.

I'll try wrapping the labels for now. It will probably fit. - Yeah, this works for now!

> Super exciting!

It is! It's certainly been a journey from where I started to where I ended up. That's for sure! But I think I had to build it to work out how to get rid of it. ("it" being the data adapter stuff)

P.S. I think that plot idea you had turned out really well. Because I don't sort the x-axis points, you can make all sorts of interesting shapes with it.

phatchman:
Closing to create a clean PR.

phatchman:
> > > For scroll key events, these are propagated upward if the scroll container is already at its min or max and can't process the event. Does it make sense to do this for other keys? If a text widget is already at end and gets a ctrl-end , what would happen if it left that event unhandled and let the parent widget handle it? What do other ui kits do?
> > 
> > 
> > I'm not sure about this. My guess is that keyboard events should not propagate to containing scroll areas, but mouse events should. That seems to be at least what browsers do?
> 
> Agree, I doubt there is anything to do here, but worth checking. I'll do some testing and see what I find.

We can definitely close this random thought off. I did a little testing of this today, both playing with applications and web sites and by trialing the propagation of events. Basically, nothing else I could find does this and hitting ctrl-end when you are a past the last char of a textbox and having it go to the last cell in the grid is not a great interaction.

phatchman:
> > I can go back at take a look that code and write up an issue for the difficulties I had with the shortcuts. I wasn't able to make much use of the "keybinds" for cut-paste etc, either. I can't remember why.
> 
> If you get a chance that would be useful - I don't know what kinds of problems people are having. Thanks!

OK I had a look at this and here are the issues I was having:
1) I needed to process events to intercept the ALT key before any widgets, but I needed to process the key after any widgets had processed. eg. if the user is in a text entry, I need to know alt is depressed, to display all of the shortcuts on the screen before anything else can swallow it. But if the 's' key is eaten by a text entry, I don't want to do the "alt-s" command.
    - I think now I could just do all of the event processing afterwards and ignored the "handled" on the ALT key event.
    - It doesn't make a lot of sense thinking about it, that ALT would be marked handled by anything. So maybe I'm not so sure why I had to move the processing of keyboard events before all the widgets.
  
2) I need to block processing of shortcuts and certain other keys when modal dialogs are being shown.
    - I think this is also a consequence of processing keys before all other widgets. I'll try and move all of the processing to afterwards.
    
Regarding the keybind issue I mentioned, it was not being able to easily do ALT on Windows and CMD on Mac. But I see there is the ctrl/cmd keybind for that, so that is also a non-issue. 
 
 So, I think maybe just inexperience with DVUI event handling and fumbling around until I got something to work. 
 
 I'll refactor that in the next version to see if there is anything remaining, but I think that covers the main issues.

david-vanderson:
Good feedback!

> 1. I needed to process events to intercept the ALT key before any widgets, but I needed to process the key after any widgets had processed. eg. if the user is in a text entry, I need to know alt is depressed,

This sounds like you want each frame to detect whether ALT is currently down?  If so, the pattern is to (before widgets) look through the events (ignoring handled) and keep state from frame to frame of whether ALT is down or not.  Does that make sense, and if so, does it sound like what you needed?  (If yes, I will add an example to the demo for this kind of thing)

> 2\. I need to block processing of shortcuts and certain other keys when modal dialogs are being shown.

This is interesting.  I'm not sure if this should be built in (dvui already blocks mouse events from going to other windows if there is a modal dialog), or if shortcuts should always use `lastFocusedIdInFrame()`, maybe combined with `focusedSubwindowId() == subwindowCurrentId()`.

I'm not sure I totally understand this situation though - were you having global keyboard shortcuts that didn't care which widget had focus?

phatchman:
> This sounds like you want each frame to detect whether ALT is currently down? If so, the pattern is to (before widgets) look through the events (ignoring handled) and keep state from frame to frame of whether ALT is down or not. Does that make sense, and if so, does it sound like what you needed? (If yes, I will add an example to the demo for this kind of thing)

Yes, this is essentially what I do. I globally keep state about whether the alt key is depressed or if shift is depressed. But definitely an example with keyboard shortcuts would be good, as I think this took me a few attempts to get something working.

> This is interesting. I'm not sure if this should be built in (dvui already blocks mouse events from going to other windows if there is a modal dialog), or if shortcuts should always use `lastFocusedIdInFrame()`, maybe combined with `focusedSubwindowId() == subwindowCurrentId()`.
> 
> I'm not sure I totally understand this situation though - were you having global keyboard shortcuts that didn't care which widget had focus?

In my app, all of the functions were buttons at the bottom of the screen (in the style of a green-screen terminal menu). So yes, these were global, unless a dialog was being displayed. This would be the same as application menu short-cuts e.g. ALT-F for the File menu should work, regardless of what has focus in the app.

So, I just tried the ALT-F in vscode and confirm it does always work, unless focus is a console window or there is a modal dialog showing. 

I don't know for non-modal dialogs as I was struggling to find an example. Most apps use side panels instead of non-modals these days. But ALT-F should definitely work when floating windows are displayed, so I expect this would extend to non-modals as they are basically the same thing.

Having an example with menu shortcuts would be good. I notice windows has started to move to the tooltip approach for shortcuts, rather than underlining the shortcut key. That seems like it would be relatively easy to implement with existing dvui, compared to underlining. And showing the full short-cut key for every menu item as-per MacOS feels quite foreign on Windows and Linux. 

(So does the "backwards" Cancel and OK buttons in the default dvui dialog function, but that's a totally different topic! It does throw me off every time though as "OK", "Cancel" ordering is so deeply engrained in my Windows brain that I sometimes can't stop myself hitting the wrong button, even as I watch myself doing it)

image

david-vanderson:
> So, I just tried the ALT-F in vscode and confirm it does always work, unless focus is a console window or there is a modal dialog showing.

I will think about how this can work.  Maybe we just need a `modalDialogShowing()` kind of function to expose if there is one?

> Having an example with menu shortcuts would be good. I notice windows has started to move to the tooltip approach for shortcuts, rather than underlining the shortcut key. That seems like it would be relatively easy to implement with existing dvui, compared to underlining. And showing the full short-cut key for every menu item as-per MacOS feels quite foreign on Windows and Linux.

Good idea!
> 
> (So does the "backwards" Cancel and OK buttons in the default dvui dialog function, but that's a totally different topic! It does throw me off every time though as "OK", "Cancel" ordering is so deeply engrained in my Windows brain that I sometimes can't stop myself hitting the wrong button, even as I watch myself doing it)

Maybe we need to do this dynamically.  We could expand what Window.init_opts.keybinds affects to include dialog button order as well.

Can you file an issue for each of these?  Thanks!  Edit: just saw the comment about the demo for shortcuts, that's good enough for that one.

PR #374, opened 2025-06-13
No more errors!

RedPhoenixQ:
Relates to #192

This removes almost all errors for the dvui abstraction and its widgets and helpers. It will log errors and use fallbacks or skip some functionality, like rendering if needed, to get through to the next frame. Fundamental functions like some direct interaction with the backend and the rendering function still have errors, as they would rarely be used by users. My reasoning is that if you need/want that low level of control you'd also want to know that something went wrong, without checking for fallback return values.

`dvui.logError` has been introduced as an error logging tool that could be extended with the functionality of reporting stack/error traces of when error occurs, probably behind a build flag. This way most error handling uses the same abstraction making it easy to make these additions. Maybe multiple versions of the function can be added for different severity levels as currently everything is `err`.

This has not been extensively tested on the error conditions, as there are too many and too hard to definitely test all of them. Although some tests using a `std.testing.FailingAllocator` validating that we are not panicking might be nice. Widget stack operations do still panic, as there is no simple way to get a valid fallback pointer for a widget.


david-vanderson (approved):
Super fantastic!  Love it, merge when ready.

I clicked on every single thing in the demo, and only noticed one thing:
* password text entry is showing me "unknown glyph" boxes instead of stars

Are you seeing that?  Can be split into a separate issue if you want me to deal with it.

In any case, I am super grateful that you slogged through the whole codebase updating it.  THANK YOU!

RedPhoenixQ:
> * password text entry is showing me "unknown glyph" boxes instead of stars

Good catch. I accidentally freed the string before displaying it.

We might want to make a tag before this change or something for users to use a newer version than 0.2 but without having to update every dvui call site. We can deal with that after merging this if we want it.

RedPhoenixQ:
I'm gonna hold of on merging right away. Tomorrow or the day after I'll double check the docs and readmes and update them.

david-vanderson:
Both good ideas - thanks!

RedPhoenixQ:
@david-vanderson @phatchman Am I good to merge this, or is there some work in progress you'd like to submit first? Realising there's going to be a lot of merge conflicts if it sits here much longer

phatchman:
@RedPhoenixQ No issues from me. Look forward to it!

david-vanderson:
@RedPhoenixQ Good to go.  I agree - let's get this in before it makes a lot of rework for you.  Thanks!

david-vanderson:
Haha I was going too fast and didn't realize it was already merged.  Way to go!

RedPhoenixQ:
I assumed it was fine since you approved earlier

PR #376, opened 2025-06-13
GridWidget: Fix header layout issue when key events are bubbled to the scroll area.

### LINKED ISSUE #375: GridWidget: Headers don't scroll correctly with arrow keys when a child checkbox has focus
## Description

When a checkbox within the grid has focus and the grid is scrolled with the arrow keys, the headers move up and down during the scrolling.

## Reproduction
1. Open the dvui Grid demo
2. Got to the "Layouts and Data" tab
3. Click in the grid to give focus.
5. Scroll up and down with the arrow keys to confirm scrolling is normal.
6. Select a checkbox
7. Scroll with the arrow keys and observe the header wobbling.    
    - When the down arrow is held, headers are displaced downwards.
    - When the up arrow is held, headers are displaced upwards.
    - When the key is released, headers return to the correct position.
7. Unselect the checkbox and scroll with the arrow keys again (headers still wobble)
8. Click elsewhere in the grid to remove focus from the checkbox.
9. Scroll with the arrow keys and observer the header behavior has returned to normal.

## Discussion

~~Just logging this for now. I still need to determine the root cause of the issue. 
It seems like the headers are reacting a frame after the scroll position has changed. 
Possible timing issue of when the scroll_info is checked?~~

The issue appears to be as follows:
1) Grid creates the column header at the current scroll_info.viewport.y and places header in the correct position.
2) Grid creates all checkboxes for the checkbox column.
3) The focused checkbox processes the key event and bubbles it to the scroll container.
4) Scroll container updates the scroll_info to scroll by 10px. However, the scroll container can't redraw with this new position until the next frame.
5) Subsequent headers are drawn with the +10 pixel offset, putting them in the wrong position for this frame.

I need to come up with a way to reliably get the screen coordinates of the top of the current scroll viewport.

phatchman:
Fixes #375

Make a copy of the scroll_info during install() so that layout is not affected by changes to the viewport during a frame.

### DIFF
--- a/src/widgets/GridWidget.zig
+++ b/src/widgets/GridWidget.zig
@@ -118,6 +118,7 @@ pub const default_col_width: f32 = 100;
 
 vbox: BoxWidget = undefined,
 scroll: ScrollAreaWidget = undefined,
+si: ScrollInfo = undefined,
 hbox: BoxWidget = undefined,
 init_opts: InitOpts = undefined,
 current_col: ?*BoxWidget = null,
@@ -175,6 +176,8 @@ pub fn install(self: *GridWidget) !void {
 
     self.scroll = ScrollAreaWidget.init(@src(), self.init_opts.scroll_opts orelse .{}, .{ .name = "GridWidgetScrollArea", .expand = .both });
     try self.scroll.install();
+    // Keep a copy of the scroll_info in case the viewport changes between column layouts.
+    self.si = self.scroll.si.*;
 
     // Lay out columns horizontally.
     self.hbox = BoxWidget.init(@src(), .{ .dir = .horizontal }, .{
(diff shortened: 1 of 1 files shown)


phatchman:
Is there any better way to do this to ensure the headers will be drawn at the top of the scroll area on screen? 

The headers need to be in the scroll container so that they can horizontal scroll. And if they need to be in the scroll container, I can't think of any other way to draw them than at y position of scroll_info.viewport.y.

david-vanderson:
> The headers need to be in the scroll container so that they can horizontal scroll. And if they need to be in the scroll container, I can't think of any other way to draw them than at y position of scroll_info.viewport.y.

This is a good point that I hadn't properly considered.  Maybe we could have multiple scrollAreas that share the same scroll_info?  Like have a horizontal-only scroll area for the headers, then pass the same scroll_info to the main grid area?

That seems like a promising thing to try.  It might end up with a frame delay for one of them (like the main grid area scrolls horizontally but it takes a frame for the headers to catch up).

That problem and this fix makes me think I got the design a bit wrong.  ScrollContainer already caches the topleft of the scroll viewport for the same reason.

Go ahead with this change, and I will file an issue and think about how to have multiple scroll areas use the same scroll_info.  Thanks!

PR #379, opened 2025-06-14
GridWidget: Fix padding issue with examples

phatchman:
The data and layouts example had inconsistent padding. Because labels and text entries have different default padding.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -4189,7 +4189,14 @@ fn gridLayouts() !void {
 
     const panel_height = 250;
     const banded: GridWidget.CellStyle.Banded = .{
-        .alt_cell_opts = .{ .color_fill = .{ .name = .fill_press }, .background = true },
+        .opts = .{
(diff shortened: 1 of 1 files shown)


phatchman:
@david-vanderson is it intended that LabelWidget has diiferent default margin and padding to TextLayoutWidget to TextEntry widget?

I ask because they don't line up when you mix Labels and Text Entries in the grid.

(I'm sure I already asked this, but can't find it)

david-vanderson:
> @david-vanderson is it intended that LabelWidget has diiferent default margin and padding to TextLayoutWidget to TextEntry widget?
> 
> I ask because they don't line up when you mix Labels and Text Entries in the grid.
> 
> (I'm sure I already asked this, but can't find it)

Nope!  Sounds like they should be the same - want to bump them up to all be the same?

phatchman:
Sure, will do. I'll play with them a bit and see what works best.

PR #380, opened 2025-06-15
SDL and raylib logs forwarded to std.log

RedPhoenixQ:
Mentioned this in https://github.com/david-vanderson/dvui/pull/336#issuecomment-2953203748

This forwards backend logs (from the external library) to std.log so that users can implement implement their own log functions to log to files for example. The scope is different for external library logs and logs from our backend implementation, allowing for filtering their levels separately.

### DIFF
--- a/src/backends/raylib.zig
+++ b/src/backends/raylib.zig
@@ -892,8 +892,18 @@ pub fn EndDrawingWaitEventTimeout(_: *RaylibBackend, timeout_micros: u32) void {
     return;
 }
 
+// Optional: windows os only
+const winapi = if (builtin.os.tag == .windows) struct {
+    extern "kernel32" fn AttachConsole(dwProcessId: std.os.windows.DWORD) std.os.windows.BOOL;
+} else struct {};
+
 pub fn main() !void {
     const app = dvui.App.get() orelse return error.DvuiAppNotDefined;
+
+    if (builtin.os.tag == .windows) { // optional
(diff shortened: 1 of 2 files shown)


RedPhoenixQ:
The raylib error is very confusing. It says `c.va_list` isn't allowed with call convention `x86_64_sysv` but the function specifically specifies `callconv(.c)`? Can only find [this old forum post](https://ziggit.dev/t/dungeon-wizard-action-roguelike-deckbuilder-in-zig/9153/6) about it where they give up on fixing it.

phatchman:
> The raylib error is very confusing. It says `c.va_list` isn't allowed with call convention `x86_64_sysv` but the function specifically specifies `callconv(.c)`? 

That error is talking about the ABI calling conventions. Apparently under the SYSV ABI, you can't pass arrays in the same ways as under Windows.  And I think even on Windows there might be issues under aarch-64.

I can't think of an easy way to do what you want in a Zig function. Zig won't play with the "c" va_start and va_end macros, and if you define it to be an actual varargs function using "...", it won't look like a normal "c" varargs function to raylib.

I think you'll have to write that callback in C and call back into Zig once you've done the vsnprintf? That's the only solution I can think of.

Edit: Confirm it also fails to build on x86-64-macos

RedPhoenixQ:
Thank you for the investigating. I guess that diminishes the value of this PR then. I'll close this with the possibility to add just the SDL function if we want

PR #383, opened 2025-06-15
Add optional stack and error return traces to error logs

RedPhoenixQ:
`-Dlog-stack-trace=12` and `-Dlog-error-trace=false` for example would enable stack traces and limit the amount of stack frames to 12 and also disable error return traces (which are enabled when stack traces are enabled). This makes is possible to locate from where the error originated without creating a panic.

This is disabled by default because of the large amount of text that would be displayed on every error.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -40,9 +40,9 @@ pub fn build(b: *std.Build) !void {
             b.option([]const u8, "image-dir", "Default directory for dvui.testing.saveImage"),
     );
     build_options.addOption(
-        u8,
+        ?u8,
         "log_stack_trace",
-        b.option(u8, "log-stack-trace", "The max number of stack frames to display in error log stack traces (32 shows almost everything)") orelse 0,
+        b.option(u8, "log-stack-trace", "The max number of stack frames to display in error log stack traces (32 shows almost everything, 0 to disable)"),
     );
     build_options.addOption(
         ?bool,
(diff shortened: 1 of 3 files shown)


david-vanderson:
> `-Dlog-stack-trace=0` and `-Dlog-error-trace=false` for example would enable stack traces and limit the amount of stack frames to 12 and also disable error return traces

Typo?  Wouldn't 0 disable stack traces?

Two questions:
* can we default the number of stack frames to a reasonable number in Debug mode, and 0 in Release modes?
* can you give an example where a user would want stack frames but not the error trace (or the other way)?

RedPhoenixQ:
> Typo? Wouldn't 0 disable stack traces?

Definitely a typo, thank you. I mention 12 later in the text, which is what I intended to type. 0 does indeed disable traces.

> * can you give an example where a user would want stack frames but not the error trace (or the other way)?

It was mostly to reduce noise and make the trace logs clearer (there is a lot of text and the delimiters between stack and error traces can be hard to read). An error return trace from within an allocator might not be useful, but a stack trace telling you what function fail would be. 

> * can we default the number of stack frames to a reasonable number in Debug mode, and 0 in Release modes?

Absolutely. There are a couple of places I'd like to change how logging works to not create to much spam, like the `Path.Builder` that would log a full stack trace on every single append that fails. Happy to fix that.

david-vanderson:
> Absolutely. There are a couple of places I'd like to change how logging works to not create to much spam, like the `Path.Builder` that would log a full stack trace on every single append that fails. Happy to fix that.

Awesome - otherwise looks good, merge when you are happy with it.  Thank you!

PR #384, opened 2025-06-16
make text layout have the padding and margin the same as labels

### LINKED ISSUE #381: Consistent widget margins and padding
## Description
The margins and padding between some widgets is inconsistent. This can be especially observed when the widgets are placed in a grid.

## Proposal
Set the following widgets to have to following padding and margin to make them consistent with other widgets.:
* margin = 4
* padding = 6
1) TextLayoutWidget
2) LabelWidget
3) dvui.checkbox()
4) dvui.radio()

## Result

See [linked](https://imgur.com/a/7MgsQXI) for all of the common dvui components placed in a grid with:
1) Default options
2) expand = both
3) border = 1

The first 2 images are with the current margin / padding and the 3rd image is the new margin / padding.

phatchman:
Fixes #381 

TextLayout default margin is removed.
TextLayout default padding set to 6.

### DIFF
--- a/src/widgets/TextLayoutWidget.zig
+++ b/src/widgets/TextLayoutWidget.zig
@@ -24,8 +24,7 @@ const TextLayoutWidget = @This();
 /// 500 if our min width is zero).
 pub var defaults: Options = .{
     .name = "TextLayout",
-    .margin = Rect.all(4),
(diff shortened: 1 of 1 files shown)


phatchman:
Updated the snapshot hashes for the new text layout padding.

PR #387, opened 2025-06-16
Mouse event mods and shift horizontal scroll

RedPhoenixQ:
This started of by adding a way for mouse events to also have access for the current modifiers, for `ctrl+click` for example. The motivating reason was that I wanted to add `shift+scroll` as horizontal scroll as it always annoyed me not to be able to horizontally scroll without using the scroll bar. Not the last received modifiers from key events is stored on the `Window`, similar to `mouse_pt` and used for `ctrl+click` and `shift+scroll`.

There was some confusion about the direction of scroll and positive and negative tick values. The docs stated that positive meant up/left but when using a horizontal scroll wheel, positive meant right. The docs have now been updated accordingly and all current code (that I found) seemed to already handle/expect this.

d59d419bf0ca6b344277d79f86101360d96cd4a2 was also found and fixed to enable this modifier caching for raylib.

### DIFF
--- a/readme-implementation.md
+++ b/readme-implementation.md
@@ -53,7 +53,6 @@ See the code for [pub fn sliderEntry](https://github.com/david-vanderson/dvui/bl
 * tab index
 * storing data from frame to frame
 * intercepting events and forwarding to child widgets
-* tracking ctrl key for ctrl-click
 * drawing a rounded rect
 
 ### One Frame At a Time
--- a/src/Event.zig
+++ b/src/Event.zig
@@ -127,6 +127,8 @@ pub const Mouse = struct {
     // mouse motion will be a touch or .none
     button: enums.Button,
 
+    mod: enums.Mod,
+
     p: dvui.Point.Physical,
     floating_win: dvui.WidgetId,
 };
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -2991,17 +2991,12 @@ pub fn scrollCanvas(comptime data: u8) void {
         dragBox.deinit();
     }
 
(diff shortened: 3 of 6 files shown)


david-vanderson:
Good idea.  This is good to merge with 2 thing:
* The doc comment now says `Positive ticks means scrolling up / scrolling right.` but the internal comment says `Invert ticks so scrolling up takes you left` - which is correct?
* `ctrlOrCommand()` should not check the target, on web we don't know the OS until runtime.  Merge as is, but please file a followup issue for this - we'll have to hook into the keybinds somehow.

Thank you!

RedPhoenixQ:
> * The doc comment now says `Positive ticks means scrolling up / scrolling right.` but the internal comment says `Invert ticks so scrolling up takes you left` - which is correct?

It's perhaps poorly worded but if scrolling up positive and that means to the right, that means that we need to invert the up scroll to make it left. Up and right are the same, we want the opposite.

david-vanderson:
> It's perhaps poorly worded but if scrolling up positive and that means to the right, that means that we need to invert the up scroll to make it left. Up and right are the same, we want the opposite.

Okay, I think I've got it.  A normal horizontal scroll positive means "go right", but we are taking a positive vertical scroll and want to make that mean "go left", so we invert it?

RedPhoenixQ:
> A normal horizontal scroll positive means "go right", but we are taking a positive vertical scroll and want to make that mean "go left", so we invert it?

Exactly

david-vanderson:
Phew!  Good to merge - thanks!

PR #392, opened 2025-06-16
Always report mouse position in physical units

### LINKED ISSUE #390: remove addEventMouseMotion(), use addEventMouseMotionPhysical()
#385 showed that `addEventMouseMotion()` is not doing the correct conversion.

My memory is that it is trying to correct for a backend that is sending us natural coordinates, but also if dvui has extra content scaling that the backend does not know about.

This was always wrong, we should have the backend always report events in physical pixels.

The tricky part here is that dvui might have extra content scaling, so the backend can't just use dvui's natural scale to convert.

RedPhoenixQ:
Closes #390

raylib does not handle highdpi yet, so this issue has not been addressed there.

This was mostly a non issue. SDL seems to work fine without any additional scaling, hinting at the fact that it might have been wrong before or my testing isn't correct. This needs to be double checked before merging.

### DIFF
--- a/src/Window.zig
+++ b/src/Window.zig
@@ -607,28 +607,13 @@ pub fn addEventTextEx(self: *Self, text: []const u8, selected: bool) std.mem.All
     return ret;
 }
 
-/// Add a mouse motion event that the mouse is now at natural pixel pt.  This
-/// is only for a mouse - for touch motion use addEventTouchMotion().
-///
-/// See `addEventMouseMotionPhysical` if you have physical pixel coords.
-///
-/// This can be called outside begin/end.  You should add all the events
-/// for a frame either before begin() or just after begin() and before
-/// calling normal dvui widgets.  end() clears the event list.
-pub fn addEventMouseMotion(self: *Self, pt: Point.Natural) std.mem.Allocator.Error!bool {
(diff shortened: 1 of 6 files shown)


david-vanderson:
Thanks!  Here's my premerge checklist so I don't forget:

- [x] modify app.zig to be able to change Window.content_scale apart from pinch-zoom

Then test below using this script:
* launch *-app with non-zero desktop content scaling
* double content_scale in app
* verify mouse hover over a button away from top-left corner still works

- [x] mac (sdl3, sdl2, raylib, web)
- [x] linux (sdl3, sdl2, raylib, web)
- [x] windows (sdl3, sdl2, raylib, dx11, web)

RedPhoenixQ:
Did something very similar while testing. Everything except raylib works on Windows. Raylib has some weird issue were the area being drawn doesn't fit the window at different content scales.

david-vanderson:
> Did something very similar while testing. Everything except raylib works on Windows. Raylib has some weird issue were the area being drawn doesn't fit the window at different content scales.

The raylib fix I did (same as the sdl one) fixed the mouse coords at least on Mac retina.  When we change dvui's content scaling, then the raylib backend gets all confused.  I'll file a separate issue for it.

david-vanderson:
Thank you!

PR #394, opened 2025-06-17
image draft for #388

### LINKED ISSUE #388: support creating image from raw rgba buffer
Im already doing this like but id love to see it inside dvui, because it depends on inner workings of dvui and dvui updates broke it multiple times: if this gets accepted im happy to open a PR.

```zig
pub fn img_tex_from_row_major_premult_rgba(rgba_normal: []u8, width: u32, height: u32) !dvui.TextureCacheEntry {
    var cw = dvui.currentWindow();
    const hash = dvui.TextureCacheEntry.hash(rgba_normal, 0);
    if (cw.texture_cache.getPtr(hash)) |tce| return tce.*;
    const alloc = cw.arena();
    const pixels: []u8 = try alloc.dupe(u8, rgba_normal);
    defer alloc.free(pixels);
    const pma = dvui.RGBAPixelsPMA.cast(pixels);
    const texture = try dvui.textureCreate(pma, width, height, .linear);
    const entry = dvui.TextureCacheEntry{ .texture = texture };
    try cw.texture_cache.put(cw.gpa, hash, entry);
    return entry;
}
fn draw(self: *ImageWidget2) !void {
    const rect = dvui.placeIn(self.wd.contentRect(), self.wd.options.min_size_contentGet(), .none, self.wd.options.gravityGet());
    const rs = self.wd.parent.screenRectScale(rect);
    if (rs.s == 0) return;
    if (dvui.clipGet().intersect(rs.r).empty()) return;

    const tce = img_tex_from_row_major_premult_rgba(self.img_bytes, self.img_width, self.img_height) catch return;
    try dvui.renderTexture(tce.texture, rs, .{ .rotation = self.wd.options.rotationGet(), .colormod = .{} });
}
```

nat3Github:
implementation draft closes #388 
- moved TextureCacheEntry related fn to TextureCacheEntry namespace as they are mostly used internally so dont have to be in dvui namespace directly they are renamed like
- `TextureCacheEntry.fromImageFile()` for stbi files
- `TextureCacheEntry.fromRgbaPixelPMA()` for raw pixel buffers
- `TextureCacheEntry.fromTvgFile()` for tvg files
- made hash fns inside `TextureCacheEntry` privat, why should they be used externally
- added `TextureCacheEntry.invalidateCachedImage` fn: this can be called manually to ensure the cache is cleared in case the Pixel buffer was updated - i find this solution most practical. hashing the whole buffer is to expensive and calling this for example after rendering a new image into pixel buffer is easy and straightforward imo
 
now im thinking why do we need renderIcon, this could be included in renderImage since an renderIcon renders a tvg image which works like an stbi/pixel buffer images

i dont know why tests are failing, locally zig build test pass.


src/dvui.zig
```diff
@@ -1102,19 +1102,125 @@ pub const TextureTarget = struct {
...
+
+        const texture = try textureCreate(.fromRGBA(pixels), @intCast(w), @intCast(h), .linear);
+
+        const entry = TextureCacheEntry{ .texture = texture };
+        try cw.texture_cache.put(cw.gpa, tex_hash, entry);
+
+        return entry;
+    }
+
+    pub fn fromRgbaPixelPMA(pma: dvui.RGBAPixelsPMA, width: u32, height: u32) !dvui.TextureCacheEntry {
```
RedPhoenixQ, line 1148: The type of the first parameter makes it clear which format the pixels should be. Also adding specific error type.

```suggestion
    pub fn fromPixels(pma: dvui.RGBAPixelsPMA, width: u32, height: u32) Backend.TextureError!dvui.TextureCacheEntry {
```

nat3Github: theoretically we could support other pixel format, this is why i chose this overly specific name. but either way i am fine. unlikely to support anything else than rgba []u8 and we would probably convert to RGBAPixelsPMA anyway

RedPhoenixQ:
The snapshot test fails in CI and has now been fixed on main. Should pass next time.

> now im thinking why do we need renderIcon, this could be included in renderImage since an renderIcon renders a tvg image which works like an stbi/pixel buffer images

Looking at it now, it would make sense for there to be an `icon`/`tvgFile` version added to the union passed to `renderImage`, also including the `IconRenderOptions` in the union type. It is a bit silly to only have the two extra lines in `renderIcon`

david-vanderson:
Looks great!  Merge when ready.

> Looking at it now, it would make sense for there to be an `icon`/`tvgFile` version added to the union passed to `renderImage`, also including the `IconRenderOptions` in the union type. It is a bit silly to only have the two extra lines in `renderIcon`

I agree, but please do it as a separate PR.  Thank you!

Issue #378, opened 2025-06-13
ScrollInfo for multiple scroll areas


david-vanderson:
@phatchman Is running into this problem with Grid.  Ideally we'd like to pull the grid headers out of the main scroll area, but we need them to scroll horizontally (locked to the grid).

Can we reuse a ScrollInfo for two different scroll areas?

If that works, do we need to draw everything using a copy of the ScrollInfo from the "start of the frame" so that scroll events don't affect (for example) the grid body after the grid headers are already drawn?

This suggests that ScrollArea/ScrollContainer might need a mechanism to accept 2 ScrollInfos (one for display, one for events).  Or some other way to separate the two.

This is a tracking issue for me to think about how to fix it.

phatchman:
This might be required to solve a minor issue I discovered with the ScrollTo event. 

A ScrollTo that results in a backward scroll, will scroll the widget to the top of the scroll area, which will then be hidden by the grid header.

david-vanderson:
Yeah we need to fix whatever needs fixing to be able to pull the header out of the scrolling area.

phatchman:
What do you think about adding a scrollTo() to the scroll_ info?

Edit: We already have scrollToOffset(), That might already do what I want? But if I'm doing anything obviously wrong below, let me know, otherwise, I can debug further.

The issue I'm having is that I want to scroll the grid based on keyboard entry, so I want to ScrollTo a particular widget. I can't add a ScrollTo event to the event queue because some widgets "unreachable" on ScrollTo events. (And would prob be added too late anyway). I've tried bubbling the event via the below, but that doesn't seem to work either from within a grid cell. (I haven't had time to trace through why yet)
```
            var scrollto = Event{ .evt = .{
                .scroll_to = .{
                    .screen_rect = if (new_focus_row.? < (current_focus_row orelse 999))
                        text.data().rectScale().r.offsetPoint(.{ .x = 0, .y = -text.data().rectScale().r.h })
                    else
                        text.data().rectScale().r,
                },
            } };
            dvui.parentGet().processEvent(&scrollto, true);
```
So I'm left with:
```
grid.scroll.scroll.processEvent(&scrollto, true);
```
Am I doing something wrong there?

~~The "scroll to" in ScrollContainer only needs the screen rect of the widget, the scroll viewport and the contentRectScale() of the scroll area. So, I think it would be feasible to the share the scrollTo() calculations between the scroll_info and the ScrollContainer.~~

[...]

PR #400, opened 2025-06-18
save rect of the focused widget

phatchman:
What do you think of something like this?

It allows me to write a generic action like below that scrolls the grid to show the currently focused widget, whenever the focused widget changes. 

So, the user would call init() before creating any columns to get the currently focused widget and then if the focus changes during any of the grid columns, scroll to the focused widget. (It probably needs some more checking that the focused widget is actually in the grid, but that is easy to add).

```
pub const ScrollToFocused = struct {
    focused_wid: ?dvui.WidgetId = null,

    pub fn init() ScrollToFocused {
        return .{ .focused_wid = dvui.focusedWidgetId() };
    }

    pub fn performAction(self: *ScrollToFocused, grid: *GridWidget) void {
        const current_focus_wid = dvui.focusedWidgetId();
        if (current_focus_wid != self.focused_wid) {
            self.focused_wid = current_focus_wid;
        } else {
            return;
        }
        const sw = dvui.currentWindow().subwindowFocused();
        if (sw.focused_widgetId != null) {
            var scroll_to: dvui.Event = .{ .evt = .{ .scroll_to = .{
                .screen_rect = sw.focused_widgetRect,
                .over_scroll = false,
            } } };
            // TODO: Add a grid.scrollTo()? Can't just take a row number as the column might not be scrolled to, so needs a rect?
            // Or a better way to bubble the event to the scroll container.
            grid.scroll.scroll.processEvent(&scroll_to, true);
        }
    }
};
```
I'm calling processEvent directly on the scroll container here, because I don't have a child widget I can bubble the event on. But I'll find a way to make that nicer.

Edit: Also add a dvui.focusedWidgetIdAndRect() or similar?

### DIFF
--- a/src/Window.zig
+++ b/src/Window.zig
@@ -143,6 +143,7 @@ pub const Subwindow = struct {
     rect: Rect = Rect{},
     rect_pixels: dvui.Rect.Physical = .{},
     focused_widgetId: ?WidgetId = null,
+    focused_widgetRect: Rect.Physical = .{},
     render_cmds: std.ArrayList(dvui.RenderCommand),
     render_cmds_after: std.ArrayList(dvui.RenderCommand),
     used: bool = true,
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -1375,6 +1375,7 @@ pub fn focusWidget(id: ?WidgetId, subwindow_id: ?WidgetId, event_num: ?u16) void
         if (swid == sw.id) {
             if (sw.focused_widgetId != id) {
                 sw.focused_widgetId = id;
+                sw.focused_widgetRect = .{};
                 if (event_num) |en| {
                     focusRemainingEvents(en, sw.id, sw.focused_widgetId);
                 }
@@ -1390,6 +1391,7 @@ pub fn focusWidget(id: ?WidgetId, subwindow_id: ?WidgetId, event_num: ?u16) void
                         while (true) : (wd = wd.parent.data()) {
                             if (wd.id == wid) {
                                 cw.last_focused_id_this_frame = wid;
+                                sw.focused_widgetRect = wd.rectScale().r;
                                 break;
                             }
 


david-vanderson:
This almost works, we need to handle the last_registered_id_this_frame and tab navigation (which generally happens at the end of a frame)

Now I'm thinking this should be built in to dvui, and any focus change should automatically attempt to scroll to show the newly focused widget?

Let me try that and see how it goes!

RedPhoenixQ:
Should things like button always add their tab index then, not just when they are visible?

david-vanderson:
> Should things like button always add their tab index then, not just when they are visible?

Yes good point - will have to make that change as well.

I'm not quite sure how it would interact with an enhancement I'm thinking about for huge scrolling areas - some kind of size-caching wrapping widget that saves us from having to run all the non-visible widgets.  But will have to figure something out there, maybe there is a way to figure out that a widget is non-visible but "close enough"?

RedPhoenixQ:
> maybe there is a way to figure out that a widget is non-visible but "close enough"?

An `almostVisible` function that pads out the clip rect by some amount and then does the check. That would give some "padding" to the visibility check but not affect anything else

david-vanderson:
> > maybe there is a way to figure out that a widget is non-visible but "close enough"?
> 
> An `amostVisible` function that pads out the clip rect by some amount and then does the check. That would give some "padding" to the visibility check but not affect anything else

Good idea!  So far I can only think of this one situation for it, but probably it should be a normal dvui function.  Sounds like the kind of thing that is generally useful.

david-vanderson:
@phatchman Does that new commit work for you?

phatchman:
> @phatchman Does that new commit work for you?

Wow! Thanks so much.

Yeah, that works for my test case. I need to expand my test a bit so that I can switch from focus within the scroll container to out of it and back, just to cover all bases. 

But it doesn't work for the grid, which isn't a problem with this patch. It is a problem with the headers being in the scroll area and covering the focused widget when tabbing backwards. When I had access to the rect, I could offset the ScrollTo screen_rect when required.

So, I'll probably have to switch my attention back to extracting the headers. It may change the API, so it's probably better to break things now anyway. 

I think I can get something working just by saving the scroll info, so that we know the header and body will be using the same x and w values. And if that doesn't work, at least we'll have a concrete use-case for #378.

PR #401, opened 2025-06-18
PanedWidget: Add init option for the ratio to be set when the widget …

foxnne:
…un-collapses.

This is cool for anything that uses the paned widget and wants to remember the paned locations and have those persist when the widget uncollapses. In my app I can just read from settings for the split ratios and then set the uncollapse ratio to the settings, and then only set the settings value if the window isn't collapsing/collapsed and the behavior is really nice.

Simple change, let me know if you want me to redo this with different names or anything.

Thanks!


david-vanderson:
Perfect - thanks!

PR #404, opened 2025-06-22
GridWidget: Extract headers from grid body plus other improvements

phatchman:
As prophesized, I'm stuck on an issue with the fit to content, and I think it might be an issue with the display of scrollbars in the scroll area? 

The first thing I noticed is that the frame count was hitting 60fps with no input, which I tracked down to the width of the content rect for the grid changing by +/-10px per frame. i.e. a scroll-bar width.

Now this only happens on "Demo Window" screen where it shows mini versions of the demos. If I actually open the demo, the FPS drops back to normal.

I've had to put an extra -10px in the example for the "Styling and sorting" demo as below to stop it. Removing the -10 demonstrates the behaviour.
Setting the vertical scrollbar to always be shown also fixes the issue, no additional -10 is required.
```
        // TODO: Why does this need -10? The columnLayoutProportional is already subtracting the scrollbar width from the content width?
        dvui.columnLayoutProportional(&.{ -1, -1 }, &local.col_widths, outer_hbox.data().contentRect().w - grid_panel_size.w - 10);
        std.debug.print("hbox_ = {d}, grid_w = {d}, widths = {d}\n", .{ outer_hbox.data().contentRect().w, grid.data().contentRect().w, local.col_widths });
```
The -10 does need to happen, but columnLayoutProportional already subtracts that.

The reason I think it might be a scroll area issue is that I saw some strange behaviour when I was playing with different methods of setting the content size for the scroll rect. In cases where the hscroll bar or vscroll bar where not at the edges of the grid content area, I'd see the horizontal and vertical scroll-bars sometimes show and sometimes disappear as I moved the mouse in, out and around the demo windows (I don't remember exactly the scenario, sorry). If I recall correctly, they should not have been showing at all.

Anyway, this isn't holding me up and I can live with the work-around for now. I'll see if I can work out that previous scenario. But if you can find some time to have a look, it would be much appreciated. 

Outside of that the grid demos appear to be working 100%. More testing and a little cleanup required, but is close.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -4018,7 +4018,7 @@ fn gridStyling() void {
         var banding: Banding = .none;
         var margin: f32 = 0;
         var padding: f32 = 0;
-
+        var col_widths: [2]f32 = @splat(0);
         const Banding = enum { none, rows, cols };
     };
 
@@ -4027,11 +4027,21 @@ fn gridStyling() void {
     const row_background = local.banding != .none or local.borders.nonZero();
 
     {
-        var grid = dvui.grid(@src(), .{ .cols = .{ .num = 2 }, .resize_rows = local.resize_rows }, .{
+        var grid = dvui.grid(@src(), .{
+            .cols = .{ .widths = &local.col_widths },
+            .resize_rows = local.resize_rows,
+            //            .scroll_opts = .{ .vertical_bar = .show },
+        }, .{
             .expand = .both,
             .background = true,
             .border = Rect.all(1),
         });
+        // Normally this would just be grid.data().contentRect().w but we want to keep the right-hand panel fixed size.
+        //dvui.columnLayoutFitContent(&.{ 0, 0 }, &local.col_widths, @floor(outer_hbox.data().contentRect().w - grid_panel_size.w));
+        // TODO: Why does this need -10? The columnLayoutProportional is already subtracting the scrollbar width from the content width?
+        dvui.columnLayoutProportional(&.{ -1, -1 }, &local.col_widths, outer_hbox.data().contentRect().w - grid_panel_size.w - 10);
+        std.debug.print("hbox_ = {d}, grid_w = {d}, widths = {d}\n", .{ outer_hbox.data().contentRect().w, grid.data().contentRect().w, local.col_widths });
+
         defer grid.deinit();
         local.resize_rows = false; // Only resize rows when needed.
 
@@ -4042,6 +4052,7 @@ fn gridStyling() void {
             .ascending, .unsorted => .{ 0, 100, 5 },
             .descending => .{ 100, 0, -5 },
         };
+
         std.debug.assert(@mod(end_temp - start_temp, interval) == 0); // Temperature range must be a multiple of interval
 
         // Manually control sorting, so that sort direction is always reversed regardless of
@@ -4215,6 +4226,7 @@ fn gridLayouts() void {
         const column_ratios = [num_cols]f32{ checkbox_w, -10, -10, -7, -15, -30 };
         const fixed_widths = [num_cols]f32{ checkbox_w, 80, 120, 80, 100, 300 };
         const equal_spacing = [num_cols]f32{ checkbox_w, -1, -1, -1, -1, -1 };
+        const fit_window = [num_cols]f32{ checkbox_w, 0, 0, 0, 0, 0 };
         var selection_state: dvui.GridColumnSelectAllState = .select_none;
         var sort_dir: GridWidget.SortDirection = .unsorted;
         var layout_style: Layout = .proportional;
@@ -4352,13 +4364,13 @@ fn gridLayouts() void {
         else
             null;
 
-        const col_widths: ?[]f32 = switch (local.layout_style) {
-            .fit_window => null,
-            else => &local.col_widths,
-        };
+        //        const col_widths: ?[]f32 = switch (local.layout_style) {
+        //            .fit_window => null,
+        //            else => &local.col_widths,
+        //        };
 
         var grid = dvui.grid(@src(), .{
-            .cols = if (col_widths) |widths| .{ .widths = widths } else .{ .num = local.col_widths.len },
+            .cols = .{ .widths = &local.col_widths },
             .scroll_opts = scroll_opts,
             .resize_rows = local.resize_rows,
         }, .{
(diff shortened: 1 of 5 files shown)


david-vanderson:
> I've had to put an extra -10px in the example for the "Styling and sorting" demo as below to stop it. Removing the -10 demonstrates the behaviour.

I *think* this was super confusing and not related to the width of the scrollbar at all.  The code was trying to calculate how much space was left for the grid's content width.  But the grid panel had a border of 1 (meaning 2, 1 on both sides), the grid itself had a border of 1, and also a padding.x = 5.  So 2 + 2 + 5 = 9 so the calculation was just missing those values - padding/border/margin are all added on top of the content size.

This tripped a refresh in BoxWidget, because if there is an expanded child but we don't give out all the space, then the assumption is something changed and we need another frame.  I'm still not entirely sure exactly what was causing the bouncing between 2 different widths.

In any case, I've pushed a commit here that shows my recommendation for how to avoid those kinds of calculations.  If you have a fixed size widget and an expanded one, add the fixed one first, then the expanded one gets the remaining space automatically.

Does that make sense?

phatchman:
Thanks so much for this and that change makes perfect sense. I'm still trying to let go of a few ingrained GUI concepts and the order of widget creation is one of them.

Now `dvui.columnLayoutProportional(&.{ -1, -1 }, &local.col_widths, grid.data().contentRect().w);` works as expected without the additional -10. This is my answer to removing the ability to expand = .horizontal the columns. 

If you don't specify a width by any method, each column will size to the widest cell found in the column. i.e. "size to content". Otherwise, you need to specify a width on at least 1 cell in the column, or provide a col_widths slice populated with the column widths via whatever algorithm you like. Resizable headers, proportional layout function, or some user supplied method. 

P.S. I can recreate the issue I saw before. It is actually that the scrollabars appear for one frame and then disappear when you move the mouse. So, there is maybe a refresh missing somewhere. If I can't track it down, I'll post the repo, but it's basically if the grid doesn't have .expand set on it.

phatchman:
OK. I'm in the middle of adding some tests as I can't easily cover everything I want to test in the Examples.

These aren't really "DOCIMG" tests, though I'll create some of those as well. Does it matter they are still being created in the zig-out\docs directory? Is there some other way I should handle these "internal" tests?

Also is there a way just to run those "dvui_sdl3" tests? i.e. the tests that actually run the code-base tests.

Edit: ~6~ 9 bugs squished already because of these new tests! 
Edit2: Something seems wrong with the ci? 
```
/opt/hostedtoolcache/zig/0.14.0/x64/lib/std/Thread.zig:558:9: error: Unsupported operating system freestanding
        @compileError("Unsupported operating system " ++ @tagName(native_os));
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/hostedtoolcache/zig/0.14.0/x64/lib/std/posix.zig:110:[33](https://github.com/david-vanderson/dvui/actions/runs/15810711492/job/44569227579?pr=404#step:5:34): error: struct 'posix.system__struct_19066' has no member named 'STDERR_FILENO'
pub const STDERR_FILENO = system.STDERR_FILENO;
                          ~~~~~~^~~~~~~~~~~~~~
/opt/hostedtoolcache/zig/0.14.0/x64/lib/std/posix.zig:49:13: note: struct declared here
    else => struct {
            ^~~~~~
/opt/hostedtoolcache/zig/0.14.0/x64/lib/std/posix.zig:1262:26: error: struct 'posix.system__struct_19066' has no member named 'write'
        const rc = system.write(fd, bytes.ptr, @min(bytes.len, max_count));
                   ~~~~~~^~~~~~
```

david-vanderson:
> These aren't really "DOCIMG" tests, though I'll create some of those as well. Does it matter they are still being created in the zig-out\docs directory? Is there some other way I should handle these "internal" tests?

Not sure I understand.  Tests with "DOCIMG" in the name are run for `zig build docs -Dgenerate-images=true`.  Do you mean you are running some other test that generates an image?  Haven't thought about that too much because we found that testing an image against a reference didn't work (can't reliably get pixel-identical images across machines/backends).

What are you trying to do with generating an image during a test?

> Also is there a way just to run those "dvui_sdl3" tests? i.e. the tests that actually run the code-base tests.

I think you want `zig build test -Dbackend=sdl3`?

> Edit: ~6~ 9 bugs squished already because of these new tests!
> Edit2: Something seems wrong with the ci?

Awesome!  Don't understand the ci failure, will look into it.

david-vanderson:
Oh - the ci failure is because wasm can't do stuff like std.debug.print.  Check for debugging statements?

phatchman:
> 
> What are you trying to do with generating an image during a test?

I doing zig build test -Dgenerate-images=true

Have a look at the latest commit, I've just added the tests. If we can't do snapshot testing, then I'm happy just to be able to generate the images, so I that I can visually check or just do a local hash comparison, even if it is not cross-platform. There are a lot of scenarios to cover with row and column orders, having headers or not and specifying inconsistent heights and widths, or not etc. that I can't easily test everything with just the examples.
> 
> > Also is there a way just to run those "dvui_sdl3" tests? i.e. the tests that actually run the code-base tests.
> 
> I think you want `zig build test -Dbackend=sdl3`?

Exactly what I wanted. Thanks!

Edit: If there is nothing existing, that is fine. I can add some more test options to allow generation of images for non DOCIMAGE tests.
I'm thinking even just using that prefix, so you can generate image / hash before and after a change and compare. Then it can still be useful when developing, even if it not during ci?

david-vanderson:
> I'm thinking even just using that prefix, so you can generate image / hash before and after a change and compare. Then it can still be useful when developing, even if it not during ci?

I'm not sure we have the best answer yet, but depending on whether you want these tests to run for everybody we have:
* `-Dimage-dir` - you can specify a directory that `testing.saveImage` will save to, even if you don't put "DOCIMG"
  * This would make sense for locally running tests where you don't want everybody generating these images.
* `-Dsnapshot-images` - if you use `testing.snapshot` then this controls if it saves an image for you to look at
  * This makes sense if you want these tests run all the time, but usually just check the hash.

Do either of those work for you?

phatchman:
For this release I'll add the tests zig file, but not include it the build. That way I can run the tests locally while I work out a better way to integrate them.

I'm pretty sure all the pieces to do what I want are already there, or at least most of them.

P.S. Did you know this is a thing? I just found out you can name the arguments in the fmt string. Don't think I've ever seen anyone use it.
```
    dvui.label(@src(), "{[col_row]}:{[col_row]}", .{ .col_row = col_row }, .{});
```

david-vanderson:
> For this release I'll add the tests zig file, but not include it the build. That way I can run the tests locally while I work out a better way to integrate them.

Good plan.

> 
> I'm pretty sure all the pieces to do what I want are already there, or at least most of them.

We can hope, but you might be the first person to seriously use the testing stuff that wasn't heavily involved in making it. That is excellent, but it might be a bit rocky, so feedback is very welcome!

> P.S. Did you know this is a thing? I just found out you can name the arguments in the fmt string. Don't think I've ever seen anyone use it.
> 
> ```
>     dvui.label(@src(), "{[col_row]}:{[col_row]}", .{ .col_row = col_row }, .{});
> ```

WHAT no I did not.  How did you find that out?

phatchman:
> WHAT no I did not. How did you find that out?

Comments at the top of fmt.zig.
```
/// - *argument* is either the numeric index or the field name of the argument that should be inserted
///   - when using a field name, you are required to enclose the field name (an identifier) in square
///     brackets, e.g. {[score]...} as opposed to the numeric index form which can be written e.g. {2...}
```

RedPhoenixQ:
> 
> -Dimage-dir - you can specify a directory that testing.saveImage will save to, even if you don't put "DOCIMG"
> This would make sense for locally running tests where you don't want everybody generating these images.
> -Dsnapshot-images - if you use testing.snapshot then this controls if it saves an image for you to look at
> This makes sense if you want these tests run all the time, but usually just check the hash.


This is the intention. Because generating images wasn't consistent enough for testing results it was kept for debugging. The DOC_IMG is just a way for dvui to filter out the tests that need to run to generate doc images. It also requires an OS window to work, so needs to be disabled in ci for example. By using saveImage and then specify the directory in the flag, it will be ignored most test runs and not slow down regular tests.

The snapshot current only test the WidgetData, like their rects and some options to make sure same options yield the same logical results. That might be useful for grid as I think would test that the rects are in the same position between test runs. So by manually verifying an image and then taking a snapshot after, you can be sure that the image content shouldn't have changed since the point you took the snapshot.

I'd be happy to take your test file and integrate it how I envisioned it and we can discuss improvements to the testing docs or api. I'm not at my usual setup so it might take a bit.

phatchman:
@RedPhoenixQ That would be really great. Thank you.

One of the things I've just added to the grid tests is the ability to test across "2 frames". Not 2 actually frames, but 2 settled frames. This is useful for testing the row and column resizing options. Didn't need to change anything to get that working.

RedPhoenixQ:
Looking at your test file. I think the only thing I'd change is saveImage -> snapshot and use the build flag to get the images. Please tell me if the snapshots don't work for whatever reason.

phatchman:
Cool. Yeah, I'm going to take a look more into the snapshots. I had to update them when I changed the textArea default margin and padding. Right now, the images themselves are more useful to me until everything settles down a bit.

PR #405, opened 2025-06-22
dx11: implement openURL

nhanb:
It surprised me when openURL() did nothing on Windows, so here's a quick fix.

Confirmed working on both Windows 11 and WINE. I have zero win32 experience, so there's no error handling on ShellExecuteW. I think it's better than nothing though!


david-vanderson:
Thank you!

PR #407, opened 2025-06-24
Grid Widget: Remove headers from body and other improvements

### LINKED ISSUE #403: Move grid headers out body scroll area.
Quick and dirty prototype [here ](https://github.com/phatchman/dvui-dev/compare/upstream_main...remove_vboxes)

~~I still need to split out the 2 scroll_infos completely, so I don't have a good handle on the scrolling smoothness yet.~~ See Edit2.

Would like some advice on the below code in the GridWidget deinit().

```
// TODO: Would rather do this by reporting widget sizes, but can't get that to work yet.
    // Also this won't interact well with virtual scrolling, which will want to set its own height.
    if (self.hscroll) |hscroll| {
        hscroll.si.virtual_size.h = self.header_height;
        hscroll.si.virtual_size.w = sumSlice(self.init_opts.col_widths.?[0..]) + 100;
    }
    if (self.bscroll) |bscroll| {
        bscroll.si.virtual_size.h = self.next_row_y;
        bscroll.si.virtual_size.w = sumSlice(self.init_opts.col_widths.?[0..]);
    }
```
When I removed the vboxes from the columns and started laying out the cells at their x/y location directly, the scroll area doesn't seem to know how high or wide it should be. 
I've tried a few ways to report this size including:
1) via widget data, 
2) setting the min size content on the scroll area

but didn't come up with anything that worked. Any suggestions?

Edit: maybe i just create a box in the scroll area with min size content set to the height and width i need? I'll try that tomorrow.

Edit2: I split out the two scroll_infos and it is not smooth when dragging slowly, headers appear to be +/- 1 pixel compared to body at times. I'm not 100% sure why yet. If both header and body have the same width and both viewports have the same 'x' position, then why would they not draw at the same place?
   - I put some checking in the deinit for ScrollContainer and the ending viewport positions are the same for both widgets. The starting viewport positions are also the same.
   - Can the body scroll further after the header has been drawn?
   - Scrolling with the scroll wheel is smooth.
   - My guess is that the body is processing some scroll events and displaying at the new position and the header does not know about this until the next frame? But if that was the case, why are the ending viewports the same?
   - We will prob need to consider some way to send scroll events from the header scroll area to the body scroll area. e.g. if someone uses a mouse wheel in the header, we want those events ideally to go to the body.
   - Also need to try removing the columnHeader, columnBody functions and just have bodyCell take a row and col. This would allow layout by col and layout by row?
   - Consider using scroll container for header over scroll area. Consider for body as well and manually place scrollbars?

phatchman:
Grid Widget: Remove headers from body and other improvements.
      - Header and body in different, linked scroll areas.
      - Ability to lay out by column, row or combination.
      - Removed the column() function.
      - Current column number is no longer stored internally.
      - Most grid functions now require column numbers to be passed as arguments.
      - General fixes and improvements.
      - Added tests

Fixes #403 
Partially implements #317

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -4020,109 +4020,20 @@ fn gridStyling() void {
         var banding: Banding = .none;
         var margin: f32 = 0;
         var padding: f32 = 0;
-
+        var col_widths: [2]f32 = @splat(0);
         const Banding = enum { none, rows, cols };
     };
 
     var outer_hbox = dvui.box(@src(), .horizontal, .{ .expand = .horizontal });
     defer outer_hbox.deinit();
-    const row_background = local.banding != .none or local.borders.nonZero();
 
-    {
-        var grid = dvui.grid(@src(), .{ .resize_rows = local.resize_rows }, .{
-            .expand = .both,
-            .background = true,
-            .border = Rect.all(1),
-        });
-        defer grid.deinit();
-        local.resize_rows = false; // Only resize rows when needed.
-
-        // Set start, end and interval based on sort direction.
-        const start_temp: i32, //
-        const end_temp: i32, //
-        const interval: i32 = switch (local.sort_dir) {
-            .ascending, .unsorted => .{ 0, 100, 5 },
-            .descending => .{ 100, 0, -5 },
-        };
-        std.debug.assert(@mod(end_temp - start_temp, interval) == 0); // Temperature range must be a multiple of interval
-
-        // Manually control sorting, so that sort direction is always reversed regardless of
-        // which column header is clicked.
-        const current_sort_dir = local.sort_dir;
-
-        const cell_opts: GridWidget.CellStyle.Banded = .{
-            .banding = switch (local.banding) {
-                .none, .rows => .rows,
-                .cols => .cols,
-            },
-
-            .cell_opts = .{
-                .border = local.borders,
-                .background = row_background,
-                .margin = Rect.all(local.margin),
-                .padding = Rect.all(local.padding),
-            },
-            .alt_cell_opts = .{
-                .border = local.borders,
-                .margin = Rect.all(local.margin),
-                .padding = Rect.all(local.padding),
-                .background = row_background,
-                // Only set the alternate fill colour if actually banding.
-                .color_fill = if (local.banding != .none) .fill_press else null,
-            },
-        };
-
-        // First column displays temperature in Celcius.
-        {
-            var col = grid.column(@src(), .{});
-            defer col.deinit();
-            // Set this column as the default sort
-            if (current_sort_dir == .unsorted) {
-                local.sort_dir = .ascending;
-                grid.colSortSet(local.sort_dir);
-            }
-
-            if (dvui.gridHeadingSortable(@src(), grid, "Celcius", &local.sort_dir, .fixed, .{})) {
-                grid.colSortSet(current_sort_dir.reverse());
-            }
-
-            var temp: i32 = start_temp;
-            var row_num: usize = 0;
-            while (temp != end_temp + interval) : ({
-                temp += interval;
-                row_num += 1;
-            }) {
-                var cell = grid.bodyCell(@src(), row_num, cell_opts.cellOptions(grid.col_num, row_num));
-                defer cell.deinit();
-                dvui.label(@src(), "{d}", .{temp}, .{ .gravity_x = 0.5, .expand = .horizontal });
-            }
-        }
-        // Second column displays temperature in Farenheight.
-        {
-            var col = grid.column(@src(), .{});
-            defer col.deinit();
-            if (dvui.gridHeadingSortable(@src(), grid, "Fahrenheit", &local.sort_dir, .fixed, .{})) {
-                grid.colSortSet(current_sort_dir.reverse());
-            }
-
-            var temp: i32 = start_temp;
-            var row_num: usize = 0;
-            while (temp != end_temp + interval) : ({
-                temp += interval;
-                row_num += 1;
-            }) {
-                var cell = grid.bodyCell(@src(), row_num, cell_opts.cellOptions(grid.col_num, row_num));
-                defer cell.deinit();
-                dvui.label(@src(), "{d}", .{@divFloor(temp * 9, 5) + 32}, .{ .gravity_x = 0.5, .expand = .horizontal });
-            }
-        }
-    }
     {
         var outer_vbox = dvui.box(@src(), .vertical, .{
             .min_size_content = grid_panel_size,
             .max_size_content = .size(grid_panel_size),
             .expand = .vertical,
             .border = Rect.all(1),
+            .gravity_x = 1.0,
         });
         defer outer_vbox.deinit();
 
@@ -4190,6 +4101,101 @@ fn gridStyling() void {
             }
         }
     }
+
+    const row_background = local.banding != .none or local.borders.nonZero();
+
+    {
+        var grid = dvui.grid(@src(), .colWidths(&local.col_widths), .{
+            .resize_rows = local.resize_rows,
+        }, .{
+            .expand = .both,
+            .background = true,
+            .border = Rect.all(1),
+        });
+        defer grid.deinit();
+
+        // Layout both columns equally, taking up the full width of the grid.
+        dvui.columnLayoutProportional(&.{ -1, -1 }, &local.col_widths, grid.data().contentRect().w);
+
+        local.resize_rows = false; // Only resize rows when needed.
+
+        // Set start, end and interval based on sort direction.
+        const start_temp: i32, //
+        const end_temp: i32, //
+        const interval: i32 = switch (local.sort_dir) {
+            .ascending, .unsorted => .{ 0, 100, 5 },
+            .descending => .{ 100, 0, -5 },
+        };
+
+        std.debug.assert(@mod(end_temp - start_temp, interval) == 0); // Temperature range must be a multiple of interval
+
+        // Manually control sorting, so that sort direction is always reversed regardless of
+        // which column header is clicked.
+        const current_sort_dir = local.sort_dir;
+
+        const cell_opts: GridWidget.CellStyle.Banded = .{
+            .banding = switch (local.banding) {
+                .none, .rows => .rows,
+                .cols => .cols,
+            },
+
+            .cell_opts = .{
+                .border = local.borders,
+                .background = row_background,
+                .margin = Rect.all(local.margin),
+                .padding = Rect.all(local.padding),
+            },
+            .alt_cell_opts = .{
+                .border = local.borders,
+                .margin = Rect.all(local.margin),
+                .padding = Rect.all(local.padding),
+                .background = row_background,
+                // Only set the alternate fill colour if actually banding.
+                .color_fill = if (local.banding != .none) .fill_press else null,
+            },
+        };
+
+        if (dvui.gridHeadingSortable(@src(), grid, 0, "Celcius", &local.sort_dir, .fixed, .{})) {
+            grid.colSortSet(0, current_sort_dir.reverse());
+        }
+
+        if (dvui.gridHeadingSortable(@src(), grid, 1, "Fahrenheit", &local.sort_dir, .fixed, .{})) {
+            grid.colSortSet(1, current_sort_dir.reverse());
+        }
+
+        // First column displays temperature in Celcius.
+        {
+            // Set this column as the default sort
+            if (current_sort_dir == .unsorted) {
+                local.sort_dir = .ascending;
+                grid.colSortSet(0, local.sort_dir);
+            }
+
+            var temp: i32 = start_temp;
+            var row_num: usize = 0;
+            while (temp != end_temp + interval) : ({
+                temp += interval;
+                row_num += 1;
+            }) {
+                var cell = grid.bodyCell(@src(), 0, row_num, cell_opts.cellOptions(0, row_num));
+                defer cell.deinit();
+                dvui.label(@src(), "{d}", .{temp}, .{ .gravity_x = 0.5, .expand = .horizontal });
+            }
+        }
+        // Second column displays temperature in Farenheight.
+        {
+            var temp: i32 = start_temp;
+            var row_num: usize = 0;
+            while (temp != end_temp + interval) : ({
+                temp += interval;
+                row_num += 1;
+            }) {
+                var cell = grid.bodyCell(@src(), 1, row_num, cell_opts.cellOptions(1, row_num));
+                defer cell.deinit();
+                dvui.label(@src(), "{d}", .{@divFloor(temp * 9, 5) + 32}, .{ .gravity_x = 0.5, .expand = .horizontal });
+            }
+        }
+    }
 }
 
 fn gridLayouts() void {
@@ -4221,6 +4227,7 @@ fn gridLayouts() void {
         const column_ratios = [num_cols]f32{ checkbox_w, -10, -10, -7, -15, -30 };
         const fixed_widths = [num_cols]f32{ checkbox_w, 80, 120, 80, 100, 300 };
         const equal_spacing = [num_cols]f32{ checkbox_w, -1, -1, -1, -1, -1 };
+        const fit_window = [num_cols]f32{ checkbox_w, 0, 0, 0, 0, 0 };
         var selection_state: dvui.GridColumnSelectAllState = .select_none;
         var sort_dir: GridWidget.SortDirection = .unsorted;
         var layout_style: Layout = .proportional;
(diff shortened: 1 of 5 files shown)


phatchman:
This is ready to go. Though I might sit on it for a bit, given the scope of the changes.

Are you OK with adding eq to Size? I can bring that locally instead, but I thought it would be nice if all comparisons have a consistent tolerance.

There is only one known issue, which is demonstrated in the test "GridWidget: col widths" test.
If this test is run without .expand, the frame never settles, and the content width changes +/-10px. I've had other instances of this behaviour that were my fault, so likely this one is as well. I doubt people will use the grid without .expand, so I'm not too concerned on that one unless it can be triggered some other way.

Thanks vm for all the support, especially with the linked scrolling. That made it much nicer to implement. I think this is looking really good now! (I thought that before too, but this is a good improvement).

Ahh, yes. It passes the test that originally started all of this renovation work. When you shift-tab to backward scroll the grid, you can now see the [focused widget](https://imgur.com/a/VfRw06p)!!

david-vanderson:
> This is ready to go. Though I might sit on it for a bit, given the scope of the changes.

Looks good to me!

> Are you OK with adding eq to Size? I can bring that locally instead, but I thought it would be nice if all comparisons have a consistent tolerance.

I'm not sure.  It looks like you are using it to detect if you need to refresh.  Can you walk me through the situation that requires that?

I also want to understand what that bug with no expand is coming from.  Might be connected, and we should be able to have a grid that properly reports a min size so it could be used in a non-scrolled environment.  I'll look at it.

david-vanderson:
> There is only one known issue, which is demonstrated in the test "GridWidget: col widths" test.
> If this test is run without .expand, the frame never settles,

I wasn't able to reproduce this quickly - I changed it to .expand = .none and the test passed.  Any details about how to reproduce this?

phatchman:
> I wasn't able to reproduce this quickly - I changed it to .expand = .none and the test passed. Any details about how to reproduce this?

Hmmm. It now works for me too, both in the test and as a standalone. I must have somehow fixed it with some other change without realising it. Sorry about that. 

Why do I do the eq check on size?
1) I use self.bbox to create a widget to report the "body" size to the scroll container.
2) The self.bbox is created with a min_size_content of self.last_size, which was calculated last frame in deinit().
4) The deinit() calculates the size of the box necessary to contain cell (0,0) and cell (max_row, num_cols).
5) If self.bbox's size is not the same as the calculated bbox size from this frame's deinit(), then a refresh is required to show the correct size next frame.

I used this method so there is a single widget that can report the size to the scroll container. I don't want to modify the virtual_size in case the user wants to control that themselves via their own scroll_info.

So, I think this eq() check is required because the grid is the only thing that knows how big that bbox should be next frame. The cells are laid out by rect instead of by the bbox, so it can't calculate its own size.

I've just pushed a commit which adds a "remove rows" test that shows the issue of removing that size check. If the number of rows this frame is less than last frame, the grid won't shrink back to the correct size until a refresh.
`zig build test -Dbackend=sdl3 -Dgenerate-images=true -Dtest-filter="widgets.GridWidget.testing.test.remove rows"`

Edit: Modified that check slightly with the last commit. It now also includes a check for init_opts.resize_cols.

david-vanderson:
> I used this method so there is a single widget that can report the size to the scroll container. I don't want to modify the virtual_size in case the user wants to control that themselves via their own scroll_info.

This is interesting.  Is there a usecase you have in mind for having the virtual size not be the same as the grid size?

phatchman:
> This is interesting. Is there a usecase you have in mind for having the virtual size not be the same as the grid size?

The most extreme example would be virtual scrolling with variable row heights. In that case the grid has no idea how big to set the virtual size and would set the size of bbox to surround the size of the rendered rows only. The user will need to pass their own scroll info and set the virtual size manually if they need that.

So, the question from me is should the grid be changing the virtual size of the body scroll container via scroll_info.virtual_size, over using the box widget to do that? (Unless I'm missing some other way to set the virtual size?)

1) I think the current approach mimics what would happen in the scroll container if the widgets were being placed "normally", rather than by rect. Normally, the bounding box would report its size to the parent scroll container, the same as here.
    - Though I think in the "normal" case, the box will know it's size a frame earlier, because the box size is calculated in deinit(), after everything is packed, rather than being set in init() with last frame's size like I do now.

2) Setting the virtual size directly, gets messy when the user passes their own scroll info. e.g. if they don't pass .given, the grid would have to change the passed in scroll_info to be .given in both directions and then set the virtual size. Or maybe it would need to take a copy of the user's scroll info, set .given and then copy the values back to the user's scroll_info in .deinit. Also handle the cases where they only pass .given for one axis.

The self.bbox approach seemed cleaner to me than potentially modifying the scroll_info passed in from the user, especially if they are relying on the values in scroll_info to be up to date before grid.deinit() is called.

phatchman:
After updating some of my old test apps to use the new api, I think I'll change how var_row_heights works. 
Currently user-supplied heights are ignored unless var_row_heights = true; (Something I mention 3 separate times in the comments, which should have been a clue that I was doing something strange!).

I want to specify row heights even if they aren't variable. In case of inconsistent heights I'll take the max as I do with widths. 
If var_row_heights = true, then same functionality as now.

(This change doesn't impact the above discussion)

Change done. Tests proved their worth.

david-vanderson:
> The most extreme example would be virtual scrolling with variable row heights. In that case the grid has no idea how big to set the virtual size and would set the size of bbox to surround the size of the rendered rows only. The user will need to pass their own scroll info and set the virtual size manually if they need that.

Still a bit confused.  Even with variable row heights, the grid still knows the total height of all the rows right?

> So, the question from me is should the grid be changing the virtual size of the body scroll container via scroll_info.virtual_size, over using the box widget to do that? (Unless I'm missing some other way to set the virtual size?)

Might be that we are missing some api needed here.

Here's how I would expect it to work at a high level (this is what ScrollContainer does):
* If .given, the grid uses that value and does not change it
* If .auto, the grid sets the min_size and virtual_size to the sum of all the row/col sizes
* .none is the same as .auto, but also the virtual_size is always set to the content size in install()

Since scrollContainer is already doing that, grid needs to communicate the virtual size to it (and then it will do the "right" thing depending on the ScrollInfo settings).  Having an empty box "bbox" of the right size sounds like a good way to do it.  Otherwise can figure out an api to directly communicate that information.

We could run "bbox" in grid.deinit() just before running deinit on the scrollContainer.  I think that addresses question 1.

Question 2 is confusing to me, I don't understand why we'd have to change the scrolling setting?  What am I missing here?

> The self.bbox approach seemed cleaner to me than potentially modifying the scroll_info passed in from the user, especially if they are relying on the values in scroll_info to be up to date before grid.deinit() is called.

I'm not sure if we should be making any guarantees about the values in scroll_info during the extend of grid (for user code outside grid).  Do you know of a situation where that is necessary?

phatchman:
> Still a bit confused. Even with variable row heights, the grid still knows the total height of all the rows right?

Not with virtual scrolling it doesn't. It only knows the height of what was actually rendered. (I should add I don't think this matters to the discussion. The user will provide a .given to scroll_info in that case, which will override whatever the grid does anyway)
> 
> Since scrollContainer is already doing that, grid needs to communicate the virtual size to it (and then it will do the "right" thing depending on the ScrollInfo settings). Having an empty box "bbox" of the right size sounds like a good way to do it. Otherwise can figure out an api to directly communicate that information.

I haven't found any reason not to use the box so far. If I do, it should be fairly easy to add something to the scroll container. probable not min_size_content?, but something similar. (Actually, should min_size_content work? I don't see it used anywhere in the scroll container. Could we use that to let the grid set a virtual size?)

> 
> We could run "bbox" in grid.deinit() just before running deinit on the scrollContainer. I think that addresses question 1.

Not sure what you mean by run "bbox"?
> 
> Question 2 is confusing to me, I don't understand why we'd have to change the scrolling setting? What am I missing here?
> 

This is assuming there is no new API to set the virtual size and we are not setting the size via the self.bbox.
If the grid were to set the virtual size directly through scroll info, then it could only do that if scroll info has .given for that axis right? So, if the user passed in a scroll info without .given, meaning they wanted the virtual size automatically calculated, the grid would have to change it from .auto to .given to be able to set the virtual size. 
So that's the bit I don't like, I'd have to muck with the user's scroll_info.
> 
> I'm not sure if we should be making any guarantees about the values in scroll_info during the extend of grid (for user code outside grid). Do you know of a situation where that is necessary?

If the user is supplying their own scroll info, they might call pageDn() or something and then expect to see the results of that in their scroll_info. 

But using the box or adding a new way to set the scroll area's virtual size avoids all of that and there wouldn't be any need to modify the user's scroll_info from the grid.

phatchman:
It sounds like that eq() on Size is probably not going to be generally useful. So, I'll take that out of Size and move it to the grid before I merge.

david-vanderson:
Thanks for bearing with me on this stuff.

> Not with virtual scrolling it doesn't. It only knows the height of what was actually rendered.

Does "virtual scrolling" mean when the user passes .given?  If the user passes .auto, then the grid has to provide a height for virtual_size.  Do you mean that height will only include rows where a cell was rendered?  But then how would you know where to put the cells?  Clearly I'm missing something here.

> I haven't found any reason not to use the box so far. 
> Not sure what you mean by run "bbox"?

I mean "bbox" as in the box you are using for calculating the min size (which is the virtual size) of the grid.  Because all the body cells are being positioned with .rect, they don't participate in the min size calculation for the box.  So they don't have to be run inside box at all.  The sole purpose of the box is to call `minSizeForChild` on ScrollContainer, so the whole of box can run just before we call ScrollContainer.deinit() (which is effectively the same as directly setting the virtual size) - the box just needs the right .min_size_content  - does that make sense?

> If the grid were to set the virtual size directly through scroll info, then it could only do that if scroll info has .given for that axis right?

No - .given means that grid won't touch the virtual_size.  .auto and .none means the user is asking the grid to set the virtual_size.  If that makes sense, is there some documentation we can work on to make this clearer?

> If the user is supplying their own scroll info, they might call pageDn() or something and then expect to see the results of that in their scroll_info.

Good point, we need to clarify what the expectations are.  Not totally sure, but my guess is that we would say that any virtual_size changes are logically made during grid.deinit(), so if they called pageDn() before that it would operate on the previous frame's virtual_size.  What do you think?

> But using the box or adding a new way to set the scroll area's virtual size avoids all of that and there wouldn't be any need to modify the user's scroll_info from the grid.

I agree, let's run with the box unless/until it doesn't work for some reason.

phatchman:
> Thanks for bearing with me on this stuff.

I'm surprised you can keep up with everything!
>Does "virtual scrolling" mean when the user passes .given? If the user passes .auto, then the grid has to provide a height for virtual_size. Do you mean that height will only include rows where a cell was rendered? But then how would you know where to put the cells? Clearly I'm missing something here.

By virtual scrolling, yes I mean the user passing .given. But they might only pass .given in one direction, so the grid still needs to manage the other direction.
- For consistent row heights, the grid can lay out based on row_number * row_height, so knows the starting offset, but doesn't know the total number of rows, only the row number of the last rendered cell.
- For variable row heights, the user must set the starting y offet using offsetRowsBy() and the grid will always place the next cell a the column immediately below the previous one, regardless of what the user passes as row_num. That is done through this_row_y and next_row_y, which are only used with variable row_heights. 

So yes, height will only include from (0,0 to the last rendered cell, but that will get overridden by the scroll_info's virtual size passed in by the user. All of the above already works and I don't think really matters to the discussion except to confirm that the user can set virtual size using .given and the grid shouldn't mess with that. Which we both agree on.

> the whole of box can run just before we call ScrollContainer.deinit() (which is effectively the same as directly setting the virtual size) - the box just needs the right .min_size_content - does that make sense?
> 
Yes! That's a great idea. It didn't occur to me that self.bbox didn't need to parent the cells.

> 
> No - .given means that grid won't touch the virtual_size. .auto and .none means the user is asking the grid to set the virtual_size. If that makes sense, is there some documentation we can work on to make this clearer?

Yep, I understand that bit I assume we are just talking past each other here. In summary I was just trying to say that:
I only know 2 ways of setting currently the virtual size on the scroll container:
1) By using a child widget to report the size.
2) By setting it directly on the scroll_info virtual_size.

Option 2 seems bad because if the user pases in a scroll_info with .auto on any axis, the grid would need to change that to .given when passing to the scroll_container so it could set the virtual size on the user's behalf.  So, doing option (1) or providing an alternative API to set virtual size is the better solution to me.

> I agree, let's run with the box unless/until it doesn't work for some reason.

Cool. I'll make the changes to when self.bbox is created/deinited and move Size.eq() and then it should be good to merge.

david-vanderson:
Excellent, that helps.  I think I was confused because why would you have to change ScrollInfo to .given?  But I get it now - you were saying that was one way to communicate the virtual_size to ScrollContainer.  Sorry about that.

Yeah always hard to clear up confusion via text but I very much appreciate it.

Good plan - thanks!

phatchman:
> Good plan - thanks!

Thinking.. Would it be better to use a WidgetData for reporting the size, instead of a box? Like was done with the DragHandle?
And then do something like:
```
        self.wd.minSizeSetAndRefresh();
        self.wd.minSizeReportToParent();
```
Functionally it is probably the same right? And I'm not using any functionality that the box provides except reporting the size. I'll give it a try and see what happens I suppose.

P.S. Moving that box processing to deinit() removes the need for the size check. Things should settle 1 frame earlier now.

david-vanderson:
> Thinking.. Would it be better to use a WidgetData for reporting the size, instead of a box? Like was done with the DragHandle?

Yes, in fact now that you say it, you can call `dvui.spacer`, it does everything you want.

> P.S. Moving that box processing to deinit() removes the need for the size check. Things should settle 1 frame earlier now.

Nice!

phatchman:
> Yes, in fact now that you say it, you can call `dvui.spacer`, it does everything you want.

Ahh, almost exactly the code I'd just written in a nice handy function! Changed. Will merge.

phatchman:
Noting one minor issue with the latest changes.

- In the tests, sometimes room is left for a vertical scrollbar on the right-hand side from 1 frame and other-times not.
- It looks like this happens when resizing, and needs another frame for the scrollbar padding to disappear?

Needs further investigation, but doubt users will actually see this unless they are resizing headers. 
They can work-around with refresh().

2nd minor issue:
- In my example with checkbox columns, it was not drawing any checkboxes on the first frame as it didn't know how many directory entries were created yet.
- This resulted in the widget id's for the other cell's changing on the 2nd frame when the length of the selections was know.
- Changing to populate the selection column last fixes this.
- Issue is that the parent id is different. Need to investigate why.
- Having widget id's change when some columns aren't populated is going to cause issues.

david-vanderson:
> * In the tests, sometimes room is left for a vertical scrollbar on the right-hand side from 1 frame and other-times not.

I just saw this yesterday in menus, so possible I caused the bug and it's unrelated to grid stuff.  I'm looking at it.

> * Having widget id's change when some columns aren't populated is going to cause issues.

I don't get this - each body cell has a unique id based off the row/col right?  How does it matter if some cells are populated?  Walk me through why the ids changed?

phatchman:
> I don't get this - each body cell has a unique id based off the row/col right? How does it matter if some cells are populated? Walk me through why the ids changed?

Yes, you are right. I don't get it either. I need to break out the debugger.
It took me a while to work out the issue I was seeing with column resizing was because the cell's parent had a different id, giving the cell a different id.

david-vanderson:
> > * In the tests, sometimes room is left for a vertical scrollbar on the right-hand side from 1 frame and other-times not.
> 
> I just saw this yesterday in menus, so possible I caused the bug and it's unrelated to grid stuff. I'm looking at it.

I've pushed a commit to fix this.  Can you verify if that fixed the issue for you?

phatchman:
> > > * In the tests, sometimes room is left for a vertical scrollbar on the right-hand side from 1 frame and other-times not.
> > 
> > 
> > I just saw this yesterday in menus, so possible I caused the bug and it's unrelated to grid stuff. I'm looking at it.
> 
> I've pushed a commit to fix this. Can you verify if that fixed the issue for you?

Yep that [fixes it](https://imgur.com/a/xVhK9FW), except for the ["vary header height" test](https://imgur.com/a/t78bH1C). I'll try and grab more details after I've figured out what this parent id problem is.

phatchman:
> > I don't get this - each body cell has a unique id based off the row/col right? How does it matter if some cells are populated? Walk me through why the ids changed?
> 
> Yes, you are right. I don't get it either. I need to break out the debugger. It took me a while to work out the issue I was seeing with column resizing was because the cell's parent had a different id, giving the cell a different id.

OK. Tracked this down. When creating the body scroll container I was using "src" instead of @src(). The body scroll container is created with the first body cell, which made its widget id dependent on the source location of where the first body cell was created. 

BTW being able to walk the widget tree certainly helped in debugging this as I needed to find which widget in the tree was the first to change wid between frames.

Fixed by #416

david-vanderson:
Great!

> BTW being able to walk the widget tree certainly helped in debugging this as I needed to find which widget in the tree was the first to change wid between frames.

This is good feedback, can you give more details?  Were you walking the tree via code, or the debug window?

phatchman:
> This is good feedback, can you give more details? Were you walking the tree via code, or the debug window?

I just put a loop in the code to print out the name, id and parent id. I've not really looked at the debug stuff or what it provides as generally I can just put a print statement anywhere in the dvui code and find out what I need. Even when I was developing that ADGUI app, I generally took that approach. 

That's not really a viable "dvui user" debugging strategy though, so I'll try out the debug facilities next time and see how it goes.

phatchman:
> Yep that [fixes it](https://imgur.com/a/xVhK9FW), except for the ["vary header height" test](https://imgur.com/a/t78bH1C). I'll try and grab more details after I've figured out what this parent id problem is.

OK If I put a dvui.refresh() in that test, it works. Here is the series of frames.
- [prior to refresh](https://imgur.com/a/kSTKA2L)
- [after refresh](https://imgur.com/a/CZ0X5Ic)

Compare that to the [vary row height](https://imgur.com/a/ZJXaQeI) test.

david-vanderson:
> That's not really a viable "dvui user" debugging strategy though, so I'll try out the debug facilities next time and see how it goes.

True, but hopefully this kind of bug is solely the responsibility of a developer, so print statements are fair game!

> OK If I put a dvui.refresh() in that test, it works. Here is the series of frames.

Good!

PR #410, opened 2025-06-26
removing bubbling

### LINKED ISSUE #346: lastFocusedIdInFrame(): alternative to bubbling events
This is a possibly crazy idea I had while chatting with @vlappa about how to handle key events.

The reason we bubble key events up the parent widget chain is to support the same kind of fallback handling that mouse events get.  We want a parent to be able to handle any event that a child widget **could have handled but decided not to**.

For mouse events this works because a parent rectangle contains its children.  So it knows when a mouse event would have been handled by a child but wasn't by checking the mouse event point is inside it's rectangle.  In the case of a child not being contained (one of the floating widgets), we don't expect an unhandled click in a floating widget to be handled in the parent.  Floating widgets generally also stop key events from bubbling through them.

For key events we didn't have any mechanism to tell if any of the child widgets were focused.  But now we do.

I want to explore if it's possible to use `lastFocusedIdInFrame()` as a replacement to bubbling events.

I'm not sure what to do with the non-key bubbleable events, but since they are widget specific, maybe those widgets maintain their own chain to handle those.  FloatingMenuWidget already does that with `popup_current`.

david-vanderson:
Implements #346 

- [X] remove key events from bubbling
- [x] remove other bubbled events (maybe widget-specific chains?)
- [x] see if we still need a widget parent chain for anything
- [X] Look over readme and readme-implementation and correct references to "bubbling"
- [ ] See if we need to add any section to the docs about how to handle children's mouse/key events

### DIFF
--- a/readme-implementation.md
+++ b/readme-implementation.md
@@ -94,7 +94,6 @@ Now the widget is the parent widget, so further widgets nested here will be chil
 * `processEvents()`
   * loop over `events()`, call `matchEvent()` for each
   * call `Event.handle()` if no other widget should process this event
-  * bubble event to parent if `Event.bubbleable()`
 
 See the Event Handling section for details.
 
@@ -130,10 +129,10 @@ There is always a single parent widget.  `parentSet()` is how a widget sets itse
 
 Each widget keeps a pointer to its parent widget, which forms a chain going back to `dvui.Window` which is the original parent.  This chain is used for:
 * `parent.screenRectScale()` translate from our child Rect (in our parent's coordinate space) to a RectScale (in screen coordinates).
-* `parent.processEvent()` bubble keyboard events, so pressing the "up" key while focused on a button can make the containing scroll area scroll.
+* `parent.minSizeForChild()` to inform the parent of the final size of a child for layout purposes.
 
 #### Opting Out of Normal Layout
-If `Options.rect` is set, the widget is directly specifying its position and size (still in parent coordinates).  In this case, it does not call `parent.rectFor()` nor `parent.minSizeForChild()`, which means it is invisible to its parent for layout purposes.  The parent can still receive bubbled events.
+If `Options.rect` is set, the widget is directly specifying its position and size (still in parent coordinates).  In this case, it does not call `parent.rectFor()` nor `parent.minSizeForChild()`, which means it is invisible to its parent for layout purposes.
 
 ### Windows and Subwindows
 `dvui.Window` maps to a single OS window.  All widgets and drawing happen in that window.
@@ -203,8 +202,6 @@ Most events are either mouse (includes touch) or keyboard:
   * have the id of the last focused widget
   * have the id of the last focused subwindow (each subwindow has a focused widget)
   * we want the focused widget to process the event
-  * if it doesn't, it bubbles the event up the parent chain
-    * example: button has focus, pressing the "up" key can bubble to a containing scroll area
 
 Special Events
 * `.focus`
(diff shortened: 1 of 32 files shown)


RedPhoenixQ:
> see if we still need a widget parent chain for anything

Won't this always be needed for child size reporting and to get parent RectScale?

david-vanderson:
> > see if we still need a widget parent chain for anything
> 
> Won't this always be needed for child size reporting and to get parent RectScale?

Yes that is right, but we might only need the immediate parent.  Possibly we could remove the `parent` field from WidgetData.  A child might only need to interact with the parent at the start and end of its life.  Does that make sense?  (I'm still unsure myself)

RedPhoenixQ:
So it could be replaced with a parent global instead, still set in widget install and deinit?

david-vanderson:
> So it could be replaced with a parent global instead, still set in widget install and deinit?

Yep I think that could be possible.  Each widget that wanted to be a parent would have to store the existing parent before installing themselves so they could put the previous parent back in deinit.  But there would be no way to go up the whole chain at once.

I'm not sure it's a win, the debugging support we get from being able to print out the whole chain might swamp the small savings.  But not sending events up the chain I think will make behavior more understandable.

david-vanderson:
The close_popup stuff looks okay with 2 comments:
* Is there a way we can have ContextWidget participate in the popup menu chain?  Maybe they need their own tiny interface?
* In `FloatingMenuWidget.deinit()` it looks suspicious that the check for a unhandled mouse .focus event got moved above the check for a normal .focus event on top of the floating menu.  I think that means a .focus event on an empty part of a floating menu would close it?

RedPhoenixQ:
I did the scroll events too and only `Event.text` remained and as far as I could see it was never bubbled anywhere. I the removed all references to event bubbling and  `Widget.processEvent` as it is no longer needed.

We might want to give some more documentation, perhaps in `implementation-readme.md` about how "bubbled" events should be handled now instead. 


> * Is there a way we can have ContextWidget participate in the popup menu chain?  Maybe they need their own tiny interface?

I considered that but wanted your input first before doing that slightly more complicated solution


> * In `FloatingMenuWidget.deinit()` it looks suspicious that the check for a unhandled mouse .focus event got moved above the check for a normal .focus event on top of the floating menu.  I think that means a .focus event on an empty part of a floating menu would close it?

Ah, I'll double check if that happens and fix it in that case.

RedPhoenixQ:
Fixed the `FloatingMenuWidget.deinit()` by just moving the widget `deinit` calls. Probably should've done that from the start

david-vanderson:
Excellent!!  The rest of the code looks good.  I will do a thorough testing in the next few days.  Thank you!

david-vanderson:
I'm working on this, but there is some strange bug that I think this introduced:

* load demo window
* go to debugging section and enable "Convert mouse events to touch"
* go to text entry section, copy-paste a bunch of text into the multiline entry
* try to "finger scroll" the multiline entry (so the whole scroll should be within the multiline
* sometimes instead of scrolling the multiline it scrolls the demo window instead (and multiline loses focus?)

This happens even when I turn off propagating the motion scroll stuff, so it's something about touch focus and motion.

I'm also modifying the sdl-standalone example `waitTime` to have like 3 fps max.  Then sometimes trying to "finger scroll" goes to the textEntry like it should, and other times it goes straight to the scrollContainer (of the demo window).

I'll figure it out, but just wanted to let you know what was going on.  The event logging (handled and unhandled) is helping a lot!

PR #412, opened 2025-06-27
icon stroke customization should not affect default color

nat3Github:
the logic around colormod + default colors for icons rendering was recently changed.
as of now if you customize stroke width the color (default black) turns white. this is counterintuitive behaviour.

this is because the check for colormod enabling checks also checks for default stroke width (null)
this PR changes the comparison logic to only check `stroke_color` and `fill_color`

note: the known 'quirk' of our implementation happens when a user tries to overwrite icons with white, this doesnt work because colormod is enabled and default color is not white. but i think we can live with that, because in this case the user should had used colormod anyways!


david-vanderson:
Good logic, looks good.  Merge when ready.  Thank you!

Issue #357, opened 2025-06-09
Ideas surrounding building a file navigation tree


foxnne:
Hello again!

I will soon start diving into further parts of rebuilding my app, and the first piece is a file explorer pane, which holds a tree files searched recursively from the root project folder. 

![Image](https://github.com/user-attachments/assets/ce04116e-b1f2-4c3d-9114-2a094d56b53b)

This is currently done in Dear Imgui like [this.](https://github.com/foxnne/pixi/blob/c0bdfa3c7cfbd73f46221ea446d0f029a09ffbef/src/editor/explorer/files.zig#L280-L290)

I would imagine DVUI's solution would be similar, in that you would initialize a tree widget, and deinit it when you want to "pop" back up a level. 

The file explorer I built using this wasn't as good as I wanted it to be, but I couldn't figure out how to implement drag and drop into the tree with Dear Imgui, though I think its possible. Ideally, DVUI's widget would allow rearranging and getting some information back from DVUI on drag and drop so you could easily make changes outside of the widget i.e. adjusting actual file names and locations. It would be really cool if you could also make tree labels rename-able using a text entry field. This can quickly get really complex.

Just wanted to get some initial thoughts down and see if anyone has already started investigating pieces of this? I'd love to start prototyping something and see how far I can get anyway.

Thanks!

RedPhoenixQ:
I've though a bit about this but haven't started on anything yet. My initial thought was to basically use `dvui.expander` and base the event handling on [web trees](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tree_role). I though about making a general "TreeWidget" which would then allow for `dvui.fileExplorer` to be built ontop, but also allowing users to put a text entry in the tree for renaming like you mentioned

david-vanderson:
This is great, and we definitely need some built-in dvui stuff for this.  Unfortunately nobody has seriously looked into this yet, so it'll be a little rough.

I think the dear imgui approach is the one we want to follow, where the dvui tree stuff doesn't know anything about the underlying data, it just gets fed one label after another.  It should (optionally?) remember which ones are expanded.  I'm confident this will work, just needs some time invested, which I can help with.

Drag and drop (both internal to the tree and external) is something I very much want to make work.  As far as I know the scroll canvas part of the demo is the furthest dvui drag and drop has been pushed so far.  We will make this work, but it might entail more substantial changes, unsure.

[...]

PR #413, opened 2025-06-27
TreeWidget: Add new widget for trees and example in reorderable secti…

foxnne:
…on of the demo

Helps #357.

Currently this crashes when you try to drag an element in the tree. I know none of the logic is there for actually handling the dragging, but I cant seem to figure out how to resolve the crashing to move forward. It's definitely my misunderstanding of how widget data is supposed to be used to pass things between children and parents, I believe. 

Anyway, I just wanted to get what I have up in a PR for now as I try to investigate, but I'm sure @david-vanderson can spot the issues faster than I can resolve them.

Screenshots below of some various styling.

Screenshot 2025-06-27 at 8 40 01 AM

Screenshot 2025-06-27 at 9 22 20 AM

Screenshot 2025-06-27 at 1 01 56 PM

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -70,6 +70,7 @@ var animating_window_show: bool = false;
 var animating_window_closing: bool = false;
 var animating_window_rect = Rect{ .x = 100, .y = 100, .w = 300, .h = 200 };
 var paned_collapsed_width: f32 = 400;
+var tree_removed_path: ?[]const u8 = null;
 
 var progress_mutex = std.Thread.Mutex{};
 var progress_val: f32 = 0.0;
@@ -2393,12 +2394,23 @@ pub fn recurseFiles(allocator: std.mem.Allocator, root_directory: []const u8, ou
                 });
                 defer branch.deinit();
 
-                if (branch.removed()) {
-                    std.log.debug("Removed: {s} from {s}", .{ abs_path, directory });
+                if (branch.floating()) {
+                    tree_removed_path = alloc.dupe(u8, abs_path) catch null;
                 }
 
                 if (branch.insertBefore()) {
-                    std.log.debug("Inserted: {s} into {s}", .{ abs_path, directory });
+                    if (tree_removed_path) |removed_path| {
+                        defer alloc.free(removed_path);
+
+                        const old_sub_path = std.fs.path.basename(removed_path);
+
+                        const new_path = try std.fs.path.joinZ(alloc, &.{ directory, old_sub_path });
+                        defer alloc.free(new_path);
+
+                        try std.fs.renameAbsolute(removed_path, new_path);
+
+                        std.log.debug("DVUI/TreeWidget: Moved {s} to {s}", .{ removed_path, new_path });
+                    }
                 }
 
                 switch (entry.kind) {
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -328,7 +328,7 @@ pub const demoKind = enum {
             .layout => "Layout",
             .text_layout => "Text Layout",
             .plots => "Plots",
-            .reorderable => "Reorderable",
+            .reorderable => "Reorder / Tree",
             .menus => "Menus / Focus",
             .scrolling => "Scrolling",
             .scroll_canvas => "Scroll Canvas",
@@ -349,7 +349,7 @@ pub const demoKind = enum {
             .layout => .{ .scale = 0.45, .offset = .{ .x = -50 } },
             .text_layout => .{ .scale = 0.45, .offset = .{} },
             .plots => .{ .scale = 0.45, .offset = .{} },
-            .reorderable => .{ .scale = 0.45, .offset = .{ .y = -200 } },
+            .reorderable => .{ .scale = 0.45, .offset = .{ } },
             .menus => .{ .scale = 0.45, .offset = .{} },
             .scrolling => .{ .scale = 0.45, .offset = .{ .x = -150, .y = 0 } },
             .scroll_canvas => .{ .scale = 0.35, .offset = .{ .y = -120 } },
@@ -2084,12 +2084,10 @@ const reorderLayout = enum {
 
 /// ![image](Examples-reorderable.png)
 pub fn reorderLists() void {
-    const g = struct {
-        var layout: reorderLayout = .vertical;
-    };
+    const uniqueId = dvui.parentGet().extendId(@src(), 0);
+    const layo = dvui.dataGetPtrDefault(null, uniqueId, "reorderLayout", reorderLayout, .horizontal);
 
-    const expander_o: dvui.ExpanderOptions = .{ .default_expanded = true };
-    if (dvui.expander(@src(), "Simple", expander_o, .{ .expand = .horizontal })) {
+    if (dvui.expander(@src(), "Simple", .{ .default_expanded = true }, .{ .expand = .horizontal })) {
         var vbox = dvui.box(@src(), .vertical, .{ .margin = .{ .x = 10 } });
         defer vbox.deinit();
 
@@ -2099,8 +2097,8 @@ pub fn reorderLists() void {
 
             const entries = [_][]const u8{ "Vertical", "Horizontal", "Flex" };
             for (0..3) |i| {
-                if (dvui.radio(@src(), @intFromEnum(g.layout) == i, entries[i], .{ .id_extra = i })) {
-                    g.layout = @enumFromInt(i);
+                if (dvui.radio(@src(), @intFromEnum(layo.*) == i, entries[i], .{ .id_extra = i })) {
+                    layo.* = @enumFromInt(i);
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


david-vanderson:
I really like the intent here, and building on Reorderable was definitely a good thing to try.

I don't think either of us likes how drag-hovering works right now:
* expand a directory
* drag a file from outside that directory to over that directory
* the whole directory is blocked out

I don't know if that means Reorderable needs to change, or Tree should do its own thing.  Here are questions:
* Can you give a few examples of drag-n-drop cases you want?
* Do any of them involve drag-n-drop outside the Tree (like dragging a file to some other place in the app?)
* Do we need to support dragging on top of entries as well as between?
  * dragging an entry onto a different collapsed branch
* When reordering, would you rather have the "space blocked out" that Reorder currently does, or something else?
  * maybe a line showing between two existing entries where the dragged one will go?

Thanks very much!

foxnne:
> * Can you give a few examples of drag-n-drop cases you want?

Really the only intent I have here is being able to drag a file or folder from one location into or out of another from within the tree. I fully expected to have to rewrite most of the logic for drag and drop since its less about "order" and more about parent/children relationships.

Also, order within the parent shouldn't matter. I need to look at how you handle other ordering things like in the grid widget and being able to sort. I haven't looked to determine yet if DVUI handles ordering the elements or if thats done on the collection outside of DVUI. For now, the example just pushes file/folders in the order zig gives them to us, which is in no particular order.

So really it would need to just be a matter of:
   1. Was the current item removed from a parent?
   2. Was the current item added to a parent?

I believe anyway.

> * Do any of them involve drag-n-drop outside the Tree (like dragging a file to some other place in the app?)

Currently I have no need for this, or being able to drag and drop from outside of the app into the filetree. Purely drag and drop within the file tree itself for now.

> * Do we need to support dragging on top of entries as well as between?
>   
>   * dragging an entry onto a different collapsed branch
> * When reordering, would you rather have the "space blocked out" that Reorder currently does, or something else?
>   
>   * maybe a line showing between two existing entries where the dragged one will go?

I think that really dragging an item within a folder should likely just pop a divider between the items to show where it would drop. In a file system, this order doesnt matter, but in other trees it could.

Dragging an item over a folder (the header button part) should expand the folder if it isnt expanded, but I dont think we need to hold a place holder of the item rect like the reorderable does. Just a divider at the first position within the folder it will be dropped into.

I'm not sure how we can report the data back though, as this wont be a flat list where a simple index of the item removed or dropped onto is enough.

david-vanderson:
Thanks for the info!  Let me do some experimentation.  We got newer drag stuff since ReorderWidget, I'll see if that works any better and report back.

foxnne:
Awesome, sounds good. I'll also continue experimenting. Appreciate you looking into this!

foxnne:
@david-vanderson After the most recent commit, I believe I have ironed out any remaining issues I could find with the TreeWidget. 

Currently supported features:

- Custom labels for each branch, just gives you an hbox, and a button that wraps the hbox
- Optional expander for each item which remembers it's expanded position, each level can have its own custom indent
- Drag and drop within the tree: `removed`, `insertBefore` and `floating` functions are provided to allow users to know when branches are drag/dropped.
- When dragging, if hovering over an un-expanded folder for 1 second, the folder will expand.

I believe this covers the use cases we outlined previously. DVUI already supports adding a context menu to any of the branches, such that a file explorer could allow new/edit/rename/etc.

The only remaining feature I want is the ability to swap out the branch label with a input text, but I believe that is outside of the scope of the TreeWidget, as the branch label is just an hbox the user fills in with any text or icons they want.

david-vanderson:
This is fantastic, just a few comments/requests before we can merge:
* for the demo, we need something that doesn't touch the filesystem (so it can work on web, and also I accidentally moved some of my files around!)
  * the filesystem thing is really cool, maybe we offer that as a convenience function?  (If it's not called then we are okay for web)
* instead of using `dvui.animation` look at `dvui.timer` - animation will max the fps, but timer only schedules a single frame in the future
* can we somehow put the removed path into the TreeWidget?

How does that sound?  Otherwise looks great!

foxnne:
This all sounds good to me, I'll get a fake filesystem demo made and make those other changes.
The only thing I'm a bit lost on is the removed path. Do you mean specifically for the file system convenience function? Or do you mean something generic in `TreeWidget`? I tried to come up with some way to be able to get the "old" info to `insertBefore` but since this is all just layout and has no actual underlying data I'm not sure how to do it.  

I think the issue is that when `insertBefore` is true we don't know if the removed entry has already occured or not.

david-vanderson:
> This all sounds good to me, I'll get a fake filesystem demo made and make those other changes.

Thank you!

> The only thing I'm a bit lost on is the removed path. Do you mean specifically for the file system convenience function? Or do you mean something generic in `TreeWidget`? I tried to come up with some way to be able to get the "old" info to `insertBefore` but since this is all just layout and has no actual underlying data I'm not sure how to do it.
> 
> I think the issue is that when `insertBefore` is true we don't know if the removed entry has already occured or not.

I was thinking of having TreeWidget own that data somehow.  The kind of thing I'm worried about that data getting out of sync, like the TreeWidget no longer thinks something is floating, but the variable is still non-null.

Let me commit something, maybe just hanging it off the TreeWidget id somehow and see what you think.

david-vanderson:
I like this better than having a global var, but it is more complicated.  I agree it didn't make as much sense to put it inside TreeWidget as I thought, since the shape of the data is based on what is being put in the tree.  What do you think?

foxnne:
> I like this better than having a global var, but it is more complicated. I agree it didn't make as much sense to put it inside TreeWidget as I thought, since the shape of the data is based on what is being put in the tree. What do you think?

I do think this works better, but I'm also totally open to reworking something if you think there is a better way to do things. 

I'll be working on the other changes today. Thanks!

RedPhoenixQ:
@foxnne I know a fix for the example, is it ok if I push it here?

foxnne:
> @foxnne I know a fix for the example, is it ok if I push it here?

@RedPhoenixQ Absolutely!

RedPhoenixQ:
Have to fix some things that didn't work out. Will still fix

RedPhoenixQ:
The main fix was using `ArrayListUnmanaged.initBuffer` to set the buffer without an allocator. The buffer gets allocated by `dataSetSlice` and kept alive. This does mean that the state does not persist when swapping views.

It's possible to use static arrays instead which would persists the changes. I'll give that a try as well

foxnne:
Awesome! I think that's a cool solution, and by swapping views, you mean just going to another location in the demo and back? I have no issue with that "resetting" the file tree, I had planned on adding a reset button anyway but that's actually simpler. (sorry away from my pc for a bit)

foxnne:
I have been working in macOS all day, and when I got home I tried to play with the example on my Windows machine and was getting a integer overflow crash. I'm not sure if this has fixed that, just wanted to let you know the example had some bug.

RedPhoenixQ:
> and when I got home I tried to play with the example on my Windows machine and was getting a integer overflow crash. I'm not sure if this has fixed that, just wanted to let you know the example had some bug.

I didn't get any crash but I didn't test it very extensively.

david-vanderson:
If I do some testing is this ready to merge?

foxnne:
> If I do some testing is this ready to merge?

Nearly forgot the timer, haha. Okay, I believe I have nothing left to add if you're good with it! 

Also I tried to replicate that issue and couldn't, it seems stable.

david-vanderson:
Very well done!  Thank you very much!

david-vanderson:
@foxnne When I write this up, would you like to be credited as "foxnne" or a different name?

foxnne:
Thats fine or my name is Colton Franklin, either works for me - I appreciate it! This was a fun one, I've really enjoyed working on it. I continue to be impressed with the ease of creating things with DVUI, its fun to use.

Also really appreciate the help with everything along the way and answering questions!

vlappa:
Nice work!

I might try this out in the coming days, some question I've already have:

Is if the drag and drop thing optional?

It is possible to 'pass' custom keybindings?

Looking at the examples I'm also wondering if the theming / look could made to be more simple looking. In the example it looks like there is some 3D effect for instance, which is not ideal for my use case. 

Looking at the demo, it's look like buttons are being used. A other option would be using a labelClick (in a box) I can imagine. 

The 'toggle' is on the right side, while a lot other tree widgets I see have them on the left side. Maybe that could be a option? 

Also a toggle between a tree view and a list view might be nice to have. See also picture below.

Just some thoughts, will dive into it deeper later. 

```
List view:

Flintstones/Fred
Flintstones/Wilma

Tree view:

    Flintstones
    - Fred
    - Wilma
```

1751705967_grim
1751706027_grim

foxnne:
> Is if the drag and drop thing optional?

It will be, a simple switch is needed to turn off the event handling for drag and drop. I've just added that in my local copy.

> It is possible to 'pass' custom keybindings?

Currently there are no keybinds associated with the file tree, though the tree itself is just a series of boxes and buttons. So I can look into adding navigation by arrow keys, or other keybinds.

> Looking at the examples I'm also wondering if the theming / look could made to be more simple looking. In the example it looks like there is some 3D effect for instance, which is not ideal for my use case.

The example just shows what all is possible. The widget itself really is just adding a vbox with a margin on the left, and then you recursively add items in the same way to build the tree. You can easily set the styling for the button (the tree entry just wraps a button, so you can add any items you'd like, any icons in any order)

> Looking at the demo, it's look like buttons are being used. A other option would be using a labelClick (in a box) I can imagine.


Like said above, the tree itself has no labels or icons. Its simply a layout tool to add the items you wish to the label button.

> The 'toggle' is on the right side, while a lot other tree widgets I see have them on the left side. Maybe that could be a option?

Same as above

> Also a toggle between a tree view and a list view might be nice to have. See also picture below.

This can be done in userland, simply branch and either draw the TreeWidget or use a box and add paths as a list.

PR #415, opened 2025-06-28
Uniform scroll speeds

### LINKED ISSUE #406: Uniform Scrolling Speed across Backends
I noticed that the scrolling speed differs strongly across backends like sdl / web / raylib thus greatly affecting the feel of the application.
- why is that?
- what is the upside of the current behaviour?
- can we make it uniform, i.e. a heuristical scaling-speed factor?
- i guess it would be nice for development if the scrolling speed was uniform across Backends when targetting for example web and sdl at the same time

RedPhoenixQ:
Closes #406

This makes scrolling uniform for all backends on my Windows machine. More testing needs to be done on other platforms. All platforms use the same scroll speed multiplier and the web backend translates to "ticks" by dividing by the smallest delta that has been seen.

The biggest questions is potential issues with devices with higher fidelity of scrolling, where one "tick" is supposed to be some fraction of a "normal" tick. Another is the `dvui.scroll_speed / 2` on macOS on SDL. Is this linked to the HIDPI? In that case it might need to be scaled by `content_scale`.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -216,6 +216,33 @@ pub fn buildBackend(backend: enums_backend.Backend, test_dvui_and_app: bool, dvu
             }
             sdl_mod.addOptions("sdl_options", sdl2_options);
 
+            // Enable smooth scrolling on mac
+            if (target.result.os.tag == .macos) {
+                // SDL hard codes this so we have to overwrite it
+                const objc_files = b.addWriteFiles();
+                const objc_file = objc_files.add("config.mm",
+                    // https://github.com/libsdl-org/SDL/issues/2176#issuecomment-2009687592
+                    \\#import <Foundation/Foundation.h>
+                    \\
+                    \\void MACOS_enable_scroll_momentum() {
+                    \\    [[NSUserDefaults standardUserDefaults]
+                    \\    setBool: YES forKey: @"AppleMomentumScrollSupported"];
+                    \\}
+                );
+                const lib = b.addLibrary(.{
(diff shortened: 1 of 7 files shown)


david-vanderson:
> Closes #406
> 
> This makes scrolling uniform for all backends on my Windows machine. More testing needs to be done on other platforms. All platforms use the same scroll speed multiplier and the web backend translates to "ticks" by dividing by the smallest delta that has been seen.

Good idea!

> 
> The biggest questions is potential issues with devices with higher fidelity of scrolling, where one "tick" is supposed to be some fraction of a "normal" tick. Another is the `dvui.scroll_speed / 2` on macOS on SDL. Is this linked to the HIDPI? In that case it might need to be scaled by `content_scale`.

Good question I will do some testing.

david-vanderson:
Awesome!  Sorry haven't gotten to testing this yet - are there particular OS/backend combinations that need testing?

RedPhoenixQ:
I haven't test linux as I don't have a linux install currently. I can launch gui apps through Windows wsl but I don't trust the behavior to be completely correct. 

What I've mostly done for testing is open the scrolling demo and scrolled 1 tick on the scroll wheel. That now results in an offset of 20 across all backends, except on MacOS where there is scroll acceleration but it "feels" the same. (Haven't tested content scales or OS scaling at all)

david-vanderson:
I've tested this some on Linux and Mac, looks great!  Merge it.  Thank you!

PR #417, opened 2025-06-28
Simplify allocations

### LINKED ISSUE #414: simplifying our allocations
We've gotten some good experience with splitting `Window.arena()` into two (plus the widget stack).  Here are things I think I've learned:
* @RedPhoenixQ showed that Zig's arena allocator doesn't reliably free things in lifo order:
  * If it has to bump for alignment, it doesn't know how to back out that bump later
  * If it frees to the start of a memory chunk, it doesn't release that memory chunk and move back to the previous
* The split of having `arena()` being for frame-lifetime stuff and `lifo()` being for temporary stuff seems good
* Less sure about widget_stack, seems weird to need to hardcode a capacity value
* In general simplicity wins when talking about memory stuff
* We want a split between Debug and Release
  * In Debug, we probably don't want to free/reuse anything during a frame, to detect use-after-free bugs
  * In Release, we (ideally) reuse as much as we can

In that light I think I want to head in this direction:
* merge widget_stack into `lifo`
* keep shrinking_arena_allocator for both `arena` and `lifo`
  * change it to not free stuff in Debug

Longer term investigations for shrinking_arena_allocator:
* can it be made to reliably free and reuse stuff in lifo order without too much complexity?
  * maybe detect at comptime the maximum alignment and always ask for that much from the underlying allocator?
  * maybe add an extra byte to each allocation with the alignment info?

@RedPhoenixQ How does this sound?

RedPhoenixQ:
Closes #414

Widget stack now uses `ShrinkingArenaAllocator`, which removes a lot of logic around reallocations and expanding. The stack cannot use `allocatorLIFO` because of widgets containing widgets in their fields (not without adding allot of extra metadata).
There will still be a warning from `Window.end` if everything wasn't freed, similar to `Window.lifo`.

In Debug we could store an `ArrayList` of all the allocations and perhaps a stacktrace to achieve better lifo debugging when widgets aren't freed correctly.

The idea of "never free" for all allocations to catch use-after-free bugs hasn't been implemented yet as there are still some pain points around how to verify that the "normal" allocations are still correct in terms of freeing order and such. If this is not decided on/implemented in this PR, it probably doesn't close #414 yet.

### DIFF
--- a/src/Window.zig
+++ b/src/Window.zig
@@ -111,10 +111,7 @@ gpa: std.mem.Allocator,
 _arena: dvui.ShrinkingArenaAllocator,
 _lifo_arena: dvui.ShrinkingArenaAllocator,
 /// Used to allocate widgets with a fixed location
-_widget_stack: std.heap.FixedBufferAllocator,
-/// The peak amount of bytes used by the widget stack since
-/// the creation of `Window`. Used to resize the stack ifs needed.
-peak_widget_stack: usize = 0,
+_widget_stack: dvui.ShrinkingArenaAllocator,
 render_target: dvui.RenderTarget = .{ .texture = null, .offset = .{} },
 end_rendering_done: bool = false,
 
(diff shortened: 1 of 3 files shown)


david-vanderson:
This looks good.  Merge when ready - thanks very much!

PR #418, opened 2025-06-28
Abstract click event handling into a reusable dvui helper

### LINKED ISSUE #345: reusable click logic
The logic in `dvui.labelClick` is identical to the logic in `ButtonWidget` and likely duplicated elsewhere.  @vlappa was asking on irc about catching clicks in other situations.

This issue is to explore if we can wrap that logic into a function so that users could easily handle clicks for arbitrary screen rects.

Maybe the initial should be restricted to a widget boundary, so that we have a WidgetData.

RedPhoenixQ:
Closes #345

This handles mouse capture, hovered state and a "click" for any rect, with the same logic as `ButtonWidget` and `labelClick` had before. The rect defaults to the border rect of the passed in widget data, which is always required to handle the events. I moved the well documented version of the event handling from `labelClick`.

This will be even simpler to understand when event bubbling is removed by #410, as there would be no more confusion with `processEvent` and this helper.

### DIFF
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -2456,7 +2456,7 @@ pub const CaptureMouse = struct {
 /// (which is what you would expect for e.g. background highlight)
 ///
 /// Only valid between `Window.begin`and `Window.end`.
-pub fn captureMouse(wd: ?*WidgetData) void {
+pub fn captureMouse(wd: ?*const WidgetData) void {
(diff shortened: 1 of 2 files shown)


david-vanderson:
Great! - Merge when ready - thanks!

RedPhoenixQ:
I'll hold of on merging in the hopes that the bubbling changes lands first to rebase these changes as it's simpler that way

PR #420, opened 2025-06-29
Improve zls compatibility

MotokoKusanagii:
This will make zls not automatically go to the web implementation when going to the definition of `dvui.backend`, but instead correctly takes you to the specified backend.

Drawbacks:
- The current implementation forces you to specify a target, even when building the examples.
`zig build sdl3-app` becomes `zig build sdl3-app -Dbackend=sdl3`.
`zig build raylib-app` becomes `zig build raylib-app -Dbackend=raylib`.
etc.
- ~In order to not make `zig build docs` panic, `back_to_build` will default to `.testing`.~

Fixing that is tricky without much code duplication but I can look into it.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -14,7 +14,10 @@ pub fn build(b: *std.Build) !void {
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
 
-    const back_to_build: enums_backend.Backend = b.option(enums_backend.Backend, "backend", "Backend to build").?;
+    var back_to_build: ?enums_backend.Backend = b.option(enums_backend.Backend, "backend", "Backend to build");
+    if (back_to_build == null) {
+        back_to_build = .sdl3;
+    }
 
     const test_step = b.step("test", "Test the dvui codebase");
     const check_step = b.step("check", "Check that the entire dvui codebase has no syntax errors");
@@ -24,6 +27,9 @@ pub fn build(b: *std.Build) !void {
     const test_filters = b.option([]const []const u8, "test-filter", "Skip tests that do not match any filter") orelse &[0][]const u8{};
 
     const generate_doc_images = b.option(bool, "generate-images", "Add this to 'docs' to generate images") orelse false;
+    if (generate_doc_images) {
+        back_to_build = .sdl2;
+    }
 
     const build_options = b.addOptions();
     build_options.addOption(
@@ -68,7 +74,7 @@ pub fn build(b: *std.Build) !void {
         .build_options = build_options,
     };
 
-    switch (back_to_build) {
+    switch (back_to_build.?) {
         .custom => {
             // For export to users who are bringing their own backend.  Use in your build.zig:
             // const dvui_mod = dvui_dep.module("dvui");
--- a/build.zig
+++ b/build.zig
@@ -14,7 +14,7 @@ pub fn build(b: *std.Build) !void {
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
 
-    const back_to_build: ?enums_backend.Backend = b.option(enums_backend.Backend, "backend", "Backend to build");
+    const back_to_build: enums_backend.Backend = b.option(enums_backend.Backend, "backend", "Backend to build").?;
 
     const test_step = b.step("test", "Test the dvui codebase");
     const check_step = b.step("check", "Check that the entire dvui codebase has no syntax errors");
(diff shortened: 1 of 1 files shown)


david-vanderson:
Good code, give me some time to play with it and see if I can keep building all the backends while maintaining the switch.

Thanks very much!

RedPhoenixQ:
Nice that you figured out how to solve it! There are two things I think need to be done in combination with this:

1. The CI testing needs to be updated to run once for each backend (which might be an improvement since not all tests will fail if only a single backend is incorrect)
2. The examples could be renamed to "app", "standalone", and "ontop" with no prefix as only one backend will be part of the build graph at a time. This makes the build command `zig build app -Dbackend=sdl3` which is slightly nicer

For point 2, perhaps the default backend should be something more universal like sdl3 so that `zig build app` will run the sdl3 app by default?

Also, the docs would fail if the `-Dgenerate-images` is passed as it looks for and sdl2 backend. Maybe the default should be swapped to sdl2/sdl3 when this flag is passed?

david-vanderson:
I'm going to try a smaller fix first, running the switch if a single backend is asked for, and then .... something else if doing all of them, maybe switch in a loop?  I'd like to keep building all the examples at once if we can.

> The CI testing needs to be updated to run once for each backend (which might be an improvement since not all tests will fail if only a single backend is incorrect)

I guess we can do this in any case right?

RedPhoenixQ:
> I guess we can do this in any case right?

True, I'll get on that soon.

MotokoKusanagii:
> 
> For point 2, perhaps the default backend should be something more universal like sdl3 so that `zig build app` will run the sdl3 app by default?
> 
> Also, the docs would fail if the `-Dgenerate-images` is passed as it looks for and sdl2 backend. Maybe the default should be swapped to sdl2/sdl3 when this flag is passed?

I just realized that setting it to `.testing` (or any other value) as default will break it again, so we can't set default values apparently.
Probably forgot to restart my lsp after my second commit.

MotokoKusanagii:
> Nice that you figured out how to solve it! There are two things I think need to be done in combination with this:
> 
>     1. The CI testing needs to be updated to run once for each backend (which might be an improvement since not all tests will fail if only a single backend is incorrect)
> 
>     2. The examples could be renamed to "app", "standalone", and "ontop" with no prefix as only one backend will be part of the build graph at a time. This makes the build command `zig build app -Dbackend=sdl3` which is slightly nicer
> 
> 
> For point 2, perhaps the default backend should be something more universal like sdl3 so that `zig build app` will run the sdl3 app by default?
> 
> Also, the docs would fail if the `-Dgenerate-images` is passed as it looks for and sdl2 backend. Maybe the default should be swapped to sdl2/sdl3 when this flag is passed?

I implemented the `zig build app -Dbackend=backend` behavior. `sdl3` is now the default backend and providing `-Dgenerate-images` will force the sdl2 backend. Let me know if I should change something!

david-vanderson:
> I just realized that setting it to `.testing` (or any other value) as default will break it again, so we can't set default values apparently.
> Probably forgot to restart my lsp after my second commit.

Do you have a sense now for what zls requires from us to make this work?  Do we know for sure that zls works with this no matter what backend you are using?

MotokoKusanagii:
I think the main contribution was switching to a `switch` statement. My theory is that `if (back_to_build == null or ..)` somehow forced ZLS to choose between multiple code paths, though I know absolutely nothing about the inner workings of ZLS.

> I just realized that setting it to `.testing` (or any other value) as the default breaks it again, so apparently we can't set default values.

Trying
`var back_to_build: enums_backend.Backend = b.option(enums_backend.Backend, "backend", "Backend to build") orelse .sdl3;`
suddenly works again, and I don't know why.

I tested the latest commit with Raylib, SDL3, and SDL2, all jumping to the correct implementation respectively. I can't test DX11 because I don't have a Windows setup on me.

RedPhoenixQ:
I can't get any of them to work. Not matter what I specify in the `b.dependency("dvui", .{ .backend = .dx11 })` it always jumps to the fallback backend, in this case sdl3. If I change the default, in dvuis build.zig, to raylib it goes there instead. How did you change the backend?

MotokoKusanagii:
> I can't get any of them to work.

Do you mean they wont compile or does only zls not work? 
Because I usually have to build the project first once for zls to correctly use the cache

RedPhoenixQ:
> Do you mean they wont compile or does only zls not work?
> Because I usually have to build the project first once for zls to correctly use the cache

Sorry bad wording. Everything compiles and zls works, it just always links to the default backend (even after compiling and restarting zls)

MotokoKusanagii:
Did you pull my latest changes ([a2206f0](https://github.com/david-vanderson/dvui/pull/420/commits/a2206f074e03b1f01550ba93a5b55ab31591e802))?
I only set it in three places:
```zig
var back_to_build: ?enums_backend.Backend = b.option(enums_backend.Backend, "backend", "Backend to build");
if (back_to_build == null) {
back_to_build = .sdl3;
}
//...
const generate_doc_images = b.option(bool, "generate-images", "Add this to 'docs' to generate images") orelse false;
if (generate_doc_images) {
back_to_build = .sdl2;
}
```
There’s no default value, except maybe the if (back_to_build == null) check, but I can't believe it would just arbitrarily take that.

RedPhoenixQ:
> There’s no default value, except maybe the if (back_to_build == null) check, but I can't believe it would just arbitrarily take that.

It does for me. I think zls runs each build.zig without considering the parameters they are called with.

MotokoKusanagii:
This is my sdl3 setup: https://github.com/MotokoKusanagii/motoko/blob/engine/build.zig#L21-L34
And my raylib setup is basically the demo

> I think zls runs each build.zig without considering the parameters they are called with.

I just don't understand why it would work for me then

RedPhoenixQ:
The only difference I can see is `.import = &.{...}` instead of `exe_mod.addImport("dvui", ...)`, but it still doesn't change for me. It always goes to sdl or whichever one is specified in the `if (back_to_build == null)` check.

I'm using the VSCode plugin if that matters for how zls gets initialized

RedPhoenixQ:
In the zls log output when starting up I'm getting:

```
2025-06-30 00:47:04.947 [info] Starting ZLS      0.14.0 @ '...\zls.exe'
2025-06-30 00:47:04.948 [info] Log File:         ...\zls\zls.log (debug)
2025-06-30 00:47:04.948 [info] Client Info:      Visual Studio Code-1.101.2
2025-06-30 00:47:04.948 [info] Autofix Mode:     source.fixall
2025-06-30 00:47:04.948 [info] added Workspace Folder: .../dvui-test
2025-06-30 00:47:04.955 [info] Set config option 'global_cache_path' to "...\\zls"
2025-06-30 00:47:04.971 [info] Loaded build file '.../dvui-test/build.zig'
2025-06-30 00:47:04.973 [info] Loaded build file '.../dvui/build.zig'
```

The last line where it loads dvuis build.zig directly was what made me think it ran each build.zig separately

MotokoKusanagii:
I just tried vscode and it took me to the default value as well. So it might be zls version specific? ..That's annoying^^

RedPhoenixQ:
Are you on a newer version than zls 0.14.0? In that case this change is still probably worth making, it just won't have the full effect until the next zls release

MotokoKusanagii:
doing `zls version` gives me 0.14.0.
I downloaded it via https://zigtools.org/zls/install/. I remember writing 0.14.1 into the version field, even though it just gave me a 0.14.0 binary. Is there any way I can check the specific commit my zls was build from?

RedPhoenixQ:
It should be the same then. I don't have "build on save" enabled if that makes a difference.

> Is there any way I can check the specific commit my zls was build from?

Don't think so.

MotokoKusanagii:
Setting zls path in vscode to my global zls does not fix it as well. 

> I don't have "build on save" enabled if that makes a difference.

I don't either

RedPhoenixQ:
If it works outside VSCode, it's still an improvement so no big deal.

MotokoKusanagii:
I use the helix editor with anyzig and a global install of zls version 0.14.0.
my language.toml is literally
```toml
[[language]]
name = "zig"
```
The editor should not really matter though, that's why I'm very confused.

The last thing I can think of is that I include my fork locally in `build.zig.zon` instead of fetching it.
```zig
.dvui = .{
   .path = "path/to/dvui",
},
```
> If it works outside VSCode, it's still an improvement so no big deal.

I'm sorry, I really thought I fixed it. I will investigate this more over the coming days!

RedPhoenixQ:
I'm currently using both zig and zls installed via the VSCode plugin.

> The last thing I can think of is that I include my fork locally in `build.zig.zon` instead of fetching it.

I did too for this testing

> I'm sorry, I really thought I fixed it. I will investigate this more over the coming days!

All good! Still some improvements, and I do prefer a switch over the if statements

MotokoKusanagii:
Does it work for you if you apply following code? Just tested it in vscode with raylib and it worked for me
```diff
diff --git a/build.zig b/build.zig
index f54a54c..16067a0 100644
--- a/build.zig
+++ b/build.zig
@@ -15,9 +15,9 @@ pub fn build(b: *std.Build) !void {
     const optimize = b.standardOptimizeOption(.{});

     var back_to_build: ?enums_backend.Backend = b.option(enums_backend.Backend, "backend", "Backend to build");
-    if (back_to_build == null) {
-        back_to_build = .sdl3;
-    }
+    // if (back_to_build == null) {
+    //     back_to_build = .sdl3;
+    // }

     const test_step = b.step("test", "Test the dvui codebase");
     const check_step = b.step("check", "Check that the entire dvui codebase has no syntax errors");
```

RedPhoenixQ:
> Does it work for you if you apply following code? Just tested it in vscode with raylib and it worked for me

Yes it does! Weird that that works. It crashes at the `.?` in the switch, so I guess it's forced to use the information from the projects build.zig through the dependency. So by default it must require a backend option.

There must be some nicer way to stop VSCode from running the build file in the first place, don't know how though. It would be nice not to have to fail the build when just running `zig build` (without `-Dbackend` or another flag we add) to make zls work.

david-vanderson:
Sounds like to improve here we have to require a `-Dbackend` flag?  Or am I misunderstanding?

RedPhoenixQ:
> Sounds like to improve here we have to require a `-Dbackend` flag? Or am I misunderstanding?

For it to work in VSCode you have to require it. I'm thinking we still provide a default value for now, for better usability, in the hopes that there is another way to make VSCode work. Perhaps `null` should indicate "all backends" like before and run the switch in a loop like you suggested (also handling the `test_dvui_and_app` variables). Hopefully that doesn't regress the improvements outside VSCode

david-vanderson:
> > Sounds like to improve here we have to require a `-Dbackend` flag? Or am I misunderstanding?
> 
> For it to work in VSCode you have to require it. I'm thinking we still provide a default value for now, for better usability, in the hopes that there is another way to make VSCode work. Perhaps `null` should indicate "all backends" like before and run the switch in a loop like you suggested (also handling the `test_dvui_and_app` variables). Hopefully that doesn't regress the improvements outside VSCode

Sounds worth trying.  I'm not sold on changing `zig build sdl3-app` to `zig build app` so I'm looking to understand what's necessary for this.

RedPhoenixQ:
> > The CI testing needs to be updated to run once for each backend (which might be an improvement since not all tests will fail if only a single backend is incorrect)
> 
> I guess we can do this in any case right?

CI is now improved in e127232e1da4b2385476b39cc917f3bf7e4e0f1b

MotokoKusanagii:
I also had the idea of removing conditional modules altogether.
Instead of relying on -Dbackend, we could define every module as name_backend (e.g., sdl2_backend).

In code, this would look like:

```zig
const dvui = @import("dvui");
const Backend = dvui.sdl3_backend; // or: const Backend = @import("sdl3_backend");
```

It would be a pretty intrusive change, and maybe not possible for reasons that aren't directly apparent to me.
If this isn’t desirable for other reasons, that’s completely fair as well.

RedPhoenixQ:
The backend currently needs to be know at comptime in "Backend.zig" to provide the interface. We could change it back to vtable through but it's not as optimal.

The bigger issue is that all the imports would have to be available for every build. This would require that some module is passed in the build script for "dx11" on Linux builds for example, which would require the build script mocking the module.

I think we want compiling/linking the backend be defined in the build script for these reasons.

david-vanderson:
To sum up:
* this improves the zls experience for all backends
* but we give up the ability for `zig build` to build all the backends at once

Is that correct?  Would we lose the zls benefit if we added an enum value `.all` so `zig build -Dbackend=all` builds them all?  If yes, can `.all` be the default?

RedPhoenixQ:
Not specifying the backend is now back to the previous behaviour of building all backends. @MotokoKusanagii could you check if this still works for you? Otherwise we can still require `.all` (or maybe `-Dall` to not pollute the `Backend` enum?) to be passed in so  at least on argument is required.

MotokoKusanagii:
> Not specifying the backend is now back to the previous behaviour of building all backends. @MotokoKusanagii could you check if this still works for you? Otherwise we can still require `.all` (or maybe `-Dall` to not pollute the `Backend` enum?) to be passed in so at least on argument is required.

Still works for me, awesome!

david-vanderson:
Awesome! @RedPhoenixQ merge when ready.  @MotokoKusanagii thank you!

RedPhoenixQ:
Update for future reference. ZLS seems to get confused because of the indirection added by `dvui.backend`. If a project using dvui imports the backend explicitly, ZLS can correctly link to it and its functions. See [this change to the dvui-demo](https://github.com/david-vanderson/dvui-demo/blob/38bf182331a100ab13ce2be233684893a44ada95/build.zig#L63)

This does not help application using `dvui.App`, but they should rarely need to call functions on the backend directly. If the backend can be changed via a build flag (ie. the `@import("backend")` can change from sdl3 to raylib via something like `-Dbackend=raylib`), then the default backend set in the build script will be the one ZLS links to.

Issue #423, opened 2025-06-30
Add safety checks for undefined widget id in data()


phatchman:
## Description
- Check the widget id is not set to undefined in common functions like data()
- Change all internal uses of wd.id to data() 
  - Note most of these won't need to go through widget vtable, so not expecting any change in performance.

phatchman:
Merged #424

phatchman:
https://github.com/ziglang/zig/issues/211 is relevant to this if it rver gets implemented.

PR #424, opened 2025-06-30
Use after deinit safety checks in data() calls.

phatchman:
## Description
 - Adds an .undef element to WidgetId enum.
 - Adds validate() to WidgetData
 - Added validation of Widget Id to all data() returns of *WidgetData by calling validate()
 - Changes all instances of self.wd to self.data() except in widget init() functions.

 ## Discussion
1. Zig really doesn't want people to compare to undefined. Encoding the undefined value directly was the cleanest of all the tricks I tried. 
2. Comparing to undefined is unchecked illegal behavior and may not work in the future. 
3. The check is in an assert, so it will only impact safe build modes.
4. The rule I decided was that self.wd is allowed in init() and data(), but everywhere else should use self.data(). The exception to this is if self.wd is assigned or created within a non init() function. Then we can assume self.wd is valid for the remainder of the function.
5. validate() returning a *WidgetData seems nice to me. But it does require a `@constCast()` to allow this chaining.
6. This is draft until I review and double-check the changes.
7. All of the demo appears to work, except for the 2 issues noted below.
8. Removing the inline from validate() gives better error messages. Better errors are good, so I'll remove the inline before merging. Maybe only use this for debug and provide an inline one that just returns the WidgetData for release modes (including safe?)
9. What deters future widget writers and users from just using wd instead of data? Not something for this PR, but worth thinking about.
10. I need to check why I needed to make data() *const Self.
```
C:\Users\phatc\AppData\Roaming\Code\User\globalStorage\ziglang.vscode-zig\zig\x86_64-windows-0.14.1\lib\std\debug.zig:550:14: 0x7ff78f94949d in assert (sdl2-standalone.exe.obj)
C:\src\dvui-dev\src\widgets\LabelWidget.zig:95:28: 0x7ff78faa58db in data (sdl2-standalone.exe.obj)       
    return self.wd.validate();
```
vs
```
C:\Users\phatc\AppData\Roaming\Code\User\globalStorage\ziglang.vscode-zig\zig\x86_64-windows-0.14.1\lib\std\debug.zig:550:14: 0x7ff78f94949d in assert (sdl2-standalone.exe.obj)
             ^
C:\src\dvui-dev\src\WidgetData.zig:327:21: 0x7ff78f9f5776 in validate (sdl2-standalone.exe.obj)
    std.debug.assert(self.id != WidgetId.undef); // Indicates a use after deinit() error.
                    ^
C:\src\dvui-dev\src\widgets\LabelWidget.zig:95:28: 0x7ff78faa58db in data (sdl2-standalone.exe.obj)       
    return self.wd.validate();
```

Implements: #423

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -4736,7 +4736,7 @@ pub const StrokeTest = struct {
     }
 
     pub fn data(self: *Self) *dvui.WidgetData {
-        return &self.wd;
+        return self.wd.validate();
     }
 
     pub fn rectFor(self: *Self, id: dvui.WidgetId, min_size: dvui.Size, e: dvui.Options.Expand, g: dvui.Options.Gravity) dvui.Rect {
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -89,7 +89,7 @@ const AnimatingDialog = struct {
         var win = FloatingWindowWidget.init(@src(), .{ .modal = modal }, .{ .id_extra = id.asUsize(), .max_size_content = .width(300) });
 
         if (dvui.firstFrame(win.data().id)) {
-            dvui.animation(win.wd.id, "rect_percent", .{ .start_val = 0.0, .end_val = 1.0, .end_time = duration, .easing = easing });
+            dvui.animation(win.data().id, "rect_percent", .{ .start_val = 0.0, .end_val = 1.0, .end_time = duration, .easing = easing });
         }
 
         const winHeight = win.data().rect.h;
@@ -148,7 +148,7 @@ const AnimatingDialog = struct {
         }
 
         if (closing) {
-            dvui.animation(win.wd.id, "rect_percent", .{ .start_val = 1.0, .end_val = 0.0, .end_time = duration, .easing = easing });
+            dvui.animation(win.data().id, "rect_percent", .{ .start_val = 1.0, .end_val = 0.0, .end_time = duration, .easing = easing });
         }
     }
 
@@ -438,7 +438,7 @@ pub fn demo() void {
                 defer box.deinit();
 
                 var options: dvui.Options = .{ .gravity_x = 0.5, .gravity_y = 1.0 };
-                if (dvui.captured(bw.wd.id)) options = options.override(.{ .color_text = .{ .color = options.color(.text_press) } });
+                if (dvui.captured(bw.data().id)) options = options.override(.{ .color_text = .{ .color = options.color(.text_press) } });
 
                 dvui.label(@src(), "{s}", .{e.name()}, options);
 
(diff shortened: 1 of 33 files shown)


phatchman:
Yay! Found my first issue:
```
var mslabel = dvui.LabelWidget.init(@src(), "{d:0>3} ms into second", .{@as(u32, @intCast(left))}, .{}, .{});
mslabel.install();
mslabel.draw();
mslabel.deinit();

if (dvui.timerDoneOrNone(mslabel.data().id)) {
    const wait = 1000 * (1000 - left);
    dvui.timer(mslabel.data().id, wait);
}
```

Unrelated issue
```
debug: Free from lifo arena failed. Somewhere between when this was allocated and this call to free there was another allocation that was not freed first. Stack trace:
C:\src\dvui-dev\src\widgets\TextLayoutWidget.zig:432:47: 0x7ff699ac4234 in format__anon_59631 (sdl2-standalone.exe.obj)
    defer if (l.ptr != fmt.ptr) cw.lifo().free(l);
                                              ^
C:\src\dvui-dev\src\Examples.zig:3760:18: 0x7ff6999f14cc in debuggingErrors (sdl2-standalone.exe.obj)     
        tl.format("Latest matched keybinding: {s}", .{g.latest_slice}, .{});
                 ^
C:\src\dvui-dev\src\Examples.zig:523:42: 0x7ff69999c9f7 in demo (sdl2-standalone.exe.obj)
            .debugging => debuggingErrors(),

....

warning(dvui): Arena was not empty at the end of the frame, 31 bytes left. Did you forget to free memory somewhere?
```

david-vanderson:
> Yay! Found my first issue:

Well done!

> Unrelated issue

Do you know how to reproduce that?

phatchman:
> Do you know how to reproduce that?

Go to Debugging->Key Bindings.
Do Ctrl-A, Ctrl-C (or the equivalents).
It's not all the time and probably isn't just those keys.
Seems to help if you alternate between them quickly?

RedPhoenixQ:
> > Do you know how to reproduce that?
> 
> Go to Debugging->Key Bindings. Do Ctrl-A, Ctrl-C (or the equivalents). It's not all the time and probably isn't just those keys. Seems to help if you alternate between them quickly?

It's was a misuse of `lifo` for a longer lived copy buffer (the copy buffer stored between calls to `TextLayoutWidget.addTextEx`). I will change is to use `arena` since it only allocates on copy anyways

david-vanderson:
Thank you both!

phatchman:
OK I've checked through this again with fresh eyes and it is good to merge.

Everywhere uses self.data() now unless it is the function that self.wd is created in. In that case I use self.wd as we can assume it will be valid for the duration of the function.

david-vanderson:
Looks good - merge when ready.  Thanks!

> 10\. I need to check why I needed to make data() *const Self.

What was the answer to this?

phatchman:
> 
> What was the answer to this?

It was only needed for MenuItemWidget.data(), because it is chained with another call so receives a const MenuItemWidget. So, it was a consequence of changing self.wd to self.data(), not because of the new validate() stuff.
```
pub fn data(self: *const MenuItemWidget) *WidgetData {
    return self.wd.validate();
}
```

PR #425, opened 2025-06-30
Unmanaged containers and removal unnecessary default undefined values 

RedPhoenixQ:
Having multiple or most fields in widgets default to `undefined` was always a bit confusing to me. It wasn't immediately obvious if the fields where set after `init`, `install` or some other function. This removes all default undefined values for fields that can be set in `init` and adds `SAFETY:` comments above the rest specifying when they are set.

Additionally, to remove some undefined values for the hash maps and array lists in `Window`, all `Window` containers have been changed to use their `...Unmanaged` variants. They all have a doc comment specifying which allocator is supposed to be used for each (which is important for users adding themes or key binds for example).

Sidenote, during this is became very obvious that most of these fields are only valid after `install` has been called. That means that `deinit` would probably panic on `undefined` values if `install` didn't get called. I realised that `WidgetData.rect_scale_cache` only gets set by `register`, which should only be called in `install`, so we can use that to detect if a widget was installed. This way we could make `deinit` safe to call even if `install` isn't called. I don't know if that is something we'd want to support or maybe the panic enforces correctness? This technique for detecting if a widget was installed can still be used by `ReorderWidget` and `FloatingTooltipWidget` though.

### DIFF
--- a/src/Font.zig
+++ b/src/Font.zig
@@ -140,14 +140,14 @@ pub const TTFBytes = struct {
     //pub const OpenDyslexicBdIt = @embedFile("fonts/OpenDyslexic/compiled/OpenDyslexic-Bold-Italic.otf");
 };
 
-pub fn initTTFBytesDatabase(allocator: std.mem.Allocator) std.mem.Allocator.Error!std.StringHashMap(dvui.FontBytesEntry) {
-    var result = std.StringHashMap(dvui.FontBytesEntry).init(allocator);
+pub fn initTTFBytesDatabase(allocator: std.mem.Allocator) std.mem.Allocator.Error!std.StringHashMapUnmanaged(dvui.FontBytesEntry) {
+    var result: std.StringHashMapUnmanaged(dvui.FontBytesEntry) = .empty;
     inline for (@typeInfo(TTFBytes).@"struct".decls) |decl| {
-        try result.putNoClobber(decl.name, dvui.FontBytesEntry{ .ttf_bytes = @field(TTFBytes, decl.name), .allocator = null });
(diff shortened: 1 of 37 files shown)


phatchman:
Like this! 
I'll rebase my data() validation changes on this once you merge. We hit a lot of the same files, but my changes are easier to apply.

david-vanderson:
Very good - thanks!  I briefly looked over the changes, but haven't tested anything.  Please give the demo a once-over test before merging.

> Sidenote, during this is became very obvious that most of these fields are only valid after `install` has been called.

This has always bugged me (the rarely used distinction between init and install), and this prompted me to reconsider it.  Maybe we could merge them and find some other mechanism to accomplish the few places it's actually used?  I will file an issue so I look into it.  Thanks!

phatchman:
> This has always bugged me (the rarely used distinction between init and install), and this prompted me to reconsider it. Maybe we could merge them and find some other mechanism to accomplish the few places it's actually used? I will file an issue so I look into it. Thanks!

It's the pointer initialization that causes the issue. Ypu could move to the model where init takes a pointer to self? Most widgets are heap allocated anyway, so instead if ret.* = init(), it would be init(ret).

Or box: BoxWidget = undefined; BoxWidget.init(box) for stack allocated.

david-vanderson:
> It's the pointer initialization that causes the issue. Ypu could move to the model where init takes a pointer to self? Most widgets are heap allocated anyway, so instead if ret.* = init(), it would be init(ret).
> 
> Or box: BoxWidget = undefined; BoxWidget.init(box) for stack allocated.

Good point, will have to see how that works.  I'm going to copy this to the new issue.

RedPhoenixQ:
All the demos work and there are no memory leaks!

PR #427, opened 2025-06-30
Themes: use .zon for quickthemes for comptime builtin themes

RedPhoenixQ:
Zig 0.14.0 allows us to import .zon files directly into a zig type. This allows for comptime themes, still defined in separate files without the need for runtime parsing (but it's still possible through `std.zon`). The json `fromString` still remains an option too.

This doesn't require any large changes to the quick themes, except making all slices const and not requiring an allocator. The major benefit is that any theme can now be comptime, which allows for easier use of the builtin themes.

For example if a user wanted to set the `dracula` theme as default when initializing the `Window`, they would have to parse and allocate the quicktheme before passing it via pointer. Now you can simply do: `&dvui.Theme.builtin.dracula`


phatchman:
Yes please! I wanted to provide a small number of custom, themes for ADGUI, but it was too much mucking around.

david-vanderson:
Sounds great!

PR #429, opened 2025-06-30
Mark mouse events during capture to "handle" all captured events

### LINKED ISSUE #422: mouse capture should work like keyboard focus
While testing #410 I stumbled across a general problem with mouse capture and multiple events.

If you have this kind of sequence of events all in the same frame:
- mouse down (widget A captures)
- mouse motion (widget A declines to process, leaves it unhandled)
- mouse up (widget A ends capture)

Now a containing widget might process the mouse motion event.  Even though the motion event happened logically during A's mouse capture, once A runs and ends capture, that information is lost.

I think this is the same problem that we solve with keyboard focus by marking the events.  So the solution would be something like:
* when a widget takes mouse capture, provide an event num, we mark all the rest of the events with that widget id
* when a new event comes in, we also mark it with the current mouse capture widget id if any
* when a widget gives up mouse capture, also provide an event num and we remove the mark for all remaining events.

RedPhoenixQ:
Closes #422

This prevents the motion event from the following pattern from being handled by some containing widget:

- mouse down (widget A captures)
- mouse motion (widget A declines to process, leaves it unhandled)
- mouse up (widget A ends capture)

This repurposes the `focus_widgetId` and `foces_windowId` from the key event focus marking. I'm considering if these should be renamed to `widget_id_mark` and `window_id_mark` along with `focusRemaining...` being renamed `markRemaining...` to distance them from focus only?

This does however make it impossible to scroll the "multiline text area" in the demo with touch within a single frame. The capture always happens on press and only 2 frames later is the capture automatically removed and the scroll area can see the event. I don't know what the correct behaviour here is.

### DIFF
--- a/src/Event.zig
+++ b/src/Event.zig
@@ -35,6 +35,14 @@ pub fn handle(self: *Event, src: std.builtin.SourceLocation, wd: *const dvui.Wid
     self.handled = true;
 }
 
+pub fn focusable(self: *const Event) bool {
+    return switch (self.evt) {
+        // Only wheel events cannot be focused/captured
+        .mouse => |me| me.action != .wheel_x and me.action != .wheel_y,
+        else => true,
+    };
+}
+
 pub const Text = struct {
     txt: []u8,
     selected: bool = false,
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -2261,7 +2261,8 @@ pub fn reorderListsAdvanced() void {
     defer vbox.deinit();
 
     if (added_idx) |ai| {
-        reorder.dragStart(ai, added_idx_p.?); // reorder grabs capture
+        // FIXME: What events should/shouldn't be marked by this?
+        reorder.dragStart(ai, added_idx_p.?, 0); // reorder grabs capture
     }
 
     var seen_non_floating = false;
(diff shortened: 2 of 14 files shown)


david-vanderson:
I still have a few details to check on, and I will figure out the naming.  But does this commit fix scrolling in the single frame for you?

RedPhoenixQ:
> But does this commit fix scrolling in the single frame for you?

Yes it does. Now if the down, motion, up sequence is done in a single frame in the text area it still scrolls!

I didn't have a complete grasp of when/where this problematic behavior would cause issues so I might have missed something. Thank you for following up!

david-vanderson:
Phew!  Yeah this one was very complicated.  I really appreciate it!

I still need to (list for my memory):
- adjust naming
- figure out if the mouse position event should be capturable, and include it if so
- go through the FIXME comments and decide what should happen
- fix a remaining bug with textlayout<->scrollContainer:
  * same finger down, motion, finger up in same frame
  * textlayout gives up capture, and scroll container correctly takes capture
  * but textlayout processes the finger up first
  * so scroll container has capture but will never see a finger up (so keeps capture across frames)

I think the right solution is to have scroll container always look for finger up events (even if it was marked for another widget or handled) and give up capture.  I'm pretty sure this is only a scroll container problem because most widgets take capture from a finger down - scroll container is in this situation because of the weird way it grabs capture from a touch motion event.  Yes - writing this down helped.

RedPhoenixQ:
@david-vanderson Did you run `zig build`, `zig build check` or `zig build test`? `zig build check` in particular I think only validates that the files parse correctly, not that the semantics are correct

david-vanderson:
> @david-vanderson Did you run `zig build`, `zig build check` or `zig build test`? `zig build check` in particular I think only validates that the files parse correctly, not that the semantics are correct

I ran `zig build` which compiled successfully.  The only thing I can think of is that some code paths weren't exercised in the demo, only via the testing `refAllDecls`.

RedPhoenixQ:
> The only thing I can think of is that some code paths weren't exercised in the demo, only via the testing `refAllDecls`.

`zig build` should build the test binary and thus reference all test path. Wierd...

PR #430, opened 2025-06-30
Color: store pixels in []Color.PMA for better type safety and ergonomics

RedPhoenixQ:
Because of the `extern` on `Color.PMA`, the layout should always match the c abi and the fields should never be reordered. This means that `Color.PMA` is safe to `@bitCast` to and from a `[@sizeOf(Color.PMA)]u8` and `@ptrCast([]Color.PMA)` to and from `[]u8` is safe, as long as the length is a multiple of `@sizeOf(Color.PMA)`.

A comptime assert has been added to verify this casting behaviour (and this can be reverted if that ever causes an error on some platform but it really shouldn't). 

This removes a lot of `* 4` when calculating the amount of pixels. It also makes creating arrays of pixels much simpler with the added color decls, see "animating frames" demo,  `ColorPickerWidget.getValueSaturationTexture` and font atlas creation.

### DIFF
--- a/src/Color.zig
+++ b/src/Color.zig
@@ -265,14 +265,48 @@ pub fn average(self: Color, other: Color) Color {
     };
 }
 
-/// A color premultiplied by alpha, mostly used for vertex colors
+/// A color premultiplied by alpha
+///
+/// This type is safe to `@bitCast` to and from a `[@sizeOf(Color.PMA)]u8`.
+/// This also means `@ptrCast` of `[]Color.PMA` to and from `[]u8` is safe,
+/// as long as the length is a multiple of `@sizeOf(Color.PMA)`.
+///
+/// The pixel format will always be RGBA.
 pub const PMA = extern struct {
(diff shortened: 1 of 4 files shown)


david-vanderson:
I like it!  Does endianness affect this at all?  Not sure how much to worry about it until/unless someone shows up with a big-endian system, not sure how well we could even test it.

RedPhoenixQ:
Endianness shouldn't affect it as there are 4 individual u8 fields and not a combined u32. The c abi lays these out in the same order since the alignment doesn't change and the u8 itself doesn't change from endianness.

> not sure how well we could even test it.

This is what the comptime assert is for. If it ever triggers then it is an issue, but it really shouldn't. 

I will add an extra assert for `@sizeOf(Color.PMA)  == 4` so no mistakes can be made there either.

david-vanderson:
Great!  Merge when ready.  Thank you!

PR #431, opened 2025-07-01
Detect preferred color scheme where possible

RedPhoenixQ:
This now takes the users preferred color scheme into account when picking a default theme. `sdl3` already changed the window decorations to match the preferred theme, at least on Windows with out current init flags, so now dvui matches the theme of the window.

A use of dvui can also use this function directly from the backend to decide on a theme to pass into `Window.init`.

Should there be a way to opt out or set the fallback value for this?

In the future there is also ways to watch for changes in the preferred color scheme. This would require some sort of hook back in to dvui, perhaps a function returning the change for a single frame or `Window` storing a `light_theme` and `dark_theme` and swapping between them?
- [win32](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regnotifychangekeyvalue)
- [sdl3](https://wiki.libsdl.org/SDL3/SDL_EVENT_SYSTEM_THEME_CHANGED)
- [web](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList#events)

ps: for win32 api reasons the error handling functions in the `dx11` backend where renamed when I discovered 2 more potential return types that would also need their own error conversion functions.

### DIFF
--- a/src/Backend.zig
+++ b/src/Backend.zig
@@ -118,6 +118,11 @@ pub fn openURL(self: Backend, url: []const u8) GenericError!void {
     return self.impl.openURL(url);
 }
 
+/// Get the preferredColorScheme if available
+pub fn preferredColorScheme(self: Backend) ?dvui.enums.ColorScheme {
+    return self.impl.preferredColorScheme();
+}
+
 /// Called by `dvui.refresh` when it is called from a background
 /// thread.  Used to wake up the gui thread.  It only has effect if you
 /// are using `dvui.Window.waitTime` or some other method of waiting until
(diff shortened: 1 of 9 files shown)


david-vanderson:
Love it!  Merge!

> Should there be a way to opt out or set the fallback value for this?

Unless you feel otherwise, I would wait until someone shows up actually needing to override it.

phatchman:
One impact of this is that the output of my grid tests are now dependent on the OS theme of the machine it is running on. Is it possible to default this to the light theme when running tests?

RedPhoenixQ:
> One impact of this is that the output of my grid tests are now dependent on the OS theme of the machine it is running on. Is it possible to default this to the light theme when running tests?

Absolutely. We didn't add that right away as we didn't know if there was a use case. I'll add it to the window config and set a default for the testing backend

PR #435, opened 2025-07-02
Add URL validation to openURL()

phatchman:
After playing with this a bit, this PR my suggested approach.

The rationale for returning a bool over an error is:
1) If the user wants more details on why the url is invalid, they should validate it themselves before sending to open. My goal is just to make sure they aren't doing anything dumb.
2) If the backend returns an error, I would need to map that into something the user could understand, so logging and retuning false seems to be the best compromise.


david-vanderson:
Merge when ready - thanks!

PR #436, opened 2025-07-03
Reuse data allocations if they are the same size

RedPhoenixQ:
This was removed in #209 but it causes large amounts of reallocations for the common pattern of `dataGet` and then later `dataSet` with the same id and key.

In the app demo the amount of trashed data in a single frame peaks at 2694 when opening the demo window. The scrolling demo also has a ~1000 data trashed every frame. With this change we go back to reusing the previous allocation as long as the size matches, otherwise it's trashed. With this change it peaks at 7 while caching and is 0 most frames.

This still prevents segmentation faults if a value or pointer was gotten earlier in the frame. The only difference in behavior is if you use `dataGetPtr` or `dataGetSlice` and then `dataSet`/`dataSetSlice` with the same size, you will see the new value updated through the pointer/slice. I don't believe this is an issue and was the working behavior before #209.

### DIFF
--- a/src/Window.zig
+++ b/src/Window.zig
@@ -1301,36 +1301,49 @@ pub fn dataSetAdvanced(self: *Self, id: WidgetId, key: []const u8, data_in: anyt
     self.data_mutex.lock();
     defer self.data_mutex.unlock();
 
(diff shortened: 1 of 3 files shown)


david-vanderson:
Looks great - merge when ready - thanks!

PR #437, opened 2025-07-03
Catch use-after-free errors with no reuse option in ShrinkingArenaAllocator 

RedPhoenixQ:
@foxnne encountered a use-after-free in the examples (7556eca79bff56756c4e3da224c87dd3bbf99972) which was not caught by `WidgetData.validate`. This indicates that reusing memory prevents the effectiveness of the undefined checks that have been implemented, as mentioned in #414.

This PR makes `ShrinkingArenaAllocator` configurable and add the `reuse_memory` option which can be disabled. This duplicates all allocations in a separate `std.heap.ArenaAllocator` that never frees anything and only returns pointers to that memory. It does still perform the allocations in the normal arena to validate lifo behaviours.

### DIFF
--- a/src/Window.zig
+++ b/src/Window.zig
@@ -125,10 +125,10 @@ capture: ?dvui.CaptureMouse = null,
 captured_last_frame: bool = false,
 
 gpa: std.mem.Allocator,
-_arena: dvui.ShrinkingArenaAllocator(.{}),
-_lifo_arena: dvui.ShrinkingArenaAllocator(.{}),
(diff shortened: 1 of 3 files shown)


david-vanderson:
Code looks good.  Is there any particular testing you want me to do?

RedPhoenixQ:
I don't think so. It successfully caught the use after free and there seems to be no issues otherwise. Maybe we merge and fix the issues if anythings comes up?

david-vanderson:
Sounds good!

PR #438, opened 2025-07-03
Common backend functions

RedPhoenixQ:
This adds a place to share common functionality between backends.

Currently only the Windows color scheme registry lookup and Windows `attachConsole` is shared. This also makes it easier for dvui users to do the `attachConsole` workaround through `dvui.Backend.Common.windowsAttachConsole`.

There is also the potential to implement color scheme check for macOS, I just haven't found a good method/snippet of objective-c code to do this.


david-vanderson:
Good idea - merge when ready.  Thanks!

PR #439, opened 2025-07-03
`image()` expose interpolation option

MotokoKusanagii:
Adds an interpolation option to `ImageInitOptions`.
The default is `.linear`, preserving existing behavior.


david-vanderson:
This is great.  Thank you!

PR #442, opened 2025-07-05
ImageSource

david-vanderson:
Consolidates previously separate parameters to image() and imageSize() into a single struct that represents a raster image plus an interpolation strategy.

- [ ] add caching strategy to ImageSource (great idea from Meiko on discord)
  * regen only on base pointer change
  * regen on any pixel change
  * regen always

- [ ] add interpolation to the hash

### DIFF
--- a/src/Backend.zig
+++ b/src/Backend.zig
@@ -69,9 +69,9 @@ pub fn drawClippedTriangles(self: Backend, texture: ?dvui.Texture, vtx: []const 
     return self.impl.drawClippedTriangles(texture, vtx, idx, clipr);
 }
 
-/// Create a `dvui.Texture` from the given `pixels` in RGBA.  The returned
-/// pointer is what will later be passed to `drawClippedTriangles`.
-pub fn textureCreate(self: Backend, pixels: [*]u8, width: u32, height: u32, interpolation: dvui.enums.TextureInterpolation) TextureError!dvui.Texture {
+/// Create a `dvui.Texture` from premultiplied alpha `pixels` in RGBA.  The
+/// returned pointer is what will later be passed to `drawClippedTriangles`.
+pub fn textureCreate(self: Backend, pixels: [*]const u8, width: u32, height: u32, interpolation: dvui.enums.TextureInterpolation) TextureError!dvui.Texture {
     return self.impl.textureCreate(pixels, width, height, interpolation);
 }
 
--- a/src/Color.zig
+++ b/src/Color.zig
@@ -354,6 +354,32 @@ pub const PMA = extern struct {
     pub fn castToColor(self: PMA) Color {
         return .{ .r = self.r, .g = self.g, .b = self.b, .a = self.a };
     }
+
+    /// Alpha multiplies `pixels` in place
+    pub fn sliceFromRGBA(pixels: []u8) []Color.PMA {
+        for (0..pixels.len / 4) |ii| {
+            const i = ii * 4;
+            const a = pixels[i + 3];
+            pixels[i + 0] = @intCast(@divTrunc(@as(u16, pixels[i + 0]) * a, 255));
+            pixels[i + 1] = @intCast(@divTrunc(@as(u16, pixels[i + 1]) * a, 255));
+            pixels[i + 2] = @intCast(@divTrunc(@as(u16, pixels[i + 2]) * a, 255));
+        }
+        return @ptrCast(pixels);
+    }
+
+    /// Unapplies the alpha multiplication in place, returning the inner slice
+    pub fn sliceToRGBA(pma_pixels: []Color.PMA) []u8 {
+        var pixels: []u8 = @ptrCast(pma_pixels);
+        for (0..pixels.len / 4) |ii| {
+            const i = ii * 4;
+            const a = pixels[i + 3];
+            if (a == 0) continue;
+            pixels[i + 0] = @intCast(@divTrunc(@as(u16, pixels[i + 0]) * 255, a));
+            pixels[i + 1] = @intCast(@divTrunc(@as(u16, pixels[i + 1]) * 255, a));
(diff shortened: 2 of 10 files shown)


RedPhoenixQ:
I like this a lot! Removing the awkwardly named `RGBAPixelsPMA` is very nice now that we can use `[]Color.PMA` to describe the same thing.

MotokoKusanagii:
In my use case, the image is guaranteed to change every frame. That’s why I think there should be a third option that regenerates the image unconditionally.

RedPhoenixQ:
Maybe a more generic option where you provide your own key. Could be some metadata that you hash from a database or in your case @MotokoKusanagii a hash of the current frame index.

Or a way to pass in your own texture directly and just use the image function for sizing and display?

MotokoKusanagii:
>Maybe a more generic option where you provide your own key. Could be some metadata that you hash from a database or in your case @MotokoKusanagii a hash of the current frame index.

That would work for me as well, and it's probably the better solution.

> Or a way to pass in your own texture directly and just use the image function for sizing and display?

In my mind, the `image()` widget is meant for cases where you don’t want to manually handle recreating textures. Maybe there could be a `texture()` widget for directly rendering textures, essentially a thin wrapper around `renderTexture()`.

RedPhoenixQ:
I my mind `renderImage` is already a very thin wrapper around `renderTexture` so one if statement calling `renderTexture` directly wouldn't be that strange. `image` is also very generic and handles both image files and raw bytes, so adding textures (which is just bytes in another form) isn't that far fetched

MotokoKusanagii:
Okay, yeah, I understand now. I somehow thought of textures and images (as in bytes in ram) as two separate concepts, but you’re right, they’re both just bytes in the end.

david-vanderson:
> Okay, yeah, I understand now. I somehow thought of textures and images (as in bytes in ram) as two separate concepts, but you’re right, they’re both just bytes in the end.

That's true, but generally textures live in gpu memory and images in cpu memory, which is why it's annoying to convert between them.

RedPhoenixQ:
In my previous use case I had a list/grid of images loaded from disk. Keeping these in RAM just to be able to call `image` was quickly getting expensive with many 100s of MB for decently sized images. 

What I did was:
- load the image
- create the texture
- put it in dvuis cache manually (with database key as the map key)
- then create my own `image` to render the texture (taking only the database key that I got from a foreign key from another query)

This was a lot of work when I could've just passed the texture to `image` and perhaps have access to a function to add key/texture to the texture cache. Something like `TextureCacheEntry.fromTextureWithKey` would be really nice.

david-vanderson:
> In my previous use case I had a list/grid of images loaded from disk. Keeping these in RAM just to be able to call `image` was quickly getting expensive with many 100s of MB for decently sized images.
> 
> What I did was:
> 
>     * load the image
> 
>     * create the texture
> 
>     * put it in dvuis cache manually (with database key as the map key)
> 
>     * then create my own `image` to render the texture (taking only the database key that I got from a foreign key from another query)
> 
> 
> This was a lot of work when I could've just passed the texture to `image` and perhaps have access to a function to add key/texture to the texture cache. Something like `TextureCacheEntry.fromTextureWithKey` would be really nice.

That is a very cool usecase.  I assume you wanted to use the texture cache because that would automatically clean up textures you weren't using?

In that case it sounds like what you need is a kind of ImageSource where you can query if the texture is in the cache before calling `image()`, and if not go grab data from the database.  Is that right?

RedPhoenixQ:
> That is a very cool usecase. I assume you wanted to use the texture cache because that would automatically clean up textures you weren't using?

Yes exactly.

> In that case it sounds like what you need is a kind of ImageSource where you can query if the texture is in the cache before calling `image()`, and if not go grab data from the database. Is that right?

I did forget the step of checking if the key was in the cache, otherwise load the file and do the rest. If the `fromTextureWithKey` function is added, a corresponding `TextureCacheEntry.hasKey` would be nice to be able to generate/load the texture if it gets uncached.

RedPhoenixQ:
So i guess what I want is

```zig
pub const ImageSource = struct {
  bytes: union(enum) {
    ...
    key: u64,
  }
  
  ...

  /// Logs error if `bytes` is not `.key`?
  pub fn hasTexture() bool { ... }   // Maybe named `hasKey`

  /// Adds a texture to the cache with the provided key
  /// 
  /// Logs error if `bytes` is not `.key`?
  pub fn addTexture(texture: Texture) void { ... }
}
```

That would allow something like:

```zig
const im_src: ImageSource = .{ .bytes = .{ .key = hash } }
if (im_src.hasTexture()) {
  ... 
  const tex = ...; // Create texture
  im_src.addTexture(text);
}
dvui.image(@src(), im_src, .{});
```

This might warrant some renaming, specifically of `bytes`.

Another option:

```zig
pub const TextureCacheEntry = struct {
  // This should always return the cache entry, but if adding to cache fails it will just not be cached for the next frame
  pub fn add(key: u64, texture: Texture) TextureCacheEntry {
    ...
    return new_entry;
  }

  pub fn get(key: u64) ?TextureCacheEntry {
    ...
    return existing_entry_or_null;
  }
}
```

Which would allow:

```zig
const key: u64 = ...; // Some key
const entry = if (TextureCacheEntry.get(key)) |entry| entry else blk: {
  ... 
  const tex = ...; // Create texture
  break :blk TextureCacheEntry.add(key, tex);
}
dvui.image(@src(),  .{ .bytes = .{ .cached = entry } }, .{});
```

RedPhoenixQ:
I've though of an alternative to this that removes `TextureCacheEntry` and moves some functions. Will make a separate PR and we can see which we prefer

david-vanderson:
@MotokoKusanagii If you get a chance please let us know if this works for you, and any feedback you have.  Thanks!

PR #445, opened 2025-07-05
ImageSource and texture cache changes

RedPhoenixQ:
This extends #442 

This adds functions to interact with the texture cache directly which means `TextureCacheEntry` is no longer needed for hashing. The creation and caching of textures has been separated so that `ImageSource` can do it's own logic while also letting users create textures and cache them as they need.

This way the user can choose to create their own textures and use the dvui cache for automatic cleanup (some more docs on `textureAddToCache` are needed before merging).

`ImageSource` has also been changed to a `union(enum)` to avoid more levels of `.{ .XXX = .{ ... } }`. The `hash` function also makes it possible for the user to invalidate the texture using `textureInvalidateCache` without some special function on `ImageSource`. If we want, each anonymous struct in the union could also be extracted and named separately.

Additionally, hashing of `interpolation` is included as well as an invalidation strategy to determine how the hash is calcualted.

### DIFF
--- a/src/Backend.zig
+++ b/src/Backend.zig
@@ -69,9 +69,9 @@ pub fn drawClippedTriangles(self: Backend, texture: ?dvui.Texture, vtx: []const 
     return self.impl.drawClippedTriangles(texture, vtx, idx, clipr);
 }
 
-/// Create a `dvui.Texture` from the given `pixels` in RGBA.  The returned
-/// pointer is what will later be passed to `drawClippedTriangles`.
-pub fn textureCreate(self: Backend, pixels: [*]u8, width: u32, height: u32, interpolation: dvui.enums.TextureInterpolation) TextureError!dvui.Texture {
+/// Create a `dvui.Texture` from premultiplied alpha `pixels` in RGBA.  The
+/// returned pointer is what will later be passed to `drawClippedTriangles`.
+pub fn textureCreate(self: Backend, pixels: [*]const u8, width: u32, height: u32, interpolation: dvui.enums.TextureInterpolation) TextureError!dvui.Texture {
     return self.impl.textureCreate(pixels, width, height, interpolation);
 }
 
--- a/src/Color.zig
+++ b/src/Color.zig
@@ -354,6 +354,32 @@ pub const PMA = extern struct {
     pub fn castToColor(self: PMA) Color {
         return .{ .r = self.r, .g = self.g, .b = self.b, .a = self.a };
     }
+
+    /// Alpha multiplies `pixels` in place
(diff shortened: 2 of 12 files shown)


david-vanderson:
This is fantastic, thanks so much!  Merge!

david-vanderson:
I'm a bit confused by ImageSource.hash when ImageSource is `.texture`, so maybe I'm missing something.  When `.texture`, then it doesn't do anything with the texture cache right?

RedPhoenixQ:
It indeed doesn't interact with the cache. I choose to hash the fields of `Texture` to have something to return from `hash`. It could perhaps just return 0.

Do you think it should mention this in the docs?

david-vanderson:
> It indeed doesn't interact with the cache. I choose to hash the fields of `Texture` to have something to return from `hash`. It could perhaps just return 0.
> 
> Do you think it should mention this in the docs?

Yeah - let's have hash return 0 with a comment and something in the doc comments for `.texture` that says it doesn't participate in the texture cache at all.  Thanks!

RedPhoenixQ:
Fixed in 54b87e6d96ed5cf968685ea815d986c969269615

nat3Github:
- i like ImageSource and that it is top level in dvui. i like the refactor of tvg and texture caching!
- i would like ImageSource.pixels.rgba to remain a []const u8 in the future since its the easiest to cast to using std.mem.sliceAsBytes, imo the user / developer should be capable to use it correctly and bugs are not subtle
- there is a typo in ImageSource.InvalidationStrategy
- have you benchmarked texture recreation vs hash recalculation with big images?

RedPhoenixQ:
> * i would like ImageSource.pixels.rgba to remain a []const u8 in the future since its the easiest to cast to using std.mem.sliceAsBytes, imo the user / developer should be capable to use it correctly and bugs are not subtle

That's good feedback. The public API could definitely remain `[]const u8` with the possibility for internal helper functions to take `[]const Color` for clarity.

> * there is a typo in ImageSource.InvalidationStrategy

Good catch, will fix.



> * have you benchmarked texture recreation vs hash recalculation with big images?

Not at all. The texture cache existed for images before I started to work with the project so I have assumed it's beneficial. There is a chance that for large images (when hashing the bytes) the difference between caching and recreating might be minimal. The `always` invalidate option does allow you to never cache the texture if you see a performance benefit.

david-vanderson:
> * have you benchmarked texture recreation vs hash recalculation with big images?

This kind of information would be useful!  I don't know very much about how expensive creating a new texture is across different backends/hardware.  My assumption is that sending the image data over to GPU ram is worse than hashing the bytes in CPU ram, but performance is tricky.

Issue #444, opened 2025-07-05
'Interface' for tooltips


vlappa:
It seems @phatchman is working on a way to be able to pass custom keybindings to his grid widget. Assuming the same thing would be possible for similar widgets in dvui, that's a great step forward I think. 

There is a similar 'problem' for tooltips I think. Now I've to copy buttonIcon function to my codebase and add something like:

```
 const Button = enum {
        Quit,
        New,
        Close,
       Save,


    };

    // disadvantage: can't have a button with two words
    const task = std.meta.stringToEnum(Button, label_str) orelse {
        return error.InvalidChoice;
    };

    if (conf.show_tooltip) {
        const tooltip_txt = switch (task) {
            .Quit => "Quit application",
            .New => "Create new session",
            .Close => "Close session",
           .Save => "Save session",

        };

        dvui.tooltip(@src(), .{ .position = .vertical, .active_rect = bw.data().borderRectScale().r }, "{s}", .{tooltip_txt}, .{});
    }
```

This was just a quick thing to make it work and it's not a big deal to copy the function. But I can imagine a possibility where you can pass some kind of struct or interface to the function which handles the tooltips.

david-vanderson:
Thanks very much for this issue!  I'm finally starting to see the pattern here.

In this case, you want to call `dvui.button()` but then call `dvui.tooltip()` with the rectangle of the button.

In the "default focus in dialogs" case, we needed to to call `dvui.button()` but then get the id of the button.

I'm thinking that maybe we need a universal way to extract the WidgetData from a high-level api.  I wonder if adding an out parameter to Options that would receive a copy of the WidgetData makes sense?

Will try that and report back.

phatchman:
> I'm thinking that maybe we need a universal way to extract the WidgetData from a high-level api. I wonder if adding an out parameter to Options that would receive a copy of the WidgetData makes sense?
> 
That would definitely help the plotting app. When there are multiple widgets in a cell, i need to know their positions/sizes, so I can know which widget was clicked when the cell is clicked. I'm using tags for this now, but it's awkward.

david-vanderson:
Good, that makes at least 3 examples, so we are on the right track!

david-vanderson:
I've pushed a new `Options.out_data` for this.  Let me know how it goes using it for tooltips and grid stuff!

phatchman:
[...]

PR #448, opened 2025-07-07
GridWidget: selection and navigation

phatchman:
This PR introduces the following:
* Grid functions now take and return Cell, instead of col_num, row_num.
* Support for Single and Multi-selection across multiple "selectables" e.g. checkboxes. 
   - Selection is "generic" in that it can work independently of the grid.
* "select all" via keyboard shortcut.
* Keyboard navigation for grid.
* Removal of "column functions" gridColumnFromSlice and gridColumnCheckbox.
* Introduces checkboxEx, which takes an optional SelectionInfo.
* Two new examples demonstrating selection and navigation.
* gridColumnHeading argument order was changed for consistency.
* var_row_heights renamed to row_height_variable.
* Numerous minor cleanups and improvements.

Partially implements #317
Had a long genesis which is recorded in #370.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -5350,7 +5350,12 @@ fn gridSelection() void {
         defer top_controls.deinit();
         dvui.labelNoFmt(@src(), "Filter (contains): ", .{}, .{ .margin = dvui.TextEntryWidget.defaults.margin });
         var text = dvui.textEntry(@src(), .{}, .{ .expand = .horizontal });
-        local.filename_filter = text.getText();
+        if (text.text_changed) {
+            local.filename_filter = text.getText();
+            local.filtering_changed = true;
+        } else {
+            local.filtering_changed = false;
+        }
         defer text.deinit();
     }
     {
@@ -5418,7 +5423,6 @@ fn gridSelection() void {
 
         if (local.row_select)
             local.highlight_style.processEvents(grid);
-        const was_filtering = local.filtering;
 
         {
             var itr: DirEntry.Iterator = .init(&directory_examples);
@@ -5488,8 +5492,6 @@ fn gridSelection() void {
                 }
             }
         }
-
-        local.filtering_changed = (was_filtering != local.filtering);
 
         if (local.selection_mode == .multi_select) {
             local.kb_select.processEvents(&local.select_all_state, grid.data());
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -4537,7 +4537,7 @@ pub fn icon_browser(src: std.builtin.SourceLocation, show_flag: *bool, comptime 
 const grid_panel_size: Size = .{ .w = 250 };
 
 pub fn grids(height: f32) void {
-    // Below is a workaround for standaline demos being placed in a scroll-area.
+    // Below is a workaround for standalone demos being placed in a scroll-area.
     // By default the scroll-area will scroll instead of the grid.
     // Fix the height of the box around the tabs to prevent this.
     // This would not be required for normal usage.
@@ -4547,14 +4547,16 @@ pub fn grids(height: f32) void {
         min_size_content = .{ .h = height };
         max_size_content = .height(height);
     }
-
     const GridType = enum {
         styling,
         layout,
         scrolling,
         row_heights,
+        selection,
+        navigation,
         const num_grids = @typeInfo(@This()).@"enum".fields.len;
     };
+
     const local = struct {
         var active_grid: GridType = .styling;
 
@@ -4564,10 +4566,12 @@ pub fn grids(height: f32) void {
 
         fn tabName(grid_type: GridType) []const u8 {
             return switch (grid_type) {
-                .styling => "Styling and sorting",
-                .layout => "Layouts and data",
-                .scrolling => "Virtual scrolling",
-                .row_heights => "Variable row heights",
+                .styling => "Styling and\nsorting",
+                .layout => "Layouts and\ndata",
+                .scrolling => "Virtual\nscrolling",
+                .row_heights => "Variable row\nheights",
+                .selection => "Selection\n ",
+                .navigation => "Keyboard\nnavigation",
             };
         }
     };
@@ -4578,7 +4582,6 @@ pub fn grids(height: f32) void {
         var tabs = dvui.TabsWidget.init(@src(), .{ .dir = .horizontal }, .{ .expand = .horizontal });
         tabs.install();
         defer tabs.deinit();
-
         for (0..GridType.num_grids) |tab_num| {
             const this_tab: GridType = @enumFromInt(tab_num);
 
@@ -4593,6 +4596,8 @@ pub fn grids(height: f32) void {
         .layout => gridLayouts(),
         .scrolling => gridVirtualScrolling(),
         .row_heights => gridVariableRowHeights(),
+        .selection => gridSelection(),
+        .navigation => gridNavigation(),
     }
 }
 
@@ -4761,7 +4766,7 @@ fn gridStyling() void {
                 temp += interval;
                 row_num += 1;
             }) {
-                var cell = grid.bodyCell(@src(), 0, row_num, cell_opts.cellOptions(0, row_num));
+                var cell = grid.bodyCell(@src(), .colRow(0, row_num), cell_opts.cellOptions(.colRow(0, row_num)));
                 defer cell.deinit();
                 dvui.label(@src(), "{d}", .{temp}, .{ .gravity_x = 0.5, .expand = .horizontal });
             }
@@ -4774,7 +4779,7 @@ fn gridStyling() void {
                 temp += interval;
                 row_num += 1;
             }) {
-                var cell = grid.bodyCell(@src(), 1, row_num, cell_opts.cellOptions(1, row_num));
+                var cell = grid.bodyCell(@src(), .colRow(1, row_num), cell_opts.cellOptions(.colRow(1, row_num)));
                 defer cell.deinit();
                 dvui.label(@src(), "{d}", .{@divFloor(temp * 9, 5) + 32}, .{ .gravity_x = 0.5, .expand = .horizontal });
             }
@@ -4812,23 +4817,13 @@ fn gridLayouts() void {
         const fixed_widths = [num_cols]f32{ checkbox_w, 80, 120, 80, 100, 300 };
         const equal_spacing = [num_cols]f32{ checkbox_w, -1, -1, -1, -1, -1 };
         const fit_window = [num_cols]f32{ checkbox_w, 0, 0, 0, 0, 0 };
-        var selection_state: dvui.GridColumnSelectAllState = .select_none;
+        var selection_state: dvui.selection.SelectAllState = .select_none;
         var sort_dir: GridWidget.SortDirection = .unsorted;
         var layout_style: Layout = .proportional;
         var h_scroll: bool = false;
         var resize_rows: bool = false;
 
         /// Create a textArea for the description so that the text can wrap.
-        fn customDescriptionColumn(src: std.builtin.SourceLocation, grid: *GridWidget, col_num: usize, data: []Car, opts: *const GridWidget.CellStyle.Banded) void {
-            for (data, 0..) |*car, row_num| {
-                var col = grid.bodyCell(src, col_num, row_num, opts.cellOptions(col_num, row_num));
-                defer col.deinit();
-                var text = dvui.textLayout(@src(), .{ .break_lines = true }, .{ .expand = .both, .background = false });
-                defer text.deinit();
-                text.addText(car.description, opts.options(col_num, row_num));
-            }
-        }
-
         const ConditionTextColor = struct {
             base_opts: *const GridWidget.CellStyle.Banded,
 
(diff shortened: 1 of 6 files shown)


phatchman:
Things to note about this PR.

1) There is a callback introduced in GridWidget.KeyboardNavigation. It is called for every "focus event" and is used to convert a mouse position into a cell. Required when user is using variable row heights or if there is more than one "focusable" widget per cell.
```
fn processEventsCustom(self: *KeyboardNavigation, grid: *GridWidget, cellConverter: fn (
        grid: *GridWidget,
        point: Point.Physical,
    ) ?Cell)
```

~2) I decided to keep the grid_cell_up and grid_cell_down as keybinds. Mostly because of the cmd/ctrl difference between OSs.~ Actually I don't use ctrl/cmd for this anymore. I'll remove those and just default to up and down arrow keys.

3) Removes all of the "column" functions i.e. gridColumnFromSlice and gridColumnCheckbox. The impacts of this can be seen in the gridLayouts() example. It requires minimal extra code per column, but gives maximum flexibility over using the old functions.

4) I created a new checkbox function named checkboxEx, similar to labelEx, rather than checkboxWithSelction or similar. I think that makes more sense for future extension. Most people won't need to pass SelectionInfo to checkbox, so breaking the existing checkbox API for everyone felt very bad. 

5) SelectionInfo vs SelectInfo (i.e. ScrollInfo vs ScrollingInfo)? I've gone with Selection for everything over Select. Select just seemed too generic. e.g. could be a type of widget, a SQL statement etc.

6) Creates a "selection" namespace. As selection can be used to control any group of selectable things I've put it in dvui.selection. I don't import the content into "dvui", so users will reference it as e.g. dvui.selection.SingleSelect rather than dvui.SingleSelect.

If you could cast your eyes over the new examples to check if I'm doing anything that's not "best practice"? I'd like the examples to showcase how you *should* use DVUI. I think I'm getting a handle on most things, but I'm sure there are still a lot of gaps.

Cheers!

david-vanderson:
> 1. There is a callback introduced in GridWidget.KeyboardNavigation. It is called for every "focus event" and is used to convert a mouse position into a cell. Required when user is using variable row heights or if there is more than one "focusable" widget per cell.

This is fine.  Maybe we'll figure out something in the future, but I don't have a better option for now.

> 3\. Removes all of the "column" functions i.e. gridColumnFromSlice and gridColumnCheckbox. The impacts of this can be seen in the gridLayouts() example. It requires minimal extra code per column, but gives maximum flexibility over using the old functions.

The code for the examples looks really good!

> 4\. I created a new checkbox function named checkboxEx, similar to labelEx, rather than checkboxWithSelction or similar. I think that makes more sense for future extension. Most people won't need to pass SelectionInfo to checkbox, so breaking the existing checkbox API for everyone felt very bad.

I think https://github.com/david-vanderson/dvui/issues/444 will fix this - it looks like you needed it to get the WidgetData back out from `checkbox`.  Doesn't hold up merging this, we can come back to it later.

> * SelectionInfo vs SelectInfo (i.e. ScrollInfo vs ScrollingInfo)? I've gone with Selection for everything over Select. Select just seemed too generic. e.g. could be a type of widget, a SQL statement etc.
> 
> * Creates a "selection" namespace. As selection can be used to control any group of selectable things I've put it in dvui.selection. I don't import the content into "dvui", so users will reference it as e.g. dvui.selection.SingleSelect rather than dvui.SingleSelect.

Naming looks good.  The keyboard nav example is great.  How complex to make the examples is something I don't have a great handle on.  Maybe we can figure out a way around using the tag `grid_focus_next`, but no reason to hold this up.

Overall looking excellent!  Merge when ready.  Thank you very much!!

phatchman:
> Naming looks good. The keyboard nav example is great. How complex to make the examples is something I don't have a great handle on. Maybe we can figure out a way around using the tag `grid_focus_next`, but no reason to hold this up.

Once we have #444, this tag will go away as well. I only need it to get the widget id of the correct widget to focus. If I could get that immediately, I'd focus directly after creating the widgets for the focused cell. Then that CellNav struct will go away as all it does is set the tag. The one thing it wouldn't let me do that I wanted to add was tabbing in to the cell to move the cursor to the end. I can do it with textEntry, just not with textEntryNumber. But I think that is a harder issue to solve unless it returns the underlying text entry as well.

On the examples. I think you have a good mix. The "basic widgets" are relatively simple, the other demos still show some advanced concepts like focus detection, linked and virtual scrolling etc. It would be nice to have some more "how to" style examples, maybe focusing on specific topics, like event handling before / after widgets, when to use the different types of "rects" etc. I'm sure they will come over time or there will be documentation equivalents. But IMHO documentation by example is far more valuable. 

The only big issue I had getting started was getting a "basic" app up and running that I could build from as there is a lot of code in the examples, spread across a few files. But I think this is pretty much solved now with the "app" example and your sample dvui application code.

For some real-world feedback. I started with DVUI, then got a bit frustrated with it, so tried IMGUI, didn't even make it to writing any code for it because everything seemed so difficult. After briefly considering using some other UI libraries, there aren't many that a) are truly cross-platform; b) will give you a consistent look and feel. and c) don't look like the widgets were designed in 1980s (slight exaggeration). But I very quickly got even more frustrated with those other libraries and came back to DVUI after about a week of exploring. The second time it pretty much clicked, when I worked out how to cut down the example and get my build file working. I haven't looked back since!

I truly believe that DVUI offers something special. It is just so nice to use compared to any other UI library I've used personally or professionally (and I've used a few). Building UIs with it is almost declarative. I used to get annoyed when I had to write a GUI, but with DVUI, the screen design just naturally flows from my brain to the code and everything just seems to make sense (once you understand it of course). Can't wait to see how far DVUI goes!

david-vanderson:
Thank you very much!  If you have any suggestions for how to (possibly) work around a future you getting frustrated initially, please let me know!

Well done!

phatchman:
> Thank you very much! If you have any suggestions for how to (possibly) work around a future you getting frustrated initially, please let me know!

Keep in mind that the only "serious" zig programming I had done prior to using DVUI was the command line version of that Altair disk application. That GUI was my first attempt at any GUI programming in zig and only my second zig project. I'm not saying it is a minority experience, rather that some of the frustration came from trying to learn so much so quickly and looking for short-cuts. (e.g. me thinking "IMGUI is more established. There are probably more examples and resources to help me, rather than me reading through DVUI source code").

1. I think the getting a basic app up and running part is already much better. Maybe it was before, and I just didn't notice the example application?

2. Maybe splitting the examples up into their own files? Being able to go to the "BasicWidgets" example and then be able to easily switch to the "Scrolling Example" when you are trying to stitch together two examples.
It also splits out the machinery for display the demos from the demo screens themselves. So a first-time user doesn't have to search for some string they see on the Example screen, they can just straight to example\BasicWidgets.zig
Would be happy to do split them up if you agree that it's useful.

3. I was struggling to debug the layout because, like all starting out projects, i mixed gui, data and logic together, obfuscating the structure. Knowing about the debug window would have helped a lot here.

The below are more "nice to have", or things I found confusing, rather than that initial frustration.

1. When to use which Rect took a long time for me to understand. Not just the Natural vs Physical, (which is much better now they are typed), but things like I want the "whole" widget I should use borderRect() vs using rect for everything, including the margins, vs contentRect() just for what fits inside the padding etc. 

2. Maybe an example the focuses solely on event processing. I didn't really understand when I should be handling events before / after widgets etc. I knew I got different results depending on when I did the processing, but didn't really understand why, even though it seems really obvious in hindsight.

3. An example on window resizing. E.g. laying out the fixed part of a window first with gravity and the variable parts later. 

4. Example with a custom theme - I think that is simpler now with recent changes, but it was never quite clear to me the proper way to do that.

5. Example with the screen rect / mouse to natural rect conversion that makes it clear why /when that happens. Maybe there is some way to visualize the difference in the demo. Again, it is a super simple concept in hindsight, but when it is another thing you have to know on top of everything else, it like "i'll just code this magic number in here for now that I know makes this all work and figure out how to get rid of it later" vs "Oh yeah! There is an example that shows me exactly how to do it properly the first time".

6. Commenting of examples. This is another thing I'd be happy to help with. More comments for examples would be useful and should focus on the "what am I doing", "why am I doing it this particular way", "what alternatives are there".  e.g. "When processing focus events, dvui.eventMatch is used instead of dvui.eventMatchSimple because ...... "

7. Maybe we put some of the examples in an "extras" if they are just demonstrations of a single concept, so as not to overwhelm people. Examples that are more "here is the code to do this" than something people would want to look at specifically.

Then we could have some really super-simple examples that focused on a single thing. Maybe? There is a line there somewhere and I don't think any of us knows exactly where that is yet. Clearly, a hundred examples are too many. No one would find anything and maintaining them would be too burdensome. But I like the idea of some examples, for when you are starting out, that are written in a "best practice way" , are very simple, and you can almost copy-paste into your code as a starting point.

The trouble with things like the grid examples is that as soon as you add options for the user to configure the example, it's no longer code that the user can copy-paste. They have to extract all of the conditional logic to find the "core" of the example.

8. I wish I had discovered the debugging features sooner. I know it is in the demo, but dvui.currentWindow().debug_window_show isn't the most obvious interface. Maybe promote this to a dvui function or make it more visible in some other way. That would have saved me some many print statements, recompile, run, more print statements when developing ADGUI.

9. I think the examples should encourage the style that defer is "always" used for widget deinit() unless just not practical. Often this just means creating a new scope so that the right widgets go in and out of scope at the same time. I think all of the use after deinit() have been from not using a defer within a scope.

10. Add a File->Exit to the examples. We should have a idiomatic way to handle closing an application. I guess, disabled for wasm.

On the positive side:

1) I know we're still figuring out where to go with namespaces, but having a lot of stuff in the functions in dvui.zig is actually really handy. I could open the file, collapse everything to functions and just walk through and see pretty much everything that dvui had to offer. So while we should continue to put things in appropriate namespaces, having that first contact point for a widget through "dvui.zig" is very valuable IMHO. 

2) The existing examples are *really* good. As much as I think we can improve them, they are really useful and one of the shining stars of DVUI. That's maybe another reason to split them out. So a user can more clearly see how little code it takes to do interesting things with DVUI.

3) The source code is generally very easy to understand and trace through. Compare the dvui source code to any other full-featured GUI library out here and I think DVUI wins easily in terms of being able to understand how it works from the source code alone. I use the source code in preference to the documentation almost all the time.

4) The support provided by you and @RedPhoenixQ in particular is truly awesome. Not scalable though. But that's a nice to have problem for the future!

Something I think would be a really nice feature for DVUI is to have a window similar to debug window, but it lets you change the widget options for the selected widget. So, this gives you an imitation of a graphical designing system, which is more for fine-tuning things like a max width, or a min size, or a color, or an expand or all those little polishing things that takes time to change in source code, recompile and re-run.

And even if the zig recompile times become almost instant, it is still useful, because getting to that specific screen or setting up that specific scenario again still takes time. Much easier to click on a widget and change the size until happy and then copy the correct widget options to clipboard to paste back into the code. I'm sure this could be improved, but something like it would save me a lot of time.

david-vanderson:
This is excellent feedback!  I'm going to read through it a few times, and then file issues for the particular things I think we can improve currently.

Totally agree about changing/overriding the widget options at runtime.  I'm pretty sure we can do this, just needs a lot of work ironing out the details.  I'm hoping to be able to use structEntry for it, but maybe a bespoke editor would be better.

Thanks very much!

RedPhoenixQ:
> Totally agree about changing/overriding the widget options at runtime. I'm pretty sure we can do this, just needs a lot of work ironing out the details. I'm hoping to be able to use structEntry for it, but maybe a bespoke editor would be better.

I'd like to look into this. Will be away for a few days but will give it a try during the weekend. I'm thinking bespoke editor with a nice way to copy the modified options so you can easily add them back into the code.

david-vanderson:
> I'd like to look into this. Will be away for a few days but will give it a try during the weekend. I'm thinking bespoke editor with a nice way to copy the modified options so you can easily add them back into the code.

Sounds awesome!

phatchman:
[Here](https://github.com/david-vanderson/dvui/compare/main...phatchman:dvui-dev:breakup_examples) is the start of splitting the first few examples out. It works especially well for calculator so far.

It didn't take long to do this, it is more just to show what it might look like. So, if it makes sense, I doubt it would take me more than a couple of hours to complete.

My thought is that a new user would appreciate being able to open up Examples/calculator.zig and basically have the whole application, without any other distractions.

Edit: OK I finished a first cut of this over lunch. No big deal if we don't want to proceed, but just quickly playing with it, I think it is an improvement.

david-vanderson:
> Edit: OK I finished a first cut of this over lunch. No big deal if we don't want to proceed, but just quickly playing with it, I think it is an improvement.

Looks good - let's go for it.  Do you need anything before merging?

phatchman:
No. I've tested all of Demo screens and didn't find anything that was not already an issue. (e.g. tree widget losing entries when you drag and drop between different "depths". But if I understand correctly, tree widget is still a work in progress).

david-vanderson:
> No. I've tested all of Demo screens and didn't find anything that was not already an issue. (e.g. tree widget losing entries when you drag and drop between different "depths". But if I understand correctly, tree widget is still a work in progress).

Good - go ahead and merge.  I did not know this - can you file an issue with instructions for reproducing it?

phatchman:
> I did not know this - can you file an issue with instructions for reproducing it?

I think this was my misunderstanding how the re-ordering works. If you open up the example and drag build.zig to above assets, it wasn't clear to me that this put it in the "assets" folder.

I now understand with that example you can't actually re-order things, only move them between folders.

What was confusing is that when I was dragging build.zig, there is a blue line above "assets", so it seems like I should be able to drag it to be above assets. I think it would have made more sense to me if all of assets was highlighted, rather than just the line above?

One area where I don't think that works well now is if you drag .gitignore, below license, that puts the file in tests. But if you drag build.zig below build.zig.zon, it reverts back to the original ordering. The UI doesn't give you any feedback that these two operations are different.

Happy to write this up as an issue.

david-vanderson:
Ah I see, yes please make an issue, we need to split "move into" vs. "move between".  Thanks!

Issue #449, opened 2025-07-07
Grid Layouts and Data demo causes constant refreshes


phatchman:
## Description

The "Layouts and Data" grid demo causes the framerate to hit 100% when horizontal scrolling is enabled

## Repro

1. Run a DVUI demo.
2. Go to the Layouts and Data Grid example.
3. Enable Horizontal Scrolling
4. Scroll to the right.
    - It happens on the 2nd click of my mouse hscroll wheel, but not the first.
5. Observe framerate.

### Actual
Screen continually refreshes.

### Expected
Frame should only refresh on new events.

PR #450, opened 2025-07-07
GridWidget: fix issue causing constant refreshes

phatchman:
- header scroll area was not being correctly synchronized with body

Fixes: #449

### DIFF
--- a/src/widgets/GridWidget.zig
+++ b/src/widgets/GridWidget.zig
@@ -295,7 +295,8 @@ pub fn deinit(self: *GridWidget) void {
     defer self.* = undefined;
     defer dvui.widgetFree(self);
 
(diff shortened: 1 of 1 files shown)


david-vanderson (commented):
That first line looks weird - it's setting self.hsi.viewport.x to itself?

phatchman:
Yeah, you are right. That first line isn't doing anything. I think I only need the second one.
In your linked scrolling example, it is doing the equivalent of:
```
    if (hsi.x != fv.x) bsi.viewport.x = hsi.viewport.x;
    if (bsi.viewport.x != fv.x) hsi.viewport.x = bsi.viewport.x;
```
I was trying to do the equivalent of that, but missed changing a hsi to a bsi on the first line.
But I don't think that first line is needed, for the grid anymore, because the header can't receive any scrolling events.
So I think I can remove that line entirely when I merge the selection and navigation PR.

PR #455, opened 2025-07-08
GridWidget: really fix scroll sync/refresh issue

phatchman:
Revert of #450, also reverted the correct fix put in #448.
This change restores the correct fix.

### DIFF
--- a/src/widgets/GridWidget.zig
+++ b/src/widgets/GridWidget.zig
@@ -324,7 +324,7 @@ pub fn deinit(self: *GridWidget) void {
     defer self.* = undefined;
     defer dvui.widgetFree(self);
 
-    if (self.hsi.viewport.x != self.frame_viewport.x) self.hsi.viewport.x = self.bsi.viewport.x;
+    if (self.bsi.viewport.x != self.frame_viewport.x) self.hsi.viewport.x = self.bsi.viewport.x;
 
     // resizing if row heights changed or a resize was requested via init options.
     if (self.resizing) {


phatchman:
It's not clear to my why this works:
```
    if (self.bsi.viewport.x != self.frame_viewport.x) self.hsi.viewport.x = self.bsi.viewport.x;
```
But this doesn't:
```
    self.hsi.viewport.x = self.bsi.viewport.x;
```

Why aren't they equivalent?

phatchman:
OK. That line never ever triggers, because hsi.viewport.x always equals frame_viewport.x.
I can remove that line completely and get the same functionality.
```
    std.debug.print("PRE {d}:{d}:{d}\n", .{ self.frame_viewport.x, self.hsi.viewport.x, self.bsi.viewport.x });
    //if (self.bsi.viewport.x != self.frame_viewport.x) self.hsi.viewport.x = self.bsi.viewport.x;
    std.debug.print("POST {d}:{d}:{d}\n", .{ self.frame_viewport.x, self.hsi.viewport.x, self.bsi.viewport.x });
```
Gives:
```
PRE 120:0:120
POST 120:0:120
PRE 140:0:140
POST 140:0:140
PRE 160:0:160
POST 160:0:160
PRE 180:0:180
POST 180:0:180
PRE 200:0:200
POST 200:0:200
PRE 200:0:200
POST 200:0:200
PRE 220:0:220
POST 220:0:220
```
I don't change the values in self.hsi at any point, except via passing "frame_viewport" to the scroll area. So I think the proper solution is removing that synchronisation line altogether? Right now, it isn't doing anything so can wait until the next grid change.

david-vanderson:
This was pretty confusing.  So if you have linked scrolling, there are (at least) two different ScrollInfo (header, body).  So we expect to need to sync their viewports, because either could be scrolled.  The logic is, only one can be scrolled in a frame, so if it's viewport changed, then set the other.

But then we added `event_rect` to ScrollContainer so that only it would get mouse events.  So only the body ScrollInfo is getting changed (by mouse events).  And `frame_viewport` overrides the header's viewport offset, meaning the only way to see it is if you turn on the scrollbar for the header.

So then the only things I can think of that would scroll the header would be a key event (some widget in the header had focus), or an explicit scrollTo thing, like tabbing through header widgets.

I'm not sure if this means we need to do anything.  Does this make sense?

phatchman:
Yeah. I'll test putting a scroll area in one of the header cells. That should be worst-case, I think?

phatchman:
So, I tried putting in a scroll area with a really long label in it, a long textLayout and a long text widget.

~~In all cases, whether using keyboard or scrollwheel, scrolling is totally contained within the header cell. Nothing is leaking over to the header or to the body.  (well in the label case, nothing scrolls, but I think that is expected)~~ Edit I needed to do a more complicated test.

Current functionality is:
1) Mouse scrolling via scroll-wheel, scrolls the internal scroll container first and then body.
2) Keyboard scrolling is contained to the header cell. pressing arrow at end of the scroll area does not propagate to body.

Playing with this a little, it does "feel right".
If I arrow down a list in the header until I reach the end, and then arrow one more time, would I expect the body to scroll? I think I'd expect it to stay on the last item in the header?

I can publish a test branch if you want to see how it feels. Not sure what we can compare it to?

david-vanderson:
> Playing with this a little, it does "feel right".

That's good enough for me - thanks for testing it!

PR #463, opened 2025-07-10
Tree: Add option to disable reordering event handling

foxnne:
As suggested [here](https://github.com/david-vanderson/dvui/pull/413#issuecomment-3038579623).

@david-vanderson Not sure if you have any opinions surrounding that comment, we could reduce the effects on the example file tree, I just felt like it showcased what all you can do. There are regular `Options` for the file tree's box, each branch's button, and the expanded and indented box where borders outline the files. I think other than the missing switch to disable reordering, the other requests are possible with various styling of the boxes and what's added to each. 

I guess it would take someone trying to achieve something specific to test that out though.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -2372,6 +2372,9 @@ pub fn reorderTree() void {
     exampleFileTree(
         @src(),
         .{
+            .enable_reordering = false,
+        },
+        .{
             .background = true,
             .border = dvui.Rect.all(1),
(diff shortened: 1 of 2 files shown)


david-vanderson:
I like the added option, but we should keep the demo enabled.  Merge when ready.  Thank you!

foxnne:
Done! Thanks, merging after CI.

david-vanderson:
Perfect thanks!

PR #468, opened 2025-07-11
Examples: add an example of how to disable widgets to Basic Widgets

phatchman:
Add an example of how to enable/disable widgets to the demo.
- Add an "enabled" checkbox for the "Gray" button.
- Only respond to events if the button is enabled.

This is one of the things I was wondering how to do when I started out.


david-vanderson:
Good idea - merge!

PR #469, opened 2025-07-11
Examples cleanup

phatchman:
I noticed a strange line in the Calculator example that didn't match modern zig.

While fixing that I made it behave a bit more like a calculator:
1) Starting a new operation before pressing = now works.
2) Round the display output to 6 decimal places, to avoid FP-precision issues. 
3) Clear current calculation if digit is pressed and there is no current operation. e.g. typing a number directly after =, N or %.
4) Replaced C-like sizeof(array)/sizeof(array[0]) with array.len

Still does not correctly display leading zeros until a non-zero digit is pressed e.g. 0.00004 does not display until the 4 is entered. But I can live with that.

Leaving this open while I look through the rest of the examples. (To be clear, I'm looking for obviously un-zig or un-dvui  code. Not trying to re-write everyone's example code).

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -826,7 +826,7 @@ const dialogs = @import("Examples/dialogs.zig").dialogs;
 const animations = @import("Examples/animations.zig").animations;
 const structUI = @import("Examples/struct_ui.zig").structUI;
 const debuggingErrors = @import("Examples/debugging.zig").debuggingErrors;
-const icon_browser = @import("Examples/icon_browser.zig").icon_browser;
+const icon_browser = @import("Examples/icon_browser.zig").iconBrowser;
 
 const grid_examples = @import("Examples/grid.zig");
 const gridStyling = grid_examples.gridStyling;
--- a/src/Examples/animations.zig
+++ b/src/Examples/animations.zig
@@ -182,16 +182,18 @@ pub fn animations() void {
         const millis = @divFloor(dvui.frameTimeNS(), 1_000_000);
         const left = @as(i32, @intCast(@rem(millis, 1000)));
 
-        var mslabel = dvui.LabelWidget.init(@src(), "{d:0>3} ms into second", .{@as(u32, @intCast(left))}, .{}, .{});
-        mslabel.install();
-        mslabel.draw();
+        {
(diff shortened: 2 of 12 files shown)


phatchman:
I figured out why the theme editor is broken. After much @compileLog()-ing.
Theme has Options, which now have a pointer to WidgetData.
WidgetData has a parent of *Widget.
Widget has a VTable that the StructUI isn't going to cope with.

Not sure what the fix is yet, but that is the issue.

Edit: Skipping the opaque and function fields doesn't help. Because Options now has a circular reference to itself. And StructUI can't cope with that. I'm going to leave it commented for now as I don't have a good solution in mind.

david-vanderson:
Looking great!  Merge when ready.

Thanks so much for tracking down the structUI issue!  I also don't know what the answer is, but that is certainly progress.

phatchman:
> Thanks so much for tracking down the structUI issue! I also don't know what the answer is, but that is certainly progress.

I think it would be reasonable to either:
1) Have some option to not follow pointers.
2) Only follow pointers 1 level deep.

How easy either of those is to do I'm not sure. Perhaps it should be comptime configurable what the pointer following behavior is.
Edit: See #472 for first attempt at this.

phatchman:
I tried all the ways I could think of to use defer deinit() with text_entry, but I think all of them made the code worse. That's because text entries are populating constant variables which are used later in the code.

1) Moving constant declarations to the top of a block and setting them to undefined.
    - They go from being const to var;
    - You have to search in the code for where it is set vs declared.
  
2) Using labeled blocks
    - Fixes the const issue.
    - You have to assign everything inside the block to temporary variables because you need access to widget id's etc before they are deinited. Then break :blk , { var1, var2, var 3 }
    - You still have to search to find where the variable is set inside the block and it will have a different name inside the block to outside it.
    
3) Last trick I used other places was to put the defer in a scope after all of the variables that need the scope are created. But that doesn't really work in this case either.
    
 I could almost live with 2 except for these temporary variables. See name_error for example. It would be used both inside any new hbox block/scope and also outside it.

I think in this case, simpler is better. And we have better tools for picking up use after deinit() than we did., so no need to force it if it makes the code less readable overall.

david-vanderson:
All good, merge when ready.  I very much appreciate the effort and restraint - it's hard to know how complicated demo code should be.  Thanks for trying a bunch of different approaches.

Thank you!

PR #472, opened 2025-07-12
Make struct UI optionally follow pointers

phatchman:
## Description
The theme editor in the demo is currently broken, because StructUI blindly follows all pointers.

This PR allows the user to control whether pointers should be followed or not.

# Discussion

The current PR is just a POC. Next steps would be to:
1) Change the logic to only disallow pointers to containers, rather than special-casing slices.
2) Try and move the follow pointers variable into "FieldOptions" and any specific field options that require it, rather than polluting all of the call signatures.
3) Investigate allowing following pointers "n" levels deep. It's not clear how to achieve this at comptime without making things ever more difficult to debug than they are now.


RedPhoenixQ:
> Investigate allowing following pointers "n" levels deep. It's not clear how to achieve this at comptime without making things ever more difficult to debug than they are now

Would it be enough to add another argument to the ptr functions that have a depth which gets decremented on each recursive call, returning at 0?

phatchman:
> Would it be enough to add another argument to the ptr functions that have a depth which gets decremented on each recursive call, returning at 0?

Hmmm. I think that would simplify my code considerably! I'll give it a try. Thanks!

phatchman:
This is really tricky to get to work because of the way the structs are re-created with new types. So, at the "fields" level it is expecting a type that has depth of 0, so you can't return something that has a depth of 1.
```
src\structEntry.zig:893:60: error: expected type 'structEntry.ArrayFieldOptions([4]u8,.{ .depth = 0, .max_depth = 2 })', found 'structEntry.ArrayFieldOptions([4]u8,.{ .depth = 1, .max_depth = 2 })
```

I know the above is not very clear, or helpful. The bottom line is that depth can't be part of the type, so it can't be an argument to the function that creates the different field options because each depth creates a different type. The values you pass into "fields" for e.g. structEntryEx() don't know what depth that field is in, so they get a depth of 0. At least that is the best I can work out right now. It is also very possible I'm missing something obvious.

The other option is the "disabled" syntax". This seem like it only works for the top level object. So, the other option I thought of was to extend this so that "data_out" can be disabled. That should get us back to the original functionality without infecting everything with a "Depth" parameter. And I think that gives more flexibility as you can chose which fields to include or ignore as per the original intention.

RedPhoenixQ:
I don't think it should go on the FieldOptions struct, just a an additional argument to `fieldWidget` and all the specialiced functions, decreasing by 1 on every recursive call to `fieldWidget`. I'll try and make my envisioned change on top of yours and see if my idea is sound at all.

phatchman:
Ok. Been thinking about different ways to do this and this is my proposal:

## Current behavior

1. Assumes all fields of atruct and all children  are included unless marked 'disabled'
2. Disabling happens at run-time. Comptime recursively follows the struct regardless of what is disabled.
3. Lots of extra code can be generated, that will be linked, but never called at runtime.

## Proposed behavior

1. Provide a comptime exclude field tuple for any struct or union.
2. Don't include or follow fields whose name is in the comptime exclude tuple, regardless of depth.
3. Fixes the exe bloat issue above.
4.Doesn't require me to manage comptime state like the depth/recurive options, so should be simpler.

So for the theme editor, the fields marked as "disabled" in the structWidgetEx call, will move to being comptime exclude params, and that *should* fix the issue. But not today..

phatchman:
> I don't think it should go on the FieldOptions struct, just a an additional argument to `fieldWidget` and all the specialiced functions, decreasing by 1 on every recursive call to `fieldWidget`. I'll try and make my envisioned change on top of yours and see if my idea is sound at all.

That FieldOptions stuff happens recurively at comptime, regardless of what happens you do in the struct widget function. 

So it ends up with Options having a circular reference back to itself, if I just handle the depth in the struct widget function.

RedPhoenixQ:
I have more insight. The struct fields also fail because of `WidgetData`, specifically because of the parent field to `Widget` which holds function pointers.

I have to fixes here:
- The struct fields should fail to compile on types it cannot handle?
- The `Theme` struct shouldn't hold `Options` directly.
  -  There should be an intermediate type just for the color options, which is what the `style_accent` fields are intended for

phatchman:
>  I'll try and make my envisioned change on top of yours and see if my idea is sound at all.

Yeah please. It has been doing my head in.

RedPhoenixQ:
> Yeah please. It has been doing my head in.

No you are right, it won't work like I hoped because of the type dependency

```bash
src\structEntry.zig:710:12: error: struct 'structEntry.StructFieldOptions(Options)' depends on itself
```

phatchman:
After doing this, it makes me glad I got rid of that comptime magic from the grid.

I think exclude is the way to go. To stop it building that big tree it of FieldOptions, even if you are only displaying a handful of fields.

phatchman:
OK pushed the latest verison with exclusions, which seems to do the trick.
So doing this, I now understand why I was getting that error with the depth-limiting approach. It is true what I was saying, but there is a second issue:

```
const color_field_options = dvui.StructFieldOptions(dvui.Color, &.{ "style_err", "style_accent" }){ .fields = .{
dvui.structEntryEx(@src(), "dvui.Theme", &.{ "style_err", "style_accent" }, dvui.Theme, dvui.themeGet(), .{
```
These two things must have the same field type, which means they need the same exclusions, which also meant previously they needed the same depth.

I want to clean this up by:
1) Turning the field names into a type / tuple / bare enums or similar so it is more similar to the fields variable.
2) Should be able to turn the isExclude into a hasField or similar? Maybe.

phatchman:
OK. I think this should be good to go now.

I feel like I'm missing a better way to do isExcluded()
The set of exclusions should be small and iterating through the excluded values for each field happens at comptime, so it's not soo bad. It's probably not worth doing anything too complex here.

david-vanderson:
I'm also a bit lost on what to do here, but this is better.  Merge when ready.  Thank you!

phatchman:
It feels like StructUI needs a different approach in general? 

Maybe it is trying to do too much? There are lots of TODO's and commented out part of that code. Seems like an 80% solution at the moment. e.g. a simple one is that displaying an array of u8 doesn't work and is marked as a TODO. Slice of u8 seems to work. Some of the memory management looks like TODO's or needs to be reviewed at least.

I get that it is nice for debugging to just throw a struct at the API and you can see / change all the values. But maybe the user should declare the fields they want displayed and how they want them displayed, rather than trying to work it all out at comptime. i.e. replace the comptime FieldOptions with something the user can provide as an argument.

The biggest downside with the current approach is the current compiler doesn't give a good enough trace when there are errors. So it shows the error in the middle of the StructUI somewhere, because it is recursing through all the fields.

But I don't actually have a good handle on the use-cases. Hopefully, someone who actually needs it can take a fresh look.

phatchman:
I think a big lesson to me from the data adapter thing was that you can write lots of code that takes control away from the user and do lots on their behalf, but then you have to cater for every possibility. Or you can find a way for the user to keep control, but make it super easy for them to do what they need.

StructUI feels a lot like data adapters on steroids.

What if struct ui was some functions that said "display this field", "with these options". Underneath it could still do different things depending on the type, e.g. display a reorderable. But this would let the user put in separators, section headings etc, because we keep the high-level layout in their control.

So basically keep the fieldWidget stuff, but it just displays a single field. So e.g. you could display a whole array and get the reorderable. Or you could loop through and show each individual field and custom control the layout. Depending on what user chooses to do.

I'll see if I can develop this idea a bit more. I can sort of see a direction.

Edit: i mean what really is the difference between that field array that is passed to structWidget vs the same number of lines of code calling a display function? The latter is much simpler to implement and debug.

david-vanderson:
I totally agree.  The whole idea was for someone who wanted a quick way to "just throw this thing onto the screen to be edited", but pretty quickly we ran into memory ownership issues.

If there's no way to process arbitrary structs, then maybe you are right that some helper functions for common fields is the way to do.

Thanks very much for pushing on this.  I'm trying to keep up but haven't had the time yet to really understand what is going on so I appreciate it!

phatchman:
Ahhh. I see all  that deep copy stuff, that was presumably meant for adding new instances of a struct?

Yeah I can see how that gets real complicated.

I'm hoping we can just have something like `displayField(instance, .field_name, . { .gravity_y = 0})`
And a similar editField. But there would need to be a few different types of edit field. Eg slice:[] const u8 might need to be handled the same or different to slice: []u8. I.e. first always needs to be set to a new allocated slice. While the second might also need allocation or instead want an edit in place. 

It's actually not too different to what is already there (which is good, cause there us a lot of good work there), just exposed to the user differently and cutting out the recursively building of all these new structs to represent the user's original struct.

And who knows we might think up some way to auto add these calls for simple structs, so you can still just throw up a Colour struct easily.

phatchman:
Minimal POC looks promising. Can display structs down to specified depth, or you can use the individual "widget" functions to do a custom layout.

Very little code change required.
```
{
    {
        var scroll = dvui.scrollArea(@src(), .{}, .{ .expand = .both });
        defer scroll.deinit();
        var al = dvui.Alignment.init();
        defer al.deinit();
        dvui.se.intFieldWidget2(@src(), &testStruct, "int1", .{}, &al);
        dvui.se.intFieldWidget2(@src(), &testStruct, "uint2", .{ .dvui_opts = .{ .gravity_x = 0.5 } }, &al);
        wholeStruct(@src(), &testStruct, 1);
        wholeStruct(@src(), &testStruct, 0);
    }
}
```
```
// TODO: Currently only supports ints and floats.
pub fn wholeStruct(src: std.builtin.SourceLocation, container: anytype, depth: usize) void {
    var al = dvui.Alignment.init();
    defer al.deinit();
    inline for (std.meta.fields(@TypeOf(container.*)), 0..) |field, i| {
        var box = dvui.box(src, .vertical, .{ .id_extra = i });
        defer box.deinit();
        switch (@typeInfo(field.type)) {
            inline .int => dvui.se.intFieldWidget2(@src(), container, field.name, .{}, &al),
            inline .float => dvui.se.floatFieldWidget2(@src(), container, field.name, .{}, &al),
            inline .@"struct" => if (depth > 0) wholeStruct(@src(), &@field(container, field.name), depth - 1),
            else => {},
        }
    }
}
```

Whatever wholeStruct becomes will have the high-level layout logic in it for how to indent structs and needs the extra types added, but outside that, I don't think a lot really needs to change, except for obliterating that recursive NamespaceFieldOptions(T) and replacing it with what wholeStruct turns into. (while wholeStruct is still recursive and comptime, it's not creating and returning types, which simplifies things quite a bit).

And I'd like to add support for display-only widgets. I can imagine that being helpful for debugging especially if you are trying to display a lot of entities.

phatchman:
So, I'm up to the allocation bits, which are fairly annoying to put it mildly.

I'm of the view that if you need to do allocation (like setting an optional pointer e.g. ?*ScrollOptions) that falls under "advanced" usage and the user has to manually call an optionalPtrWidget() fn and pass in a pre-allocated version of the pointer's child that is used if the optional goes from null to set. The user checks the return val of optionalPtrWidget() to know if that passed in value was used or not.

The API will return to the user whether that pre-allocated instance was used or if it can be freed. In the case of optional pointers, it just needs to be a single pre-allocated instance that is either used or not. In other cases, user can get away with swapping between 2 pre-allocated instances, a "currently used" and a "use this next". These could even be global vars, no allocation needed.

This keeps allocations in the user's control at the expense of them not being able to auto-generate the struct. But I think we can solve that by saying auto-generate everything except these 2 optional pointer fields, and then the user manually creates the final two fields, along with whatever addition allocation logic is required. And if they don't customize the fields, the default struct function will display the optional as read-only.

So, in summary, my position is that if allocation might need to happen, the user needs to pre-allocate it, and pass the pre-allocated value in, just in case.  

The downside is there is an extra allocation that may/not be used and it pushes some complexity back to the user to manage. But if you are displaying a struct full of optional pointers to things, I think the user just has to do some extra work. If someone is pushing huge structs through this where the allocations matter, they can submit PR's to make it work for that case ;)

None of the above is needed for non-pointer optionals, which get a separate displayWidget fn that doesn't need pre-allocation.

Hope you can follow the above and it makes some sort of sense. Pointers will work the same way. One version for just always use the same thing as last time and another for something pre-allocated if it changes.

Oh! And we have the start of a new feature. Basically, you can have a read-only struct, but register an init() function to set the struct values with. I'm not going to include that yet, was just playing around with what was possible to do with the modified architecture.

david-vanderson:
That sounds great!  What happens if the optional pointer starts filled in but then structUI changes it to null?  Is there a way to know how to deallocate that?  Or can we pass that back to the user somehow?

I remember we couldn't figure out how strings should work.  Like if you wanted to change the name in a Font, we didn't know whether to deallocate it (because it might have been a static string).

I've been contemplating changing font name from []const u8 to [50]u8 or something, so this question of who owns the memory of those strings is resolved.  But that doesn't solve the general structUI problem.

phatchman:
> That sounds great! What happens if the optional pointer starts filled in but then structUI changes it to null? Is there a way to know how to deallocate that? Or can we pass that back to the user somehow?

The plan is for the user to free as well. Worst case is the user can check the struct themselves, but it would be nicer to let them know via the api if I can.

> I remember we couldn't figure out how strings should work. Like if you wanted to change the name in a Font, we didn't know whether to deallocate it (because it might have been a static string).

That's why I'm pretty sure  it has to be the user who does all the memory management. We often have no idea when to free, and you can get into cases where you only want to free sometimes. 

Imagining trying to display a linked list, adding / removing / moving elements. And the elements have a mix of static and dynamically allocated strings. Pointers back to self and all sorts of things that I don't really want to deal with directly.

phatchman:
The code might not be that elegant, but we now have a way to edit a value in a static string []const u8.

```
        var buf = gpa.alloc(u8, 50) catch return;
        buf = dvui.se.textFieldWidgetBuf(@src(), "slice5", &testStruct.slice5, .{}, buf, &al);
        if (!first_change) {
            gpa.free(buf);
        } else {
            first_change = false;
        }
```

The first_change nonsense is because testStruct is initialized with `slice5: []const u8 = "ABCDEF",`. I'm hoping to think of a cleaner way, but at least it is now possible.

But the main concept is that textFieldWidgetBuf returns a pointer to the slice that is no longer being used. The user can then work out if that is something they allocated or not. If buf is *always* allocated, then there is a nice pattern for this:
```
        var buf = gpa.alloc(u8, 50) catch return;
        gpa.free(dvui.se.textFieldWidgetBuf(@src(), "slice5", &testStruct.slice5, .{}, buf, &al));
```

We still have the normal textFieldWidget, and it will do in-place editing if it can. Otherwise it will display as read-only.

For optionals, the optionalFieldWidgetPtr returns an optional pointer to the field. So, if that optional value is returned, you know that it can be freed (i.e. the optional has been set to null.) or probably you will just re-initialize it to some default values, so it can be used next time.  This way we give a little bit of help to the user, in that they don't have to separately keep track of the allocated pointer.

I realize there are way too many optionals in the above paragraph, so maybe this is clearer:
```
var testStruct: ?*TestStruct = gpa.create(TestStruct);
testStruct.* = .{};
if (dvui.se.optionalFieldPtr(@src(), "test", &testStruct, .{}, &al) | result_ptr | {
    gpa.destroy(result_ptr);
}
```
Or alternatively keep testStruct around and re-use it for next time the optional is required again

Not quite happy with the optional stuff yet as you don't get control over what the optional is displaying. Right now, I can't customize how "testStruct" is displayed if I want ot get the "optional" display functionality.

Edit: Oh actually I can, and it is obvious in hindsight. An optional should behave like an "expander". Instead of the optional having to know how to display the struct, it just displays if the optional is set or not. The user then has to make a separate call to display the optional struct / enum / whatever. That also gets rid of *all* the memory management from an optional. The user can allocate or not based on whether the optional is set.

david-vanderson:
This is looking really good.  It also seems like a higher-level wrapper api could be created that handled it all for you in the case where all the allocations/deallocations are on the same Allocator (and all the strings were known to have already been allocated on that same Allocator), at least up to some kind of complexity point.

phatchman:
This is what that "optional" idea above looks like. I'm not sure if returning the vbox will stay, but it does let us keep the two parts together.

```
if (dvui.se.optionalFieldWidget2(@src(), "slice_opt10", &testStruct.slice_opt10, .{}, &al)) |optional_box| {
    defer optional_box.deinit();
    if (testStruct.slice_opt10 == null)
        testStruct.slice_opt10 = testStruct.slice7;
    processWidget(@src(), "", &testStruct.slice_opt10.?, &al); // This displays a text entry box for the []u8 slice.
} else {
    testStruct.slice_opt10 = null; // optional widget doesn't set to null by itself in case it needs to be freed by user.
}
```

This gives the user full control over how to initialize the optional, whether that means allocating or not. In this case I want the u8 slice to point back to one of the internal u8 arrays when the optional is selected. 

The default wrapper function will handle all non-pointer optionals automatically as long as there is default initialization available for the optional type. Pointer optionals will be read-only.

The upside of the new architecture is that -freference-trace now shows you the line causing the error. That still might not be that helpful in a lot of cases, but it is at least an improvement.

I'm going to experiment with doing some others in this style. That way we give the user full access to all the building blocks, but still provide progressively higher-level helper functions built on top of the primitive ones.

Edit: Yeah the default wrapper for this works as well:
```
inline .optional => |opt| {
    if (dvui.se.optionalFieldWidget2(@src(), field.name, &@field(container, field.name), .{}, &al)) |hbox| {
        defer hbox.deinit();
        if (@field(container, field.name) == null) {
            @field(container, field.name) = defaultValue(opt.child); // If there is no default value, it will remain null.
        }
        if (@field(container, field.name) != null) {
            processWidget(@src(), "", &@field(container, field.name).?, &al);
        } else { // TODO: Doesn't have a default value. Make it read-only }
    } else {
        @field(container, field.name) = null;
         dvui.label(@src(), "{s} is null", .{field.name}, .{ .id_extra = i }); // TODO: Make this nicer formatting.
    }
}
```

phatchman:
I'm a little torn how "developer focused" to make the struct UI. 

For example:
1) Do we want to distinguish in the UI whether something is an embedded Rect, or whether it is a pointer to a Rect?
2) Do we want to show a null optional as "null" or just show an un-checked checkbox?

My instinct is that this is catering mostly to developers, so should closely map to the struct, using the struct's field names and normal zig terms. But the existing struct ui seems to be focused more on an end user experience to some extent. e.g.  lets you rename fields to friendlier display names, doesn't distinguish pointers etc. 

I'm thinking my model use-case would be a game dev, putting up widgets for the entitles in their system, they might have an id, name, position, velocity vector, hit_box rect etc and pointers to some common game state. 
- There are some fields like ID, that are system assigned and must be read-only.
- Some pointers should be followed read-only, e.g. to common state.
- Can allow creation / deletion of entities
- Creation of new entities requires some sort of initialization. e.g. get an id and register with the entity list.
- Allows modification of Entity attributes
- Allows modification of display options (optionals)

Any thoughts? Are we mostly aiming for devs, or a half-way where we'd expect this sometimes as an end-user interaction tool?

Also, I'm dropping the re-orderable part of arrays. It's only implemented for `mutate_value_and_realloc` anyway and while I could make this work for plain old arrays of ints, it won't work for anything that can't be copied easily. e.g. has internal pointers, indexes in other array that need updating etc. I might come back to this in a phase 2 and offer some custom version of the slice/array widget that does allow re-ordering, but I suspect that use-case is pretty niche. But it's clear to me that any moving has to be done by the user, just like any creation or memory management.

Edit: Actually, the above use-case could be a fun demo for this. Do some simple animated boxes that can collide or something and the user can add / remove boxes, change their velocity etc. As long as it doesn't end up being too complex for a example. But I don't think it will.

david-vanderson:
I imagine this as completely developer focused.

> I'm thinking my model use-case would be a game dev, putting up widgets for the entitles in their system, they might have an id, name, position, velocity vector, hit_box rect etc and pointers to some common game state.

This is my mental model as well - a developer is going fast and wants a quick and dirty way to tweak some values in a random struct.

Given that the design requirements would be something like:
* see how useful it can be for arbitrary structs
* okay if certain things can't be done
* probably better to limit customizability in favor of just saying "write your own editor for that"
  * this is where your helper functions could be very useful, as a half-way place for someone who wants some customization but is still in quick-and-dirty development

Dropping the reordering arrays sounds good to me.

Good idea about the demo - some array of structs that are all data, no pointers, maybe an editable string name (but no reallocation, just editing in place)

PR #475, opened 2025-07-13
Don't use `Options` for `Theme` styles

RedPhoenixQ:
This is as partial fix for #472

Most of the issue in #472 originated from cyclical pointers in `Options` introduced by the `data_out` field. There is no good reason for `Theme` to hold `Options` directly in the first place, except maybe duplicated field definitions. This adds a new type for only color styles that the `Theme` can use, which allows the `themeEditor` to work as before. This is not a true fix for the issues in #472, but it's an improvement to the types of `Theme`.


david-vanderson:
Good!  Merge when ready.  Thanks!

PR #478, opened 2025-07-14
Options editor

RedPhoenixQ:
Discussed in https://github.com/david-vanderson/dvui/pull/448#issuecomment-3053269240

This adds an `Options` editor to the debug window. To make this reasonable the debug window and it's related fields have been moved to a separate file.

### DIFF
--- a/src/Debug.zig
+++ b/src/Debug.zig
@@ -259,26 +259,59 @@ pub fn show(self: *Debug) void {
 
         var it = self.options_override.iterator();
         var i: usize = 0;
+        var remove_override_id: ?dvui.WidgetId = null;
         while (it.next()) |entry| : (i += 1) {
             const id = entry.key_ptr.*;
             const options, const src = entry.value_ptr.*;
 
-            var button = dvui.ButtonWidget.init(@src(), .{}, .{ .id_extra = i, .expand = .horizontal });
-            button.install();
-            defer button.deinit();
-            button.processEvents();
-            button.drawBackground();
+            const row = dvui.box(@src(), .horizontal, .{ .expand = .horizontal });
+            defer row.deinit();
 
-            if (button.clicked()) copyOptionsToClipboard(src, id, options);
+            var reset_wd: dvui.WidgetData = undefined;
+            if (dvui.buttonIcon(@src(), "Reset Option Override", dvui.entypo.back, .{}, .{}, .{
+                .gravity_y = 0.5,
+                .data_out = &reset_wd,
+            })) {
+                remove_override_id = id;
+            }
+            dvui.tooltip(@src(), .{
+                .active_rect = reset_wd.borderRectScale().r,
+                .position = .vertical,
+            }, "Remove the override", .{}, .{});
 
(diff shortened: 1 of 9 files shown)


RedPhoenixQ:
By the way, the `data_out` field is really nice for adding tooltips to widgets created through the `dvui.X` helpers.

david-vanderson:
This is awesome!  I did some quick testing and took a brief look at the code.  Merge when ready.

One thing I noticed is that it's relatively easy to get confused when editing the options for a `button` and not realize that it doesn't flow through to the embedded `label`.  I'm not sure what kinds of things we can do about this, but will think about it.  Some quick knee-jerk thoughts:
* change widget code to flow options through better?
* modify the widget selection to somehow preference "outer-most" widgets for compound ones?
  * helps prevent adjusting options to some internal widget that is already constrained by the outer-most one

But in any case this is a great step in the right direction.  Well done!

RedPhoenixQ:
> * change widget code to flow options through better?

Would it make sense if `WidgetData` took the options from the parent and called `override` with only the styles perhaps? That way colors would be inherited.

david-vanderson:
> > * change widget code to flow options through better?
> 
> Would it make sense if `WidgetData` took the options from the parent and called `override` with only the styles perhaps? That way colors would be inherited.

Interesting idea.  `Options.strip()` is theoretically doing that, but I suspect I didn't consistently do it with the compound widgets.  But just looking at `button()` it looks like that should work.  Oh I see - the override happens at the WidgetData level, which doesn't affect the Options that are passed to the label.

Okay good insight - maybe a subset of Options should automatically inherit from the parent.  Will ponder.

phatchman:
Just checked this out. Really cool!

david-vanderson:
I think we lost a little-known feature of the debug window.  If you clicked "Debug Under Mouse (until click)" and then clicked on a widget, you used to see a stack of all the widgets that were under the mouse when it was clicked - they showed up under "Event Logging".

I'm still seeing the list while hovering, but when I click on a widget the list goes away.  @RedPhoenixQ Do you see that too?

RedPhoenixQ:
> I'm still seeing the list while hovering, but when I click on a widget the list goes away. @RedPhoenixQ Do you see that too?

No, it works for me. I can click and then scroll through and select from the scrollable list bellow "Event Logging". Will look into why it could fail.

RedPhoenixQ:
I don't see how it could fail. I have a `target` enum that is set to `.mouse_until_click`. When a click is seen it's set to `.quitting` and the "mouse over stack" only gets cleared if the target is `.mouse_...`.

david-vanderson:
Wow that was subtle.  @RedPhoenixQ Is is still working for you?

I only saw the problem on my linux machine, because we don't use callbacks on linux:
* Window.begin() clears the debug stack
* then we add all the events, which sets debug_target to .quitting
* then we run the frame, but nothing got added because .quitting wasn't considered to be `mouse()`

RedPhoenixQ:
Still working!

Oh, because the order of `Window.begin` and adding of events are different the click event set quitting before the reset. That is subtle. Good catch!

PR #479, opened 2025-07-14
Add API to change cursor visibility

### LINKED ISSUE #476: Request: Ability to hide cursor
Currently you can change the cursor to other system cursors but you can't hide it, for the use case of either a game or to draw a custom cursor through DVUI's `image` or similar.

RedPhoenixQ:
Closes #476

This adds a new API to `Backend` to change the visibility of the cursor.

All backends have been tested on my Windows machine, shouldn't be any issues on other platforms.


david-vanderson:
Excellent!  Merge when ready.  Thank you!

Issue #232, opened 2025-03-29
consistent focus highlight color


david-vanderson:
On discord curtisdas pointed out that there was no way to set a radio circle's active color.  I pushed a commit to make it use the Options accent color.

When testing (like `.color_accent = <yellow>`) then the radio circle turns yellow, but also the focus highlight turns yellow.

This is making me think that the focus highlight color should probably not be sourced from the Options sent to each widget, but instead sourced from the current Theme directly.  That way it stays consistent even if an individual widget is using a different accent color.

Also this frees up widgets to use the Options accent color as a general purpose auxiliary color.

RedPhoenixQ:
That would make it impossible to change the focus color, or disable the focus highlight or disable it (by making the highlight transparent) if that is needed. Another option would be to add `.color_focus` to `Options`, but not the theme. This way `.color_focus` can use the theme `.color_accent` by default, but it can be overridden on a per widget basis.

david-vanderson:
There should still be two options even with this:
* change the Theme color before/after the widget
* use the lower-level widget api and customize the focus drawing (I realize we likely have widgets that need work in this area)

I'm also not sure how easy to make overriding the focus color on a per-widget basis, so I'm leaning towards making it a bit harder.

Do you think wanting a different focus color for some widgets is common enough for us to make it easy?  Or do you think these methods are too difficult?

RedPhoenixQ:
> * change the Theme color before/after the widget

This would solve any use case I though of, and it's inconvenient enough to not be used in general or by accident.

> I'm also not sure how easy to make overriding the focus color on a per-widget basis, so I'm leaning towards making it a bit harder.

This sounds very reasonable to me. I'm now on board with the original idea. Will make a PR in the coming days

david-vanderson:
Awesome thanks!

PR #481, opened 2025-07-15
text layout: allow for setting a background color per line of text

RedPhoenixQ:
Addresses #480

### DIFF
--- a/src/Examples/text_layout.zig
+++ b/src/Examples/text_layout.zig
@@ -73,8 +73,8 @@ pub fn layoutText() void {
         tl.addText("Heading\n", .{ .font_style = .heading });
 
         tl.addText("Here ", .{ .font_style = .title, .color_text = .{ .color = .{ .r = 100, .b = 100 } } });
-        tl.addText("is some ", .{ .font_style = .title_2, .color_text = .{ .color = .{ .b = 100, .g = 100 } } });
-        tl.addText("ugly text ", .{ .font_style = .title_1, .color_text = .{ .color = .{ .r = 100, .g = 100 } } });
+        tl.addText("is some ", .{ .font_style = .title_2, .color_text = .{ .color = .{ .b = 100, .g = 100 } }, .color_fill = .green });
+        tl.addText("ugly text ", .{ .font_style = .title_1, .color_text = .{ .color = .{ .r = 100, .g = 100 } }, .color_fill = .teal });
         tl.addText("that shows styling.", .{ .font_style = .caption, .color_text = .{ .color = .{ .r = 100, .g = 50, .b = 50 } } });
     }
 }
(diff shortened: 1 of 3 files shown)


david-vanderson:
This is good, but shows that we probably need to separate the color used for background from the color used for selected text.  I can think of two options:
* use the background color of TextLayout for selected text
  * less granular control, the idea is selected text all looks the same, overrides the colors from `addText`
* change selected text to do some alpha blended highlighting
  * more control, can see styling through selection, but selection is less obvious
  * only uses a single color, the accent from the theme I think?

I'm leaning towards trying the second.  Any opinions?

RedPhoenixQ:
Good point. Didn't realise the selection background also used the `.fill` color.  

> * use the background color of TextLayout for selected text
>   
>   * less granular control, the idea is selected text all looks the same, overrides the colors from `addText`
> * change selected text to do some alpha blended highlighting
>   
>   * more control, can see styling through selection, but selection is less obvious
>   * only uses a single color, the accent from the theme I think?
> 
> I'm leaning towards trying the second. Any opinions?

Agreed, second option using the theme sounds good, with some blending. That would be very similar to #232 where the focus color now is taken from theme directly and we would be making the same tradeoffs here.

RedPhoenixQ:
The blending options doesn't work very well because the selection might extent to a space character which in not included in the background coloring. This makes the space get colored with the modified accent color and which is different from all other lines. You can also partially select an area with a background color, so modifying the background color instead also results in inconsistencies.

I have a third simple option for using the accent color that I will push here. Maybe always using the accent color is enough.

RedPhoenixQ:
I we want the blending behaviour we need to add logic to separate the background into the selected part and the non selected part. Feels unnecessarily complicated

david-vanderson:
> I we want the blending behaviour we need to add logic to separate the background into the selected part and the non selected part. Feels unnecessarily complicated

I don't understand this - I was imagining rendering all the text as unselected, and then compositing on top a partially transparent selection rectangle.  Maybe that is what you tried, and I don't get how it didn't work.

In any case, this is an improvement, so merge it, and I'll play around with the compositing later.

Thank you!

RedPhoenixQ:
> I was imagining rendering all the text as unselected, and then compositing on top a partially transparent selection rectangle. Maybe that is what you tried, and I don't get how it didn't work.

I does but it makes the normal selection much dimmer, to the point where some themes are bad contrast for the text.

david-vanderson:
> > I was imagining rendering all the text as unselected, and then compositing on top a partially transparent selection rectangle. Maybe that is what you tried, and I don't get how it didn't work.
> 
> I does but it makes the normal selection much dimmer, to the point where some themes are bad contrast for the text.

Makes sense, let's move forward with your solution.  Thanks!

PR #485, opened 2025-07-16
Window: Only prevent y-scrolling if shift is the only modifier pressed

foxnne:
Hello again!

The motivation behind this was allowing zooming when `ctrl/cmd` hotkey is pressed, but also allowing a faster zoom if `shift` is also pressed.

I noticed that this is currently impossible, as `mouse_y` scroll events are not sent at all if `shift` is pressed, even if it's not the only modifier.

Please let me know if this isn't the correct fix for this scenario!

Thanks,


RedPhoenixQ:
This seems very reasonable! Will look at why CI is failing

foxnne:
This is an interesting test failure for this commit. Was this already failing beforehand? I'd love to help out if I can.

RedPhoenixQ:
The CI was unrelated and should be fixed on main now

foxnne:
I just pulled main and I seem to still be getting an error running `zig build test -Dbackend=testing`

```
test
└─ run testing-app 2/3 passed, 1 failed
error: 'app.test.snapshot' failed: /Users/foxnne/dev/proj/dvui-dev/src/testing.zig:299:9: 0x102d33613 in snapshot__anon_142171 (testing-app)
        return SnapshotError.SnapshotsDidNotMatch;
        ^
/Users/foxnne/dev/proj/dvui-dev/examples/app.zig:183:5: 0x102d3499b in test.snapshot (testing-app)
    try t.snapshot(@src(), frame);
    ^
 ```

File: src/Window.zig [hunk 1]
```diff
@@ -750,7 +750,7 @@ pub fn addEventMouseWheel(self: *Self, ticks: f32, dir: dvui.enums.Direction) st
         .evt = .{
             .mouse = .{
                 .action = if (dir == .vertical)
-                    if (!self.modifiers.shift())
+                    if (!self.modifiers.shiftOnly() or self.modifiers == .none)
```
RedPhoenixQ, line 753: The if statement should probably be inverted. The extra `or` threw me of a bit.

```zig
if (self.modifiers.shiftOnly())
// wheel x here
else 
// default wheel y here
```

RedPhoenixQ:
> I just pulled main and I seem to still be getting an error running `zig build test -Dbackend=testing`

I'm not, how annoying. Just ignore the snapshot error for now, will look into it

src/Window.zig
foxnne: Ah, `shiftOnly` returns true if if the modifiers are `.none`, I've added just a simple check to handle that case. Let me know if you'd like to handle this another way.

RedPhoenixQ: Ah good find

RedPhoenixQ:
> I'm not, how annoying. Just ignore the snapshot error for now, will look into it

There was another commit that changed the snapshot in between my update and you merging. Should be good now

RedPhoenixQ:
Thank you for the changes! Everything looks good now so I think we are ready to merge

foxnne:
Thank you! If you want, I'll merge when CI finishes. Do you typically squash merge or just create a merge commit? Just wanted to make sure for future reference.

RedPhoenixQ:
We don't usually squash. You could if the changes are small enough to be in one commit, otherwise you could squash/merge commits with a rebase and force-push to cleanup the commits. The commit history in main is already pretty messy so this is fine without any cleanup

david-vanderson:
Excellent thank you both!

PR #489, opened 2025-07-17
refactor Event Handling, consistent API around Window.extra_frames_needed?

nat3Github:
this PR will be an attempt to untangle Event handling logic from Window. WIP
status:
- refactor: all fields that are touched by input event handling (the `addEventXYZ` functions) are now in a seperate struct called EventState
- Window has a field `event_state: EventState`
- changed event list to be a fixed Capacity List, the Capacity is configurable by the User. this has the benefit of bounded memory usage and bounded worst case performance at the cost of more memory allocated upfront.
- EventState uses its own arena for duped strings that is reset with `EventState.endFrame(...)`

Reasoning: 
- this makes it easier to understand what state in `Window` is actually coupled to input event handling
- it seperates Window logic from Event handling logic which makes Event Handling more flexibel

Discussion
- I came across a lot of widgets that read fields of the Window struct or even modify them in their logic
- I think it would be beneficial to provide an api for the things these widgets need, to uncouple them from the Event Handling / Window Implementation, this also would make it easier to understand what these widgets doing with respect to the Event/Window state

### DIFF
--- a/src/Event.zig
+++ b/src/Event.zig
@@ -27,7 +27,7 @@ evt: union(enum) {
 /// matched this event, using `dvui.matchEvent` or similar.
 /// This makes it possible to see which widget handled the event.
 pub fn handle(self: *Event, src: std.builtin.SourceLocation, wd: *const dvui.WidgetData) void {
-    if (dvui.currentWindow().debug.logEvents(null)) {
+    if (dvui.currentWindow().debug().logEvents(null)) {
         var action: []const u8 = "";
         switch (self.evt) {
             .mouse => action = @tagName(self.evt.mouse.action),
--- a/src/Examples/debugging.zig
+++ b/src/Examples/debugging.zig
@@ -12,7 +12,7 @@ pub fn debuggingErrors() void {
     dvui.label(@src(), "- text, icons, and images rounded to nearest pixel", .{}, .{ .margin = .{ .x = 10 } });
     dvui.label(@src(), "- text rendered at the closest smaller font (not stretched)", .{}, .{ .margin = .{ .x = 10 } });
 
-    _ = dvui.checkbox(@src(), &dvui.currentWindow().debug.touch_simulate_events, "Convert mouse events to touch", .{});
+    _ = dvui.checkbox(@src(), &dvui.currentWindow().debug().touch_simulate_events, "Convert mouse events to touch", .{});
     dvui.label(@src(), "- mouse drag will scroll", .{}, .{ .margin = .{ .x = 10 } });
     dvui.label(@src(), "- mouse click in text layout/entry shows touch draggables and menu", .{}, .{ .margin = .{ .x = 10 } });
 
--- a/src/Examples/scroll_canvas.zig
+++ b/src/Examples/scroll_canvas.zig
@@ -372,7 +372,7 @@ pub fn scrollCanvas() void {
 
         if (!done) {
             // still dragging, draw a half-opaque box to show we are dragging
-            const dr = Rect.Physical.fromPoint(dvui.currentWindow().mouse_pt.diff(.{ .x = 10, .y = 10 })).toSize(.all(20));
+            const dr = Rect.Physical.fromPoint(dvui.currentWindow().mouse_pt().diff(.{ .x = 10, .y = 10 })).toSize(.all(20));
             dr.fill(.{}, .{ .color = dvui.Color.lime.opacity(0.5) });
         }
     }
--- a/src/WidgetData.zig
+++ b/src/WidgetData.zig
@@ -27,7 +27,7 @@ rect_scale: ?RectScale = null,
 pub fn init(src: std.builtin.SourceLocation, init_options: InitOptions, opts: Options) WidgetData {
     const parent = dvui.parentGet();
     const id = parent.extendId(src, opts.idExtra());
-    const options = if (dvui.currentWindow().debug.options_override.get(id)) |val| val.@"0" else opts;
+    const options = if (dvui.currentWindow().debug().options_override.get(id)) |val| val.@"0" else opts;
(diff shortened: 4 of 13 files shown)


RedPhoenixQ:
I'm generally for splitting data into struct but I think this is a little too much. The event state now takes ownership of the subwindows, widget data (whose rect is not very separated from the `rect_pixels` which are really closely related) and natural_scale which are used in many other places not directly linked to event handling. 

I would prefer just a struct holding `events`, `event_num`, `mouse_pt`, `mouse_pt_prev`, `modifiers` and `capture` without moving any of the window functions out of the Window struct (perhaps moving the `addEvent...` functions but instead taking already scaled mouse points for example, so that `EventState` only ever deals in `Physical` values). 

> * I came across a lot of widgets that read fields of the Window struct or even modify them in their logic

I don't think the alternative is preferrable (not allowing/discouraging access to any fields). We would either have a function named the same thing as the current field that returns a mutable pointer, which is not better IMO. The other alternative would be "getter" and "setter" functions for everything anyone would want to modify on the Window. This is more code and more work, and when the API to modify something doesn't exist people would still reach down to the field directly and modify it to solve their use case.

> * I think it would be beneficial to provide an api for the things these widgets need, to uncouple them from the Event Handling / Window Implementation, this also would make it easier to understand what these widgets doing with respect to the Event/Window state

We could make common operations more convenient, like detecting if a child widget called `refresh` similar to the `lastFocusedWidgetInFrame` API. A function to put a widget in the "debug" state (with the red borders) should also be a function.

nat3Github:
> I'm generally for splitting data into struct but I think this is a little too much. The event state now takes ownership of the subwindows, widget data (whose rect is not very separated from the rect_pixels which are really closely related) and natural_scale which are used in many other places not directly linked to event handling.

i agree its a lot. just shows how many things in window are intertwined with event handling. 

>I would prefer just a struct holding events, event_num, mouse_pt, mouse_pt_prev, modifiers and capture without moving any of the window functions out of the Window struct (perhaps moving the addEvent... functions but instead taking already scaled mouse points for example, so that EventState only ever deals in Physical values).

good idea!

nat3Github:
some questions came up:
- what is the boolean return value for the addEventXYZ functions?
- why do we need a previous_window in Window
- how do subwindows work with regards to event handling

david-vanderson:
> some questions came up:
> 
>     * what is the boolean return value for the addEventXYZ functions?
It is for applications that are only using dvui for floating windows - the return value tells the application whether the event is being handled by dvui or not.
> 
>     * why do we need a previous_window in Window
To support nested Window begin/end (for multiple OS windows), but I don't know if that will ever be used.
> 
>     * how do subwindows work with regards to event handling
`Window.windowFor()` uses retained subwindow information to tag mouse events with the correct subwindow.  The tag is checked in `dvui.eventMatch()`

nat3Github:
I tried to refactor EventState with events, event_num, mouse_pt, mouse_pt_prev, modifiers and capture and use the global dvui.current_window when available instead but unfortunately it broke something / I did not get it working.

my current understanding of the event handling is:
- input events from sdl get translated to dvui.Events, these are just events
- as they are added based on the window state - target_windowId, target_widgetId, num, floating_win get assigned
- the above is currently one step, but it should be able to make it two steps?
- widgets as they are called can modify the events for the frame to affect handling or other widgets?

is it generally possible to seperate events from the window state, so to say events are just POD the state is queried from the window?

david-vanderson:
A lot of this is open questions that we don't have answers to:

> I tried to refactor EventState with events, event_num, mouse_pt, mouse_pt_prev, modifiers and capture and use the global dvui.current_window when available instead but unfortunately it broke something / I did not get it working.
> 
> my current understanding of the event handling is:
> 
>     * input events from sdl get translated to dvui.Events, these are just events
> 
>     * as they are added based on the window state - target_windowId, target_widgetId, num, floating_win get assigned
> 
>     * the above is currently one step, but it should be able to make it two steps?
Maybe?  Those values:
* num is a way for for `focusWidget` and `captureMouse` to adjust events "after" the event that caused the focus/capture
* target_windowId/widgetId are for focus/capture, these are also what get adjusted in `focusWidget`/`captureMouse`
* floating_win is set when the event is added based on retained subwindow information

I don't know what two steps you are thinking of.
> 
>     * widgets as they are called can modify the events for the frame to affect handling or other widgets?
Yes in the limited way via `focusWidget`/`captureMouse` and marking events as "handled".
> 
> 
> is it generally possible to seperate events from the window state, so to say events are just POD the state is queried from the window?

This is an open question, I don't know. Maybe there is a way to drop the target_windowId/widgetId and put "am I the focused widget" logic into `eventMatch`?  That is a good idea but I haven't thought about it enough to know whether it would work.

david-vanderson:
> This is an open question, I don't know. Maybe there is a way to drop the target_windowId/widgetId and put "am I the focused widget" logic into `eventMatch`? That is a good idea but I haven't thought about it enough to know whether it would work.

Thought about it a bit.  Here are two scenarios that could break it, if the focus/capture tags where removed from Event.

Mouse Scenario, the event list looks like (button A is processed first):
- mouse down over button B
- mouse up on button B
- mouse move
- mouse down over button A

When button A runs, it captures the mouse.  When button B runs, it knows a different widget has focus, so doesn't process any events, so we miss the button B click.

The keyboard scenario is similar, something like:
- key events for textEntry B
- mouse click on textEntry A

textEntry A runs first, the click focuses it.  When B runs, it would not have focus and drop those key events.

nat3Github:
closing this because its outdated

vlappa:
> closing this because its outdated

Could you clarify why this is outdated?

nat3Github:
>Could you clarify why this is outdated?

- out of sync with main
- im not convinced by these changes 
- goal unclear
- im currently looking in to other things.

Issue #490, opened 2025-07-19
add theming demo


david-vanderson:
squirl on discord asked for a demo that showed more theming stuff.

The demo already shows switching to different built-in themes, so this demo would be (I think) more of a "edit the current theme" so people can see what kinds of things can be changed.

phatchman:
- Add a custom theme to the list of selectable themes.
- Document how to start the application with the chosen theme.
  -  e.g. can you only set the theme between window begin and end or can it be statically set at compile time?
- Cloning an existing theme and changing a few items.

RedPhoenixQ:
I also have some other input on themes in DVUI. There are 2 things that I currently find a bit of about how themes are handled.

1. Why is there a hashmap of themes on the `Window`? It's unlikely that any app would change it's theme on the fly and in that case they could curate a potentially comptime-known list of themes themselves. The only reason for `Window` to have this map is `Theme.picker` which could easily take a list of themes instead, or not exist at all and just be made into an example for users to copy.
3. The `Window.InitOptions` takes a `Theme` by pointer and then dereferences it immediately. This just make the ownership and lifetime of the passed `Theme` struct confusing, especially since it's not `const` (a user might assume that a pointer to their struct will be returned from `dvui.themeGet()`). Just changing it to an optional value fixes this.

david-vanderson:
> I also have some other input on themes in DVUI. There are 2 things that I currently find a bit of about how themes are handled.
> 
>     1. Why is there a hashmap of themes on the `Window`? It's unlikely that any app would change it's theme on the fly and in that case they could curate a potentially comptime-known list of themes themselves. The only reason for `Window` to have this map is `Theme.picker` which could easily take a list of themes instead, or not exist at all and just be made into an example for users to copy.

Might not be the best.  I was thinking maybe we could have a theme picker that shows a few widgets in each theme.  Also there is a use-case for something like a game that loads a theme from a file on start, so I'm wary of requiring it to be comptime known.
> 
>     2. The `Window.InitOptions` takes a `Theme` by pointer and then dereferences it immediately. This just make the ownership and lifetime of the passed `Theme` struct confusing, especially since it's not `const` (a user might assume that a pointer to their struct will be returned from `dvui.themeGet()`). Just changing it to an optional value fixes this.

Yes this was a mistake.

I'm also thinking of changing Font so that it has an array of u8 (like [30:0]u8) instead of a slice, so that ownership is more obvious.

RedPhoenixQ:
[...]

PR #494, opened 2025-07-26
Theme rework

RedPhoenixQ:
This is related to #490 and does some improvements to Themes and Fonts which will make the future example and documentation simpler and clearer.

This includes 3 major changes:
1. `dvui.App` can now pass `Window.InitOptions` to the window, allowing for setting the default Theme, and other options, before the init function.
2. The theme hashmap on `Window` has been removed in favour of application specific handling of storing multiple themes. Most applications will not make use of the hashmap and themes that always were populated.
3. `Font` now has an id based on the hash of its name. The builtin font names have named id enum variants for improved debugging. The runtime name has been moved to `Window.font_bytes` to make use of its preexisting memory management for the font bytes. Additionally the name slice is now safe to include in `FontCacheEntry` thanks to its lifetime being the same as the font file itself, which makes debugging the font cache simpler.

Perhaps `dvui.themeSet` should take an owned `Theme` instead of the current `*const Theme` with a doc comment mentioning that it will be copied?


david-vanderson (approved):

PR #495, opened 2025-07-27
Theming editor example

### LINKED ISSUE #490: add theming demo

RedPhoenixQ:
Closes #490

Examples-theming

This adds an example that allows for editing the theme and optionally applying it to the entire window. I'm thinking that the doc improvements for how to set a theme will be in a separate PR.

Anything else that should be added to the preview ui page to the right? I think most of the options are shown but perhaps there is a better way to show all the theme fields effects more concisely.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -14,6 +14,7 @@ pub const demoKind = enum {
     calculator,
     text_entry,
     styling,
+    theming,
     layout,
     text_layout,
     plots,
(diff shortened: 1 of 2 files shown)


david-vanderson:

PR #497, opened 2025-07-28
Theme docs

RedPhoenixQ:
Final additions for #490

This adds a couple of changes related to themes:
- Makes the existing comments in `Theme` proper doc comments
- Add doc comment to `Window.theme` field that describes how to set it and how it's allowed to be changed
- Update the readme to explain how default themes are chosen and to not use the theme hashmap that was removed in #494
- Add the code to set the theme to the standalone and app examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,4 +302,6 @@ if (theme_dark) {
 ```
 The theme's color_accent is also used to show keyboard focus.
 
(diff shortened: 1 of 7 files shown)


david-vanderson:
Looks great.  Merge when ready.  Thank you!

PR #501, opened 2025-08-02
add azem project to readme

thuvasooriya:
hope i didn't mess up any formatting guidelines.
the app is very barebones but will help some people get an idea of what's possible ig.


david-vanderson:
Thanks!

PR #503, opened 2025-08-03
Id refactor

RedPhoenixQ:
`WidgetId` was used for more than just widgets and the code to translate from `u64` to `WidgetId` was very messy.

This renames `WidgetId` to just `Id` and moves `hashSrc` (renamed to `extendId` as that is the intended use) and `hashIdKey` (renamed to `update` to match the hash update function which it just calls directly) to it. This makes `Id` more generally useful and the removed the distinction of when you get and id and when you get just a `u64` hash.


david-vanderson:

Issue #506, opened 2025-08-04
Menu Operation Impacted by Recent Changes


br07h3rj:
I updated dvui a few days ago. After this, the menus stopped following the mouse, i.e. I can open a menu, but, to get the menu next to it to open, I have to click it, not hover over it. None of the floatingMenus open on hover for me now.

I'm guessing that this is what did it, though I'm not 100% sure:
https://github.com/david-vanderson/dvui/commit/395b7c6b6eefb4032cb4a8ac817679ff28ddd7f7

Is there a new way of making this happen? Should it already be working and I just need to adjust my implementation?

P.S. This is a great project, by the way. Thanks for all the hard work!

david-vanderson:
Thanks!  You are probably right.

The new intended operation is that if you click and hold the mouse down, then menus will open on hover.  So you should be able to "mouse down, hover to a menu item, and mouse up to activate".

If the menu is opened by a full click (down and up), then a full click is required to activate a menu item.

Is that the behavior you are seeing?

br07h3rj:
Yes, that is what I'm seeing now. It actually hadn't occurred to me that this behavior would be intended. The issue I wanted to address was the fact that the top level menuItemLabels act that way now too. Say we have a menubar with File, Edit, View and so forth. If I full click to open File, then Edit and View won't open without a click, so all menus and submenus require their own click unless we go Mac-style and hold down the mouse button the whole time. This is counter to normal PC use (Linux and Windows), isn't it? The menus seem to be working great except for this.

I know you're trying to juggle mouse operation, keyboard operation, touchscreen operation, and drag-drop features (and probably some more) all in one pile. No worries. I love the project and look forward to seeing where it is in a year or two. You're getting close to completing the best visual backend / widget system Zig is going to have. Because of the backend design, the potential is really great. This is the GUI software I want to work with going forward. I just need the menus to work like people expect. 😄 If there's a way I can solve this myself, please let me know.

david-vanderson:
> Say we have a menubar with File, Edit, View and so forth. If I full click to open File, then Edit and View won't open without a click

Alright I now understand what you mean, and you're right it should work that way.  I just didn't quite hit the mark during the earlier fix (which was mainly about not grabbing focus on a random menu/dropdown hover).

Thanks very much for letting me know!

So the menus need to track if any part of the menu has been opened, and if so, then do the "focus-on-hover" thing.  Does that sound right to you?

RedPhoenixQ:
[...]

PR #507, opened 2025-08-05
Make Menu behaviours match common UIs

RedPhoenixQ:
Closes: #506

I hope this fixes the behaviour that @br07h3rj observed in a satisfactory way. In summary you still need one down press to active a submenu, but once it's active only hovering is needed so swap between siblings.

This does not automatically open menus on hover, which was intentionally removed in 395b7c6, but allows for any menu that is already active to moved between available submenus on hover.

### DIFF
--- a/src/Examples/menus.zig
+++ b/src/Examples/menus.zig
@@ -61,6 +61,7 @@ pub fn menus() void {
                 _ = dvui.menuItemLabel(@src(), "Dummy Long", .{}, .{ .expand = .horizontal });
                 _ = dvui.menuItemLabel(@src(), "Dummy Super Long", .{}, .{ .expand = .horizontal });
             }
+            if (dvui.menuItemLabel(@src(), "Log", .{}, .{ .expand = .horizontal })) |_| {}
         }
 
         dvui.labelNoFmt(@src(), "Right click for a context menu", .{}, .{ .gravity_x = 1.0 });
--- a/src/WidgetData.zig
+++ b/src/WidgetData.zig
@@ -92,6 +92,10 @@ pub fn register(self: *WidgetData) void {
         }
     }
 
+    if (self.id == dvui.focusedWidgetIdInCurrentSubwindow()) {
+        cw.last_focused_id_in_subwindow = self.id;
+    }
+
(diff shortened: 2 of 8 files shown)


RedPhoenixQ:
@br07h3rj in response to [your comment with additional behaviours](https://github.com/david-vanderson/dvui/issues/506#issuecomment-3156744493)

1. Only root requires hover e909156.
2. Clicking on a root menu item will close its submenu if open 8abb513
3. Submenus closing upon hovering another item works after a795c22 and e909156. Just moving the mouse away from the submenu doesn't currently close it and I'm not sure if it should. It stays open until you hover a menu item lower in a lower submenu or if you click outside/move focus outside the menu

0281641 cleans up the root toggleable hovering behaviour. If there is no submenu active, hover does nothing, but with an active submenu hovering will swap between them and move over non-submenu items as well.

david-vanderson:
I did some testing here, and at least for me:
* clicking on an open submenu in the root doesn't close it (and we think it should right?)
* moving the mouse from a submenu item towards the floating submenu but it happens to go across another menu item focuses that menu item (it shouldn't, the logic around "_mi_mouse_down" is wrong now)
* context menus don't focus on highlight (but they probably should, even though they don't have a "more root" menu)

My plan is:
* remove all the "_mi_mouse_down" stuff
* replace root menu stuff with "are we in a FloatingMenuWidget" logic
* if not in a FloatingMenuWidget, then have a persistent bit of state that says if focus should follow hover (that is toggled on via a submenu click, and off via any click I think?)

david-vanderson:
That had a lot more weird corner cases than I was expecting, especially since we use the same code for dropdowns and suggestions.  But I think it's good to go now.

@RedPhoenixQ I can't reproduce the test failure locally, can you?

RedPhoenixQ:
> @RedPhoenixQ I can't reproduce the test failure locally, can you?

Sorry, found the underlying issue! I had falsely assumed that the saved hash hex string would be the same length, but if it was shorter then a byte of undefined memory would be included when parsing the hash. Fixed by 9b743164be117f83858319a60cde7b1693dfae5d.

david-vanderson:
Nice thanks!

PR #509, opened 2025-08-06
PanedWidget handle multiple widgets per pane

### LINKED ISSUE #500: PanedWidget expects only a single widget in each pane?
@foxnne ran into a problem with code like this:
```zig
pub fn draw() !void {
    drawTools() catch {};
    var paned = dvui.paned(@src(), .{
        .direction = .vertical,
        .collapsed_size = 300,
        .handle_size = 2,
        .handle_dynamic = .{},
    }, .{ .expand = .both, .background = true });
    defer paned.deinit();

    if (paned.showFirst()) {
        drawLayers() catch {};
    }

    if (paned.showSecond()) {
        dvui.labelNoFmt(@src(), "COLORS", .{}, .{ .font_style = .title });
    }
}
```

Where the content in `drawLayers()` was showing up in the second pane.  Wrapping a box around drawLayers() works around it, so my guess is that PanedWidget goes wrong if there are multiple widgets in a pane.

Will investigate.

RedPhoenixQ:
Closes #500

`PanedWidget` now uses `dvui.BasicLayout`, similar to `Window` and `ScaleWidget`, to stack multiple widgets vertically in each pane. This removed the need for a wrapping box for multiple child widget.

### DIFF
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -7434,15 +7434,15 @@ pub const BasicLayout = struct {
                 if (e.isVertical()) {
                     self.seen_expanded = true;
                 }
-                r.y = self.pos;
-                r.h = @max(0, r.h - r.y);
+                r.y += self.pos;
+                r.h = @max(0, r.h - self.pos);
             },
             .horizontal => {
                 if (e.isHorizontal()) {
                     self.seen_expanded = true;
                 }
-                r.x = self.pos;
-                r.w = @max(0, r.w - r.x);
+                r.x += self.pos;
+                r.w = @max(0, r.w - self.pos);
             },
         }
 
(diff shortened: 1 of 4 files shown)


src/dvui.zig
```diff
@@ -7434,15 +7434,15 @@ pub const BasicLayout = struct {
                 if (e.isVertical()) {
                     self.seen_expanded = true;
                 }
-                r.y = self.pos;
-                r.h = @max(0, r.h - r.y);
+                r.y += self.pos;
```
david-vanderson, line 7437: This looks a bit weird to me but don't have time to dig into it yet.  Just wanted to let you know - thanks!

foxnne:
Thank you both very much for looking into this. Today I borrowed the new `PanedWidget` from @RedPhoenixQ 's branch to re-implement my layers panel with this local change. Everything looks so much better and the code is far less complex. Please let me know if the `BasicLayout` changes wont work, I'll need to rethink how I'm approaching things.

PR #510, opened 2025-08-07
Paned autofit first pane to size of children

RedPhoenixQ:
This was requested by @foxnne and [showcased here](https://discord.com/channels/605571803288698900/1402035407138586806). It makes automatically setting a split ratio on creation much simpler. The autofit can be configured in the init options or called manually whenever to reset the split ration to fit again.

Currently it only works for vertical layout, which is where is makes the most sense. Applying the same method on width didn't work, not 100% sure why.

Sizing to the second pane could be possible but much harder, perhaps just inverting the visible placement of the first and second pane would be the simples solution (only needing to invert the offsets in `rectFor`).

### DIFF
--- a/src/Examples/layout.zig
+++ b/src/Examples/layout.zig
@@ -292,6 +292,7 @@ pub fn layout() void {
 
     dvui.label(@src(), "Auto-Fit Panes", .{}, .{});
     {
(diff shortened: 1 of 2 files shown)


david-vanderson:
Looks great - merge when ready.  I will look into making it work for horizontal panes.

Thank you!

PR #514, opened 2025-08-13
Update TabsWidget.zig

br07h3rj:
Added tab accent border control to TabsWidget. Users should now be able to turn off drawing the tab accent border by passing ".draw_focus = false" in the init options of TabsWidget.init(). Default set as "true" so that previous behavior does not change.

### DIFF
--- a/src/widgets/TabsWidget.zig
+++ b/src/widgets/TabsWidget.zig
@@ -16,6 +16,7 @@ pub var defaults: Options = .{
 
 pub const InitOptions = struct {
     dir: dvui.enums.Direction = .horizontal,
+    draw_accent : bool = true,
 };
 
 pub fn init(src: std.builtin.SourceLocation, init_opts: InitOptions, opts: Options) TabsWidget {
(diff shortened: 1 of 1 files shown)


david-vanderson:
Can you change the name to "draw_focus"?  I screwed up initially and said "accent" everywhere but we are trying to move to "focus" for this stuff.

(I would have made the change but I don't have permission to push to this branch)

br07h3rj:
I made the change and invited you and RedPhoenixQ to access my fork. I've been coding in some way or another for about 25 years. This is my first time using Github. I hate it. :smiling_face_with_tear:

RedPhoenixQ:
Whenever you make a PR you can choose if the collaborators in this repo can add commits to it. That would probably be enough for us to add changes to one of your WIP branches. You can also make a draft PR for stuff that isn't ready yet but you want to share to discuss solutions/implementations, where you can choose to allow us to add commits as well.

br07h3rj:
Good to know. I'll get the hang of it before long.

david-vanderson:
> I made the change and invited you and RedPhoenixQ to access my fork. I've been coding in some way or another for about 25 years. This is my first time using Github. I hate it. 🥲

I'm in a similar boat, we'll get through it!  Thanks for bearing with it.

PR #515, opened 2025-08-14
Custom Rect Function

br07h3rj:
Can we please add something like this for lazy people like me? I made one to use in developing my project to cut down on typing. It's basically Rect.all, except you pass it four values instead of one. Thoughts?

### DIFF
--- a/src/Rect.zig
+++ b/src/Rect.zig
@@ -55,6 +55,10 @@ pub fn RectType(comptime units: dvui.enums.Units) type {
             return .{ .x = v, .y = v, .w = v, .h = v };
         }
 
+        pub fn customRect(x: f32, y: f32, w: f32, h: f32) Self {
+            return .{ .x = x, .y = y, .w = w, .h = h };
+        }
+
         pub fn fromPoint(p: PointType) Self {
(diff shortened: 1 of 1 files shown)


br07h3rj:
I was going to name it "custom," which still compiled fine, but I was unsure about potential for conflicts in the future.

david-vanderson:
How about "rect"?  Then code reads like this: `.corner_radius = .rect(0, 2, 7, 100)`

Or could do "xywh": `.corner_radius = .xywh(0, 2, 7, 100)`

Or "init" I guess: `.corner_radius = .init(0, 2, 7, 100)`

br07h3rj:
I would love to use ".rect", but that gives lots of compilation errors because it shadows const declarations in a bunch of other places in Rect.zig. I think I've come up with the perfect name. It's short, descriptive, and easy to remember: getRect().

Edit: Now that I think about it more, maybe makeRect() would be more descriptive?

david-vanderson:
> I would love to use ".rect",

Then let's do it - how does that look?

br07h3rj:
Looks great to me!

david-vanderson:
Thanks!

PR #517, opened 2025-08-14
Removed Auto-Select First Entry in Menu

br07h3rj:
I found one last thing I don't like about the menu behavior. (There's always one more.) Auto-selecting the first entry in the menu is an unusual feature and removing it doesn't seem to break anything as far as I can tell. Is there a reason this is needed?


br07h3rj:
Note: I just checked and keyboard navigation is uneffected.

david-vanderson:
I added that to make navigating into deep submenus faster, but you're right, we shouldn't do this.

david-vanderson:
Thank you!

PR #520, opened 2025-08-17
Better cross widget dragging

### LINKED ISSUE #516: cross-widget drag from Reorderable
@foxnne needs the ability to drag something from a Reorderable (or Tree) to something else (which might be a Reorderable or Tree, but also could be anything).

The demo will be enhanced to show dragging from a Reorderable or Tree to a box or tabs or something (and back?).

My first thought was to add a "dragging_name" option to each widget.  But then I thought maybe an application could have multiple overlapping sets of things that drag between them, so a single dragging name might not be enough.

So current plan is to try and expose when a drag starts, so that user code can optionally set the dragging name and give up mouse capture.  Still might need a "dragging_name" option on something like Reorderable/Tree so that it can render/respond to something that it can receive.

Will try that and report.

david-vanderson:
Fixes #516

When @foxnne tried to do cross-widget dragging for real, we didn't have enough support.  This adds:
* eventMatch() dragging_name parameter
  * used for opting in to a named (cross-widget) drag
* FloatingWidget mouse_events option
  * if a floating widget is being used to visually show a cross-widget drag, then mouse events need to go through it to the subwindow below
  * this allows a widget under the floating one to receive mouse position and release to end the drag

I've updated the scroll_canvas demo.

Still working on adding a demo showing drag-n-drop from a Tree to another widget, and between two Reorderables.

### DIFF
--- a/src/Event.zig
+++ b/src/Event.zig
@@ -21,6 +21,15 @@ evt: union(enum) {
     text: Text,
 },
 
+pub fn format(self: *const Event, comptime _: []const u8, _: std.fmt.FormatOptions, writer: anytype) !void {
+    try std.fmt.format(writer, "Event({d}){{{s} ", .{ self.num, @tagName(self.evt) });
+    switch (self.evt) {
+        .mouse => |me| try std.fmt.format(writer, "{s}}}", .{@tagName(me.action)}),
+        .key => |ke| try std.fmt.format(writer, "{s}}}", .{@tagName(ke.action)}),
+        .text => try std.fmt.format(writer, "}}", .{}),
+    }
+}
+
 /// Mark the event as handled
 ///
 /// In general, the `dvui.WidgetData` passed here should be the same one that
@@ -28,13 +37,7 @@ evt: union(enum) {
 /// This makes it possible to see which widget handled the event.
 pub fn handle(self: *Event, src: std.builtin.SourceLocation, wd: *const dvui.WidgetData) void {
     if (dvui.currentWindow().debug.logEvents(null)) {
-        var action: []const u8 = "";
-        switch (self.evt) {
-            .mouse => action = @tagName(self.evt.mouse.action),
-            .key => action = @tagName(self.evt.key.action),
-            else => {},
-        }
-        dvui.log.debug("{s}:{d} {s} {s} event (num {d}) handled by {s} ({x})", .{ src.file, src.line, @tagName(self.evt), action, self.num, wd.options.name orelse "???", wd.id });
+        dvui.log.debug("{s}:{d} {} handled by {s} ({x})", .{ src.file, src.line, self, wd.options.name orelse "???", wd.id });
     }
     self.handled = true;
 }
--- a/src/Examples/reorder_tree.zig
+++ b/src/Examples/reorder_tree.zig
@@ -246,7 +246,7 @@ pub fn reorderListsAdvanced(cross_drag: bool) void {
 
         dvui.label(@src(), "Drag to add : {d}", .{g.strings_len}, .{});
 
-        if (dvui.ReorderWidget.draggable(@src(), .{ .top_left = hbox2.data().rectScale().r.topLeft() }, .{ .expand = .vertical, .gravity_x = 1.0, .min_size_content = dvui.Size.all(22), .gravity_y = 0.5 })) |p| {
+        if (dvui.ReorderWidget.draggable(@src(), .{ .rect = hbox2.data().rectScale().r }, .{ .expand = .vertical, .gravity_x = 1.0, .min_size_content = dvui.Size.all(22), .gravity_y = 0.5 })) |p| {
             // add to list, but will be removed if not dropped onto a list slot
             g.strings[g.strings_len] = g.strings_template[g.strings_len];
             added_idx = g.strings_len;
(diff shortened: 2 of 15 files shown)


david-vanderson:
@foxnne Do you have a recommended way to visually communicate (during a cross-widget drag):
1. "This area can be dropped onto"
2. "You are hovering over this drop area"

In the updated scroll_canvas demo I'm doing 1 with a border color, and 2 with a semi-opaque fill, but I just made that up.

I suspect that people basing apps off this code will generally follow what we do, so what should we do?

foxnne:
> In the updated scroll_canvas demo I'm doing 1 with a border color, and 2 with a semi-opaque fill, but I just made that up.
> 
> I suspect that people basing apps off this code will generally follow what we do, so what should we do?

Sorry, been trying to think on this, and I just quickly went through several other apps or os windows dragging and dropping. It seems in most cases theres no visual indication that an area can be dropped on to until its hovered. In my attempt I didn't either, but like you did showed a semi-transparent highlight color fill over the droppable area when its hovered.

I think the outline/border is a good approach and is subtle enough to not be intrusive when hovering. I think ideally it would almost behave like the current draw focus that outlines the control in the highlight color, where droppable regions are outlined, and when hovered they fill in with a semi-transparent fill?

david-vanderson:
Thanks for looking into it!

> I think the outline/border is a good approach and is subtle enough to not be intrusive when hovering. I think ideally it would almost behave like the current draw focus that outlines the control in the highlight color, where droppable regions are outlined, and when hovered they fill in with a semi-transparent fill?

That is a great idea, will run with it - thanks!

david-vanderson:
I think the last remaining part I need to do here is rework Reorderable and Tree to split drag start from drag hover.  Right now they both assume that if a drag is happening they started it.

foxnne:
Awesome! I'll be trying the new demo and digging into this very soon!

My ultimate goal would be drag and drop from file tree or tab bar to either editor to split or join an existing pane or directly into the other tab bar reorderable. 

I think the demo having this kind of interaction is really awesome too!

david-vanderson:
> My ultimate goal would be drag and drop from file tree or tab bar to either editor to split or join an existing pane or directly into the other tab bar reorderable.

I think it's turning out pretty well, and we will definitely be able to support that stuff.  I hope to rework Reorderable and Tree tomorrow.

> 
> I think the demo having this kind of interaction is really awesome too!

Gotta flaunt it!

foxnne:
> I think it's turning out pretty well, and we will definitely be able to support that stuff. I hope to rework Reorderable and Tree tomorrow.

Amazing! Im currently trying to update `Reorderable` following the same changes to `Tree`, but for some reason I cannot get it to work in my app and I'm not sure what I'm doing wrong... I might just have to wait on your changes and see what I've got mixed up haha.

david-vanderson:
> > I think it's turning out pretty well, and we will definitely be able to support that stuff. I hope to rework Reorderable and Tree tomorrow.
> 
> Amazing! Im currently trying to update `Reorderable` following the same changes to `Tree`, but for some reason I cannot get it to work in my app and I'm not sure what I'm doing wrong... I might just have to wait on your changes and see what I've got mixed up haha.

Yeah hang tight, once I get it working I'll writeup on this PR the things I learned and a rough draft of how to do this stuff.

david-vanderson:
Got the first very rough version going.  Can drag items from the simple reorder list to the advanced reorder list.

Things I've learned:
* I need to add an offset and size to the drag info, so anyone can figure out from the current mouse position the screen rect of the thing being dragged.
* mouse release on a widget is what triggers the end of the drag, but the source widget only finds out the next time it runs

So the general strategy is:
* the app must record info about the thing being dragged (in ReorderWidget that is from reorderable.floating())
* on mouse release, reorderable reports insertBefore()
  * if it also reports removed() then the item was dragged and dropped within the same widget
  * otherwise it came from a different widget
  * the app can either only do the insertion, or also remove the item from wherever it came from (and must wipe the app-recorded info)
* when the source widget realizes the drag has ended, it can (optionally) remove the dragged item (the app recorded this info before)
  * in some cases a reorderable might still report removed() but the app can detect that its recorded info was wiped and do nothing

I need to clean a lot of this up, so it's not ready yet.

foxnne:
Awesome! Great work and sorry this has been complex and I'm sure taken a good bit of time. Thank you so much for all your effort!

Please let me know if you'd like me to take a stab at anything or if there's anything I can do to help, I'm getting up to speed with your code now and adapting my project as you go so I can test things and try to learn how it works.

david-vanderson:
I think this is mostly there now, so please ask questions.  A bit annoying to directly manipulate some stuff on dvui.Window, but we can clean that up later.

One thing that's a bit weird is that Reorder will detect a mouse release during a drag if the dragged item (drag offset plus drag size) is above the Reorder, even if the mouse is not.  I'm worried that might cause problems if someone is trying to drag an item that hovers over multiple drop sites at the same time.  Will have to see.

I haven't updated TreeWidget yet, but it should go similar to ReorderWidget.

I might not be able to work on this for a week or so, but will answer questions as I can.  Then I'll look at it again before merging.

Thanks!

foxnne:
Okay, so I've been able to get my project updated to use this latest code and it's so awesome, being able to drag and drop like this is really cool. One change I've had to make is like you mentioned earlier, it currently checks for an overlap with the dragged item rect. 

I would recommend that we change the intersect test to `rs.r.contains(dvui.currentWindow().mouse_pt)` so that regardless of the offset, we always test under the mouse position. That way users can specify an offset if they want and it still not affect the drop position. This is how Safari, VSCode, ghostty, etc behave. The tab gets either centered or keeps the offset from the drag start but the drop target remains under the mouse position.

I'm updating this branch on my fork with that change but let me know if you'd like me to push it here as well. 

Thanks again, this is awesome!

foxnne:
My apologies, I meant that to be pushed to my branch, and I'm honestly a bit unsure how I have a branch from the DVUI project in my fork, but I guess I'm misunderstanding how branches work. If this change is unwanted, hopefully it can be easily reverted.

david-vanderson:
> I would recommend that we change the intersect test to `rs.r.contains(dvui.currentWindow().mouse_pt)`

Good - let's go forward with that.  It can also simplify the Reorder event handling, which I'll do before merging.

The reason I did that was when originally showing reorderable lists to people, they expected that if the dragged item was visually anywhere over a valid drop place it should work.  But it makes it much more complicated, so we should do the simple thing until/unless somebody really needs it.

Thanks!

PR #524, opened 2025-08-26
web: implement refresh

### LINKED ISSUE #523: dvui.refresh doesn't request a new frame in web backend
[backends.web.refresh implementation](https://github.com/david-vanderson/dvui/blob/a145573253c8a304824bd3aa1b36a1a69ce80c07/src/backends/web.zig#L715)

I have some code that affects the UI asynchronously, but apparently unless you e.g. move the mouse, a new frame won't be drawn in the web backend. I wanted to use `dvui.refresh` to force a new frame to draw after the async work completes. Is it supposed to work this way? Is there an alternative?

RedPhoenixQ:
Closes #523

This allows for calling `dvui.refresh` with an explicit `Window` parameter when using the web backend. If you need to request a frame from JavaScript, then use `dvui_instance.requestRender()`

### DIFF
--- a/src/backends/web.zig
+++ b/src/backends/web.zig
@@ -83,6 +83,7 @@ pub const wasm = if (!builtin.is_test) struct {
         return undefined;
     }
     pub fn wasm_sleep(_: u32) void {}
+    pub fn wasm_refresh() void {}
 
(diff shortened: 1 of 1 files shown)


david-vanderson:
Thank you!

Issue #488, opened 2025-07-17
Struct UI Enhancements


phatchman:
## Description

The current StructUI has some limitations:
1) FieldOptions are built recursively for the displayed struct and all of its children, even those fields are never displayed
    - This causes issues with displaying self-referential data types, like lists. (or dvui.Options)
2) Memory management is largely an unfinished problem
    - It's not always clear when to allocate memory or when to free it as these are based on user-defined rules unknown to dvui.
3) Limited customization options
    - Some things can be modified through "options", but if it isn't in options, it can't be changed.
4) Unable to mark data as read-only. i.e. not all values in a struct are free to be modified.
5) Lower-level API is not exposed.
    - Users must use one of the structWidget family of functions or not use structUI at all.
6) It is generally hard to debug without `@compileLog` statements.
    - There are two lots of comptime things happening in partallel. 1) For the FieldOptions and 2) For the widget display code.
    - It is very difficult to debug the FieldOptions issues as they become nested to the extent that Zig can't give a good error trace.

## Proposal

Restructure Struct UI so that:
1) Remove FieldOptions and NamespaceFieldOptions, which cause the recursive type issues.
    - Keep individual options for each type of widget, but do not propagate options to sub-fields.
2) Move all memory management to the user. This includes:
    - editing slices to const u8
    - memory allocation for optional pointers becoming set.
    - Moving memory for entity re-ordering etc.
    - Initializing any created entity.
    - Any memory allocation or free directly associated with the struct or it's children.
3) Expose the lower-level widget display functions for basic types, such as ints, float, bool, enum.
4) Provide a mid-level API that will provide default display logic for collections of basic types, such as arrays, slices, optionals etc.
5) Provide a high-level API that will recursively display nested structs to a specified depth.
6) Ability to exclude fields from display
7) Ability to make fields read-only
8) Ability to add / remove elements from a data structure
    - User will be responsible for performing managing add / remove and all initialization required.
9) Allow mixing of low-level and high-level API's
     - For example, display 90% of the struct using the high-level API, and use the low/mid-level APIs to display the few custom fields required.
10) Perform comptime param checking as soon as possible to give better error messages
    - Make sure to include field names in any error messages. This makes a big difference to tracking down issues.

  
## Discussion

A POC has been created demonstrating most of the principles above and solves a lot of the issues that stalled further progress on StructUI. Fortunately, a lot of the existjng code can be re-used, just structured differently.

david-vanderson:
Awesome!!

[...]

PR #526, opened 2025-08-27
struct_ui rework

phatchman:
Implements #488 

## Description ##

Overhaul of the struct_ui functionality. Highlights are:
1) Field options are now per type, rather than per field. This simplifies the options processing considerably.
2) Decomposed the API into ~3 levels of publicly accessible functions
    - Individual widget display functions (e.g. numberFieldWidget)
    - Field display functions that call one or more widget display functions ( e.g. displayNumber)
    - Container display functions to display a struct or union. (e.g. displayStruct)
3) Improved compile-time errors messages.
4) Renamed from structUI to struct_ui and generally conform to zig naming standards.
5) Adds a "depth" parameter to the struct display, so that the user can control how many levels of nested structs to display. A depth of 0 will show all non-struct fields, including pointers, optionals or unions. Depth is only incremented when processing a struct fields.

Hopefully the new comptime code is simpler to understand and easier to use.
 
 ## Discussion / TODO ##
 
 This is really just to get something out there showing the reworked API and some usage examples. There are still a few things to clean up, but no changes to the API are expected.
 
1. I ended up deleting the data adapter changes (deja vu). They do work quite well and are useful, but:
   - I'm not sure how much use struct_ui will ultimately get and whether it is worth adding complexity until we know there is a need for it.
   - Just from a usability standpoint, it adds another comptime path that the user needs to get right, requires passing pointers to adapters, rather than copies of adapters, which all needs comptime checking with nice error messages to make it manageable. i.e. it's a surprising amount of code to add for just being able to modify a string. (well not just that, but mostly that).
   - If there is a need, it is easy to bring it back, maybe in a simpler form. I'll keep the branch alive.
2. As a result, strings are handled in a bit of a hacky, but simple way. They get assigned to the TextWidget's internal dynamically allocated buffer. If the struct should live longer than the lifetime of the window (or outside the window begin/end??), the user will have to dupe the strings. Hopefully that is an OK compromise for now.
3. The theme editor in demos works using the new APIs.
4. ~~I've temporarily hijacked sdl-standalone for some struct tests.~~
5. ~~There are some things that still need fixing, like optionals for strings, but it's working for all struct combinations I've managed to think up so far.~~
6. ~~There is a lot more comptime error checking to be done. I've found you need to check the `type` and `anytype` parameters at each level to get any sort of comptime debuggability.~~
7. ~~The UI needs a bit of a cleanup for alignment and display of optionals and unions. But I'm trying to keep it to a fairly minimal look based around expanders grouping together fields in a container.~~
8. Probably won't expose any of it via dvui.zig functions. People can just import the dvui.struct_ui namespace and use the functions directly. [Note now exposes one function named structUI, to make sure the displayStruct is wrapped in a box to ensure a unique WidgetId.]
10. Potentially make changes to the options so fields can be totally skipped, rather than just "not displayed". - Done
11. ~~displayStruct doesn't take @src(), so if someone does a displayStruct on the same type more than once, they will get duplicate id errors. So maybe it is worth exposing just his one via a dvui.zig function to wrap the call in a box. I could make it take @src(), and pass it everywhere, but after creating the first box, it is no longer needed, everything else will get a unique src based on the comptime expansion for each type.~~
12. Add an example showing an advanced use of the API, e.g. for an optional pointer to a union containing a struct or similar with 3 levels of nesting, with the custom handling for the optional pointer in the 2nd level.
13. ~~Implement read-only display for all widget functions.~~

## Theme editor 
image

### DIFF
--- a/src/Examples/struct_ui.zig
+++ b/src/Examples/struct_ui.zig
@@ -1,16 +1,23 @@
-/// ![image](Examples-struct_ui.png)
+///![image](Examples-struct_ui.png)
 pub fn structUI() void {
     var b = dvui.box(@src(), .{}, .{ .expand = .horizontal, .margin = .{ .x = 10 } });
     defer b.deinit();
 
     const Top = struct {
         const TopChild = struct {
-            a_dir: dvui.enums.Direction = undefined,
+            a_dir: dvui.enums.Direction = .vertical,
+        };
+
+        const SomeUnion = union(enum) {
+            enum_field: enum { one, two, three, four },
+            number_field: f32,
+            struct_field: TopChild,
         };
 
         const init_data = [_]TopChild{ .{ .a_dir = .vertical }, .{ .a_dir = .horizontal } };
         var mut_array = init_data;
         var ptr: TopChild = TopChild{ .a_dir = .horizontal };
+        var str: [7]u8 = .{ 'e', 'd', 'i', 't', ' ', 'm', 'e' };
 
         a_u8: u8 = 1,
         a_f32: f32 = 2.0,
@@ -19,62 +26,50 @@ pub fn structUI() void {
         a_bool: bool = false,
         a_ptr: *TopChild = undefined,
         a_struct: TopChild = .{ .a_dir = .vertical },
-        a_str: []const u8 = &[_]u8{0} ** 20,
+        a_str_const: []const u8 = &[_]u8{'$'} ** 20,
+        a_str_var: []u8 = &str,
         a_slice: []TopChild = undefined,
+        a_union: SomeUnion,
         an_array: [4]u8 = .{ 1, 2, 3, 4 },
 
-        var instance: @This() = .{ .a_slice = &mut_array, .a_ptr = &ptr };
+        var instance: @This() = .{ .a_slice = &mut_array, .a_ptr = &ptr, .a_union = .{ .enum_field = .three } };
     };
 
     dvui.label(@src(), "Show UI elements for all fields of a struct:", .{}, .{});
     {
-        dvui.structEntryAlloc(@src(), dvui.currentWindow().gpa, Top, .{}, &Top.instance, .{ .margin = .{ .x = 10 } });
+        dvui.structUI(@src(), "Top.Instance", &Top.instance, 1, .{});
     }
 
-    //if (dvui.expander(@src(), "Edit Current Theme", .{}, .{ .expand = .horizontal })) {
-    //    themeEditor();
-    //}
+    if (dvui.expander(@src(), "Edit Current Theme", .{}, .{ .expand = .horizontal })) {
+        themeEditor();
+    }
 }
 
-/// ![image](Examples-themeEditor.png)
+var font_buf: [50]u8 = @splat('z');
+
+///![image](Examples-themeEditor.png)
 pub fn themeEditor() void {
     var b2 = dvui.box(@src(), .{}, .{ .expand = .horizontal, .margin = .{ .x = 10 } });
     defer b2.deinit();
 
-    //const color_field_options = dvui.StructFieldOptions(dvui.Color, .{ .style_err, .style_accent }){ .fields = .{
-    //    .r = .{ .min = 0, .max = 255, .widget_type = .slider },
-    //    .g = .{ .min = 0, .max = 255, .widget_type = .slider },
-    //    .b = .{ .min = 0, .max = 255, .widget_type = .slider },
-    //    .a = .{ .disabled = true },
-    //} };
-
-    //dvui.structEntryEx(@src(), "dvui.Theme", dvui.Theme, .{ .style_err, .style_accent }, dvui.themeGet(), .{
-    //    .use_expander = false,
-    //    .label_override = "",
-    //    .fields = .{
-    //        .name = .{ .disabled = true },
-    //        .dark = .{ .widget_type = .toggle },
-    //        .font_body = .{ .disabled = true },
-    //        .font_heading = .{ .disabled = true },
-    //        .font_caption = .{ .disabled = true },
-    //        .font_caption_heading = .{ .disabled = true },
-    //        .font_title = .{ .disabled = true },
-    //        .font_title_1 = .{ .disabled = true },
-    //        .font_title_2 = .{ .disabled = true },
-    //        .font_title_3 = .{ .disabled = true },
-    //        .font_title_4 = .{ .disabled = true },
-    //        .color_accent = color_field_options,
-    //        .color_err = color_field_options,
-    //        .color_text = color_field_options,
-    //        .color_text_press = color_field_options,
-    //        .color_fill = color_field_options,
-    //        .color_fill_window = color_field_options,
-    //        .color_fill_control = color_field_options,
-    //        .color_fill_hover = color_field_options,
-    //        .color_fill_press = color_field_options,
-    //        .color_border = color_field_options,
-    //    },
-    //});
+    // Initialize with just the r, g and b fields. a will not be displayed.
+    // Each time a new colour struct is instantiated it will use the supplied defaults for r, g, b and a.
+    const color_options: dvui.struct_ui.StructOptions(dvui.Color) = .init(.{
+        .r = .{ .number = .{ .min = 0, .max = 255, .widget_type = .slider } },
+        .g = .{ .number = .{ .min = 0, .max = 255, .widget_type = .slider } },
+        .b = .{ .number = .{ .min = 0, .max = 255, .widget_type = .slider } },
+    }, .{ .r = 127, .g = 127, .b = 127, .a = 255 });
+    const theme: *dvui.Theme = &dvui.currentWindow().theme; // Want a pointer to the actual theme, not a copy.
+    if (dvui.struct_ui.displayStruct(
+        "Theme",
+        theme,
+        2,
+        .default,
+        .{color_options},
+        null,
+    )) |box| {
+        box.deinit();
+    }
 }
 
 test {
@@ -104,7 +99,7 @@ test "DOCIMG struct_ui" {
 //
 //    const frame = struct {
 //        fn frame() !dvui.App.Result {
-//            var box = dvui.box(@src(), .{}, .{ .expand = .both, .background = true, .color_fill = .fill_window });
+//            var box = dvui.box(@src(), .vertical, .{ .expand = .both, .background = true, .color_fill = .fill_window });
 //            defer box.deinit();
 //            themeEditor();
 //            return .ok;
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -94,12 +94,12 @@ pub const TextLayoutWidget = widgets.TextLayoutWidget;
 pub const TreeWidget = widgets.TreeWidget;
 pub const VirtualParentWidget = widgets.VirtualParentWidget;
 pub const GridWidget = widgets.GridWidget;
-const se = @import("structEntry.zig");
-pub const structEntry = se.structEntry;
-pub const structEntryEx = se.structEntryEx;
-pub const structEntryAlloc = se.structEntryAlloc;
-pub const structEntryExAlloc = se.structEntryExAlloc;
-pub const StructFieldOptions = se.StructFieldOptions;
+pub const struct_ui = @import("struct_ui.zig");
+//pub const structEntry = se.structEntry;
+//pub const structEntryEx = se.structEntryEx;
+//pub const structEntryAlloc = se.structEntryAlloc;
+//pub const structEntryExAlloc = se.structEntryExAlloc;
+//pub const StructFieldOptions = se.StructFieldOptions;
 
 pub const enums = @import("enums.zig");
 pub const easing = @import("easing.zig");
@@ -7578,6 +7578,23 @@ pub const BasicLayout = struct {
         return self.min_size_children;
     }
 };
+
+/// Display a struct and allow the user to edit values
+///
+/// Refer to struct_ui.zig for full API.
+/// Call StructOptions(T) to to create display options for the struct or use .{} for defaults.
+/// See struct_ui.displayStruct for more details.
+///
+/// IMPORTANT NOTE:
+/// Any modifyable string slice fields are assigned to the TextWidget's internal string buffer.
+/// If the lifetime of the struct will outlive the lifetime of the TextWidget's window, you must make sure to
+/// dupe any modified strings before the widnow is deinitialized.
+pub fn structUI(src: std.builtin.SourceLocation, comptime field_name: []const u8, struct_ptr: anytype, comptime depth: usize, struct_options: anytype) void {
+    var vbox = dvui.box(src, .{ .dir = .vertical }, .{ .expand = .horizontal });
+    defer vbox.deinit();
+    const struct_box = struct_ui.displayStruct(field_name, struct_ptr, depth, .default, struct_options, null);
+    if (struct_box) |b| b.deinit();
+}
 
 test {
     //std.debug.print("DVUI test\n", .{});
--- a/src/struct_ui.zig
+++ b/src/struct_ui.zig
@@ -1,0 +1,1131 @@
+const std = @import("std");
+const dvui = @import("dvui.zig");
+
+/// Field options control whether and how fields are displayed.
+///
+/// Use TextFieldOptions for any array or slice of u8 you want ot display as a string.
+/// Use NumberFieldOptions for any numbers, allowing setting of min and max ranges and other options
+/// Use StandardFieldOptions (.default) for all other fields.
+///
+/// All FieldOptions types must provide:
+/// - display: DisplayMode
+/// - label: ?[]const u8
+pub const FieldOptions = union(enum) {
+    /// Control if the field should be displayed and if it is editable.
+    const DisplayMode = enum { none, read_only, read_write };
+    standard: StandardFieldOptions,
+    number: NumberFieldOptions,
+    text: TextFieldOptions,
+
+    pub const default: FieldOptions = .{ .standard = .{} };
+
+    pub fn displayMode(self: FieldOptions) DisplayMode {
+        return switch (self) {
+            inline else => |fo| fo.display,
+        };
+    }
+
+    pub fn displayLabel(self: FieldOptions, field_name: []const u8) []const u8 {
+        return switch (self) {
+            inline else => |fo| fo.label orelse field_name,
+        };
+    }
+
+    pub fn labelSet(self: *FieldOptions, field_label: []const u8) void {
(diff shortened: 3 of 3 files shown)


david-vanderson:
Excellent!  Will dig into it!

phatchman:
Can anyone thing of a good way of testing comptime error messages? 

I need to exercise all the error paths at least once. Outside of just checking the errors look OK, there could be syntax errors, bugs etc in the error handling paths that don't get found until that code is run during compile time.

Right now I have a fn that contains lots of calls with invalid parameters and comment out all the lines except the current one I'm testing. But that's not really sustainable outside of a 1-off test.

I can think of complicated ways to do it using conditional compilation and the build system, but that seems like a huge overkill to test some error messages. I need comptime tests or something.

phatchman:
> Excellent! Will dig into it!

Should be relatively close now. 

Needs :-
1)  better doc comments
2) I'd like to make it so you can omit FieldOptions for a field and that will prevent the field from being processed at all. So if there are unsupported types, like untagged unions, or many item pointers, then can be totally ommitted.
3) Better examples.
4) Need to restore ability to do custom display for struct and union fields. (i.e. displayStruct returns box for additional fields to be displayed in)

david-vanderson:
> Can anyone thing of a good way of testing comptime error messages?

Sadly no: https://github.com/ziglang/zig/issues/513

Looking at the zig source code, they have the build runner spit out a "tmp.zig" with the test case, run it, and capture the compiler output.

I think I'd keep doing what you are doing.  I really appreciate the commitment to testing!

phatchman:
OK. I think just now time to do the finishing touches and get a nicer demo of the features.

I've also come to quite like using it. While developing, structs are constantly changing adding, removing, moving etc. Being able to just say "display this struct" and have it adapt as you change your code is actually quite nice. Then once it has settled down you can consider making a dedicated interface that is more tailored.

There's quite a few more options and display types we can add in the future, max string lengths, should the struct default to expanded, etc. But they should all fit fairly easily into this framework.

sdl-standalone has a couple of examples of handling custom fields. e.g. handling data_out for options:

In the code below, I'm creating way more options than you normally would, e.g. providing default and min, max values for Options.MaxSize. But notice data_out is set to not display in the options. (I could also omit that field from the options struct as an alternative). 

Then the displayStruct() returns a box, that allows the data out field to be displayed at the end of the struct and set a valid pointer to data_out if the user makes it not null.

Everyone shares the same dvui.Alignment, so it displays exactly as if it was performed via displayStruct()
It also gives a way for the user to put buttons or other widgets in the display if they want full customization.

Removing .parent from the WidgetData options is again, just a test. It's not required, but shows that you can initialize a struct will all defaults and then just change or remove the ones you need. 

```
    var max_size_opts: dvui.struct_ui.StructOptions(dvui.Options.MaxSize) = .initWithDefaults(.{}, .{ .h = 100, .w = 100 });
    max_size_opts.field_options.put(.w, .{ .number = .{ .min = 1, .max = dvui.max_float_safe } });
    max_size_opts.field_options.put(.h, .{ .number = .{ .min = 1, .max = dvui.max_float_safe } });

    const font_opts: dvui.struct_ui.StructOptions(dvui.Font) = .initWithDefaults(.{}, .{ .size = 10, .id = .Vera });
    const options_options: dvui.struct_ui.StructOptions(dvui.Options) = .initWithDefaults(.{
        .name = .{ .text = .{ .display = .read_write } },
        .tag = .{ .text = .{ .display = .read_write } },
        .data_out = .{ .standard = .{ .display = .none } },
    }, null);
    const color_options: dvui.struct_ui.StructOptions(dvui.Color) = .init(.{
        .r = .{ .number = .{ .widget_type = .slider } },
        .g = .{ .number = .{ .widget_type = .slider } },
        .b = .{ .number = .{ .widget_type = .slider } },
    }, .{ .r = 127, .g = 127, .b = 127, .a = 255 });

    var widget_data_options: dvui.struct_ui.StructOptions(dvui.WidgetData) = .initWithDefaults(.{}, null);
    widget_data_options.field_options.remove(.parent);

    var al: dvui.Alignment = .init(@src(), 0);
    defer al.deinit();
    const box = dvui.struct_ui.displayStruct(
        "dvui.Options",
        &dvui_opts,
        1,
        .default,
        .{ options_options, max_size_opts, font_opts, color_options },
        &al,
    );
    if (box) |b| {
        defer b.deinit();
        dvui.struct_ui.displayOptional(
            "data_out",
            &dvui_opts.data_out,
            3,
            .default,
            .{ options_options, max_size_opts, font_opts, color_options },
            &al,
            &data_out,
        );
    }
}
```

In the second one, I'm manually handling a union, so I can set default values for that union. The displayStruct() is perfectly capable of displaying this union, but I've chosen to do it manually as an example.

Displaying the optional int pointer is the same tecnique as above, a default value is provided as the last parameter to displayOptional().

Displaying the union, it uses the low-level union widget to get the selection from the user and then if the selection has changed, the code initializes the union with custom init values, then displays the union field.

```
if (dvui.struct_ui.displayStruct("test_struct", &ts, 1, .{ .standard = .{} }, .{TestStruct.structui_options}, &alignment)) |box| {
    defer box.deinit();

    dvui.struct_ui.displayOptional("o_ptr", &ts.o_ptr, 1, .{ .number = .{} }, .{}, &alignment, &ts_int);
    const selected_tag = dvui.struct_ui.unionFieldWidget(@src(), "union", &ts.union4, .default);
    switch (selected_tag) {
        .c1 => {
            if (ts.union4 != .c1)
                ts.union4 = .{ .c1 = .{ .value = 44 } };
            dvui.struct_ui.displayField("c1", &ts.union4.c1, 1, .default, .{}, &alignment);
        },
        .c2 => {
            if (ts.union4 != .c2)
                ts.union4 = .{ .c2 = .{ .value2 = 55 } };
            dvui.struct_ui.displayField("c2", &ts.union4.c2, 1, .default, .{}, &alignment);
        },
    }
}
```
This also shows another pattern we could do. dvui structs could supply a const declaration named struct_ui_options which provides the information to display that struct nicely. 

So that's basically how users can handle fields that the normal displayStruct() won't, including for memory allocation and deallocation.

phatchman:
This should be good to merge now, if you are good with it. But no rush to review.

The "just display something" function is dvui.strucUI(), while the more advanced API is in the dvui.struct_ui namespace. I'm open to better naming.

Also let me know if you see any issues with the way I'm handling the strings with assigning the slices to the string widget's buffer. There is an alternative displayStringBuf() that can be used if the user wants to manually pass in a buffer for the text widget to use instead. 

I want to add a better example of custom field handling to the demo, but will do that as a separate PR.
The UI could be made a little nicer in places, but again, that can come as a separate PR.

david-vanderson:
This is looking really good!

> Also let me know if you see any issues with the way I'm handling the strings with assigning the slices to the string widget's buffer. There is an alternative displayStringBuf() that can be used if the user wants to manually pass in a buffer for the text widget to use instead.

That is a good idea, and I like pushing the ownership of the string back onto the usage code.  Maybe I missed it, but how would the usage code know if it needed to dupe/deal with a changed string?

phatchman:
> That is a good idea, and I like pushing the ownership of the string back onto the usage code. Maybe I missed it, but how would the usage code know if it needed to dupe/deal with a changed string?

Yeah, there is no easy way without comparing before/after pointers. It only matters if the struct will out-live the window, but I think that would be most of the time with "on-top" style interfaces where I'm guessing this would be most used.

My thoughts were the user would always dup strings before the window closes. But I now realise that doesn't really work if the strings are already heap allocated. You wouldn't know which of the original strings to free unless you know which have changed.

So options are:
1) User needs to compare before/after pointers and free themselves.
2) User always needs to use the displayStringBuf versions if the if they don't want to remember pointers
3) I could make the API always change the string to point to the TextWidget's buffer even if the string doesn't change. That would mean:
    - After the first display of the struct, all original heap allocated strings can be freed.
    - After the last display of the struct (before the window is closed), all strings should be duped.

I think (3) sounds promising as the user can handle all the strings at once, rather than checking if each is modified. But, I can think of a future issue with it. If we provide an option that allows some structs to be displayed collapsed, the user wouldn't know if some child struct was not displayed and the textWidget is not created for it. But that's for a future me to work out.

Let me put together a small demo with heap allocated strings and see what it looks like.

phatchman:
What do you think about something like the below: (Yes, I'll try and come up with a better name for this)
```
var sw: StringWrangler = .{};
```

```
fn gui_frame() void {
            if (!sw.initialized) {
                sw.init(&top_string.top_str);
            }
            dvui.structUI(@src(), "top_string", &top_string, 2, .{ top_options, base_options });
            if (sw.changed()) {
                std.debug.print("changed\n", .{});
                sw.wrangle(gpa) catch |err| @panic(@errorName(err));
            }
}

const TopString = struct {
    top_str: []const u8 = "top_default",
    base_str: ?BaseString = .{},
};
```
For each string, you create a "Wrangler" that:
1) Keeps track of the original pointer, so knows if it has changed.
2) Knows if the string is statically or dynamically allocated (user will also be able to set this during init())
3) Will dup the string only if it has changed. (Implemented in the wrangle() fn)
4) Has a deinit() so if the string was dynamically allocated, it gets freed at the end.

In the example above, you probably wouldn't "wrangle" on each change, only when the window is closed. The wrangler needs to have the same lifetime as the struct, so when you open the window again, you'd use the same wranglers from previous to retain the state about whether the strings are dynamically allocated.

```
pub const StringWrangler = struct {
    initialized: bool = false,
    dynamically_allocated: bool = false,
    str_ptr: ?*[]const u8 = null,
    str_slice: ?[]const u8 = null,

    pub fn init(self: *StringWrangler, str_ptr: *[]const u8) void {
        self.initialized = true;
        self.str_ptr = str_ptr;
        self.str_slice = str_ptr.*;
    }

    pub fn changed(self: StringWrangler) bool {
        if (self.initialized) {
            return self.str_ptr.?.*.ptr != self.str_slice.?.ptr;
        } else {
            return false;
        }
    }

    pub fn wrangle(self: *StringWrangler, allocator: std.mem.Allocator) !void {
        if (self.initialized and self.changed()) {
            if (self.dynamically_allocated) {
                allocator.free(self.str_slice.?);
            }
            self.str_slice = try allocator.dupe(u8, self.str_ptr.?.*);
            self.dynamically_allocated = true;
        }
    }

    pub fn deinit(self: *StringWrangler, allocator: std.mem.Allocator) void {
        if (self.initialized and self.dynamically_allocated) {
            allocator.free(self.str_slice.?);
        }
    }
};
```

It's basically the "remember the original pointer address" solution, but with some helpers to make it nicer.

There is still a problem with optional structs that have strings with default values, because the string would keep switching from statically to heap allocated to not existing as the ipuser edits the string / nulls the optional. But I think I'm going to put that in the "too hard basket". The user will have to handle those manually with a buffer.

david-vanderson:
Wow this is a hard problem.  It feels like we're trying to do too much automatically with not enough information.

Maybe we should push reallocating strings into a future problem.  Right now we can show read-only strings, and read-write in-place strings right?

I think this structUI stuff is very useful and don't want to hold you up, so I say let's merge this and punt on heap reallocation for now.

What do you think?

phatchman:
> Wow this is a hard problem. It feels like we're trying to do too much automatically with not enough information.

Basically yes. It's not having enough information and not being able to easily communicate back to the user from a recursive function.

In "summary". 

1) struct_ui can't tell if a string is heap, stack or statically allocated. The user has to tell us.
2) Because of the desired recursive nature of displayStruct, it is doing things on behalf of the user like instantiating optionals and union members when it can. This makes it hard for the user to intervene just for a single field buried in the hierarchy.
3) struct_ui can't easily communicate back to the user what has changed or if allocations were performed. I've looked at several alternatives here
  - Return some sort of list of changes and/or allocations from displayStruct(). The issue here is that you have to come up with some sort of structure that mirrors all of the fields in the struct and children. Then you have optionals and unions, which means that structure is dynamic.
  - Putting something the user controls in between struct_ui and the struct. i.e. the data adapter approach. This is a lot of code and even then, it can't cover all cases because of the point above. struct_ui  can instantiate things without the user's knowledge.
  - Using the field_options to take an allocator or keep track of changed fields. The issue is that I deliberately changed the field_options to be per-type to simplify and remove the recursive following of fields that might never be displayed. So, there is no way to say which Font instance was instantiated in Options or Theme, just that an instance of Font instance was.
   - Making the user always handle read-write strings manually. They can already do this if the simple functionality doesn't work for them.
   - Rebuilding a parallel comptime structure that keeps pointers to fields in the real struct and annotates each string with whether it is heap allocated. This is very difficult to debug at comptime because all errors are on the duplicated struct, not the underlying struct, so error traces will never point to the user's code. This was the one of the annoying problems with the old FieldOptions.
   - Using a callback whenever a string changes and let the user handle the assignment. This is a simpler version of the data adapater approach. But there is going to be a lot of complex logic on the user's side. If they get passed a pointer to an instance of a Font string (when fonts had strings for names) they also have to know that the Font got set to null last time, so this is potentially a brand new static string they haven't ever seen before. So, we'd also need to then add a callback when optionals are created or unions initialized. The user would need to manage a parallel set of state.
   - make displayStruct not recursive. Anytime the user wants to display a field containing a struct, the have to call displayStruct on that field, otherwise it is ignored. This is basically a replacement for the "depth" option. This kills the simplicity of "display struct" and imagine displaying something like Theme and the number of optional child structs. You may as well just call the underlying widget display functions directly in that case and forget displayStruct alltogether.
   - Probably lots of other less successful things I've forgotten.

At this point I've tried every angle I can think of. And while I can make the data adapter stuff work for most cases, as discussed, I think the maintenance burden of that is too high for too little pay off.

> Maybe we should push reallocating strings into a future problem. Right now we can show read-only strings, and read-write in-place strings right?

Yes., you can actually modify any string, as long as it is a slice and not an array, because it just re-assigns the slice pointer to the new buffer. By default, anything that has const u8 elements is displayed as read only and var u8 elements as read-write, but the user can override that. But this means that both types of strings have the same allocation issue if they are modified.

You can in-place edit u8 arrays as strings, but it's not really recommended unless you are prepared to deal with 0 termination in your array. That's why I don't do in-place editing for var strings either. I want the string to grow and shrink as needed and not use null termination.

> 
> I think this structUI stuff is very useful and don't want to hold you up, so I say let's merge this and punt on heap reallocation for now.
> 
> What do you think?

Unless you can see a better way of handling var strings that doesn't require changing what the slice points to, I'd probably merge what we have today and then refine the string wrangler/handler when I do the PR to add some more advanced demos. 

I think the string wrangler/helper isn't too bad. It doesn't impact the rest of the struct_ui code and does work for all of the cases that can feasibly be solved and most of all it is easy to understand how it works. If it ultimately ends up not being useful, deleting it won't have any impact. 

Also, it can potentially be improved in the future with some function that walks the hierarchy each frame and collects all of the strings and changes for you.

phatchman:
See [this example](https://github.com/phatchman/dvui-dev/blob/f4a32affdebed9d242db7f98e2ba5e007dff4a58/src/Examples/struct_ui.zig) for a simple string handling demo where the strings can outlast the lifetime of the window. 

There's a buffer display version and the "StringHelper" dynamically allocated version.

I've not added a new dialog for this, yet, but the code is essentially the same, except that duplicateIfChanged() would be called before window close instead of each frame. Either way works though.

There is also some bug with the deinit / init in the demo that I still need to work out.

david-vanderson:
> Unless you can see a better way of handling var strings that doesn't require changing what the slice points to, I'd probably merge what we have today and then refine the string wrangler/handler when I do the PR to add some more advanced demos.

Nope.  Let's merge this!

I very much appreciate you laying out all the different approaches.

> See [this example](https://github.com/phatchman/dvui-dev/blob/f4a32affdebed9d242db7f98e2ba5e007dff4a58/src/Examples/struct_ui.zig) for a simple string handling demo where the strings can outlast the lifetime of the window.

I like how that is looking, need to spend more time playing with it.  I don't quite have the logic about when to set `str_ptr` vs `str_slice` straight in my head.

> There is also some bug with the deinit / init in the demo that I still need to work out.

One thing that has gotten me a few times is that all the demo code runs twice - once for the button you click on, and again for the real thing, so any globals might be being fought over.  I've been trying to slowly change them to all use dataGet/dataSet to deal with this.

phatchman:
> I like how that is looking, need to spend more time playing with it. I don't quite have the logic about when to set `str_ptr` vs `str_slice` straight in my head.
> 

They can definitely do with better naming!

- str_ptr is a pointer to the slice field and this value shouldn't change for the lifetime of the struct.
- str_slice a copy of the slice pointed to by str_ptr. (i.e. a copy of the .ptr and .len). 
- str_slice is updated each time a duplication is made, so any future changes can be detected.

I compare the current ptr at str_ptr.*.ptr with the str_slice.ptr to see if it has changed and then dup if it has. But I suspect there is a bug due to this line being incomplete.
```            
self.str_slice = try allocator.dupe(u8, self.str_ptr.*);
```
It should also be setting the duped string back to the underlying field (self.ptr.*) to store it back into the struct, not just remembering it internally.


re: Demos running twice. Thanks, I'll check that.

phatchman:
Actually, this seems quite hard to test.  I put the editing of the struct in a floating window for the demo and the realised that's not really the test. It has to be an actual dvui.Window.

As far as I understand, the buffer backing text widget won't be deallocated until deinit() is called on the Window. So, unless you are actually deinting and re-creating the window to display the structs, instead of just not showing the window, the text widget buffer backing the strings will always be valid. Am I right?

Maybe the string duplication really is a niche case?

To put an example in the demo for this seems tricky, I can do it with a floating window and put some comments in about how it's not needed for floating windows I suppose.

david-vanderson:
> As far as I understand, the buffer backing text widget won't be deallocated until deinit() is called on the Window.

The buffer is allocated using dataGet/dataSet right?  That stuff should go away as soon as there is a frame where it's not referenced (the id/name combo is not used in any dataGet or dataSet).  Are you not seeing that?

> Maybe the string duplication really is a niche case?

I think this is probably the case.  Especially for a general-purpose struct viewer.

Given the enormous difficulty we've had so far with memory allocation, I suspect a solution might involve making some tradeoff, like saying "editing slices with structUI is going to leak memory" (so okay for development/one-off stuff).

I very much appreciate the time and effort you've put into this - thank you!

phatchman:
> The buffer is allocated using dataGet/dataSet right? That stuff should go away as soon as there is a frame where it's not referenced (the id/name combo is not used in any dataGet or dataSet). Are you not seeing that?
> 

Yeah, I am. 1) I missed the code that does the freeing and 2) I needed a more elaborate test-case to cause it to fail. 

That means anything that is not always displayed (e.g. anything in an expander) has the same problem, which means all strings will have the problem because I display structs in expanders.

Hmm. OK. I think it needs a better solution then.

PR #529, opened 2025-08-29
updates for zig 0.15.1

david-vanderson:
* only enough to get past the initial dependency issues
* now starting to get the actual code errors

### DIFF
--- a/src/backends/web.zig
+++ b/src/backends/web.zig
@@ -771,9 +771,6 @@ pub fn getNumberOfFilesAvailable(id: dvui.Id) usize {
 
 /// A `std.Io.Writer` wrapper for interacting with the JavaScript console.
 ///
-/// The console writer will work with buffer size of 0, but it will force
-/// a call from Wasm to JavaScript on every write.
-///
 /// The drain function is infallible, meaning that using `catch unreachable`
 /// on the writer functions should be safe (unless the `std.Io.Writer`
 /// implementation can throw an error explicitly)
@@ -784,7 +781,12 @@ pub fn getNumberOfFilesAvailable(id: dvui.Id) usize {
 pub const Console = struct {
     writer: std.Io.Writer,
 
+    /// The minimum recommended buffer size.
+    pub const min_buffer_size: usize = 128;
+
+    /// Assserts that `buffer.len >= Console.min_buffer_size`
     pub fn init(buffer: []u8) Console {
+        std.debug.assert(buffer.len >= Console.min_buffer_size);
         return .{ .writer = .{
             .buffer = buffer,
             .vtable = &.{
--- a/src/backends/web.zig
+++ b/src/backends/web.zig
@@ -28,7 +28,7 @@ const EventTemp = struct {
     float2: f32,
 };
 
-pub var event_temps = std.ArrayList(EventTemp).init(gpa);
+// pub var event_temps = std.ArrayList(EventTemp).init(gpa);
 
 pub const wasm = if (!builtin.is_test) struct {
     pub extern "dvui" fn wasm_about_webgl2() u8;
@@ -140,7 +140,7 @@ pub const wasm = if (!builtin.is_test) struct {
 };
 
 export fn dvui_c_alloc(size: usize) ?*anyopaque {
-    const buffer = gpa.alignedAlloc(u8, 8, size + 8) catch {
+    const buffer = gpa.alignedAlloc(u8, .@"8", size + 8) catch {
         //log.debug("dvui_c_alloc {d} failed", .{size});
         return null;
     };
--- a/src/backends/web.zig
+++ b/src/backends/web.zig
@@ -150,7 +150,7 @@ export fn dvui_c_alloc(size: usize) ?*anyopaque {
 }
 
 pub export fn dvui_c_free(ptr: ?*anyopaque) void {
-    const buffer = @as([*]align(8) u8, @alignCast(@ptrCast(ptr orelse return))) - 8;
+    const buffer = @as([*]align(8) u8, @ptrCast(@alignCast(ptr orelse return))) - 8;
     const len = std.mem.readInt(u64, buffer[0..@sizeOf(u64)], builtin.cpu.arch.endian());
     //log.debug("dvui_c_free {?*} {d}", .{ ptr, len - 8 });
 
--- a/src/backends/web.zig
+++ b/src/backends/web.zig
@@ -778,14 +778,6 @@ comptime {
     }
 }
 
-const WriteError = error{};
-const LogWriter = std.io.Writer(void, WriteError, writeLog);
-
-fn writeLog(_: void, msg: []const u8) WriteError!usize {
-    WebBackend.wasm.wasm_log_write(msg.ptr, msg.len);
-    return msg.len;
-}
-
 pub fn logFn(
     comptime message_level: std.log.Level,
     comptime scope: @Type(.enum_literal),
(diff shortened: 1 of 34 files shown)


RedPhoenixQ:
Thank you for fixing the CI. The error wasn't very helpful (just error code 1) and zig on WSL has some issues interacting with the Windows file system so I could test it locally.

david-vanderson:
Premerge checklist:

- [x] tag release for anyone using zig 0.14.1
- [x] get z2d PR https://github.com/vancluever/z2d/pull/147 merged
- [ ] get zig-lib-svg2tvg PR https://github.com/nat3Github/zig-lib-svg2tvg/pull/6 merged
- [ ] update dvui's build.zig.zon to point to updated zig-lib-svg2tvg

Can update zig-lib-svg2tvg later.

david-vanderson:
> Thank you for fixing the CI. The error wasn't very helpful (just error code 1) and zig on WSL has some issues interacting with the Windows file system so I could test it locally.

Of course!  I was also stumped by the error for a while, but I think we are hitting some error inside zig's native x86_64 backend, so the fix was to specify .use_llvm = true.

I would like to send this error upstream but right now I have no idea how to provide any extra information.  I'll just file an issue for it and come back to it later.

RedPhoenixQ:
For a future change, what do you think of replacing most returned allocated slices (like `Picture.png`, `pngEncode` and perhaps functions like `dvui.clipboardText`) instead take a `std.Io.Writer` (removing the need to take and allocator)? Would allow more control for the user, like directly writing the png to a file etc.

david-vanderson:
> For a future change, what do you think of replacing most returned allocated slices (like `Picture.png`, `pngEncode` and perhaps functions like `dvui.clipboardText`) instead take a `std.Io.Writer` (removing the need to take and allocator)? Would allow more control for the user, like directly writing the png to a file etc.

Yeah that sounds great!  I don't understand the new Io stuff enough yet (especially around when you have to call flush), but it should be better than what we have now.

david-vanderson:
I think this is ready to go, does anyone know of any potential blockers?

If I don't here anything in a day or 2 I'll merge.

src/backends/web.zig
```diff
@@ -801,7 +793,9 @@ pub fn logFn(
     const prefix2 = if (scope == .default) ": " else "(" ++ @tagName(scope) ++ "): ";
     const msg = level_txt ++ prefix2 ++ format ++ "\n";
 
-    (LogWriter{ .context = {} }).print(msg, args) catch return;
+    const buf = std.fmt.allocPrint(gpa, msg, args) catch "dvui log OOM";
+    defer gpa.free(buf);
```
knightpp, line 797: Wouldn't this do something wrong if allocation fails, but gpa freed the static string?

david-vanderson: Whoops - yeah - I'll fix it.  Thanks!

RedPhoenixQ: Working on the console Writer rn

RedPhoenixQ:
I was just fixing the same utf8 error! Was using the error logs to test my wasm console printing

david-vanderson:
> I was just fixing the same utf8 error! Was using the error logs to test my wasm console printing

Haha whoops.  I can't believe I made that error in the first place, and then that it compiled.  Thanks goodness for having the example in the debugging part of the demo!

RedPhoenixQ:
Ah, 82d39ff8556d9a4431bb140030e6ae79ddff4bc9 worked because of the missing `.{...}` around the `fmtUtf8()`. It took is as `fmt("{s}", .{ .data = "OUR STRING AS FIRST ARG" })`, which is technically correct

david-vanderson:
The best kind of correct!  (in this case worst)

RedPhoenixQ:
0cfbc7d5a2687487da0db5384b7c510a7b7ebf4e now adds a `std.Io.Writer` wrapper for the javascript console. As you call `print` or `write` the writers buffer gets filled up, without calling the javascript `wasm_log_write` function for each step. When the buffer is full, the `drain` function gets called to empty the buffer to it's final location (and it can write other large slices of data at once with this function directly). I've made it a bit overkill trying to use the buffer as much as possible, avoiding calling into the javascript side. Finally you call `flushAtLevel` which first calls `flush` on the writer pushing all the buffered data to the javascript side, and then flushes the completed string to the console.

RedPhoenixQ:
3d11dd7b08357b63127bd6bbfe4159b2617879f9 makes the Console writer much simpler. The interface allows the `drain` function to only write part of the data, so that's what it does now. It filles the buffer, flushes out the buffer, then return the amount it could consume (which could even be 0 if the buffer was already filled). `std.Io.Writer.writeSplatAll` will handle calling out drain function in a loop as needed to write all the data eventually.

phatchman:
Just built this and have two issues so far:

1) Not all of the grid demos are displaying correctly. The Layouts and Data one goes into a refresh loop.
     - I'll see if I can work out what's going on with this.
     - Edit: This is unrelated. It broken on main as well. #536
  
2) I'm getting build errors on Windows. Example of some messages below.
    - Dx11 won't build because of the callconv not being updated in the zigwin32 dependency, so I guess those are known errors?
    - And the fmt.allocPrintZ error looks to be the same as above, just the package not being fully updated yet?
    - The only other issue is the wasm_log_write error when building zig web-test.

```
examples\web-test.zig:25:20: error: struct 'web.wasm__struct_18319' has no member named 'wasm_log_write'
    WebBackend.wasm.wasm_log_write(buf.ptr, buf.len);
    ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
```
```
C:\Users\phatc\AppData\Local\zig\p\zigwin32-25.0.28-preview-AAAAAHsJ-wPA4nREAzT_OOkF6gLrornNuHqREfHDADoS\win32\graphics\direct3d.zig:567:45: error: root source file struct 'os.windows' has no member named 'WINAPI'
        ) callconv(@import("std").os.windows.WINAPI) ?*anyopaque,
                   ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
C:\Users\phatc\AppData\Roaming\Code\User\globalStorage\ziglang.vscode-zig\zig\aarch64-windows-0.15.1\lib\std\os\windows.zig:1:1: note: struct declared here
//! This file contains thin wrappers around Windows-specific APIs, with these
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
referenced by:
    VTable: C:\Users\phatc\AppData\Local\zig\p\zigwin32-25.0.28-preview-AAAAAHsJ-wPA4nREAzT_OOkF6gLrornNuHqREfHDADoS\win32\graphics\direct3d.zig:563:31
    win32.graphics.direct3d.ID3DBlob: C:\Users\phatc\AppData\Local\zig\p\zigwin32-25.0.28-preview-AAAAAHsJ-wPA4nREAzT_OOkF6gLrornNuHqREfHDADoS\win32\graphics\direct3d.zig:572:20
    8 reference(s) hidden; use '-freference-trace=10' to see all references
```
```
C:\Users\phatc\AppData\Local\zig\p\zigwin32-25.0.28-preview-AAAAAHsJ-wPA4nREAzT_OOkF6gLrornNuHqREfHDADoS\win32\zig.zig:268:56: error: root source file struct 'fmt' has no member named 'allocPrintZ'
                const msg_z: [:0]const u8 = if (std.fmt.allocPrintZ(
                                                ~~~~~~~^~~~~~~~~~~~
C:\Users\phatc\AppData\Roaming\Code\User\globalStorage\ziglang.vscode-zig\zig\aarch64-windows-0.15.1\lib\std\fmt.zig:1:1: note: struct declared here
//! String formatting and parsing.
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

RedPhoenixQ:
The win32 dependency issue for dx11 is know and should be the only blocker for dx11 to build.

> * The only other issue is the wasm_log_write error when building zig web-test.

That was an oversight. We have test stubbs for all the external wasm functions and the name wasn't updated there, fixed in  3d11dd7b08357b63127bd6bbfe4159b2617879f9.

PR #537, opened 2025-09-01
Adapt PNG encoder to `std.Io.Writer`

RedPhoenixQ:
Depends on #529

This removes the need to buffer the entire png in memory when writing to a file. The plot example has been updated accordingly.

Using `std.Io.Writer` for `Backend.textureReadTarget` isn't really an improvement because many of the underlying graphics apis still need a complete buffer to write in to, which would mean an additional memcpy. The current solution provides optimal buffering with regards to the graphics api compatibility.

### DIFF
--- a/src/Examples/plots.zig
+++ b/src/Examples/plots.zig
@@ -75,31 +75,38 @@ pub fn plots() void {
         p.stop();
         defer p.deinit();
 
-        const arena = dvui.currentWindow().lifo();
+        if (dvui.wasm) blk: {
+            var writer = std.Io.Writer.Allocating.init(dvui.currentWindow().arena());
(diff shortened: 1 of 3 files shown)


david-vanderson:
Very cool!

RedPhoenixQ:
@david-vanderson With #529 merged, can I merge this too?

david-vanderson:
> @david-vanderson With #529 merged, can I merge this too?

Yes sorry!  I don't know why it's listing all the commits that were part of #529 as well, but I guess we can always fix it if it merges badly.  Go for it!

RedPhoenixQ:
One rebase later and there should be no issues

PR #542, opened 2025-09-06
Replace stateful showCursor with Cursor.hidden

### LINKED ISSUE #540: How to handle stateful mouse visibility?
Hello!

This may just be an issue with my understanding of how to handle this, but I'm hitting an issue that I don't quite know how to fix.

Since cursor visibility is stateful, I had tried just setting the cursor visibility to true every editor tick, and then run my editor code which has an opportunity to set the visibility back to false if need be, but that each frame the mouse defaults to visible. This made things easy to handle for multiple canvases as the mouse may exit a canvas and need to be set back to visible.

The issue however is that on Windows it seems just setting the visibility makes the mouse "flicker" visible even when its not supposed to be. I think its because setting it to visible immediately sets it visible, and then setting it back to invisible shows the flicker each frame as the mouse moves. 

This is not the case on macOS, I'm not sure on Linux.

TLDR: Setting mouse visibility to `true` and then back to `false` in a single frame on Windows causes the mouse to "flicker", which makes it hard to combine the statefulness of the mouse visibility with general immediate mode code.

RedPhoenixQ:
Closes #540

This requires that most backends cache the previous state of the cursor, which most already did. Most the backends `showCursor` functions still remain, but they're no longer intended to be called directly and is handled by the `setCursor` system. The stateful-ness of each backend is now contained to the backend itself and the user just needs to call `dvui.setCursor`, with `dvui.showCursor` being removed, and the correct state will be presented at the end of the frame.

### DIFF
--- a/src/Backend.zig
+++ b/src/Backend.zig
@@ -135,13 +135,6 @@ pub fn preferredColorScheme(self: Backend) ?dvui.enums.ColorScheme {
     return self.impl.preferredColorScheme();
 }
 
-/// Show/hide the cursor.
-///
-/// Returns the previous state of the cursor, `true` meaning shown
-pub fn cursorShow(self: Backend, value: ?bool) GenericError!bool {
-    return self.impl.cursorShow(value);
-}
-
 /// Called by `dvui.refresh` when it is called from a background
 /// thread.  Used to wake up the gui thread.  It only has effect if you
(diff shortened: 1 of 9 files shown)


david-vanderson:
Awesome - merge when ready.  Thank you!

PR #543, opened 2025-09-07
work around ziglang/zig#25180 to allow use of self-hosted backends

silversquirl:


david-vanderson:
This makes sense to me - @RedPhoenixQ do you agree?

RedPhoenixQ:
Yep looks good. I don't know if there is a need to change it back? It's a one-time call so it shouldn't matter either way

RedPhoenixQ:
Seems to be some issues with the tests. Can't reproduce on Windows. See [this comment](https://github.com/david-vanderson/dvui/pull/544#issuecomment-3263961359)

david-vanderson:
I fixed the two issues that came up in the CI:
* actual bug in findUtf8Start (when used on undefined)
* hashing the bytes of null in Options.hash

I think both are related to the byte representation of undefined and null changing between the llvm and zig native backends.  Which makes sense.

david-vanderson:
Thank you!

PR #544, opened 2025-09-07
Don't install test artifacts

silversquirl:
there's no real reason to, and it just clutters up the `zig-out` directory.
even worse, because the test artifacts for examples were named the same thing as the example itself, they overwrote each other, meaning it was unpredictable which one you'd get in eg. `zig-out/bin/sdl3-app` - quite confusing if you simply clone dvui, run `zig build`, and expect to be able to run the examples from `zig-out/bin`!

this PR stops installing the test artifacts, and also renames them to `test-thing` rather than simply `thing`, to avoid conflicts

depends on #543

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -510,7 +510,7 @@ const DvuiModuleOptions = struct {
     build_options: *std.Build.Step.Options,
 
     fn addChecks(self: *const @This(), mod: *std.Build.Module, name: []const u8) void {
-        const tests = self.b.addTest(.{ .root_module = mod, .name = name, .filters = self.test_filters, .use_lld = self.use_lld, .use_llvm = true });
+        const tests = self.b.addTest(.{ .root_module = mod, .name = name, .filters = self.test_filters, .use_lld = self.use_lld });
         self.b.installArtifact(tests); // Compile check on default install step
         if (self.check_step) |step| {
             step.dependOn(&tests.step);
@@ -523,7 +523,6 @@ const DvuiModuleOptions = struct {
                 .name = name,
                 .filters = self.test_filters,
                 .use_lld = self.use_lld,
-                .use_llvm = true,
             });
             step.dependOn(&self.b.addRunArtifact(tests).step);
         }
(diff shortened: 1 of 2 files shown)


RedPhoenixQ (dismissed):
Looks good. There is indeed no reason to install tests. Thank you!

RedPhoenixQ:
There seems to be some issue with the CI. It never finishes and there are some arena alloc errors

RedPhoenixQ:
> There seems to be some issue with the CI. It never finishes and there are some arena alloc errors

Can't reproduce locally. WSL and zig 0.15 don't cooperate currently so I can't test on Linux. @david-vanderson could you run `zig build test -Dbackend=testing` on Linux on this PR and see if you get the same errors?

david-vanderson:
> > There seems to be some issue with the CI. It never finishes and there are some arena alloc errors
> 
> Can't reproduce locally. WSL and zig 0.15 don't cooperate currently so I can't test on Linux. @david-vanderson could you run `zig build test -Dbackend=testing` on Linux on this PR and see if you get the same errors?

Yes I am able to reproduce those, and unfortunately they go away if we use the llvm backend.  So not directly this change, but rather related to #543 

I don't know much else right now.

david-vanderson:
@silversquirl I'm looking to merge this and deal with the zig backend stuff separately.

One thing I noticed is that before we called `b.addTest()` twice - once for `zig build check` and once for `zig build test`.  This was to make sure `zig build check` didn't try to link/install anything.  See https://ziggit.dev/t/zig-build-check-and-tests/9807 for me figuring this out.

Did you change that on purpose because you know something changed?

phatchman:
> Can't reproduce locally. WSL and zig 0.15 don't cooperate currently so I can't test on Linux. @david-vanderson could you run `zig build test -Dbackend=testing` on Linux on this PR and see if you get the same errors?

I can build under WSL as long as I'm compiling on a linux fs, rather than a mounted Windows fs.

david-vanderson:
@silversquirl ping for this:

> @silversquirl I'm looking to merge this and deal with the zig backend stuff separately.
> 
> One thing I noticed is that before we called `b.addTest()` twice - once for `zig build check` and once for `zig build test`. This was to make sure `zig build check` didn't try to link/install anything. See https://ziggit.dev/t/zig-build-check-and-tests/9807 for me figuring this out.
> 
> Did you change that on purpose because you know something changed?

david-vanderson:
This functionality got merged as part of #606 - please reopen if there is anything else needed.  Thank you!

phatchman:
Yep. Perfect. And works with the in-progress changes I have to correctly reads out the labels of each option.

Issue #545, opened 2025-09-07
memory allocation and freeing widgets


david-vanderson:
@silversquirl gave this patch on discord for better debugging.

[safety_check_alloc.patch](https://github.com/user-attachments/files/22198537/safety_check_alloc.patch)

This issue is for me to work that patch into dvui, and see if that gives any insight into the problems we are seeing running the tests with zig 0.15.1 native x86_64 backend on linux.

We also discussed changing how widgets free their memory.  Right now they always free even if they weren't heap allocated, and we should probably move towards widgets knowing whether they should free or not.  Need to figure that out.

RedPhoenixQ:
```patch
diff --git a/src/shrinking_arena_allocator.zig b/src/shrinking_arena_allocator.zig
index 2d49e7bb..ba757515 100644
--- a/src/shrinking_arena_allocator.zig
+++ b/src/shrinking_arena_allocator.zig
@@ -262,7 +262,7 @@ pub fn ShrinkingArenaAllocator(comptime opts: InitOptions) type {

                 if (has_never_free and !self.never_free.arena.allocator().rawResize(in_memory, alignment, new_len, ret_addr)) {
                     // reset the size in the normal arena, which should always succeed       
-                    std.debug.assert(self.arena.allocator().rawResize(memory, alignment, memory.len, ret_addr));
+                    std.debug.assert((self.arena.allocator().rawResize(memory.ptr[0..new_len], alignment, memory.len, ret_addr));
                     // log.debug("A false negative resize/remap failure was introduced", .{});
                     self.never_free.failed_remap = true;
                     return false;
```

Adding this to the patch removes all errors about frees with different sizes. Only applies in debug where we double allocate to "never free".

The issue of multiple calls to free comes from our handling of alignment. We make multiple attempts to free with different alignments. This can probably be changed to store the alignment of the last item freed to know where the allocation bellow should've been aligned to. Will see if I can get that to work. Just skipping the multiple attempt loop in with the debug allocator patch removes all errors.

david-vanderson:
Awesome will add that to the patch.  Thank you!

RedPhoenixQ:
> Awesome will add that to the patch. Thank you!

I'm working on a branch with this fix and the alignment fix

david-vanderson:
> > Awesome will add that to the patch. Thank you!
> 
> I'm working on a branch with this fix and the alignment fix

Amazing!

RedPhoenixQ:
[...]

PR #546, opened 2025-09-07
Store if widget was allocated to conditionally free them

RedPhoenixQ:
Relates to #545

Some widgets that don't have or for some reason can't use `WidgetData` to store this information does so in their `InitOptions`.

Reorder widget is one of the exceptions where it has a `WidgetData` but still uses `InitOptions` because the `WidgetData` gets overwritten at multiple points.


src/dvui.zig
```diff
@@ -4539,7 +4541,7 @@ pub fn animate(src: std.builtin.SourceLocation, init_opts: AnimateWidget.InitOpt
 ///
 /// Only valid between `Window.begin`and `Window.end`.
 pub fn dropdown(src: std.builtin.SourceLocation, entries: []const []const u8, choice: *usize, opts: Options) bool {
-    var dd = dvui.DropdownWidget.init(src, .{ .selected_index = choice.*, .label = entries[choice.*] }, opts);
+    var dd = dvui.DropdownWidget.init(src, .{ .selected_index = choice.*, .label = entries[choice.*], .was_allocated_on_widget_stack = true }, opts);
```
david-vanderson: This looks wrong to me - we should only set this if we called widgetAlloc right?

david-vanderson (commented):
Looks great - please double-check that one spot and then commit when ready.  Thank you very much!

src/dvui.zig
RedPhoenixQ: Yeah, I just caught that with the patch from #545. Fixed in the most recent force push

PR #548, opened 2025-09-08
ShrinkingArenaAllocator:  Handle alignment padding when freeing

RedPhoenixQ:
Relates to #545

This fixes the issue of varying alignments making calls to free fail. We can now guarantee to only need  one call to free.

Because of svg2tvg using a 16 byte aligned buffer we are forced to set the maximum to 16, unless we create a wrapper arena allocator everywhere we call into the library (which isn't unreasonable). The max alignment can be easily changed in that case.

### DIFF
--- a/src/shrinking_arena_allocator.zig
+++ b/src/shrinking_arena_allocator.zig
@@ -170,7 +170,10 @@ pub fn ShrinkingArenaAllocator(comptime opts: InitOptions) type {
 
             // Extend the allocated length to the previous freed allocation
             // to account for padding between allocations
-            const index_adjustment = if (self.last_freed_ptr) |ptr| ptr -| (@intFromPtr(memory.ptr) + memory.len) else 0;
+            const index_adjustment = if (self.last_freed_ptr) |ptr|
+                @min(ptr -| (@intFromPtr(memory.ptr) + memory.len), Alignment.@"32".toByteUnits())
+            else
+                0;
             if (self.arena.state.end_index > index_adjustment) {
                 self.arena.state.end_index -= index_adjustment;
             }
--- a/src/shrinking_arena_allocator.zig
+++ b/src/shrinking_arena_allocator.zig
@@ -46,6 +46,7 @@ pub fn ShrinkingArenaAllocator(comptime opts: InitOptions) type {
         arena: ArenaAllocator,
         peak_usage: usize = 0,
         current_usage: usize = 0,
+        last_freed_ptr: ?usize = null,
 
         never_free: if (has_never_free) NeverFree else void = undefined,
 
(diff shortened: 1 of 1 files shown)


RedPhoenixQ:
This is not done. Resize can request a length that is not aligned, which messes up the current logic. Might get to fixing that tomorrow

RedPhoenixQ:
The arena now handles padding by adjusting the end_index based on the difference from the last call to free. If anything was allocated since the last free, no adjustment is applied.

Perhaps the adjustment needs to be limited to some smaller value, currently it's unlimited. What could happen, when not using it as a lifo allocator is the following:

- Allocate three things in a row
- Free the last item
- Then free the first item, skipping the second.

In the scenario above the second item would become invalid, and potentially overwritten in future allocations, because then end_index will be adjusted to free the first item. If we limit the alignment adjustment to 32 then this could still happen if you allocate and free items whose size are less than that. 

This currently fails on the widget stack for this exact reason and is fixed by #546, which should be merged at the same time

david-vanderson:
This is good - please merge when ready.

I think in addition we should take a second to reassess our memory handling strategy.  You've put in a great deal of work and we've both learned a lot, and I am greatly appreciative of that!

It's also pushed us to tighten up our memory game, fixing all the use-after-free widget issues.  And #546 is another step forward.

Given what we know now, do we still think the benefits of doing custom memory management are worth the costs?

The benefits are (am I missing any?):
* lower peak memory usage (do you think this is still less than 1MB in savings?)
* shrinking the arena after a peak
* custom logging
* debugging aids

I'm particularly nervous about myself being able to troubleshoot any issues.

What are your thoughts here?

RedPhoenixQ:
> The benefits are (am I missing any?):
>
> * lower peak memory usage (do you think this is still less than 1MB in savings?)
> * shrinking the arena after a peak
> * custom logging
> * debugging aids
> 
> I'm particularly nervous about myself being able to troubleshoot any issues.
> 
> What are your thoughts here?

You might be right. The benefits are basically what you describe, it's just a shame that the stack like nature of the widgets can't result in stack like memory usage. The only difference is `On` vs `Olog(n)` (where n is the number of widgets/allocations) scaling when freeing as a stack vs not (would still be in the MB for larger programs I believe). Using an arena with a lifetime of the entire frame is a perfect use case for an arena an simple memory management model.

We could absolutely go back to `std.heap.ArenaAllocator` for all use cases, but it might be worth keeping the separate allocators as many frees will still succeed from the `lifo` arena and the work of separating them is already done. Otherwise, allocating a 1MB string for printing and then rendering it would almost guarantee other allocations above it and not being able to free. The multiple allocators removes most of that limitation. 

I think if we move `Window.render_commands` so use the `gpa` allocator, the `arena` would be in full control of the user and `lifo` used for rendering and text formatting (by label and textLayout).

TLDR; The separate arenas are good, but they can be `std.heap.ArenaAllocator` with no requirement to free

david-vanderson:
> I think if we move `Window.render_commands` so use the `gpa` allocator, the `arena` would be in full control of the user and `lifo` used for rendering and text formatting (by label and textLayout).
> 
> TLDR; The separate arenas are good, but they can be `std.heap.ArenaAllocator` with no requirement to free

Got it - thanks very much!

PR #553, opened 2025-09-11
struct_ui - String handling revision

phatchman:
## Description
With the previous string handling, if a struct was not displayed (i.e. expander was collapsed), the backing string memory would be freed and potentially corrupted.

With this change a global string map is created for all struct_ui string fields. This map keeps track of all memory allocations, and will free any previously heap allocated memory for that string before allocating more.

1. The user can indicate whether the original string was heap-allocated via the TextFieldOptions. 
1. The user can clean up any leftover strings by using struct_ui.deinit(gpa). Note this requires an allocator to be passed as I expect it to be called outside of win.begin()/win.end().
1. By default currentWindow().gpa is used , but an alternate allocator can be supplied for all struct_ui string allocations via the global struct_ui.string_allocator variable.

## Limitations
- Does not currently provide a way for the user to know if a field has been modified. This could be implemented via a lookup on string_map if required.
- A single allocator must be used for all strings. If a different allocator is required, the user will need to call the textFieldWidget directly and pass in an allocator or buffer.
- All strings are freed at once in deinit(). If the user needs something more, they will need to implement that themselves via the StringHelper pattern or similar.

## Discussion

This change essentially moves the StringHelper logic into textFieldWidget(). While it would be possible to make the string_map per-type, I suspect a global one is enough for most usages and it makes cleanup simple with a single deinit().

I've added a small test to the struct_ui example that provides editing in a pop-up window. Just editing wasn't enough to trigger the issue, you would have to edit, go to another demo and then come back. 

My initial solution was for the TextFieldOptions to keep track of the allocations. But that means the FieldOptions have to keep state and be passed around as pointers. But that would have given more flexibility to use different allocators per type.

Implements #488


phatchman:
Should be good to go. String handling looks good and added some new demos.

But I have run into an issue with the app.zig examples, which I'm not sure how to solve properly.

The allocator for the apps is created in the backends and there doesn't seem to be a way to set that or access it outside of dvui.currentWindow().

I need the allocator that was used in Window.init to do the struct_ui cleanup in AppDeinit(). For now, I'm storing a copy from inside window begin/end to use for the deinit().
```
pub fn AppFrame() !dvui.App.Result {
    if (gpa == null) {
        gpa = dvui.currentWindow().gpa;
    }
    return frame();
}
```

Otherwise, I've not found any issues with this approach. The string handling is totally transparent outside of the need to call struct_ui.deinit(). It does potentially waste some memory for strings that are no longer used, but I can't see that being a real-world concern.

I think this is good enough now to remove the experimental label.

david-vanderson:
This is fantastic!  Merge when ready.

2 followups:
* maybe we should have `dvui.Window.deinit()` always call `struct_ui.deinit()`?  We could say (at least for now) that struct_ui always uses Window.gpa and calls deinit, and then if someone needs to use a different allocator, we can figure that out?
* I'm a bit confused by the dynamic_strings and buffered_strings popups in the example.  Is raw_buffer only touched in the buffered_strings popup?  If so can we remove it from the dynamic_strings part?

phatchman:
> 
> * maybe we should have `dvui.Window.deinit()` always call `struct_ui.deinit()`?  We could say (at least for now) that struct_ui always uses Window.gpa and calls deinit, and then if someone needs to use a different allocator, we can figure that out?

The one potential downside is if anyone needs the string to live beyond Window.deinit(). Is there any realistic way you can see that would happen? Even with an on-top implementations, I assume it would just be showing and hiding the window, rather than doing init/deinit?

> * I'm a bit confused by the dynamic_strings and buffered_strings popups in the example.  Is raw_buffer only touched in the buffered_strings popup?  If so can we remove it from the dynamic_strings part?

Yeah. I was trying to show that one re-used the buffer and the other didn't. Removing it does the same thing.

david-vanderson:
> > * maybe we should have `dvui.Window.deinit()` always call `struct_ui.deinit()`?  We could say (at least for now) that struct_ui always uses Window.gpa and calls deinit, and then if someone needs to use a different allocator, we can figure that out?
> 
> The one potential downside is if anyone needs the string to live beyond Window.deinit(). Is there any realistic way you can see that would happen? Even with an on-top implementations, I assume it would just be showing and hiding the window, rather than doing init/deinit?

I can't imagine that happening - exactly like you say.  I think we should assume this will never happen and deal with it if/when it does.

> 
> > * I'm a bit confused by the dynamic_strings and buffered_strings popups in the example.  Is raw_buffer only touched in the buffered_strings popup?  If so can we remove it from the dynamic_strings part?
> 
> Yeah. I was trying to show that one re-used the buffer and the other didn't. Removing it does the same thing.

Okay makes sense!

phatchman:
Note for future improvements:

It should really keep track of whether each string was originally heap allocated or not and only free strings that weren't originally heap allocated. i.e. we will assume the user will free these strings as they originally allocated them.

The work-around is for the user to remove the pointer to the string from the struct_ui.string_map prior to struct_ui.deinit() being called. Given there is an easy work-around, will leave this until/unless it becomes an issue.

david-vanderson:
Thank you very much!!

PR #554, opened 2025-09-11
WidgetData: Only use fade if scale is 1

foxnne:
Hi! 

Please let me know if this is incorrect, but I wanted to take a look at using fade only if its a low-dpi screen, as this fixes the border showing 1 px when it shouldn't, like we spoke about a while back on the tree widget. 

Thanks!


david-vanderson:
Good idea - let's try it!

PR #555, opened 2025-09-11
JPG encoder

### LINKED ISSUE #551: encoding jpegs
@foxnne would like to encode images to jpg - it's in stb_image_write already, so we just need to expose it.

We should see if we can do the same resolution thing we do for PNGs.

RedPhoenixQ:
Closes #551

This mostly mirrors the `PNGEncoder`, but has `init` functions because it was cleaner when there were more options to configure (both quality and density). The format is also simpler so a simpler strategy for handling the density overwriting was implemented that just waits for all bytes to be in the buffer and then replaces the density values.

`dvui.Picture` also gains the ability to output a jpg.

### DIFF
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -7552,13 +7552,18 @@ pub const JPGEncoder = struct {
     /// Bytes remaining to fill the buffer with the header
     ///
     /// Set to 0 to skip custom density header
-    bytes_remaining_of_jfif_app0_segment: u32 = jfif_app0_segment,
+    bytes_remaining_of_jfif_app0_segment: u32 = start_of_jfif_file,
 
-    pub const min_buffer_size = jfif_app0_segment;
+    pub const min_buffer_size = start_of_jfif_file;
     // https://en.wikipedia.org/wiki/JPEG_File_Interchange_Format#JFIF_APP0_marker_segment
     // Assumes stbi jpg will always return image files starting with SOI (FF D8) followed by JFIF-APP0 (FF E0 ...)
     /// The number of bytes from the start of a JFIF file to the end of the density values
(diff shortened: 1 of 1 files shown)


foxnne:
@RedPhoenixQ thank you so much for this!

RedPhoenixQ:
I think the new `std.Io.Writer` is great! Fun to solve how to make good use of the buffering capabilities and knowing that it can optimally write directly to file or similar if you want.

david-vanderson:
This is excellent!

I'm having trouble testing it on my retina mac (see screenshot).  The preview shows twice as big as it should be, and the preview app says:
X Density: 36,864
Y Density: 1

Screenshot 2025-09-11 at 6 32 22 PM

I haven't looked into it in detail, but does this mean anything to you?

RedPhoenixQ:
> This is excellent!
> 
> I'm having trouble testing it on my retina mac (see screenshot). The preview shows twice as big as it should be, and the preview app says: X Density: 36,864 Y Density: 1
> 
> Screenshot 2025-09-11 at 6 32 22 PM
> I haven't looked into it in detail, but does this mean anything to you?

I was version 1.2 in all my testing so I made an assumption stbi would alway output the same order of bytes. Seams like that wasn't right and in overwriting the wrong bytes.

Gonna have to do some more proper parsing looking for the segment with the density values.

RedPhoenixQ:
@david-vanderson I pushed a fix, turns out the offset was 2 off and I had  the wrong endianness. The jpg I was testing on must have been stale when I checked the results, file explorer caching or something... Could you double check that it works now?

david-vanderson:
Looks great!  Merge when ready.  Thank you!!

Screenshot 2025-09-13 at 10 11 14 AM

PR #558, opened 2025-09-14
minor fixes to saveImage() and struct_ui

phatchman:
- Fix issue with -generate-images=true failing with file not found 
- struct_ui now uses an AutoHashMap for strings instead of the array version


RedPhoenixQ:
Thank you for the testing fix!

PR #561, opened 2025-09-16
Example: reorder_tree: Persist open branch state when tree is not pro…

### LINKED ISSUE #560: [Tree Widget] Remember which branches were open
Hi! 

I wanted to go ahead and create an issue, though I'm about to start investigating. If anyone has some info on what would be best for doing this, please let me know! 

This is annoying currently, but I punted on this when I was writing the original tree widget. I'm not sure how to go about it in a "generic" way, I will try first implementing in my own project by trying to save the file paths of expanded items. I'm not sure of the feasibility of making it work generically as the tree could drastically change from one frame to the next. 

The naive approach of just grabbing a unique id prior to recursively iterating, and setting the file path as data on that id seems not to work for some reason, investigating now.

foxnne:
…cessed

Please let me know if you'd rather go any other way on this, for now it just felt most appropriate that if an application wishes to persist that data, that the application is responsible for storing that state and passing it along to `expanded` in the branch's `init_options`. 

Since the tree is largely very generic right now, I think trying to wrap in persistent state into the widget just complicates things where-as handling it externally is quite simple.

Please let me know of any changes you'd like to make or any additional comments to make things more clear!

Closes #560.

### DIFF
--- a/src/Examples/reorder_tree.zig
+++ b/src/Examples/reorder_tree.zig
@@ -46,6 +46,10 @@ const g_advanced = struct {
 var g_cross_drag_from: ?enum { simple, advanced } = null;
 var g_cross_drag_item: ?usize = null;
 
+var reorder_tree_buffer: [2048]u8 = undefined;
+var reorder_tree_fba = std.heap.FixedBufferAllocator.init(&reorder_tree_buffer);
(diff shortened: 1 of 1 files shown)


david-vanderson:

PR #562, opened 2025-09-16
TreeWidget: Add `init_option`s to branch, adding a vertical ov…

foxnne:
…ershoot animation

This is just for fun, but I think it makes the tree widget feel interactive and fun, and replays the animation when the tree resets. I'd love to include this in my app if it's an acceptable change. 

Edit: Wanted to go ahead and make this easy to customize, so now it accepts a duration and easing for the expansion animation


david-vanderson:
Very cool!  Merge when ready.  Thank you!

PR #563, opened 2025-09-16
textLayout partial processing

david-vanderson:
Gives textLayout the ability to only process what it needs for display.  Tentative api:

```zig
    var tl = dvui.textLayout(@src(), .{ .break_lines = false }, .{ .expand = .both });
    if (tl.bytesNeeded()) |bn| {
        tl.addText(lorem[bn.start..bn.end], .{});
    } else {
        tl.addText(lorem, .{});
    }
```

Notes:
* need to deal with copy()
* need to do something if we fail to allocate for new ByteHeights
* need a way for the app to invalidate (tell textLayout the text has changed)
* need to invalidate automatically if our width changes and we are break_lines = true
  * need to test when break_lines = true
* need to test with touch

During testing I ran into our `max_float_safe` which is the max height of any widget.  Meaning you can load enough text to where you can't scroll to the bottom.

Maybe it's time for dvui to upgrade all the sizing stuff from f32 to f64?

### DIFF
--- a/src/widgets/TextLayoutWidget.zig
+++ b/src/widgets/TextLayoutWidget.zig
@@ -845,6 +845,11 @@ fn selMoveText(self: *TextLayoutWidget, txt: []const u8, start_idx: usize) void 
             }
 
             while (self.cursor_seen and wlr.count > 0) {
+                // do this first, so if we break out of the loop but never see
+                // more text we still scroll to cursor
+                self.scroll_to_cursor_next_frame = true;
+                dvui.refresh(null, @src(), self.data().id);
+
                 switch (wlr.scratch_kind) {
                     .punc => {
                         // skipping over punc
@@ -871,9 +876,6 @@ fn selMoveText(self: *TextLayoutWidget, txt: []const u8, start_idx: usize) void 
                         }
                     },
                 }
-
-                self.scroll_to_cursor_next_frame = true;
-                dvui.refresh(null, @src(), self.data().id);
             }
         },
     }
@@ -975,12 +977,6 @@ fn cursorSeen(self: *TextLayoutWidget) void {
                 dvui.dataSet(null, self.data().id, "_sel_move_cursor_updown_select", cud.select);
                 dvui.refresh(null, @src(), self.data().id);
 
-                // scroll up/down to where we want the cursor, don't need
-                // overscroll because we are staying within the current text
-                dvui.scrollTo(.{
-                    .screen_rect = self.screenRectScale(cr_new.outset(self.data().options.paddingGet())).r,
-                });
-
                 // even though we scrolled to where we thought the cursor would
                 // be, we might have moved up from a long line to a short one
                 // and need to scroll horizontally
@@ -1033,9 +1029,42 @@ pub fn bytesNeeded(self: *TextLayoutWidget, edit_start: usize, edit_end: usize, 
     const Context = struct { height: f32, byte: usize };
     var context: Context = .{ .height = vr.y, .byte = edit_start };
     var sel_end: usize = edit_end;
+    var end_height = vr.y + vr.h;
+
     if (self.copy_sel) |sel| {
-        context.byte = sel.start;
-        sel_end = sel.end;
+        context.byte = @min(context.byte, sel.start);
+        sel_end = @max(sel_end, sel.end);
+    }
+
+    var include_cursor = self.scroll_to_cursor;
+
+    // if we are moving the cursor, need to process the text around where we are moving it
+    switch (self.sel_move) {
+        .none => {},
+        .mouse => {}, // all in visible region
+        .expand_pt => |*ep| {
+            switch (ep.which) {
+                .word, .line => {}, // all in visible region
+                .home, .end => include_cursor = true,
+            }
+        },
+        .char_left_right => include_cursor = true,
+        .cursor_updown => |*cud| {
+            if (cud.pt) |p| {
+                // found cursor last frame, need to include p this frame
+                context.height = @min(context.height, p.y);
+                end_height = @max(end_height, p.y);
+            } else {
+                // we are looking for the cursor to move from
+                include_cursor = true;
+            }
+        },
+        .word_left_right => include_cursor = true,
+    }
+
+    if (include_cursor) {
+        context.byte = @min(context.byte, self.selection.cursor);
+        sel_end = @max(sel_end, self.selection.cursor);
     }
 
     // binary search for the start
@@ -1059,10 +1088,24 @@ pub fn bytesNeeded(self: *TextLayoutWidget, edit_start: usize, edit_end: usize, 
         self.insert_pt.y = startBH.height;
         self.bytes_seen = start_byte;
 
-        if (!self.cursor_seen and (self.selection.cursor < self.bytes_seen)) {
+        if (!include_cursor and (self.selection.cursor < self.bytes_seen)) {
+            std.debug.assert(self.cursor_seen == false);
             self.cursor_rect = Rect{ .x = self.insert_pt.x, .y = self.insert_pt.y, .w = 1, .h = 10 };
-            self.scroll_to_cursor = false;
             self.cursorSeen();
+        }
+
+        switch (self.sel_move) {
+            .word_left_right => |*wlr| {
+                if (wlr.count < 0) {
+                    // update default so that if someone does tons of word left in
+                    // the same frame (so they move to before we started processing
+                    // text, they will only go back to this index (instead of 0)
+                    for (&wlr.word_start_idx) |*i| {
+                        i.* = start_byte;
+                    }
+                }
+            },
+            else => {},
         }
 
         //std.debug.print("setting min height to {d}\n", .{self.insert_pt.y});
@@ -1077,7 +1120,7 @@ pub fn bytesNeeded(self: *TextLayoutWidget, edit_start: usize, edit_end: usize, 
 
     // linear scan for the end (but not the final)
     for (self.byte_heights[first_past_height .. self.byte_heights.len - 1], first_past_height..) |bh, i| {
-        if (bh.height >= (vr.y + vr.h) and bh.byte > sel_end) {
+        if (bh.height >= end_height and bh.byte > sel_end) {
             //std.debug.print("found end {d} {d} bh height {d} vr {d} {d} {d}\n", .{ i, self.byte_heights.len, bh.height, vr.y, vr.h, vr.y + vr.h });
             end_byte = bh.byte;
 
@@ -1096,7 +1139,7 @@ pub fn bytesNeeded(self: *TextLayoutWidget, edit_start: usize, edit_end: usize, 
         end_byte -= @intCast(-edit_added);
     }
 
-    //std.debug.print("bytesNeeded {d} {d} {d}\n", .{ start_byte, end_byte, edit_added });
+    //std.debug.print("bytesNeeded end {d} {d} {d}\n", .{ start_byte, end_byte, edit_added });
 
     return .{ .start = start_byte, .end = end_byte };
 }


RedPhoenixQ:
If you are adding text in multiple parts, it would be quite annoying to figure out were you need to split and that logic needs to be included for every call to `addText`.

Would it be enough for the text layout to render all bytes on the first render (to figure out total height) while also keeping track of the intersection into the clip area? On the next frame it can skip bytes (doing no extra work) until it reaches the first bytes/line that would be visible and continue until it's outside the clip area, skipping the rest. This would be transparent to the caller.
 
If the clipping or parent content size changes between frames, we'd need to reset and process all text again to figure out our size. We could probably skip calling `renderText` when we know we are outside the clip area. I don't know what would be a good solution to invalidate the text bytes, perhaps a dedicated function or hashing bytes before the first visible line (but that would cause a one frame delay...).

I can experiment a little bit with some caching for text sizing, maybe using a bloom filter like you mentioned.

david-vanderson:
> On the next frame it can skip bytes (doing no extra work) until it reaches the first bytes/line that would be visible and continue until it's outside the clip area, skipping the rest. This would be transparent to the caller.

Good idea!  I will try this.

RedPhoenixQ:
> I can experiment a little bit with some caching for text sizing, maybe using a bloom filter like you mentioned.

I took a deeper look at the kerning changes you did recently and I'm pretty sure the complexity and overhead of the hashing in a bloom filter won't be worth it. I also don't have the tool to do proper measurements at this level right now

phatchman:
> I also don't have the tool to do proper measurements at this level right now.

I've been using the performance profiler from Visual Studio 2022 Community Edition.

phatchman:
> During testing I ran into our `max_float_safe` which is the max height of any widget. Meaning you can load enough text to where you can't scroll to the bottom.
> 
> Maybe it's time for dvui to upgrade all the sizing stuff from f32 to f64?

We exceed this in the virtual scrolling grid demo already. Luckily, I've not seen any problems doing that so far, outside of layout issues due to fp imprecision. Assuming rows are 25px tall, in theory the grid can only contain 40k rows before exceeding `max_float_safe`.

david-vanderson:
> > During testing I ran into our `max_float_safe` which is the max height of any widget. Meaning you can load enough text to where you can't scroll to the bottom.
> > Maybe it's time for dvui to upgrade all the sizing stuff from f32 to f64?
> 
> We exceed this in the virtual scrolling grid demo already. Luckily, I've not seen any problems doing that so far, outside of layout issues due to fp imprecision. Assuming rows are 25px tall, in theory the grid can only contain 40k rows before exceeding `max_float_safe`.

I think we could bump `max_float_safe` up to 2 million and still have 0.25 precision (Edit: I think it's actually 0.125 below 2 million).  With a normal text height being about 20 logical pixels, that gives us 100k lines max.  Does that sound reasonable?

david-vanderson:
@RedPhoenixQ Thanks for the api suggestion!  Now there is a `.cache_layout = true` init option.  Seems better, what do you think?

RedPhoenixQ:
> @RedPhoenixQ Thanks for the api suggestion! Now there is a `.cache_layout = true` init option. Seems better, what do you think?

Very nice. Can the byte_heights slice be removed with this change? My initial idea was to not have to stora a dynamic amount of metadata.

david-vanderson:
> > @RedPhoenixQ Thanks for the api suggestion! Now there is a `.cache_layout = true` init option. Seems better, what do you think?
> 
> Very nice. Can the byte_heights slice be removed with this change? My initial idea was to not have to stora a dynamic amount of metadata.

I don't think so?  This approach needs regular ByteHeight samples, which will be some fraction of the length of the text given.  The only alternative I can think of would be to fix the max number of ByteHeights, but usually you don't need that many.

Did you have some other idea?

phatchman:
With the latest commit I get some really nice results for my 50k line test:
| Release Mode| Kerning| FPS|
|------------|--------------|----|
| Debug | Y | 6 |
| Debug | N |60 |
| ReleaseFast | Y | 60 | 
| ReleaseFast | N | 60 | 

With kerning off, the majority of the time is in Font.zig: `    var s = fce.textSizeRaw(text, options) catch return .{ .w = 10, .h = 10 };` So I doubt you can do much better for the non-kerning case.

RedPhoenixQ:
No, I think my idea was flawed. Something along the lines of "if the clip area changes, recalculate all text until we reach the new clip area and remember only the point we start to intersect". That will be a lot more re-processing of text and more work if you are further down in the text area. If it would work, you'd only need to store the offset at which the visible text starts and keep rendering until you are outside the clip area

david-vanderson:
> With the latest commit I get some really nice results for my 50k line test:

Wohoo!

> No, I think my idea was flawed.

I initially started with that plan as well, but realized scrolling needed to be fast.  Good to have a double-check!

RedPhoenixQ:
Adding the kerning cache to `LabelWidget` would get us pretty far I think in terms of limiting the impact of kerning. The cache should at least cut the impact in half. Perhaps it's worth making the kerning buffer larger (and not having to splat 0 over all of it, textSizeEx could set a 0 terminator instead)?

Both stb and FreeType have some heuristic to determine if the font has kerning at all. FreeType has a flag and stb has a `kern` pointer that will be null. This could be used to overrider the global overside so to speak, so there is zero cost for a non-kerning font

david-vanderson:
> Adding the kerning cache to `LabelWidget` would get us pretty far I think in terms of limiting the impact of kerning. The cache should at least cut the impact in half. Perhaps it's worth making the kerning buffer larger (and not having to splat 0 over all of it, textSizeEx could set a 0 terminator instead)?
> 
> Both stb and FreeType have some heuristic to determine if the font has kerning at all. FreeType has a flag and stb has a `kern` pointer that will be null. This could be used to overrider the global overside so to speak, so there is zero cost for a non-kerning font

Good idea - I am still looking into trying to dump the kerning tables so we can put them into our own datastructure.

RedPhoenixQ:
> I am still looking into trying to dump the kerning tables so we can put them into our own datastructure.

Stb has a nice function for it but it really seems like FreeType doesn't have that.

david-vanderson:
> > I am still looking into trying to dump the kerning tables so we can put them into our own datastructure.
> 
> Stb has a nice function for it but it really seems like FreeType doesn't have that.

Oh good!  In that case I can use stb to dump it even if we are using FreeType for rasterization.

david-vanderson:
> > > I am still looking into trying to dump the kerning tables so we can put them into our own datastructure.
> > 
> > 
> > Stb has a nice function for it but it really seems like FreeType doesn't have that.
> 
> Oh good! In that case I can use stb to dump it even if we are using FreeType for rasterization.

This turned out to be less than I had hoped for.  6edbe9d24bf241f4bc4259891424a9bf4022e433 is how far I got with it.  It seemed like maybe it could save 1-2 fps in the "large doc" with 100 copies.

But just trying to get the kerning table into a nice form is very annoying - because the kerning table uses glyph indexes, not codepoints.  So we have to make a reverse map.

So I'm giving up on this (kerning improvments) and going back to the partial rendering.

david-vanderson:
@RedPhoenixQ @phatchman I'd appreciate it if you could test this for a few minutes and tell me how it goes.  Run a demo, go to the text entry section, and click the new "Large Doc" button, then play around with the stuff in that window.

At least for me the performance when editing is decent, and much better than previous.  I think the remaining slowness is related to buffer management (like when you delete a character, we move all the rest of the characters up in the whole buffer).  Can look into better backing datastructures later.

RedPhoenixQ:
> At least for me the performance when editing is decent, and much better than previous. I think the remaining slowness is related to buffer management (like when you delete a character, we move all the rest of the characters up in the whole buffer). Can look into better backing datastructures later.

On sdl3-app (debug build) on Windows 11 while mashing the keyboard at the top bottom and middle of the entry (same-ish fps for all locations) I get:

- 100 copies: 60fps (vsync limited)
- 200 copies: ~40fps
- 350 copies: ~20fps
- 600 copies: ~10fps

I did run into an assert while changing the amount of copies. I can't reproduce it unfortunately.
 https://github.com/david-vanderson/dvui/blob/93d2ab6be3ffa7aa62d83e3380a27e5172ad78d2/src/widgets/TextLayoutWidget.zig#L1512 

It should be efficient to add content at the end of the array, but it still lowers the fps for about the same amount as adding content at the top. I think the large cache invalidation after each modification is a large part of the issue. I haven't kept up to date with these changes but some partial invalidation of the the caches would be a large improvement I think (or other improvement in not having to process the text above the scroll area, even when editing the text bellow.)

david-vanderson:
Thank you!

> I did run into an assert while changing the amount of copies. I can't reproduce it unfortunately.

I'm able to reproduce it sometimes.  On it.

> It should be efficient to add content at the end of the array, but it still lowers the fps for about the same amount as adding content at the top. I think the large cache invalidation after each modification is a large part of the issue. I haven't kept up to date with these changes but some partial invalidation of the the caches would be a large improvement I think (or other improvement in not having to process the text above the scroll area, even when editing the text bellow.)

Wow I really pulled defeat from the jaws of victory.  The "cache_bad" stuff I added at the last minute to make the example easier, and totally messed it up.  Will fix.

david-vanderson:
I think I've addressed both of those, and now for me with sdl3-app Debug and 1000 copies (generating that takes some time) I can max out at 60fps when mashing the keyboard at the top or bottom.

Let me know if you see the same performance!

RedPhoenixQ:
> I can max out at 60fps when mashing the keyboard at the top or bottom.
> 
> Let me know if you see the same performance!

Wow! Massive difference. I can't make it go bellow 60 no matter where/what I type with 1000 copies. Disabling vsync gives me ~200 fps while refreshing and rendering 1000 copies and ~180-190 fps while typing, at the top or bottom. Even after manually pasting in 4000 copies it still doesn't go bellow 180 fps (although it does take more than a second when pasting the 1000 copies, but still). I'd call that really solid performance.

RedPhoenixQ:
I assume other backing datastructures would help with the reallocation/copying that's being done when pasting (i.e when it's forced to reallocate). That's gonna be a large rework though, especially if we want to make sure the default backing array case stays fast.

david-vanderson:
I'm also very happy with the performance!

> I assume other backing datastructures would help with the reallocation/copying that's being done when pasting (i.e when it's forced to reallocate). That's gonna be a large rework though, especially if we want to make sure the default backing array case stays fast.

I agree.  It's probably worth looking at the code, because I wasn't thinking about performance when I wrote that stuff.  But that is a problem for another day.

phatchman:
> I'm also very happy with the performance!

I've not been able to test performance properly yet (except that it looks really good), but I did notice that while ctrl-home and ctrl-end put the cursor in the correct position, the textbox is not scrolling to the new cursor position correctly. For ctrl-home it doesn;t scroll at all and from ctrl-end it doesn't scroll enough.

Also, I think pg-up/pg-dn support should be added if we are catering to larger amounts of text.

phatchman:
Just did some testing now. With 1000 copies, I get a solid 60fps even in Debug builds. 

See some profiling info I collected, from scrolling, editing and deleting text at start, middle and end of textarea. 
Note that the initial population of the textArea is excluded from these results.

image

I'm not sure there is that much useful info in the trace below, except to show that there is no prominent hot path anymore.

image

david-vanderson:
Thanks for the testing!
 
> I've not been able to test performance properly yet (except that it looks really good), but I did notice that while ctrl-home and ctrl-end put the cursor in the correct position, the textbox is not scrolling to the new cursor position correctly. For ctrl-home it doesn;t scroll at all and from ctrl-end it doesn't scroll enough.

You are right - something about the scroll to cursor code isn't quite working - will investigate.

> 
> Also, I think pg-up/pg-dn support should be added if we are catering to larger amounts of text.

Good point - this works in textLayout, but somehow doesn't in textEntry - will fix.

> I'm not sure there is that much useful info in the trace below, except to show that there is no prominent hot path anymore.

Excellent!

david-vanderson:
I think I've addressed all the problems I know about.  If anyone has time to test that would be appreciated, but if not, I'll merge this in a day.

phatchman:
On my x86 Windows 11 machine, I get the following debug messages occasionally:

`debug(dvui): TextLayoutWidget forcing cache_layout false due to break_lines change`

I'm not on the original machine any more so can;t check, but on my home laptop I get the error:
`debug(dvui): TextLayoutWidget forcing cache_layout false due to scale change`

Is that expected?

~~Edit: You can probably ignore this. The only way I've been able to replicate it is by clicking on the "break lines" checkbox. I must have accidently pressed it while testing dragging and resizing the window.~~

Edit2: No it happens when the window gets resized smaller. In fact, it behaves quite weirdly with the text changing size. Notice the font differences between the two screenshots below. I know they look like they are taken at different scales, but they are not. All of the font sizes change, including the menu labels, floating window headings etc.

As I resize the window, the font keeps switching between the smaller and larger one. I need to see if I can recreate this on another OS.

image

image

phatchman:
~~I don't get the message or strange resizing on Linux or MacOS, only on Windows using sdl3-app. Not present in sdl3-standalone.~~ This is not true, I get it on all 3, it is just much more prominent on Windows. But only with sdl3-app using callbacks.

I think I can sort of see what is happening. On my MacOS VM when I resize the window, it rescales the content of the window until you let go of the mouse and then it only refreshes new content when you let go of the mouse. 

Actually, now I'm occasionally getting this behaviour on MacOS as well. Yes it seems to be when I move the mouse really fast while resizing. But I don't get the `break_lines` or `scale change` message on MacOS or Linux no matter what I try.

I can maybe see something similar on Linux if I resize big and small really quickly. Just for a frame or 2? Though I'm not certain it is related to speed. Rate limiting the FPS doesn't seem to affect it and it looks like there are certain window sizes that cause the rescaling to occur and then once you are outside that range of sizes it jumps back to normal until the next "scaling" range. So it flips back and forth between scaling and redrawing.

But it seems some sort of rescaling on resize is fighting with the drawing callback? Does that make any sense?

david-vanderson:
> On my x86 Windows 11 machine, I get the following debug messages occasionally:
> 
> `debug(dvui): TextLayoutWidget forcing cache_layout false due to break_lines change`
> 
> I'm not on the original machine any more so can;t check, but on my home laptop I get the error: `debug(dvui): TextLayoutWidget forcing cache_layout false due to scale change`
> 
> Is that expected?

Yes and no, meaning I'm not sure what to do here.  textLayout can sometimes know that cache_layout should be false (meaning the text sizing is not the same as last frame):
* scale changes
* break_lines changes
* width change while break_lines is true

I think these are situations where we want textLayout to force cache_layout false (as opposed to erroring and treating it as a application bug).  But should we output a debug message?

On one hand it's telling developers why things got slow.  But on the other it reads like they did something wrong.  Any ideas here?

david-vanderson:
> But it seems some sort of rescaling on resize is fighting with the drawing callback? Does that make any sense?

That is very weird - I can imagine a bug in the callback code where for a frame we have mismatched physical and natural pixel sizes, so the natural scale is off.  I will try to reproduce.

foxnne:
I'm having trouble replicating this text sizing issue, is it just within the text layout large document window? I've been resizing windows and I can't seem to make this happen.

Outside of that, I can confirm another test point that 1000 copies runs well on a macbook air (m2), locked 60 fps regardless of scroll position.

david-vanderson:
> I'm having trouble replicating this text sizing issue, is it just within the text layout large document window? I've been resizing windows and I can't seem to make this happen.
> 
> Outside of that, I can confirm another test point that 1000 copies runs well on a macbook air (m2), locked 60 fps regardless of scroll position.

I was unable to replicate on Linux or Mac.  Rereading @phatchman's comments, my best guess is that on Mac @phatchman is seeing the (expected as of now and what I see) resizing visual thing where the mac OS stretches the content to fit the new size before we get a chance to rerender.  That would explain the visual stuff without the "forcing cache_layout false" debug output.  (Note that if you turn off sdl3-callbacks, then the stretching happens while the dvui app is frozen until you release the mouse during resize)

As for Windows, I've been getting my windows machine up and running, so will be able to test there soon.  If resizing the OS window (while the textLayout large doc demo window is visible) produces those debug logs, then something weird is going on.

I still suspect we are getting mismatched physical and natural coords for a frame or two.  Hopefully I'll be able to reproduce it.

david-vanderson:
Phew!  I can reproduce the windows scale jumping issue - split out into #579 

It happens for me even on main, so I'm going to merge this and followup on that issue separately.

phatchman:
Confirming that fixes it for me as well.

PR #565, opened 2025-09-17
TreeWidget: Fix animation/box options, AnimateWidget: Add `.none` option to disable animation

foxnne:
This was an oversight from my previous PR. Please let me know if any other options need to be forwarded to the animation, but I think we want it to mostly remain empty other than expansion and let the underlying expander box handle the options.

### DIFF
--- a/src/Examples/animations.zig
+++ b/src/Examples/animations.zig
@@ -95,7 +95,7 @@ pub fn animations() void {
 
             dvui.labelNoFmt(@src(), "Animate", .{}, .{ .gravity_y = 0.5 });
 
-            _ = dvui.dropdown(@src(), &.{ "alpha", "horizontal", "vertical" }, &global.animation_choice, .{});
+            _ = dvui.dropdown(@src(), &.{ "alpha", "horizontal", "vertical", "none" }, &global.animation_choice, .{});
 
             dvui.labelNoFmt(@src(), "easing", .{}, .{ .gravity_y = 0.5 });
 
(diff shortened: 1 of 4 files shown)


david-vanderson:
Good catch.  I think using `strip()` like this is probably not what you want, because that'll kill any gravity/min/max/margin/border/padding that the app provided.

Sorry about this - optionally wrapping a widget in another (AnimateWidget) is something that I don't have a clear strategy for.  And `strip()` is a bit weird because it zeros out some defaults (this could probably be better once we sort out the Theme stuff).

In general you want to use the passed options on the outer widget, and use `strip()` for the inner widget, but probably also make the inner widget `.expand = .both` so that it takes all the space.

In this case, maybe we should always have the AnimateWidget wrap it, but enhance AnimateWidget with a `.kind = .none` that disables the animation parts.  That way the widget hierarchy is always the same, just passing different options to AnimateWidget.  Does that sound good?

foxnne:
So the options passed in here will be for the expanded box, and I guess you're saying that instead of even using the "expander box" we will just replace that with the animation widget, and let that style things? And then just disable the animation with a new option?

david-vanderson:
> So the options passed in here will be for the expanded box, and I guess you're saying that instead of even using the "expander box" we will just replace that with the animation widget, and let that style things? And then just disable the animation with a new option?

Almost, but I think we need the expander box, because BoxWidget can handle multiple children, but AnimateWidget only expects a single child widget.

So I think we always want AnimateWidget (this is styled) with the expander box inside (using `strip()`).  Then disabling the animation with a new option like you say.  Does that make sense?

foxnne:
Ah! Thank you for clearing that up, I'll make that change and add the new option to the AnimateWidget.

david-vanderson:
> Ah! Thank you for clearing that up, I'll make that change and add the new option to the AnimateWidget.

Thank you!

foxnne:
Alright, please let me know if I've missed anything or if this is incorrect, this works with the demo and my app correctly, both animated and without animation.

david-vanderson:
> Alright, please let me know if I've missed anything or if this is incorrect, this works with the demo and my app correctly, both animated and without animation.

Looks great - I added a check to not create a starting animation if .none.  If that looks okay to you, then merge when ready!

Issue #566, opened 2025-09-17
How to get Event.Mouse struct when widget is clicked?


rasp2006:
I want to check what mouse button was used.

RedPhoenixQ:
There is a related issue in #532 that also needs the mouse button information to be made available. If you want a quick solution right now, I would copy the `dvui.clicked` helper and modify it to return the mouse button that was used instead of just a `bool`

Issue #532, opened 2025-08-29
middle clicking textLayout URL in web demo doesn't open link in new tab


MichaelBelousov:
If you open the [Text Layout](https://david-vanderson.github.io/demo/) demo on the web, left clicking the  blue link text will open a link in a new tab. Middle clicking will not do anything.

Typically, a left click navigates the current tab to it (which not everyone will want so would be nice to disable that behavior), and middle click will open in a new tab.

david-vanderson:
Good point.  We need:
* `openURL` to grow an option for whether to open the url in the current page vs new tab/page
  * does this mean anything when not in a web context?
* `labelClick` to differentiate between which mouse button is causing the click

Those should be the easy part, and once that is working, then updating `textLayout` to do similar.

david-vanderson:
I still need to implement this for textLayout.

david-vanderson:
@MichaelBelousov Does this solve your issue?

Commit summary:

dvui.link() and TextLayoutWidget.addLink() both package up the normal
link clicking behavior:
* link is colored as the current theme's focus color (maybe that should
  change in the future)
* left click (or touch click) calls dvui.openURL with new_window = false
* middle click or ctrl/cmd-click calls dvui.openURL with new_window =
  true

The web backend is currently the only backend that cares about
new_window.  It also has a flag to override new_window to true for cases
where you don't want users to accidentally replace the webapp with a
different page.

MichaelBelousov:
Sounds good but on second thought, it's kind of weird for a DVUI app to use a link that stays in the current tab? I mean yes that's what the browser does but I imagine most people using dvui on the web is so they can reuse the UI on another platform where the link will not close the current app. I could be wrong about that though!

I'm a bit unsure of what should be the default for left click but I'm glad middle mouse click is now built in!

david-vanderson:
I also had that thought, so I've pushed a commit to make that the default.  Great!

Weirdly, a few times during testing I left-clicked a link that moved me out of the demo to a new page, then clicked the back button and got back to the same state as before (like somehow the browser had kept the demo alive in the background).  But it wasn't repeatable.

MichaelBelousov:
I think I've had a similar feeling! There may be some interesting browser spec somewhere haha

PR #568, opened 2025-09-19
Event.Mouse function

rasp2006:


src/dvui.zig
```diff
@@ -3500,6 +3500,85 @@ pub const ClickOptions = struct {
...
+                        // if the release was within our border, the click is successful
+                        if (click_rect.contains(me.p)) {
+
+                            // if the user interacts successfully with a
+                            // widget, it usually means part of the GUI is
+                            // changing, so the convention is to call refresh
+                            // so the user doesn't have to remember
+                            dvui.refresh(null, @src(), wd.id);
+
+                            return me;
```
RedPhoenixQ, line 3541: Returning here exits the handler early so some potential events from the same frame will get missed. That can put the mouse capture/drag in an invalid state.

RedPhoenixQ:
It would be nice if dvui didn't have to duplicate the entire function. You could change the original `clicked` function to `clickedEx` that returns `struct { bool, ?Event.Mouse }` and add `var mouse_event: ?Event.Mouse = null;` bellow the `is_clicked` and set it during the event handling. Then `clicked` can use `clickedEx` but just return the boolean. 

```zig
pub fn clickedEx(wd: *const WidgetData, opts: ClickOptions) struct { bool, ?Event.Mouse } {
    var is_clicked = false;
    var mouse_event: ?Event.Mouse = null;

   // ... event handling loop here. Set `mouse_event` when clicked gets set to `true`
   
   return .{ is_clicked, mouse_event };
}

pub fn clicked(wd: *const WidgetData, opts: ClickOptions) bool {
    const is_clicked, _ = clickedEx(wd, opts);
    return is_clicked;
}
```

rasp2006:
Can we return some key event data?

RedPhoenixQ:
It could return `struct { bool, union(enum) { mouse: Event.Mouse, key: Event.Key } }` sure.

rasp2006:
I think that evt field in Event struct should became a defined type, not just an anonymous union

rasp2006:
But i don't know how to integrate it in other event related code, i wanted to add a feature to foxnne/pixi, and now i'm here, exploring the depths of this project

david-vanderson:
> But i don't know how to integrate it in other event related code, i wanted to add a feature to foxnne/pixi, and now i'm here, exploring the depths of this project

Welcome!  Let's get what you need for this first, can talk about modifying Event.evt after.  Does this code do what you need right now?  Is it enough to know which mouse button (or that it was a key event)?  Should a key activation be the same as the left mouse button?

rasp2006:
No it's not, I accidentally compiled and tested the code from the wrong VSCode instance (wrong branch), it's not building, it seems that this code  requires the creation of a defined type, because zig treats every instance of this anonymous union declaration as a separate type, and does not allow implicit conversion between them.
Or i'm just silly

rasp2006:
Now it's fixed, i would close #566 after testing it in the pixi project

RedPhoenixQ:
Looking good! Thank you

rasp2006:
#532

rasp2006:
dx11 build is broken, but it looks like it's not my fault.

RedPhoenixQ:
Yeah, win32 dependency hasn't updated to zig 0.15. Don't worry about it

david-vanderson:
Thank you!

PR #570, opened 2025-09-19
Fix clickedEx

rasp2006:
Function was hard coded to register only left clicks, it seems to work now, at least in the pixi editor

### DIFF
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -3500,6 +3500,9 @@ pub const ClickOptions = struct {
     ///
     /// Defaults to the border rect of the `WidgetData`
     rect: ?Rect.Physical = null,
+
+    /// Which mouse buttons to react to.
+    buttons: enum { pointer, any } = .pointer,
 };
 
 pub fn clickedEx(wd: *const WidgetData, opts: ClickOptions) ?Event.EventTypes {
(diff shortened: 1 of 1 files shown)


david-vanderson:
Looks good, but give me some time to look through the other places using `pointer()` to see if they need adjustment.  Thanks!

david-vanderson:
Places using `pointer()`:
* okay to be any button:
  * `dvui.Window`: debug until click
* should be only left or touch:
  * scroll canvas demo: dragging canvas/boxes
  * reorder/tree demo: dragging tree elements
  * `slider` and `sliderEntry`: to drag the slider
  * textLayout: drag-select text
  * dropdown: select items
  * button: click (by default)
  * paned: drag the sash
* unsure:
  * `dvui.Window`: generating `.focus` events
  * `windowHeader`: raises the subwindow

Given this I think we should leave `pointer()` the way it is but have `clickedEx` respond to any mouse button, and then have `clicked` only return true for `pointer()` mouse events (and key events).  Will make that change now.

david-vanderson:
> Is this part correct? I assume it should be `.right` and not `.eight`
> 
> ```
> const start = @intFromEnum(Button.left);
> const end = @intFromEnum(Button.eight);
> ```

Theoretically mice can have lots of buttons, and they are named left/right/middle/four/five/six/seven/eight - I guess people couldn't agree on the position of the ones past the first three!

david-vanderson:
I decided against checking the mouse button in `clicked()` because then even with the middle/right buttons we would capture the mouse, which made ButtonWidget show as pressed (so you expected something to happen).  So I went with "passing an option to clickedEx()".

@rasp2006 How does that look?

rasp2006:
Ok

rasp2006:
lgtm

david-vanderson:
That refactor seems like it is the same, but it means that even if you are only checking for mouse left clicks, the ButtonWidget will still capture the mouse with any button.  Capturing triggers the "pressed" color, so it looks like the button is being clicked.

Yeah it's a subtle thing, but I don't want to give it up accidentally.  Did you intend for that to happen?

rasp2006:
I did not intend that, but I don't think this is a completely undesirable effect, for me it's just proves what i was saying in #571, this function seems overloaded with functionality, it still can check for capturing but widget must decide for himself what to do with this information, why function that checks for clicks, checking for hovering and changing cursors?

david-vanderson:
> I did not intend that, but I don't think this is a completely undesirable effect, for me it's just proves what i was saying in #571, this function seems overloaded with functionality, it still can check for capturing but widget must decide for himself what to do with this information, why function that checks for clicks, checking for hovering and changing cursors?

That is a valid concern, but separate from this.  I think for this purpose, if a button is only going to respond to left clicks, then it should not show as pressed for other mouse buttons.  Please revert the refactor and then we can merge this.

Thanks!

david-vanderson:
Thank you!

PR #575, opened 2025-09-22
Reorganization

rasp2006:
I also wanted to split dvui.zig file into multiple, but it turned out to be more difficult than I thought.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -691,7 +691,7 @@ fn addWebExample(
     const cb = b.addExecutable(.{
         .name = "cacheBuster",
         .root_module = b.createModule(.{
-            .root_source_file = b.path("src/utils/cacheBuster.zig"),
+            .root_source_file = b.path("src/cacheBuster.zig"),
             .target = b.graph.host,
         }),
     });


rasp2006:
Mb, i haven't checked web build.

RedPhoenixQ:
I'm for the folder structure, but splitting out things from the dvui file is difficult. I made an attempt at the `dvui.data...` functions but it was really unclear where to draw the line and how/what to re-export. Would recommend leaving that alone.

All of the bellow are ideas that can be done in separate PRs if we what them. This PR is good as is, easier to do it gradually imo.

I've though about many of the "helper" functions like `dvui.box` could live in their respective widget file (`dvui.box` -> widgets/BoxWidget.zig, etc). I would make it hander to get an overview of what's available though as you'd have to jump between files more. Not really for it.

Things I think could be extracted from "dvui.zig" (but all current declaration remain as re-exports):
- PNGEncoder/JPGEncoder -> utils/...?
- FontCache and it's helper functions/structs -> utils/... OR fonts/...? not sure here.
- Rendering functions/structs (`Triangles`, `renderImage`, `ImageSource`, `renderText`, `renderTexture`, `renderTriangles`) -> one big-ish utils/rendering.zig OR multiple files (rendering.zig for all `render...` functions, Triangles.zig and `ImageSource.zig`), whichever is nicer

Potential ones that I'm less sure about:
- `dialog`/`toast` and their helper structs and functions

rasp2006:
Are you spying on me?
That's exactly what i was doing, i created render/render.zig file and tried to move all rendering types and utils in it,
created files for input.zig, font.zig, moved widget related helpers to the corresponding widgets.
But it broke to much code, and most importantly it broke user api, just understanding this huge file (Which is bigger then some single header c libs i worked with :), was hard, and i definitely don't know to to fix all of this after splitting.
But i think this should be done and as soon as possible, even if it breaks user api, i can't imagine expanding or refactoring something in this project with this blob of code that contains half of the project code.
This technical debt would just grow bigger and bring more problems.

david-vanderson:
> But i think this should be done and as soon as possible, even if it breaks user api,

I'm not willing to break the api just for internal reorganization.
* What does "ds" mean?
* I don't understand putting things under "types" - these are normal files and can live where they are.
* moving stuff we vendor (possibly with patches) into external makes sense
* utils is hard - cachebuster is only at build time, the others are normal runtime stuff

RedPhoenixQ:
> Are you spying on me?

lol, great minds think alike

> i created render/render.zig file and tried to move all rendering types and utils in it,
> created files for input.zig, font.zig, moved widget related helpers to the corresponding widgets.
> But it broke to much code, and most importantly it broke user api,

It's shouldn't have to break anything if you re-export everything in dvui.zig. Your issues are partially why I think making many smaller changes will be better/easier.

> But i think this should be done and as soon as possible, even if it breaks user api, i can't imagine expanding or refactoring something in this project with this blob of code that contains half of the project code.
> This technical debt would just grow bigger and bring more problems.

I've also wanted there to be more namespacing (and thus simple file splitting) for the functions in dvui.zig, especially those that are already pseudo-namespaced like `dvui.dataXXX`, `dvui.textureXXX` or `dvui.renderXXX`. I we are going to break api, it love for it to become `dvui.data.XXX` etc. The texture apis for example could probably move into the `Texture` struct and so on.

david-vanderson:
> I've also wanted there to be more namespacing (and thus simple file splitting) for the functions in dvui.zig, especially those that are already pseudo-namespaced like `dvui.dataXXX`, `dvui.textureXXX` or `dvui.renderXXX`. I we are going to break api, it love for it to become `dvui.data.XXX` etc. The texture apis for example could probably move into the `Texture` struct and so on.

I could see doing this kind of thing, but we'd have to be careful about it, maybe try to do deprecation compile errors.

RedPhoenixQ:
> I could see doing this kind of thing, but we'd have to be careful about it, maybe try to do deprecation compile errors.

I think as a first step we could re-export everything with the old api:

```zig
// dvui.zig
pub fn dataGet(...) ...
```

```zig
// data.zig
pub fn get(...) ...

// dvui.zig
pub const dataGet = data.get;
```

That way we can add deprecation comments to these apis if we want and users can start using `dvui.data.get` incrementally.

Same solution would apply to moving function into the `Texture` struct and also `FontCache`, i.e `pub const textureCreate = Texture.create;`

rasp2006:
>     * What does "ds" mean?
Data structures

phatchman:
What problem are you trying to solve with this re-org? 

There is a lot of benefits to discovery in having the core dvui apis available in one place.

rasp2006:
> What problem are you trying to solve with this re-org?
> 
> There is a lot of benefits to discovery in having the core dvui apis available in one place.
@phatchman 


Personally, I have no problem with the fact that dvui.zig file contains all api,
but it contains not only them but also 7000 lines of unordered random structures, utils and widgets, that can be placed in ther personal files.

rasp2006:
Since @david-vanderson didn't like moving files into separate directories, I'll revert the commit and think about it later or ask @RedPhoenixQ for help and his thoughts.

david-vanderson:
Thank you!

PR #576, opened 2025-09-23
label: ellipsize text that would otherwise be cutoff

### LINKED ISSUE #534: overflow options for max_size_content
Inspired by CSS obviously, it would be nice if when you specify `max_size_content`, you could choose an overflow option like adding an ellipsis (...) instead of just clipping.

david-vanderson:
fixes #534 

@MichaelBelousov This makes all labels everywhere always do this.  Can you think of a situation where you would not want this?

### DIFF
--- a/src/Examples/basic_widgets.zig
+++ b/src/Examples/basic_widgets.zig
@@ -126,7 +126,7 @@ pub fn basicWidgets() void {
 
         const entries = [_][]const u8{ "First", "Second", "Third is a really long one that doesn't fit" };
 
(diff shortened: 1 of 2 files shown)


david-vanderson:
@MichaelBelousov ping in case it got missed.

MichaelBelousov:
> @MichaelBelousov ping in case it got missed.

Yes I did see and forget this! I'll think of this tonight

MichaelBelousov:
Thinking about it, I'd prefer it be an option.
It's an OK default though if that's easier. I think it's rarely going to be an issue to over-ellipsize

david-vanderson:
I committed this directly to main due to the refactoring that happened in between.

PR #577, opened 2025-09-23
Refactor dvui.zig to separate files (no api break)

RedPhoenixQ:
Mentioned here https://github.com/david-vanderson/dvui/pull/575#issuecomment-3320714184

I'd like to merge this incremental change asap and create follow up PRs because of the annoying nature of merging moved files.

I'm fully prepared to rebase/merge these changes with anything that comes from #575 (or rebase that PR). I've not cared about folder structure here yet and have currently gone for the simple low hanging fruit. There are many things left but the size of dvui.zig has already decreased by 2000 loc (~25%).

I'm also considering a larger change where the logic in functions operating on `currentWindow()` would move into the Window so there is an explicit `self: Window` parameter in these functions. The functions in dvui would then just call the function on `currentWindow()`. Only downside to this right now is the need to duplicate the declaration and arguments and that the docs can't easily be linked together (we don't want to duplicate them all, but preferably we'd want to write them on the implementation, but then the most common case of using the `dvui.XXX` version wouldn't get them...)

An extension of the above case would be things like `Texture.Cache` which would operate on itself and not rely on `currentWindow()` and the `Window` would hold and instance of `Texture.Cache`. Removing the requirement of "Must be called between `Window.begin` and `Window.end`" would be a nice change imo. Less hidden logic/global variables for things that don't need them, like the texture cache.


david-vanderson:
Looks good - merge it!

> I'm also considering a larger change where the logic in functions operating on `currentWindow()` would move into the Window so there is an explicit `self: Window` parameter in these functions. The functions in dvui would then just call the function on `currentWindow()`. Only downside to this right now is the need to duplicate the declaration and arguments and that the docs can't easily be linked together (we don't want to duplicate them all, but preferably we'd want to write them on the implementation, but then the most common case of using the `dvui.XXX` version wouldn't get them...)

Interesting - will need to explore this more, but I do like the idea of reducing the things that are restricted to being between begin/end if the api still looks okay.

david-vanderson:
Thank you!

rasp2006:
Wow, great work @RedPhoenixQ, i thought i would have to fight this monster myself :).

RedPhoenixQ:
The Window class is going to be the biggest hurdle, at least while mostly maintaining the api. I'll tinker on it

PR #580, opened 2025-09-24
TreeWidget: No longer rely on setting border non-uniform...

foxnne:
...(`.{ .x = 1 }`) for the expander box border, and instead draw a stroke

The rounded edge can be controlled with the expander's border opts, zero radius will draw straight lines. This should now look basically the same except the ends of the arcs wont "fade" like they did before, but I really don't think thats a detail worth worrying about.

Screenshot 2025-09-24 at 1 57 34 PM

### DIFF
--- a/src/Examples/reorder_tree.zig
+++ b/src/Examples/reorder_tree.zig
@@ -393,6 +393,7 @@ pub fn reorderTree(uniqueId: dvui.Id) void {
             .background = true,
             .border = dvui.Rect.all(1),
             .padding = dvui.Rect.all(4),
+            .corner_radius = dvui.Rect.all(4),
         },
         .{
             .padding = dvui.Rect.all(1),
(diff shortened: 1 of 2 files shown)


david-vanderson:

You motivated me to try again with the non-uniform borders, so we'll see if I get anywhere this time!

foxnne:
Haha just make sure it's not for my sake! Its one of those things that seems like it would be simple and ends up being complex I think, because I think the current border is drawn by drawing a border color full rect fill expanded border thickness before drawing the actual background rect. 

hmm... What if we just inset the border rect by default? Specified border sizes above 0 become their actual thickness.  You could I guess just inset by fade distance if the border is 0

david-vanderson:
Great idea!  I've merged that idea into main.  Does that look okay to you?

david-vanderson:
(I hit another dead end with trying to do variable thickness strokes, but still learning!)

foxnne:
> Great idea! I've merged that idea into main. Does that look okay to you?

Screenshot 2025-09-25 at 10 40 59 AM

I just tested by forcing on fade and it seems to look great, and I no longer see the edges during motion. The only issue i can think of is slightly less of the arcs will be drawn when fade is on (since the end of the rounded edge is inset), but I don't think thats a big issue at all. Looks great! I'll take a look at this on my windows machine with 1x scale soon, but this should be the fix. Closing this PR as its no longer necessary

david-vanderson:
Awesome thanks!

Issue #559, opened 2025-09-14
Help migrating some dvui.Options code to 0.3.0


phatchman:
I have an application that creates some "reversed" buttons. The following code used to work:
```
    const reversed = dvui.Options{
        .color_text = .{ .name = .fill_window },
        .color_fill = .{ .name = .text },
        .color_fill_hover = .{ .name = .fill_press },
        .color_fill_press = .{ .name = .fill_hover },
        .expand = .horizontal,
        .margin = Rect{ .x = 2, .w = 2, .y = 2, .h = 0 },
        .corner_radius = Rect.all(0),
        .background = true,
    };
```
Now that fill_press and fill_hover are removed and instead calculated shades based on the button state, how would I implement this now? I don't see a way to be able to swap the hover and press colors? Or customize them at all any more.

I can't change the whole theme as I want some buttons in the theme colors and others with the reversed colors and that wouldn't work anyway as I have created specific hover and press colors to suit the UI. Just making them just brighter or darker shades won't work for me.

Any ideas?

Thanks.

RedPhoenixQ:
As a temporary work around you can modify the theme before and after a button to change only that button. 

There was a lot of discussion around how to handle this, and perhaps we need to reintroduce the hover and press colors to Options. Seems like the simplest fix. Perhaps all fields from `Theme.ColorStyle` should be available on options for more granular control of each widget. 

There was also some work making these options available on the widgets `InitOptions`. The motivation behind that was that most widgets didn't need/use these options, so the ones that did could expose them separately. This still might be the way to go, we just need to add it to the widget it applies to.

david-vanderson:
Sorry about it - this is an area we don't have a great answer for right now.

> As a temporary work around you can modify the theme before and after a button to change only that button.

This is the best we have right now.

We've thought of 3 different ways to deal with this:
* add more colors back to Options
  * problem here is where do we stop?
  * in your case not clear it works because we no longer have the ColorOrName indirection

* add widget-specific colors to InitOptions
  * problem here is possibly a lot of widget-specific colors
  * could do this in addition to something else
  * how common is color_press and color_hover

* add `*Theme` to Options
  * problem here is you copy/make a whole theme just to change a few colors

@phatchman any ideas or feedback regarding these possibilities?

phatchman:
@RedPhoenixQ 
I'll try that. I don't see why it wouldn't work. 

[...]

PR #582, opened 2025-09-25
Theme Cleanup 2

david-vanderson:
Based on the excellent feedback from our last attempt at fixing Theme #498 

My thoughts here:
* removing hover/press colors was a mistake, they are useful enough
  * need to think about how widgets communicate colors to children
  * Options.stylesOnly) and ButtonWidget.colors()
* saying "temporarily change the theme and back" doesn't seem like something people like or expect
  * I think it is still a valid solution for when an app wants a different them for a whole section of the UI
  * should try having a theme pointer in Options
  * will have to see how this interacts with compound widgets
* address need to produce stuff like "reversed color buttons" (#559)

Goals for this PR:
- [x] add back color_fill_hover and color_fill_press to Options
- [x] add something to the demo that changes color_fill_hover/color_fill_press/color_text_hover/color_text_press
- [x] add a `*Theme` to Options
  * if given, the widget does not touch the global theme at all
  * make a part in the demo that has a button for each theme

Going to focus just on the colors for now.  Will deal with fonts later.

### DIFF
--- a/src/Examples/basic_widgets.zig
+++ b/src/Examples/basic_widgets.zig
@@ -281,7 +281,9 @@ pub fn dropdownAdvanced() void {
     };
 
     var theme = dvui.themeGet();
+    theme.control.fill_hover = dvui.Color.purple;
     theme.highlight.fill = dvui.Color.purple;
+    theme.control.text_hover = dvui.Color.red;
     theme.highlight.text = dvui.Color.red;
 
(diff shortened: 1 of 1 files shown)


br07h3rj:
Can we have color_text_hover and color_text_press too? I think that would complete the box set for me.

david-vanderson:
> Can we have color_text_hover and color_text_press too? I think that would complete the box set for me.

Yes those will be in there.  We don't have anything that exercises them though, so I'll put that on the list.

br07h3rj:
I was looking at the file changes and about to delete my comment when you responded. My mistake. Thanks for the options. I plan on making use of all of those in implementation. :smile:

PR #584, opened 2025-09-27
Refactor data and texture cache

RedPhoenixQ:
This splits out the old `datas` and the texture cache into separate files that no longer have any logic in `dvui.zig` or depends on `dvui.currentWindow()`. Both of these structs handle their "trash" and any potential locking.

Separating other parts of `Window` and the `dvui.zig` file will get harder as there are some functions that depend on random fields. For example, splitting out events and the event adding functions would still require some reference back to the `debug` field on `Window` to add debug touch events. 

One solution to the splitting issue is that the other separate files use `const win = @fieldParentPtr("events", self)` to still allow this behaviour but still separating the different responsibilities. I think this is a decent solution for the event list, subwindows (with `RenderCommand` that needs some window fields), font cache (already takes `*Window` in it's init function so this isn't that far from reality) and dragging (needs the window natural scale) .

The goal is to make `dvui.zig` just an entry point of the api that calls the more specific apis with some conveniences, like using `currentWindow()` for you. It should not include the implementations for these functions as these should be grouped into some logical place with functions within the same domain etc. The widget helper functions, `dvui.box` etc, should also be moved to their respective widgets files (probably in a nested struct called `Helpers` or similar, potentially in a separate file, to not mix them with the widgets api).

### DIFF
--- a/src/Texture.zig
+++ b/src/Texture.zig
@@ -24,55 +24,73 @@ pub const Target = struct {
 };
 
 pub const Cache = struct {
+    cache: Storage = .empty,
+    /// Used to defer destroying textures until the next call to `reset` or `deinit`
+    trash: Trash = .empty,
+
     pub const Storage = dvui.TrackingAutoHashMap(Key, Texture, .get_and_put);
+    pub const Trash = std.ArrayListUnmanaged(dvui.Texture);
+
     pub const Key = u64;
 
-    /// Gets a texture from the internal texture cache. If a texture
-    /// isn't used for one frame it gets removed from the cache and
-    /// destroyed.
-    ///
-    /// If you want to lazily create a texture, you could do:
-    /// ```zig
-    /// const texture = dvui.textureGetCached(key) orelse blk: {
-    ///     const texture = ...; // Create your texture here
-    ///     dvui.textureAddToCache(key, texture);
-    ///     break :blk texture;
-    /// }
-    /// ```
-    ///
-    /// Only valid between `Window.begin`and `Window.end`.
-    pub fn get(key: Key) ?Texture {
-        const cw = dvui.currentWindow();
-        return cw.texture_cache.get(key);
+    pub fn get(self: *Cache, key: Key) ?Texture {
+        return self.cache.get(key);
     }
 
     /// Add a texture to the cache. This is useful if you want to load
     /// and image from disk, create a texture from it and then unload
     /// it from memory. The texture will remain in the cache as long
-    /// as it's key is accessed at least once per frame.
-    ///
-    /// Only valid between `Window.begin`and `Window.end`.
-    pub fn add(key: Key, texture: Texture) void {
-        const cw = dvui.currentWindow();
-        const prev = cw.texture_cache.fetchPut(cw.gpa, key, texture) catch |err| {
-            dvui.logError(@src(), err, "Could not add texture with key {x} to cache", .{key});
-            return;
-        };
+    /// as it's key is accessed at least once per call to `reset`.
+    pub fn add(self: *Cache, gpa: std.mem.Allocator, key: Key, texture: Texture) std.mem.Allocator.Error!void {
+        try self.trash.ensureUnusedCapacity(gpa, 1);
+        const prev = try self.cache.fetchPut(gpa, key, texture);
         if (prev) |kv| {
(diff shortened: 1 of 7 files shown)


src/Texture.zig
```diff
@@ -353,14 +371,15 @@ pub fn fromTarget(target: Target) TextureError!Texture {
...
-    cw.texture_trash.append(cw.arena(), texture) catch |err| {
+    cw.texture_cache.trash.append(cw.gpa, texture) catch |err| {
         dvui.log.err("Texture destroyLater got {any}\n", .{err});
     };
 }
 
 const std = @import("std");
 const dvui = @import("dvui.zig");
 
+const Backend = dvui.Size;
```
david-vanderson, line 382: Typo?

david-vanderson (commented):
I'm okay with this.  Merge when ready.  Thank you!

src/Texture.zig
RedPhoenixQ: Definitely. Thanks for caching!

david-vanderson:
> This splits out the old `datas` and the texture cache into separate files that no longer have any logic in `dvui.zig` or depends on `dvui.currentWindow()`. Both of these structs handle their "trash" and any potential locking.
> 
> Separating other parts of `Window` and the `dvui.zig` file will get harder as there are some functions that depend on random fields. For example, splitting out events and the event adding functions would still require some reference back to the `debug` field on `Window` to add debug touch events.
> 
> One solution to the splitting issue is that the other separate files use `const win = @fieldParentPtr("events", self)` to still allow this behaviour but still separating the different responsibilities. I think this is a decent solution for the event list, subwindows (with `RenderCommand` that needs some window fields), font cache (already takes `*Window` in it's init function so this isn't that far from reality) and dragging (needs the window natural scale) .

I might not be understanding this, but it seems like if we are depending on @fieldParentPtr we might as well take *Window.  But maybe an example would help.

> 
> The goal is to make `dvui.zig` just an entry point of the api that calls the more specific apis with some conveniences, like using `currentWindow()` for you. It should not include the implementations for these functions as these should be grouped into some logical place with functions within the same domain etc. The widget helper functions, `dvui.box` etc, should also be moved to their respective widgets files (probably in a nested struct called `Helpers` or similar, potentially in a separate file, to not mix them with the widgets api).

I'm not sold on this goal in general.  Specifically I think dvui.zig is a good place for all the high level widget functions.  But other subsystems I could be convinced to move.

RedPhoenixQ:
I'm working on a separate PR for the "widget helpers" as I've been calling them, separating the out to their respective widgets. It makes it much easier to find all the "functions that create widgets". We can discuss more there.

> Specifically I think dvui.zig is a good place for all the high level widget functions

I think we mostly agree here, I was a little extreme in my wording. Things like `widgetAlloc`, `firstFrame`, `mouseTotalMotion`, `clicked` and `logError` might belong in `dvui.zig`. They are very generic helpers that don't belong to any larger grouping. In general I feel like any set of functions with the same prefix (`dragXXX`, `dialoagXXX`, `toastXXX`, etc) should probably be grouped in the code, either as a struct or separate file. I don't mind the re-exporting to not break the public api, but more grouping makes it easier to find related functions.


> if we are depending on @fieldParentPtr we might as well take *Window. But maybe an example would help.

This is a good point. Maybe that's just clearer to take `*Window`. The @fieldParentPtr solution was just my first idea for neatly splitting a "god class" into multiple files/structs/groups.

david-vanderson:
> I'm working on a separate PR for the "widget helpers" as I've been calling them, separating the out to their respective widgets. It makes it much easier to find all the "functions that create widgets". We can discuss more there.

Cool.

> 
> > Specifically I think dvui.zig is a good place for all the high level widget functions
> 
> I think we mostly agree here, I was a little extreme in my wording. Things like `widgetAlloc`, `firstFrame`, `mouseTotalMotion`, `clicked` and `logError` might belong in `dvui.zig`. They are very generic helpers that don't belong to any larger grouping. In general I feel like any set of functions with the same prefix (`dragXXX`, `dialoagXXX`, `toastXXX`, etc) should probably be grouped in the code, either as a struct or separate file. I don't mind the re-exporting to not break the public api, but more grouping makes it easier to find related functions.
> 

Yeah we mostly agree.  Definitely stuff that shares the same prefix (dragXXX) are good targets.

PR #585, opened 2025-09-27
Refactor widget helpers

RedPhoenixQ:
This moves the widget helpers from `dvui.zig` into their respective widgets files in a nested `Helpers` struct (if you have a better name I'm down to change it). This patterns groups all the widgets and their "widget helpers" compactly in the `dvui.zig` file, making it more clear which function create UI and which are for utility.

I'm not 100% sure what to do about functions like `spinner`, `expander`, `checkbox`, etc that don't have a corresponding widget. I'm thinking of making a file for generic helper like these and list them at the end of the widget section. `slider` has a lot of variations that could fit in their own file, perhaps event called `SliderWidget`, like `SuggestionsWidget` that isn't a true widget.


david-vanderson:
Sadly I don't think this improves our code organization.  I can kind of see how it makes sense to put functions alongside the widgets they go with, but on the other hand all the high level functions go together, so it makes sense to have them all in dvui.zig.

That also reinforces that those are the primary starting point for people just starting out, which I think is hurt by putting them in separate files.

So I'm leaning against this.

RedPhoenixQ:
> but on the other hand all the high level functions go together, so it makes sense to have them all in dvui.zig.

Ok that's fair. Would it be ok if all of these helper function where in a second file, still re-exported? That way the `dvui.zig` file isn't so daunting

david-vanderson:
> > but on the other hand all the high level functions go together, so it makes sense to have them all in dvui.zig.
> 
> Ok that's fair. Would it be ok if all of these helper function where in a second file, still re-exported? That way the `dvui.zig` file isn't so daunting

dvui.zig is "only" about 6k lines.  I think the right move is to deal with the other subsystems where it makes sense to consolidate them, and then reassess?

RedPhoenixQ:
> I think the right move is to deal with the other subsystems where it makes sense to consolidate them, and then reassess?

Sound like a plan!

PR #586, opened 2025-09-27
refactor: move dialogs and toasts to a separate file

RedPhoenixQ:
This implements toasts as a specialized kind of dialog which removes a lot of duplicate logic.

The toasts and dialog helpers and their related display functions have been left in `dvui.zig` because it fits better in the group of widget helpers rather than along the storage for dialogs.


david-vanderson:

PR #587, opened 2025-09-28
refactor: move native dialogs to a separate file

RedPhoenixQ:
The c import was not moved because it's recommended to only have one `@cImport` to reduce the calls to Clang.


david-vanderson:
Good!  Merge when ready.  Thank you!

PR #589, opened 2025-09-28
refactor: move dragging logic to separate file

RedPhoenixQ:
This is the first instance were some other, unrelated, field from the `Window` struct was needed. The natural scale is needed for preDrag to make it more concistent on different scales. I've decided to take it as a parameter for now, but this isn't that nice of a solution.


david-vanderson:
I agree, but I can't think of a better solution, so we can go with this for now.  Merge when ready.  Thanks!

PR #591, opened 2025-10-01
TrackingAutoHashMap: remove unused values with an iterator

RedPhoenixQ:
The current std hashmap doesn't invalidate pointers when removing keys, instead just marking the slot as empty. This means that we do not have to allocate to remove the unused entries, which removes a point of failure and simplifies the reset logic.


david-vanderson:
Whoops pushed that commit to the wrong branch sorry.

RedPhoenixQ:
No problem. Sorry for the refactoring issue

david-vanderson:
This looks good - merge when ready!

PR #592, opened 2025-10-01
Refactor Font and FontCacheEntry

RedPhoenixQ:
This moves the previous `FontCacheEntry` to `font.zig` and creates a proper storage struct to manage the font files and the glyph/atlas cache. This gets the `dvui.zig` file down bellow 5000 lines.

To avoid using a `*Window` argument or `currentWindow`, most functions takes an allocator and a `Backend` to manage the internal hash maps and the texture atlas.

I've not done much in terms of renaming functions so if there is any old name you'd like to be changed in this refactor, feel free to tell me.

### DIFF
--- a/src/Font.zig
+++ b/src/Font.zig
@@ -301,6 +301,8 @@ pub const Cache = struct {
             if (font.id == Font.default_font_id) {
                 @panic("Default font could not be loaded");
             }
+            // Remove the invalid font cache entry
+            self.cache.map.removeByPtr(entry.key_ptr);
             return self.getOrCreate(gpa, font.switchFont(Font.default_font_id));
         };
         //log.debug("- size {d} ascent {d} height {d}", .{ font.size, entry.ascent, entry.height });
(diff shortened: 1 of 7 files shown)


src/Font.zig
```diff
@@ -176,14 +188,13 @@ pub const FontId = enum(u64) {
...
+        for (@typeInfo(builtin).@"struct".decls) |decl| {
             std.debug.assert(map.get(decl.name) == FontId.fromName(decl.name));
         }
     }
 };
 
-// functionality for accessing builtin fonts
-pub const TTFBytes = struct {
+pub const builtin = struct {
     pub const InvalidFontFile = "This is a very invalid font file";
```
david-vanderson, line 198: Why do we have this?  It's showing up in the font selection dropdowns (both in the text entry and theming part of the demo).  In main it logs errors, and in this PR it panics if you choose it.

david-vanderson:

PR #594, opened 2025-10-06
AccessKit Integration: Draft 

phatchman:
This probably good enough now to take a look at.

Currently Implements:
- Adding most of the basic widgets to the accessibility tree.
- Handles click, set value and set numeric value actions

Core Changes:
- Adds an optional "accesskit_node" to WidgetData. This is used to find the nearest ancestor that has an accessibility node to use as a parent in the accessibility tree.
- Adds AccessibilityOptions to dvui.Options
- Creates events at the start of each frame in respond to accessibility actions.

Main TODOs:
- [ ] Work out where dvui.access_kit should really live. i.e. in Window or external.
- [ ] Move all of the access kit calls from sdl-standalone to the appropriate places (e.g. window.being() and window.end())
- [ ] Extend to additional widgets and dvui functions
- [ ] Support additional actions such as scrolling
- [ ] Add support for other operating systems and backends.
- [ ] Integrate properly into build system

Concerns:
- Events are being created manually and added to the event list for the window. Everything appears to be working fine by directly creating the events the way I am doing in addAllEvents, but it is worth reviewing.
- Very difficult to test this stuff without using 3rd party tools. Needs thought on how we'd do this. I can see things breaking and no one realising it until it gets reported.
- The action handling is a little convoluted, but I think it's not too bad:
   - When an action is received it is added to `action_requests` to be processed at end of frame
   - At end of frame I generate the events for start of the next frame to `self.events`. I do this because all of the accesskit nodes have been created at that stage and I can retrieve their bounds and any other properties I need to raise the events. Once the nodes are pushed to the accesskit tree, they are no longer valid.
   - At the start of the next frame, the events from the previous frame are added to `window.events`.


DataTriny:
I have built the sdl3-app example with `zig build sdl3-app`. While I can see the window showing, no accessibility tree seem to initialize. Is there something more I should do or should I use another example app to try this out?

Side note: I'm on Windows and I had to copy accesskit.dll inside the accesskit subdirectory for the app to compile, and then copy it to zig-out/bin for it to open.

First time compiling some Zig code I think, pardon me if I'm missing something.

david-vanderson:
This is an excellent start.  Thank you very much!

I've read through the code, and I think you are on the right track.  I want to pitch in and try out some stuff, but I'm having trouble just getting started.

My understanding is that for now we are only doing `zig build sdl3-standalone` and only on Windows.  So I went to https://github.com/AccessKit/accesskit-c/releases and downloaded the 0.17.1 release.  I tried the various things under `lib/windows/x86_64/...` but either got linking errors or runtime errors on startup (error 53)

Can you give me some info about what I need to do to get this running?

DataTriny:
@david-vanderson I'm on a 64bit machine using the MSVC toolchain so I copied accesskit.dll from lib/windows/x86_64/msvc/shared into the accesskit folder and also into zig-out/bin. Using the SDL3 standalone example I can indeed see that some widgets are picked up by my screen reader.

DataTriny:
Here's how it looks inside Accessibility Insights for Windows:

Screenshot of the tool Accessibility Insights for Windows showing the accessibility tree produced by the sdl3-standalone example from dvui

david-vanderson:
Ah thank you!

Ok for my notes, this is what I did (on Windows 11 x86_64):
* download most recent accesskit release from https://github.com/AccessKit/accesskit-c/releases
* find in that zip the files in `lib/windows/x86_64/msvc/shared`
  * copy them to `dvui/accesskit`
  * also copy them to `dvui/zig-out/bin`
* run `zig build sdl3-standalone` (example runs)
* download accessibility insights from https://accessibilityinsights.io/downloads/
  * run it and mouse over parts of the sdl3-standalone example that is running

Note: it also worked for me to use the accesskit release files under `lib/windows/x86_64/mingw/shared`.  Don't know what the implications are.

DataTriny:
> Note: it also worked for me to use the accesskit release files under lib/windows/x86_64/mingw/shared. Don't know what the implications are.

They're built using different Rust toolchains:

- `lib/windows/x86_64/msvc/` contains binaries built using the x86_64-pc-windows-msvc Rust toolchain (MSVC compiler)
- `lib/windows/x86_64/mingw/` contains binaries built using the x86_64-pc-windows-gnu Rust toolchain (gcc compiler)

Binaries inside the mingw folder can be used to cross-compile Windows applications from Linux for instance or from MSYS2 with the [mingw-w64-zig](https://packages.msys2.org/base/mingw-w64-zig) package, I'm not sure they are fully interchangeable.

phatchman:
Yeah, sorry. I still need to fully integrate with the build system. sdl3-standalone is the correct one.

I've been linking the msvc static lib and it seems that works fine regardless of which libc zig links to. Also not sure of the implications of doing that, but have not seen any issues. 

@david-vanderson If you want to try the actions, select a button, slider or text box in Acessibility Insights. Then in the bottom right panel, you will see a label that you have to click on to expand. From there you can see the current values for the widget and run any actions.

Also, It should only be a couple of line change to get macos working. If that makes easier for you?

david-vanderson:
> Yeah, sorry. I still need to fully integrate with the build system. sdl3-standalone is the correct one.
> 
> I've been linking the msvc static lib and it seems that works fine regardless of which libc zig links to. Also not sure of the implications of doing that, but have not seen any issues.

When I copied either msvc or mingw static libs into dvui/accesskit then I got linker errors.  Is there something else I need to do?

> 
> @david-vanderson If you want to try the actions, select a button, slider or text box in Acessibility Insights. Then in the bottom right panel, you will see a label that you have to click on to expand. From there you can see the current values for the widget and run any actions.

Thank you!

phatchman:
> When I copied either msvc or mingw static libs into dvui/accesskit then I got linker errors. Is there something else I need to do?
> 

No, I just copied the lib. I built my accesskit  static lib from the accesskit-c repository with `cargo build --release`. I never tried using the dlls, so it's a big surprise to me that they work!

Cargo should build the same lib you are using. I'll have to try this later and see if I get the same linker issues.

phatchman:
@david-vanderson Was this the linker error?

```
error: lld-link: undefined symbol: _fltused
    note: referenced by accesskit.lib(compiler_builtins-793a3abeda5f8f32.compiler_builtins.3809829c2ae9b216-cgu.212.rcgu.o)
    note: referenced by accesskit.lib(accesskit.accesskit.e3324c1cffb2fc36-cgu.0.rcgu.o)
error: the following command failed with 1 compilation errors:
```
If so, that is indeed an issue with incompatible libc compilations. There is some info here: https://github.com/ziglang/zig/issues/24148.

We could just define that symbol as per compiler_rt and it should work as far as I know. The symbol just needs to be present in the exe.

I think ultimately, we want to try and build the access kit static libs with `zig cc`, so that we can match targets exactly. It gets built with CMake, so it shouldn't be that hard. But I didn't manage to get it working when I tried previously.

david-vanderson:
> @david-vanderson Was this the linker error?
> 
> ```
> error: lld-link: undefined symbol: _fltused
>     note: referenced by accesskit.lib(compiler_builtins-793a3abeda5f8f32.compiler_builtins.3809829c2ae9b216-cgu.212.rcgu.o)
>     note: referenced by accesskit.lib(accesskit.accesskit.e3324c1cffb2fc36-cgu.0.rcgu.o)
> error: the following command failed with 1 compilation errors:
> ```
> 
> If so, that is indeed an issue with incompatible libc compilations. There is some info here: [ziglang/zig#24148](https://github.com/ziglang/zig/issues/24148).
> 
> We could just define that symbol as per compiler_rt and it should work as far as I know. The symbol just needs to be present in the exe.
> 
> I think ultimately, we want to try and build the access kit static libs with `zig cc`, so that we can match targets exactly. It gets built with CMake, so it shouldn't be that hard. But I didn't manage to get it working when I tried previously.

Yes that was it!  I agree that makes sense as a path to shoot for, but if it's hard to get there we can stick with the prebuilt shared libraries for at least some time.

phatchman:
The latest push should support MacOS, so hopefully that helps. I just copied the MacOS x86_64 static lib into the accesskit directory and did `zig build`

However, I had to comment out the below as it was causing a panic due to an Optional unwrap in the rust code. The example code is for SDL2, so maybe SDL3 uses a different class name?
I would assume you should still be able to see the accessibility tree , even without that call.
```
        //ak.accesskit_macos_add_focus_forwarder_to_window_class("SDLWindow");
```
Frustratingly, the MacOS Accessibility Inspector and other tools I'd need seems to need an apple developer account to download and install it as part of the xcode developer tools. I've avoided doing this in the past as I'm not a big fan of their license agreement.

I'll see if I can find any good alternatives, otherwise I'll take another look and see if I can stomach Apple's agreement.

phatchman:
Do you think we can do anything to remove those dupeZ's? It's just to null terminate the strings before sending to accesskit. It is using the arena, so it's not as bad as it could be. Just thinking how bad it would be for large amounts of text.

It might not be obvious, but nodes are only created after an external app activates accessibility. So users not using a screen reader won't have any performance overhead.

DataTriny:
> Do you think we can do anything to remove those dupeZ's? It's just to null terminate the strings before sending to accesskit.

Someone else is already taking care of this for most places, see https://github.com/AccessKit/accesskit-c/pull/53.

We may change the API to allow this everywhere, there would be breaking changes around `accesskit_tree` and `accesskit_custom_action` though.

DataTriny:
> However, I had to comment out the below as it was causing a panic due to an Optional unwrap in the rust code. The example code is for SDL2, so maybe SDL3 uses a different class name?
> I would assume you should still be able to see the accessibility tree , even without that call.

The tree will be there, but there should be significant delay when handling input from user, unless SDL3 have changed the way they configure the window, in which case this call wouldn't be necessary.

DataTriny:
> I'll see if I can find any good alternatives, otherwise I'll take another look and see if I can stomach Apple's agreement.

I sometimes use https://github.com/pyatom/pyatom to reverse engineer macOS UIs, that's not a graphical tool I know, but it mostly do the job for me.

phatchman:
> > Do you think we can do anything to remove those dupeZ's? It's just to null terminate the strings before sending to accesskit.
> 
> Someone else is already taking care of this for most places, see [AccessKit/accesskit-c#53](https://github.com/AccessKit/accesskit-c/pull/53).

Yes, being able to pass a length with the pointer would be perfect!

And thanks very much. I'll give pyatom a try.

phatchman:
@david-vanderson Below is the list of actions I plan on supporting. This should give some idea of the events we might need.

And I just realised that I've messed up the commit history. You probably will need to force pull. Let me clean it up before you push anything. I tried a quick rebase, but I need to sit down and redo it properly. Edit: I ended up recreating it as a single commit. You'll need to force pull. Sorry.

|Action|DVUI Support|Comment|
|-------|---------------|-----------|
|CLICK|Y|Already supported|
|FOCUS|Y|Existing event for this|
|BLUR|N|We could technically support this, but not sure it is that useful? |
|COLLAPSE|N|Can just use click for this|
|EXPAND|N|Can just use click for this|
|CUSTOM_ACTION|N||
|DECREMENT|N||
|INCREMENT|N||
|HIDE_TOOLTIP|N||
|SHOW_TOOLTIP|N||
|REPLACE_SELECTED_TEXT|Y|Needs new event|
|SCROLL_DOWN|Y|? We don't have scroll events anymore|
|SCROLL_LEFT|Y|? We don't have scroll events anymore|
|SCROLL_RIGHT|Y|? We don't have scroll events anymore|
|SCROLL_UP|Y|? We don't have scroll events anymore|
|SCROLL_INTO_VIEW|N|There isn’t any accesskit property that allows you to set a widget as invisible. So I don’t see how we can support this, even if we did add nodes for invisible widgets.|
|SCROLL_TO_POINT|N|Not current supported in DVUI?|
|SET_SCROLL_OFFSET|Y|? We don't have scroll events anymore|
|SET_TEXT_SELECTION|Y|Needs new event|
|SET_SEQUENTIAL_FOCUS_NAVIGATION_STARTING_POINT|N|Not sure how much this is used or what it is for? But we will already support setting focus|
|SET_VALUE|Y|Almost already supported using .text event. But needs to be able to clear the previous value for things like text widgets. Not sure how this is supposed to play with SET_TEXT_SELECTION. Needs some testing.
|SHOW_CONTEXT_MENU|N||
|DATA_CUSTOM_ACTION|N||
|DATA_VALUE|N||
|DATA_NUMERIC_VALUE|N||
|DATA_SCROLL_UNIT|N||
|DATA_SCROLL_HINT|N||
|DATA_SCROLL_TO_POINT|N||
|DATA_SET_SCROLL_OFFSET|N||
|DATA_SET_TEXT_SELECTION|N||
||

phatchman:
@DataTriny I'm struggling to make scrollbars read-write.

The code below seems to set all the values I expect to see in AccessibilityInsights, except that the scrollbar is read-only so I can't send actions to it.

Any ideas on what might I'm missing? Thanks.

```
            .horizontal => {
                dvui.AccessKit.nodeSetLabel(ak_node, "Horizontal");
                dvui.AccessKit.nodeClearReadOnly(ak_node);
                dvui.AccessKit.nodeSetOrientation(ak_node, dvui.AccessKit.c.ACCESSKIT_ORIENTATION_HORIZONTAL);
                dvui.AccessKit.nodeSetNumericValue(ak_node, self.si.viewport.x);
                dvui.AccessKit.nodeSetMinNumericValue(ak_node, 0);
                dvui.AccessKit.nodeSetMaxNumericValue(ak_node, self.si.virtual_size.w);
                dvui.AccessKit.nodeSetNumericValueStep(ak_node, 1);
                dvui.AccessKit.nodeSetNumericValueJump(ak_node, 100);

                dvui.AccessKit.nodeAddAction(ak_node, dvui.AccessKit.Action.SCROLL_LEFT);
                dvui.AccessKit.nodeAddAction(ak_node, dvui.AccessKit.Action.SCROLL_RIGHT);
                dvui.AccessKit.nodeAddAction(ak_node, dvui.AccessKit.Action.SET_SCROLL_OFFSET);
                dvui.AccessKit.nodeAddAction(ak_node, dvui.AccessKit.Action.SCROLL_TO_POINT);
                dvui.AccessKit.nodeAddAction(ak_node, dvui.AccessKit.Action.SET_VALUE);
            },
```

david-vanderson:
> @david-vanderson Below is the list of actions I plan on supporting. This should give some idea of the events we might need.

Awesome.  Thanks for the heads up I'll force pull.

Event thoughts:
* blur - I agree, should probably punt on this for now
* for the scroll events I'm sure there is something we can do - could put a scroll type event where the receiving widget calls `scrollTo` or a new scrollOffset that works similarly
* I think scroll_into_view would be for widgets that are outside the clipping region (scrolled off the screen).  Does that mean we should add those to the tree?  We support that in a hacky way right now by checking `Window.scroll_to_focused` in WidgetData.register
* scroll_to_point - depending on how this works a dummy event that calls `scrollTo` seems like it fits the bill
* context menu and tooltip stuff I'm sure we can do but will have to figure out how to thread that through
* what are the DATA ones?

DataTriny:
> I'm struggling to make scrollbars read-write.

Does the node support the `CLICK` action? I was able to find a vertical scrollbar in the example but I can't even select it inside Accessibility Insights. Also, unless you really want the scrollbar to be scrolled to the left and the right, I think the actions you are looking for are increment and decrement, to change the scrollbar value (note that it doesn't matter on Windows where these two actions aren't supported).

DataTriny:
> I think scroll_into_view would be for widgets that are outside the clipping region

You're right, `accesskit_node_set_clips_children` should be called on containers that may only display a subregion of their content.

> what are the DATA ones?

These enum variants aren't actions, they come from the `accesskit_action_data` union to specify the type of additional information linked to the received action.

DataTriny:
@phatchman Maybe this isn't implemented yet: when I use the Tab key I can see visually that the buttons get the focus, however the focus doesn't seem to be reflected in the accessibility tree as my screen reader doesn't report anything.

david-vanderson:
@DataTriny I'm trying to wrap my head around when the functions get called.

My current understanding is:
* we call `accesskit_windows_subclassing_adapter_new` during app startup
  * `activation_handler` (callback) will be called on the same thread at some point in the future (when some accessibility feature first requests info?)
  * `action_handler` (callback) will be called (from unknown thread) any time an accessibility feature wants to perform an action
* we call `accesskit_windows_subclassing_adapter_update_if_active` each frame
  * `update_factory` (callback) is called immediately from the same thread to allow us to populate the tree
    * this callback will never be called outside of the extent of `accesskit_windows_subclassing_adapter_update_if_active`

Does that sound right?

DataTriny:
@david-vanderson You're mostly right.

- Keep in mind that the activation handler callback may be called on another thread on another platform than Windows.
- the update factory callback passed to `accesskit_windows_subclassing_adapter_update_if_active` will never be called if no assistive technology is running on the system at the time the function is called. To put it another way, the callback passed to this function will only be called if the activation handler callback was previously called. Therefore it is important to put as much computation related to tree building inside the callback so that users who don't need what AccessKit provides don't have to pay the performance cost.

phatchman:
@DataTriny 
> > I'm struggling to make scrollbars read-write.
> 
> Does the node support the `CLICK` action? I was able to find a vertical scrollbar in the example but I can't even select it inside Accessibility Insights. Also, unless you really want the scrollbar to be scrolled to the left and the right, I think the actions you are looking for are increment and decrement, to change the scrollbar value (note that it doesn't matter on Windows where these two actions aren't supported).

The code you were running just added scrollbars as a node, but it didn't set any properties or actions. I've just pushed some code that adds all the actions and properties that I could think of as relevant (including click), but it is still showing up as read-only for me.

> Maybe this isn't implemented yet: when I use the Tab key I can see visually that the buttons get the focus, however the focus doesn't seem to be reflected in the accessibility tree as my screen reader doesn't report anything.

Correct, that is a TODO. The focus node is currently always set to the root node. Easy fix.

>You're right, accesskit_node_set_clips_children should be called on containers that may only display a subregion of their content.

OK. 

Is there something similar for floating windows / widgets? I noticed that AccessibilityInsights is able to pick up widgets hidden behind the floating windows. Is there something I should be doing to stop that?

Edit: I missed this function in the api: `accesskit_node_is_hidden`. Is this what I need to be using?

Thanks again for your help.

phatchman:
> Does that sound right?

Once we're done, I'll put together a README or add something to the existing one on how the accessibility works and how to use it in DVUI. 

Things like the lifetime of the acesskit_node and your questions about what gets called when and from what thread took me a little while and some experimentation to fully understand. 

Also, can you take a look at the TextLayoutWidget changes in the latest commit? It appears to be only adding visible text to the tree, which was not by design. I believe DOCUMENT is the most appropriate role for this and tracks with other applications I've looked at.

I'm not 100% sure what we want to do for large text. I was thinking initially we just have an arbitrary cut-off of 1MB or something and drop any text after that point? Then if someone really has a large amount of text that they want to make accessible, they will need break it up as a document using the underling AccessKit Api and handle any paging etc manually. (I've not found any application so far that handles large amounts of text and accessibility well. So, I'm not sure what other good alternatives are).

For reference, vscode doesn't make the code editing pane accessible, but notepad and notepad++ do as Documents. (Which is more relevant to TextEntryWidget). But anything else that displays large amounts of read-only text (web browser, book reader) represents those as documents as best I can tell.

phatchman:
> Therefore it is important to put as much computation related to tree building inside the callback so that users who don't need what AccessKit provides don't have to pay the performance cost.

I don't quite do it like that, but it should do the same thing in a different way. 

All the callback does is create the update tree and push the already created nodes to the tree. I keep an `initialized` flag to tell whether access kit is initialized (i.e. was a request made to build the initial tree). I believe on Linux, there is a way to deactivate the accessibility, which will just do some cleanup and set initialized to false.

If that initialized flag is not set, no accessibility nodes are ever created.  If accessibility is enabled during the middle of the frame (i.e. the nodes at the start of the frame never get created because initialized was false), I force a refresh() to occur so we get another frame containing the full tree.

david-vanderson:
> Also, can you take a look at the TextLayoutWidget changes in the latest commit? It appears to be only adding visible text to the tree, which was not by design. I believe DOCUMENT is the most appropriate role for this and tracks with other applications I've looked at.

Yeah I can figure that out at some point, no problem.  Will think about it once I have the rest of this figured out.  I'm starting to get it, and it's looking pretty good!

Something happened with the commits where it's including backwards parts of other unrelated commits (like to src/Debug.zig) - let me know if you want me to manually separate those out and force push this branch.

phatchman:
> Something happened with the commits where it's including backwards parts of other unrelated commits (like to src/Debug.zig) - let me know if you want me to manually separate those out and force push this branch.

Doh! I can rebase my local changes to master if easier? But it won't affect me if you want to do that as, I do the dev on a different branch.

david-vanderson:
> Doh! I can rebase my local changes to master if easier? But it won't affect me if you want to do that as, I do the dev on a different branch.

Give that a shot - thanks!

phatchman:
Should be good now?  I'm not sure what I did before, but it wasn't good!

There was some manual merging required for dvui.zig, but I applied my changes on top of main. So, if anything got missed, it will be my changes.

david-vanderson:
> Should be good now? I'm not sure what I did before, but it wasn't good!
> 
> There was some manual merging required for dvui.zig, but I applied my changes on top of main. So, if anything got missed, it will be my changes.

I'm also confused.  That cleared most of it up, but a few recent commits to main got pulled in reversed (some in ScrollContainer and some in Menu and MenuItem).  Don't worry about it for now - let's get this into better shape and we can deal with the merge stuff once we are ready to merge it.

DataTriny:
Just compiled the latest commits from here and I can reproduce the scrollbar issue. I'm now pretty sure that it's a bug on our side, sorry.

Documents and floating windows aren't yet supported. Documents are going to require a lot of work in particular, please don't count on this for the near future.

phatchman:
> Documents and floating windows aren't yet supported. Documents are going to require a lot of work in particular, please don't count on this for the near future.

Would it be better to use DIALOG instead of WINDOW?

Edit: DIALOG doesn't seem to make a difference but isn't coming up as a DIALOG, it is showing as a pane. So, I'll probably stick with WINDOW for now unless there is a reason to change it.

The two issues I've noticed are:

1) With floating windows accessible widgets behind the top window can be selected.
2) With drop-downs (including menus), if there is an accessible widget behind the drop-down, it can be selected instead of the items in the drop-down.

@david-vanderson The best I can think of to resolve this is to check the bounds of each node against the rects of all the sub-windows and clip if it intersects with a subwindow that is not its own? 

I believe that would also fix (2) as drop-downs and menus are also subwindows.

david-vanderson:
> @david-vanderson The best I can think of to resolve this is to check the bounds of each node against the rects of all the sub-windows and clip if it intersects with a subwindow that is not its own?
> 
> I believe that would also fix (2) as drop-downs and menus are also subwindows.

@DataTriny What's your take on this problem (widgets in subwindows on top of other widgets)?
* To be clear "subwindow" is dvui terminology for something that looks and acts like a window but only exists inside of dvui.
* Are we using AccessKit incorrectly?
* Should we expect (now or in the future) to be able to have some widgets (representing subwindows) prevent/obscure widgets "underneath"?
* Or do we need to try and solve this on dvui's side (by somehow clipping/removing widgets from the tree if they are obscured by a subwindow on top of them)?

DataTriny:
No, dialogs aren't supported either, although they should arrive sooner than floating windows.

We ultimately plan to support as much features as are exposed by the various OS accessibility stacks, but we are far from it. I can't tell you right now how we are going to deal with masking widgets with a smaller Z-index.

DataTriny:
I tried building the bindings with zig cc, I couldn't do it with the MSVC toolchain. I got linking errors when creating the bindings with the gnu toolchain, and I got linking errors when building the SDL example with the gnullvm toolchain. Getting this to work would be the most reliable option for distributing AccessKit with dvui I think.

david-vanderson:
> No, dialogs aren't supported either, although they should arrive sooner than floating windows.
> 
> We ultimately plan to support as much features as are exposed by the various OS accessibility stacks, but we are far from it. I can't tell you right now how we are going to deal with masking widgets with a smaller Z-index.

Sounds good, we can experiment on dvui's side for now.

> I tried building the bindings with zig cc, I couldn't do it with the MSVC toolchain. I got linking errors when creating the bindings with the gnu toolchain, and I got linking errors when building the SDL example with the gnullvm toolchain. Getting this to work would be the most reliable option for distributing AccessKit with dvui I think.

Thanks for trying it.  I also think that is our best bet long term, but it doesn't need to hold up merging this.

I want to clean up some more of the dvui code and get it merged into main, and then we can keep improving it there.

Thank you!

phatchman:
The below sort of works well, but I'm lacking some knowledge to make it work correctly for multiple floating windows that overlap. I'm using this in createNode to set hidden on the node if it is obscured. 

The checking for the corners logic works well. It still allows visible parts widgets behind the floating window to be selected if than are wider or taller than the floating window (i.e. all 4 corners are oustide the bounds of the floating window) . We could clip a widget is partially obscured, but I was trying for something simple.

The issue is when multiple floating windows overlap. I can't work out how to tell which one is in front. I tried checking if the subwindow overlapped with the focused subwindow, but that doesn't work for FloatingMenus as they aren't focused (as best I can tell). There is also the stay_above_parent field, but that is not set for floating menus either).

But everything else seems to get hidden as required. 

I thought about just not creating the nodes for anything obscured, but there is one small benefit to hiding the node. In the case where we have a button with an image and label. The button may be obscured, but the image might not be. The node would then be created for the image, but it will get the wrong parent.

```
pub fn widgetObscured(self: *AccessKit, wd: *dvui.WidgetData) bool {
    const rect = wd.borderRectScale().r;
    const own_sw_id = self.window.subwindows.windowFor(rect.topLeft());
    const focused_sw_id = self.window.subwindows.focused_id;

    // Cannot be obscured if the widget's subwindow has focus.
    if (own_sw_id == focused_sw_id) return false;

    // 1.. Don't check main window as it contains everything.
    for (self.window.subwindows.stack.items[1..]) |*sw| {
        if (sw.rect_pixels.contains(rect.topLeft()) or
            sw.rect_pixels.contains(rect.bottomRight()) or
            sw.rect_pixels.contains(rect.topRight()) or
            sw.rect_pixels.contains(rect.bottomRight()))
        {
            return true;
        }
    }
    return false;
}
```

It at least helps a lot with my drop-down testing, so I can actually highlight the options, rather than the slider underneath.

phatchman:
@david-vanderson My last commit should properly integrate accesskit into the build process. It downloads from github and then links as a static library for Windows and MacOS. (I've not tested MacOS yet though).

I've also fixed the static linking error for windows x86_64.

If you can give it a test and see if it works for you.

Edit: The error messages could be a bit better. I'll fix those up.

Note to self: the logic for export if the _fltused symbol us wrong.

DataTriny:
@phatchman The downloading and compilation went well for me. However, the sdl3-standalone example is now completely broken if accessibility is enabled before the app starts. It happens when I have my screen reader running, but I am also able to reproduce if I open Accessibility Insights before launching the example. The window stays black and the app does not respond at all.

DataTriny:
The regression seem to have been introduced in 6a973ff00eab642ab7c590c63fecac302ad8be93 as the app behaves fine before that.

DataTriny:
Reading at the changes, I think I understand what is going on. The root node was apparently changed to have a role of generic container, which is not supported. Reading at the commit message, I'm guessing there is misunderstanding on how to use AccessKit in multi-window apps. You have to have one AccessKit adapter per native window, so one accessibility tree for each native window.

phatchman:
@david-vanderson My intention was to remove newFrame and have window add itsrelf as the root node (using nodeCreate) with the role WINDOW. That should keep the previous functionality, but stores the root node in Window's wd intead,

david-vanderson:
> Reading at the changes, I think I understand what is going on. The root node was apparently changed to have a role of generic container, which is not supported. Reading at the commit message, I'm guessing there is misunderstanding on how to use AccessKit in multi-window apps. You have to have one AccessKit adapter per native window, so one accessibility tree for each native window.

Aha yes this is definitely my fault - thanks for clearing that up, will fix!

david-vanderson:
> @david-vanderson My last commit should properly integrate accesskit into the build process. It downloads from github and then links as a static library for Windows and MacOS. (I've not tested MacOS yet though).
> 
> I've also fixed the static linking error for windows x86_64.
> 
> If you can give it a test and see if it works for you.

Works for me!

phatchman:
@david-vanderson Do you know if there is a way to tell if we are linking to libc from within the Zig code? The changes I made to enable WASM to build really should test if .link_libc = true instead.

Edit: Ahh don't worry. I found it in builtin. I still get it confused with std.builtin.

I'm hoping the C API mappings I've done in AccessKit.zig aren't going to be a pain to maintain. But without them you type accesskit at least three times for each function call.
```
    AccessKit.c.accesskit_node_new(ACCESSKIT_ROLE_WINDOW);
```

phatchman:
> That cleared most of it up, but a few recent commits to main got pulled in reversed (some in ScrollContainer and some in Menu and MenuItem). 

I've been through the changes and am fairly confident I have everything restored in this last commit. I'll do a double-check tomorrow to confirm.

P.S. i noticed the new .style options when reviewing this. I really like it and think it makes for a very consistent way to know what style a widget uses and also to override it. I haven't had a chance to try it yet though.

david-vanderson:
> Edit: Ahh don't worry. I found it in builtin. I still get it confused with std.builtin.

I was also confused by that a while back and dug into it - as far as I know they are the same.

> 
> I'm hoping the C API mappings I've done in AccessKit.zig aren't going to be a pain to maintain. But without them you type accesskit at least three times for each function call.

No problem - if maintaining them is annoying we can switch.

phatchman:
> No problem - if maintaining them is annoying we can switch.

Thoughts on this [code generator](https://gist.github.com/phatchman/000b1882611ee03c803b5c1e7b9c28e3) for the mappings?

It's a question of which is easier to maintain. But it turned out much less complicated than I initially thought and looks viable to me.
Given the bindings are being auto-generated on the rust side, there should not be much change required over time.

~~It generates something that can be pasted into AccessKit.zig and compiles with 3 exceptions:~~
~~1) Multi-part Enum names e.g. ActionData enum is incorrectly named with only the first part e.g. Action.~~
~~2) There are some decls that map to 'true' and 'false' that need to be changed to something that is not a keyword.  (This is only a problem, if we use lower snake_case for the enum identifiers).~~
~~3) The asU8() fn is not added for the Role enum.~~

~~But these are easy fixes.~~

Edit: It now produces output that can just be pasted at the end of AccessKit.zig and supports both ~~libc and non-libc~~ accesskit and non accesskit builds. There could be some lurking bugs, but it looks good to me.

Not sure it is really worth putting into the build to generate this. I'd probably just leave it as some stand-alone that can be compiled with zig build-exe and just run when we upgrade to a new version. Or maybe `zig build accesskit-mappings`?

Also, I've gone through and compared to main again. I found 1 line I'd missed in FloatingWindowWidget, which I've just pushed a commit for, but all the others look good. Please check again before you merge. Sorry for the mess.

phatchman:
Now I've got the ability to generate the api mappings, I can put all the enum values in lower case to make them more Zig.
Do you want me to hold off on that? It would touch a lot of files.

david-vanderson:
> Now I've got the ability to generate the api mappings, I can put all the enum values in lower case to make them more Zig. Do you want me to hold off on that? It would touch a lot of files.

Nah - that sounds like the way to go, so better to make the change now than later.  Go for it!

phatchman:
@DataTriny For some reason, even though everything looks good in accessibility insights, the Windows screen read can't read or select anything on the sdl3-standalone app. 

In the PR branch, it just reads the windows title. In my local branch where I've stopped our BoxWidget being added to the accessibility tree (it used to be a 'generic container', it reads out the window title and then says "Menu".

I've compared what our tree looks like in Insights compared to ones where the voice reader is working and I don't see anything different. Any suggestions you can think of would be appreciated.

The node tree we create for the main window is below. It looks OK to me? Menu is parented to Window, with menu items below it. The scroll view is also parented to the window and conntains the other widgets.

```
window (d6882e0b)
├─ menu (80fb88db)
│  ├─ menu_item (bd93dbc4)
│  │  └─ label (43a906ce)
│  └─ menu_item (67e1a097)
│     └─ label (d46d5e3e)
└─ scroll_view (97e96de1)
   ├─ document (4bb362f2)
   ├─ document (b7ac22a2)
   ├─ button (b504a939)
   │  └─ label (f377aae4)
   ├─ button (79741076)
   │  └─ label (6f3d1efc)
   ├─ button (65033c47)
   │  └─ label (1dfe7820)
   ├─ button (f6df7665)
   │  └─ label (cd1343d4)
   ├─ label (7f36dd84)
   └─ button (8e15d458)
      └─ label (21af0d38) 
```

DataTriny:
Hi @phatchman, I have tried your latest changes and everything seem to behave as expected on my side with both the Narrator and NVDA, on Windows 11.

> the Windows screen read can't read or select anything on the sdl3-standalone app.
> In the PR branch, it just reads the windows title.

Could you please clarify the actions you are performing once the window opens?

I don't recall if "Read and interact with the screen using a mouse" is enabled by default in Narrator's settings. If you're trying to use the Tab key for instance, I think the "focus" field of tree updates still isn't updated so you won't get any keyboard navigation. Also (and I know many UI toolkits miss this point), it is strongly advised to set the keyboard focus somewhere when the window opens, usually on the first focusable widget unless the UI author says otherwise.

Hope this helps.

DataTriny:
Once basic keyboard navigation is dealt with, having text inputs working would be great. Here's how they work:

- You pick the most appropriate role for your input, let's go with `ACCESSKIT_ROLE_MULTILINE_TEXT_INPUT`.
- For each span of text inside the input, you create a child node with role `ACCESSKIT_ROLE_TEXT_RUN`. In the case of simple text (rich text isn't supported yet) this means one child for each paragraph.
- On these children you set `accesskit_node_set_character_lengths`, `accesskit_node_set_character_positions`, `accesskit_node_set_character_widths`, `accesskit_node_set_text_direction` and `accesskit_node_set_word_lengths`.
- On the parent input node, use `accesskit_node_set_text_selection` to indicate where the caret/selection currently is.
- No need to call `accesskit_node_set_value` on the input node, it will be computed from the text runs.

phatchman:
@DataTriny Ah.. thanks. Yes, the selection with keyboard was to do with setting focus. I'd forgotten about that. 

~~Just checking there is no any way to get the parent of a Node via the C API? It's not a problem, would just make the setting focus bit of the code a little easier.~~ Found a better wsy to do it.

I had started experimenting with text runs, but haven't properly worked it out yet.  @david-vanderson I think you might be better to assess how we'd implement the character and word positions etc when you get to that part. We could use .text_run for both text layout and multi_line text entry. There are some good examples in the AccessKit tests [here](https://github.com/AccessKit/accesskit/blob/main/consumer/src/text.rs#L1220).

edit: The last commit should have fixes to the focus and various other updates. This makes things work much better with the narrator!

@david-vanderson This change makes the assumption that anything that can be focused will be in the accessibility tree. I think that should be valid.

phatchman:
@david-vanderson regarding nodeLabelFor(), I have a case, where I think that doesn't work.

See the advanced drop-down example:
```
if (g.choice) |c| {
    dvui.label(@src(), "Dropdown Choice {d}", .{c}, .{ .gravity_y = 0.5, .padding = .{ .x = 6, .w = 6 } });
} else {
    dvui.label(@src(), "Advanced Dropdown", .{}, .{ .gravity_y = 0.5, .padding = .{ .x = 6, .w = 6 } });
}
```
Those labels need to be used to label the drop-down, but I don't have a good way to do that because I don't have their Ids here. I'm going to experiment with ways to automatically pick up the first child label and use that if not already set. But even if that works, the user still needs a way to set the labelled by on the dropdown.

Any issues with introducing `for` into the `label` union to cover this case, rather than the nodeLabelFor function? 

Aside from supporting actions, the basic widgets seem to be functional. There is more testing and tweaking, but they seem usable.
I want to get the label text and text area text being read out and then I think I'm happy that the basic 'read' functionality is there and we've got something that is useful enough for someone to include.

Edit: With the latest commit you should be able to do a screen read (CAPS-R on Windows Narrator) on the Basic Widgets page and have it read the correct things. The shortcut for starting / stopping narrator is ctrl-win-enter. I also ran into a similar problem with labelling of the text box on the basic widgets page. That part needs some more thought.

It doesn't read the 'document' text from the home screen, so I'll just treat text layouts as labels for now, until we work out how to do the text run thing.

I think it makes sense for me to make any further widget changes on a local branch, so you can get something in shape to merge.

david-vanderson:
> Any issues with introducing `for` into the `label` union to cover this case, rather than the nodeLabelFor function?

Go for it - sounds like a reasonable need.

> 
> Aside from supporting actions, the basic widgets seem to be functional. There is more testing and tweaking, but they seem usable. I want to get the label text and text area text being read out and then I think I'm happy that the basic 'read' functionality is there and we've got something that is useful enough for someone to include.

Awesome!  I'll get to it soon!

phatchman:
I added some more labeling options. Ideally, I'd like a better naming for the .prev and .next, but I can't think of anything at the moment. 

If you are wondering about `.by` => `.by_id`, it is because of the keyword clash with `.for`, so it had to be `.for_id`.

`.label = .{ .label_widget =  .next } }` is what I came up with to allow advanced drop down to work without the user having to specify a labeling widget. If no label is supplied, the drop-down will be labelled by the next created node with role = .label. 


It's simple and there could be cases where it picks the wrong label, but in the vast majority of cases, it will be correct and better than having no label. And the user can override in the cases it is wrong.

.prev helps with those cases where you don't have the labelling widget's id. e.g. basic_widget's text entry.

> Awesome! I'll get to it soon!

No rush from my side. For this branch, I plan to do one more commit to clean up the `catch @panic ("TODO");` lines / assertions etc and do some testing with actions on the basic widgets. If there is anything else needed, let me know.

david-vanderson:
@phatchman I think this is ready to be merged and further work can happen in main or in followup PRs.  Is there anything else you wanted to do before merging?

If not, I'm going to look through everything one last time to check for potential merge conflicts and then merge it.

@phatchman and @DataTriny THANK YOU so much for getting this to this point and helping me as I blunder through understanding it.  This is important and I am very grateful!

phatchman:
Yeah. I'm happy to merge once you've reviewed.

phatchman:
Awesome. Excited about this getting in. Thanks for getting it into such good shape.

One bug I'm struggling to work out the correct solution for. For drop downs / menus, when they are dropped down by using the keyboard, I don't think any widget gets focus? The focus ends up being set to the main window, because nothing is focused. If you then press the down-arrow, it will highlight the first item.

If you use the Narrator and try and drop-down the advanced drop down or the File/Edit menus with the space bar, you will see the issue. And it is a similar the same issue when you drop it down via an action, the widget loses focus and shifts to the main window.

Does it make sense to change it so that if there's no currently selected value, the selection / focus default to the first value? That way something is always selected and the reader can read it out as soon as the menu expands? The normal drop-down works because it always focuses the currently selected value.

david-vanderson:
Yes I see that.  We need to change something, not totally sure what.  Will experiment.

david-vanderson:
@phatchman Can you verify if https://github.com/david-vanderson/dvui/commit/b685eab618c02edd2e96b3a58f8daa3f1436a6e6 fixes the menus/dropdowns issue you saw?

phatchman:
Yes, works great.

PR #596, opened 2025-10-08
Add ellipsized field to track ellipsizing.

br07h3rj:
Mind if we add an "ellipsized" field to show whether some label is being trimmed or not? I'm not sure if this is how you'd like it implemented, but it works for me so that I can do things like:

```
    var filename = LabelWidget.initNoFmt (@src(), entry.name, .{}, .{ .id_extra = dir_num });
    filename.install();
    filename.draw();

    if (filename.ellipsized) { std.debug.print ("Ellipsized: {s}\n", .{ entry.name }); }

    filename.deinit();
```
And it will dynamically show me which filenames are being trimmed as I drag the window edge to make it wider and narrower.

### DIFF
--- a/src/widgets/LabelWidget.zig
+++ b/src/widgets/LabelWidget.zig
@@ -29,6 +29,7 @@ label_str: []const u8,
 /// An allocator to free `label_str` on `deinit`
 allocator: ?std.mem.Allocator,
 init_options: InitOptions,
+ellipsized: bool,
 
 pub fn init(src: std.builtin.SourceLocation, comptime fmt: []const u8, args: anytype, init_opts: InitOptions, opts: Options) LabelWidget {
     comptime if (!std.unicode.utf8ValidateSlice(fmt)) @compileError("Format strings must be valid utf-8");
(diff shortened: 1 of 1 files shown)


david-vanderson:
Good idea - resolve @RedPhoenixQ review and he can merge when you're done.  Thanks!

br07h3rj:
You're right. I'm just slow to remove anything somebody else added to somebody else's codebase, so I tend to add more than subtract. Maybe that's a bad habit now that I'm on GitHub. I'll pay more attention to that on here. :smile:

david-vanderson:
> You're right. I'm just slow to remove anything somebody else added to somebody else's codebase, so I tend to add more than subtract. Maybe that's a bad habit now that I'm on GitHub. I'll pay more attention to that on here. 😄

That is a totally reasonable approach!  Might be more a reflection of how new dvui is than anything else.

br07h3rj:
Okay. I think that's it. I'm still having trouble figuring out GitHub.

david-vanderson:
Thank you!

PR #601, opened 2025-10-11
SDL: integrate sdl logging with zigs std.log scopes

RedPhoenixQ:
This enables us to get information about the sdl version and the current render being used directly from sdls logs when debugging. Also allows for easy capturing of the sdl logs along with the application logs to a file or with any other custom zig `std.options.logFn`.

The log level will be set to the same value as the `.SDLBackend`, unless specifically overwritten.

Something similar can be set up for raylib, probably not for dx11 (at least not as easily).


david-vanderson:
Awesome!  Merge when ready.  Thank you!

PR #602, opened 2025-10-11
Raylib: integrate raylib logging with zigs std.log

RedPhoenixQ:
Mirrors #601

Raylibs logging callback relies on the print function from `<stdio.h>` so we make a call to `vsnprintf` to a buffer and then write that out. It's unfortunate that long messages might be truncated but the buffer is quite large.


RedPhoenixQ:
Before this is merged, someone with access to a working linux machine should test that the callback doesn't fail at runtime with the workaround for `c.va_list` from d5282e6.

david-vanderson:
I've confirmed this works on my linux machine and my mac machine.  Merge when ready.  Thank you!

PR #603, opened 2025-10-12
tracked hashmap: add optional delayed reset

RedPhoenixQ:
This is an idea that makes rapid-ish showing and hiding of widgets feel snappier, and also reduce the need to re-create textures.

We already stored a boolean for if an item was used, which now has been upgraded to a count. When the reset is delayed we count up every time the item wasn't used and only remove the items once we reach some threashhold (arbitrarily set at 1000 for now, which is a handful of seconds at 60-240Hz).

This makes the debug web build feel very snappy when rapidly switching between the examples in the demo window. The main scenario being optimized for are items that open and close often in a short time. It also allows for the multi-frame animation example to keep the textures cached, which would be very impactful for larger/more complex animations.

I've not thoroughly thought through or tested all scenarios where this might cause issues. One thing would be a menu that opens and closes fast would keep the data set with `dvui.dataSet` between the opening and closing. Also, `dvui.firstFrame` wouldn't always trigger with this system, but the tracking hashmap could track this state.

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -74,6 +74,7 @@ pub const demoKind = enum {
 };
 
 pub var demo_active: demoKind = .basic_widgets;
+var invalidate_demo: ?demoKind = null;
 pub const demo_window_tag = "dvui_example_window";
 
 pub fn demo() void {
@@ -131,7 +132,6 @@ pub fn demo() void {
 
         var fbox = dvui.flexbox(@src(), .{}, .{ .expand = .both, .min_size_content = .width(width), .corner_radius = .{ .w = 5, .h = 5 } });
         defer fbox.deinit();
-
         inline for (0..@typeInfo(demoKind).@"enum".fields.len) |i| {
             const e = @as(demoKind, @enumFromInt(i));
             var bw = dvui.ButtonWidget.init(@src(), .{}, .{
@@ -152,7 +152,9 @@ pub fn demo() void {
             const use_cache = true;
             var cache: *dvui.CacheWidget = undefined;
             if (use_cache) {
-                cache = dvui.cache(@src(), .{ .invalidate = invalidate }, .{ .expand = .both });
+                cache = dvui.cache(@src(), .{ .invalidate = invalidate or (invalidate_demo == e) }, .{ .expand = .both });
+                // Only invalidate once
+                if (invalidate_demo == e) invalidate_demo = null;
             }
             if (!use_cache or cache.uncached()) {
                 const box = dvui.box(@src(), .{}, .{ .expand = .both });
@@ -201,6 +203,7 @@ pub fn demo() void {
             bw.drawFocus();
 
             if (bw.clicked()) {
+                if (demo_active != e) invalidate_demo = demo_active;
                 demo_active = e;
                 if (paned.collapsed()) {
                     paned.animateSplit(0.0);
@@ -216,6 +219,7 @@ pub fn demo() void {
             defer hbox.deinit();
 
             if (paned.collapsed() and dvui.button(@src(), "Back to Demos", .{}, .{ .min_size_content = .{ .h = 30 }, .tag = "dvui_demo_window_back" })) {
+                invalidate_demo = demo_active;
                 paned.animateSplit(1.0);
             }
 
--- a/src/Font.zig
+++ b/src/Font.zig
@@ -226,7 +226,7 @@ pub const Cache = struct {
     database: std.AutoHashMapUnmanaged(FontId, TTFEntry) = .empty,
     cache: dvui.TrackingAutoHashMap(u64, Entry, .{ .tracking = .get_and_put }) = .empty,
 
-    pub const default_timeout = dvui.RemovalTimeout.from_micros(std.time.us_per_min);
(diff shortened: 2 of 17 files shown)


david-vanderson:
This is very interesting, I will think about it.  You are right that the implications are different for each usecase.  We could also think about attacking it from the other direction - specify at the widget level which things are likely to be reused.  Hmmmm.

david-vanderson:
After thinking about it a bunch, I think we need more specific experience before deploying this generally.  For system reasoning, it's useful to have a general default of "if we don't see/touch it for a frame, it gets freed".

For the demo window case, the problem is a stutter when we have to regen all the demo button textures.  We know the user is going to go back to them "soon".

My guess is this implies 2 things:
* the delay should be expressed in time, not frames (similar to animations maybe, which would give a max delay of about half an hour)
* each texture might want a different delay
  * demo buttons are very likely to be seen again after a few minutes delay
  * icon browser textures likely in a short window, like 10 seconds (although they are fast enough to generate that it's probably not worth it)

Also worth thinking of other ways to deal with this situation.  Maybe we could rework the demo so it always refreshes the demo button textures even if they aren't being shown?  We should at least try that to understand how annoying it would be.

A separate solution would be to spread out the texture generation (so do only one button per frame).  That would also fix the stutter on the very first frame the buttons are shown.

RedPhoenixQ:
I haven't had that much time lately but will pick this up sometime this week. Thank you for taking the time to think through this with me.

> * the delay should be expressed in time, not frames (similar to animations maybe, which would give a max delay of about half an hour)
> * each texture might want a different delay
>   
>   * demo buttons are very likely to be seen again after a few minutes delay
>   * icon browser textures likely in a short window, like 10 seconds (although they are fast enough to generate that it's probably not worth it)

I definitely agree on moving the delay to the caller. Expressing it in time is probably fine, we can attach a timestamp in the future and make the `reset` take in the current time. 

> Also worth thinking of other ways to deal with this situation. Maybe we could rework the demo so it always refreshes the demo button textures even if they aren't being shown? We should at least try that to understand how annoying it would be.

Agreed. Maybe just some additional convenience functions would make this approach simple enough

> A separate solution would be to spread out the texture generation (so do only one button per frame). That would also fix the stutter on the very first frame the buttons are shown.

I'd prefer not to do that as it would look like dvui wasn't performing very well (in ReleaseFast it wouldn't need the delay).

j-h-a:
The issue is that the framework can't guess when/what to rasterize, or when to invalidate that cache. So the only way I know of to handle this is that the framework provides some simple functionality, and the app must use it correctly. On iOS it does this pretty well, so that's worth looking up.

I don't have time to double-check against the docs right now (and the implementation detail was never in the docs anyway), so this might be a little wrong as it's from memory...

* There was a `shouldRasterize` flag on the View's underlying `CALayer`, and there was a `needsRedraw` flag on the View.
* The built in properties relating to display/presentation would set `needsRedraw` for you when you changed them.
* If you used any custom drawing you'd need to set it yourself when something changed.
* Whenever `needsRedaw` was set, it would bubble backwards up the hierarchy and invalidate any rasterized parents.
* When a view was off-screen, its rasterized buffer would be kept, but the framework might decide to remove it, knowing it's a cache and can be rebuilt if needed. I think this used the `didReceiveMemoryWarning` to do this - if the view/layer knew it was currently off-screen it would drop the cache on memory warning.

In any case - I think this is something which can only be achieved by coordination between the framework and the app using it.

david-vanderson:
Thanks for the great information!

> In any case - I think this is something which can only be achieved by coordination between the framework and the app using it.

I agree, and this makes me think a good step to try would be adding a boolean flag to just the texture stuff that means either:
* junk this texture if not used for a frame (default, and matches current behavior)
* keep this texture forever (user code is responsible for junking)

That would solve the immediate "demo button" issue, and we can get some experience with it until more complicated usecases show up.  This is on my list to try.

david-vanderson:
> That would solve the immediate "demo button" issue, and we can get some experience with it until more complicated usecases show up. This is on my list to try.

I've merged an initial solution to this.  It was much more complicated than expected:
* we can't just keep the textures, we also must keep related data
* the demo wants to get rid of the textures/data when the demo window stops being shown
  * this ties nicely into `dvui.dataSetDeinitFunction()` - a data that is being freed can call `retainClear` on the rest
* datas can be accessed from multiple threads, so have to deal with that when calling deinit functions
* data deinit functions are called during `Window.deinit()` so need to have a current_window set then

All in all I think this is a decent direction to go in, but I'm worried it's too complicated.  But I can't think of anything else without pushing a lot of complexity into usage code.  So we'll run with it for a while and see how it goes.

It at least does fix the stutter when going back and forth in the demo window.

RedPhoenixQ:
Thank you for picking up this partially abandoned PR. I haven't done much work with dvui in a while. 

The deinit function solution provides much of the same functionality and more, maybe allowing for storing array lists easier with the dvui.data api. Thanks again

david-vanderson:
> Thank you for picking up this partially abandoned PR. I haven't done much work with dvui in a while.

No problem!  I totally understand - your giant contributions are super appreciated and you are always welcome!

PR #606, opened 2025-10-16
AccessKit: Support more backends

phatchman:
@RedPhoenixQ Might need your help with the DX11 one. If you have any insights on the others, also appreciated.

SDL2 now works on x86_64 Windows, but not aarch64. There is some duplicate mingw symbol, which is weird since we are linking to an msvc library. I'll look into that one more, but likely an upstream issue that I need to raise.

Raylib compiles, but fails to link with these duplicate symbol errors. Not sure if there is any way to get Zig to prefix the Raylib symbols so they don't clash?
```
error: lld-link: duplicate symbol: CloseWindow
    note: defined at C:\Users\phatc\AppData\Local\zig\p\raylib-5.5.0-whq8uLZNxwQbz-EkzfvEctRpeTg8Nk5WZKm6lJ-J8SQt\src\rcore.c:745
    note:            raylib.lib(rcore.obj)
    note: defined at accesskit.lib(user32.dll)
error: lld-link: duplicate symbol: ShowCursor
    note: defined at C:\Users\phatc\AppData\Local\zig\p\raylib-5.5.0-whq8uLZNxwQbz-EkzfvEctRpeTg8Nk5WZKm6lJ-J8SQt\src\platforms\rcore_desktop_glfw.c:1040
    note:            raylib.lib(rcore.obj)
    note: defined at accesskit.lib(user32.dll)
```
The good news is that if I rename those 2 symbols in the raylib source, accesskit works as expected. And we do have the option of dynamic linking if we can't get the static linking working easily.

DX11 - The issue is that accessing self.backend.impl for the dx11 backend segfaults. Specifically it segfaults on line 948 of Window.zig
```  if (dvui.accesskit_enabled and self.backend.impl.ak_initialize_in_begin) {```

I don't quite understand how the HWND / Context thing is used there as both a window handle and a self. But something about it doesn't seem to work when setting windows.impl with that "pointer". I guess it relies on dx11 not having any fields, and I've added one.

Edit: Ahh I figured it out. So, it can't have any members, which means the new ak_initialize_in_begin flag that was introduced won't work. I'll use a work-around for now and we can review later.

### CODE REVIEW

**Comment by RedPhoenixQ** [comment]
> DX11 - The issue is that accessing self.backend.impl for the dx11 backend segfaults. Specifically it segfaults on line 948 of Window.zig
> ` if (dvui.accesskit_enabled and self.backend.impl.ak_initialize_in_begin) {`

On dx11, self.backend.impl is just a pointer casted version of the HWNDfrom win32. It doesn't hold the state itself so we have to go through an attached USER_DATA pointer. 

> I don't quite understand how the HWND / Context thing is used there as both a window handle and a self. But something about it doesn't seem to work when setting windows.impl with that "pointer". I guess it relies on dx11 not having any fields, and I've added one.

Exactly, *Dx11Backend is just HWND and thus has no fields but can be used to get a pointer to the fields.


@phatchman  I'm working on moving more of the backend-specific logic to the `Backend` api and I'll push the changes to this branch.

**Comment by RedPhoenixQ** [comment]
The `_fltused` windows workaround didn't work for me so I haven't been able to test the sdl and raylib builds.

**Comment by david-vanderson** [comment]
I think this is an improvement and we should merge it.  I don't know what to do about the raylib duplicate symbols.

**Comment by phatchman** [comment]
> The `_fltused` windows workaround didn't work for me so I haven't been able to test the sdl and raylib builds.

Can you try building with -Dtarget=x86_64-windows ?

**Comment by RedPhoenixQ** [comment]
> Can you try building with -Dtarget=x86_64-windows ?

Still getting errors from the examples, not from the dvui test builds though

```
test
└─ run test dx11-ontop-test
   └─ compile test dx11-ontop-test Debug x86_64-windows 1 errors
error: lld-link: undefined symbol: _fltused
    note: referenced by accesskit.lib(compiler_builtins-793a3abeda5f8f32.compiler_builtins.3809829c2ae9b216-cgu.212.rcgu.o)
error: the following command failed with 1 compilation errors: 
    ...
test
└─ run test dx11-standalone-test
   └─ compile test dx11-standalone-test Debug x86_64-windows 1 errors
error: lld-link: undefined symbol: _fltused
    note: referenced by accesskit.lib(compiler_builtins-793a3abeda5f8f32.compiler_builtins.3809829c2ae9b216-cgu.212.rcgu.o)
error: the following command failed with 1 compilation errors: 
    ...
...
Build Summary: 10/15 steps succeeded; 2 failed; 143/212 tests passed; 69 skipped
test transitive failure
├─ run test dx11-standalone-test transitive failure
│  └─ compile test dx11-standalone-test Debug x86_64-windows 1 errors
└─ run test dx11-ontop-test transitive failure
   └─ compile test dx11-ontop-test Debug x86_64-windows 1 errors
```

**Comment by DataTriny** [comment]
Hi folks!

Ideally, the DirectX11 backend should use `accesskit_windows_adapter` instead of `accesskit_windows_subclassing_adapter`. The "subclassing" adapters are built on-top of hacks needed when you don't control the event loop (Windows) or can't directly respond to the callbacks on the view (macOS). The DirectX11 backend however, handles the messages posted to the window in its `wndProc` function and therefore can handle the `WM_GETOBJECT` message that is the central piece of accessibility on Windows. We have an example demonstrating how to properly handle accessibility inside a Win32 window [here](https://github.com/AccessKit/accesskit-c/blob/main/examples/windows/hello_world.c). This path also does not force you to create the window hidden before you create the adapter, as long as you can catch the first `WM_GETOBJECT` message.

**Comment by phatchman** [comment]
> Still getting errors from the examples, not from the dvui test builds though
> 
That is so wierd. 

On both my Win10 and Win11 x86_64 pcs, everything builds fine with accesskit enabled, except for wasm and raylib. I don't need to specify a target.  And the same if I cross-compile from aarch64 to x86_64. I tried under both 0.15.1 and 0.15.2.

So not sure what the difference could be? Hard to see it being a cache issue, but worth a try to clear it.
```
PS C:\src\dvui-dev> zig env
.{
    .zig_exe = "...",
    .lib_dir = "...",
    .std_dir = "...",
    .global_cache_dir = "...",
    .version = "0.15.1",
    .target = "x86_64-windows.win11_dt...win11_dt-gnu",
    .env = .{
        .ZIG_GLOBAL_CACHE_DIR = null,
        .ZIG_LOCAL_CACHE_DIR = null,
        .ZIG_LIB_DIR = null,
        .ZIG_LIBC = null,
        .ZIG_BUILD_RUNNER = null,
        .ZIG_VERBOSE_LINK = null,
        .ZIG_VERBOSE_CC = null,
        .ZIG_BTRFS_WORKAROUND = null,
        .ZIG_DEBUG_CMD = null,
        .CC = null,
        .NO_COLOR = null,
        .CLICOLOR_FORCE = null,
        .XDG_CACHE_HOME = null,
        .HOME = null,
    },
}
```

**Comment by RedPhoenixQ** [comment]
Running `zig build dx11-standalone -Daccesskit` directly works, but `zig build test -Daccesskit` fails with the error above. 

Also I get (cannot get any more info from this error):
```bash
$ zig build dx11-ontop -Daccesskit
dx11-ontop
└─ run exe dx11-ontop failure
error: the following command exited with error code 9:
"...\\zig-out\\bin\\dx11-ontop.exe"

Build Summary: 7/9 steps succeeded; 1 failed
dx11-ontop transitive failure
└─ run exe dx11-ontop failure

error: the following build command failed with exit code 1:
```

The window flashes on the screen before exiting, so it did compile and run. I don't have any debugger setup and I'm not getting any logging.

phatchman:
> The window flashes on the screen before exiting, so it did compile and run. I don't have any debugger setup and I'm not getting any logging.

Same. 

Maybe we disable accesskit for ontop builds for now. For on-top we'd potentially have two different accessibility trees for the application (one for the native application and one for DVUI windows). I've not really looked into if that is something that can be supported, but I doubt people will use accesskit often for on-top style applications?

Though I just tested and the narrator does work for sdl2 and sdl3 on top, so that's probably not the issue. I'll try and trace it using a debugger later.

phatchman:
> I don't know what to do about the raylib duplicate symbols.

I did a bit of research on this I don't think there is an easy answer right now. There is no way to rename or drop the symbol on windows without using llvm-objcopy or equivalent platform-specific tools.

There are a few requests on the raylib issues to allow prefixes to the API names but none are accepted. There are forks of raylib that do apply a prefix to avoid a collision and a shell script to do the same thing.

But, ultimately I think we want the accesskit static lib to dynamically link to user32 and that will avoid this issue. So when/if we get to vendoring our own accesskit-c libs, we can look at that further.

I've got a work-around for the issue with sdl2 on aarch64.

phatchman:
This introduces an option for dynamic linking, so raylib users can have a solution.
Want to see if I can 1) get DX11 working on the non-subclassing adapter and 2) track down the dx11 on-top crash below:
(not looked into it yet). 3) The cImport error for wasm.
```
Process 29856 stopped
* thread #1, stop reason = Exception 0xc0000409 encountered at address 0x7ff747d91a4c
    frame #0: 0x00007ff747d91a4c dx11-ontop.exe`std::panicking::rust_panic [inlined] std::sys::pal::windows::abort_internal at library\std\src\sys\pal\windows\mod.rs:346
```
```
* thread #1, stop reason = Exception 0xc0000409 encountered at address 0x7ffab6834048
  * frame #0: 0x00007ffab6834048 accesskit.dll`accesskit_rect_is_empty + 116952
    frame #1: 0x00007ffab6832fc4 accesskit.dll`accesskit_rect_is_empty + 112724
    frame #2: 0x00007ffab68386e8 accesskit.dll`accesskit_rect_is_empty + 135032
    frame #3: 0x00007ffab6838684 accesskit.dll`accesskit_rect_is_empty + 134932
    frame #4: 0x00007ffab6838d18 accesskit.dll`accesskit_rect_is_empty + 136616
    frame #5: 0x00007ffab683bd9c accesskit.dll`accesskit_rect_is_empty + 149036
    frame #6: 0x00007ffab6813290 accesskit.dll`accesskit_windows_subclassing_adapter_new + 412
    frame #7: 0x00007ff6397f628c dx11-ontop.exe at AccessKit.zig:54
```

DataTriny:
If the *on-top examples are responsible for creating the window, then they are also responsible for installing the AccessKit adapters into it. If they are meant to showcase how dvui can render into a window it doesn't own, potentially alongside other stuff the user could draw around or on-top of dvui, then that is more complicated in terms of accessibility. dvui would have to expose its accessibility tree and let the user merge it with their own. Another option is the concept of "sub trees" that would be part of AccessKit itself, but that is a difficult problem that we haven't figured out yet. I would find it reasonable to document that dvui accessibility only works when it owns the window for now.

david-vanderson:
@DataTriny I agree - the ontop examples are really for people using dvui in a "debugging stuff ontop of a game" kind of situation.  We shouldn't put much effort into getting accesskit working easily there until/unless somebody shows up with a real usecase and we can figure it out then.

david-vanderson:
> Running `zig build dx11-standalone -Daccesskit` directly works, but `zig build test -Daccesskit` fails with the error above.

I think this is because when compiling for tests it's not including the _fltused export.  I'm not sure how to fix it though.  I also get that error when running `zig.exe build -Daccesskit=static test`

phatchman:
> I think this is because when compiling for tests it's not including the _fltused export. I'm not sure how to fix it though. I also get that error when running `zig.exe build -Daccesskit=static test`

We can just disable accesskit for tests? I'll try this later.

phatchman:
The last commit should make the DX11 work with the non-subclassing adapter. 

@RedPhoenixQ A couple of things:

1) I pretty much just copied the logic from the accesskit-c windows  example. Could you please sanity check?
2) dx11-ontop is now super broken. It won't run even without accesskit linked and displays a Panic about an incorrect alignment in a message box. I'm done for today, so if you have some ideas, it would be appreciated. Otherwise, I'll look at it next time.

Also are you happy with how I'm setting / using global.pending_close? I need something to stop the WM_GETOBJECT from calling into accesskit when the hwnd is no longer valid.

RedPhoenixQ:
> 1. I pretty much just copied the logic from the accesskit-c windows  example. Could you please sanity check?

The `WM_GETOBJECT` looks good to me. I don't have a way to test that accesskit is actually working, other than it not crashing and logging that is initialized.

> 2\. dx11-ontop is now super broken. It won't run even without accesskit linked and displays a Panic about an incorrect alignment in a message box. I'm done for today, so if you have some ideas, it would be appreciated. Otherwise, I'll look at it next time.

Unrelated issue that surfaced now, fixed in 40ed3f6f8f54778d19742460e1bbf46295f77ebd.

I think this is pretty good now. I don't think the test need to run with accesskit enabled, perhaps they should even be completely disabled when it is?

DataTriny:
With latest changes from @RedPhoenixQ I am able to compile and navigate the accessibility tree from the dx11-standalone example. Just one major issue: screen readers won't announce the window when it opens if the adapter is told that the window doesn't have the focus. Doing Alt-Tab twice correctly announces the window and the focused widget inside.

DataTriny:
I'd be curious to know what the testing backend is used for exactly. I'm working on a new adapter that would store manually verified tree updates and check for regressions based on those saved files. Is this something you'd be interested in? This library doesn't depend on a windowing layer or even on any operating system's accessibility stack and can run on a CI.

RedPhoenixQ:
> I'd be curious to know what the testing backend is used for exactly. I'm working on a new adapter that would store manually verified tree updates and check for regressions based on those saved files. Is this something you'd be interested in? This library doesn't depend on a windowing layer or even on any operating system's accessibility stack and can run on a CI.

That is essentially the same purpose that the testing backend serves. It's just a shell that exposes the same api as a real backend. It doesn't do that much in terms of verification itself but it lets the rest of the dvui code run without an OS dependency. We use it if CI for example.

We do have some very basic snapshot testing capabilities and it sounds like the new adapter you mentioned would slot nicely into that. I myself would like something like that to verify that the implementation works as I don't have/know how to set up a proper screen reader for testing.

RedPhoenixQ:
> Just one major issue: screen readers won't announce the window when it opens if the adapter is told that the window doesn't have the focus. Doing Alt-Tab twice correctly announces the window and the focused widget inside.

I changed our default behaviour to assume we are focused in 73b4248aa277dfcdaced8215a11ab1b7be7e8f34. I'm unsure what win32 api we could use to verify this but hopefully this default is fine.

DataTriny:
@RedPhoenixQ Assuming the window is focused when it opens is reasonable, if dvui doesn't allow creating the window hidden or minimized.

src/WidgetData.zig
```diff
@@ -376,7 +376,9 @@ pub fn iterator(self: *const WidgetData) ParentIterator {
     return .{ .wd = self };
 }
 
-pub inline fn accesskit_node(self: *WidgetData) ?*dvui.AccessKit.Node {
+// NOTE: `inline` is required so that null will be returned at comptime. Otherwise other
```
david-vanderson: Sorry about that - I should have put a comment like that when I wrote it.

RedPhoenixQ: No problem! Was not too hard to find

phatchman:
`zig build -Daccesskit=shared` now builds without error on Windows and MacOS
-static fails only raylib for Windows because of the reasons previously mentioned, but works on MacOS.

~~I can't find a way to get the window pointer on SDL2 for MacOS?~~ Never mind. It's in the accesskit-c sample.
Raylib should be working on MacOS, but I can't test as it fails trying to use opengl on my vm.
All on-top's look to be good as well. No changes needed.

Happy for this to be merged. 

Will add the Linux support after the widget support is completed.

phatchman:
> The `WM_GETOBJECT` looks good to me. I don't have a way to test that accesskit is actually working, other than it not crashing and logging that is initialized.
> 
On windows, press ctrl-win-enter to turn the narrator on or off.
Then click in the window, use tab etc and check the reader reads something.
CAPS-r to read the whole screen.

david-vanderson:
Awesome thank you!!

PR #608, opened 2025-10-17
textlayout: enforce valid utf-8 before handling caching

### LINKED ISSUE #607: `TextLayoutWidget` panics on invalid UTF-8 text when `.cache_layout = true`
I found this when trying to dump some mixed test/binary data to screen for debugging purposes.  
This isn't an issue for me now (calling `dvui.toUtf8()` first is a better approach), but it took me a while to realize what I'd done wrong.

```zig
pub fn appFrame() !dvui.App.Result {
    var text_entry = dvui.textLayout(@src(), .{ .cache_layout = true }, .{});
    defer text_entry.deinit();
    text_entry.addText("test string with invalid UTF-8: \xc3", .{});
    return .ok;
}
```

```
debug(dvui): TextLayoutWidget forcing cache_layout false due to width change while break_lines
thread 14906088 panic: reached unreachable code
/Users/jamie/.zvm/0.15.2/lib/std/debug.zig:559:14: 0x101d6e043 in assert (dvui_test)
             ^
/Users/jamie/.cache/zig/p/dvui-0.4.0-dev-AQFJmXGQ1QC8SDZ6N9lNLrYY6iygwO-PwrjFhNi-DLzp/src/widgets/TextLayoutWidget.zig:1620:25: 0x101e355c3 in addTextDone (dvui_test)
        std.debug.assert(self.cache_layout_bytes_seen == self.bytes_seen);
                        ^
/Users/jamie/.cache/zig/p/dvui-0.4.0-dev-AQFJmXGQ1QC8SDZ6N9lNLrYY6iygwO-PwrjFhNi-DLzp/src/widgets/TextLayoutWidget.zig:2104:25: 0x101e18c0f in deinit (dvui_test)
        self.addTextDone(.{});
                        ^
/Users/jamie/dev/dvui-test/src/test.zig:10:28: 0x101e0c8af in appFrame (dvui_test)
    defer text_entry.deinit();
```

RedPhoenixQ:
Fixes #607

Some parts of this function was using the non-validated input text. Now everything should use the valid utf-8, with a small performance penalty for caching that now still has to validate the utf-8. This fixes the assert trigger at https://github.com/RedPhoenixQ/dvui/blob/textlayout-use-utf8/src/widgets/TextLayoutWidget.zig#L1620.

### DIFF
--- a/src/Examples/debugging.zig
+++ b/src/Examples/debugging.zig
@@ -40,6 +40,9 @@ pub fn debuggingErrors() void {
         var b = dvui.box(@src(), .{}, .{ .expand = .horizontal, .margin = .{ .x = 10 } });
         defer b.deinit();
         dvui.labelNoFmt(@src(), "this \xFFtext\xFF includes some \xFF invalid utf-8\xFF\xFF\xFF which is replaced with \xFF", .{}, .{});
+        const tl = dvui.textLayout(@src(), .{ .cache_layout = true }, .{});
+        defer tl.deinit();
+        tl.addText("Some \xFFinvalid utf-8 \xc3 in a text layout", .{});
     }
 
     if (dvui.expander(@src(), "Scroll child after expanded child (will log error)", .{}, .{ .expand = .horizontal })) {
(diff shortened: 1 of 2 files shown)


david-vanderson:
Awesome - merge when ready.  Thank you!

PR #610, opened 2025-10-23
AccessKit: additional widget support

phatchman:
Supports most DVUI widgets.
- Add readme-accessibility.md
- Add 'replace' option to the text event.
  * Used by accesskit actions to replace existing values e.g. when updating a textbox
- Support for new widgets:
    * toasts
    * suggestions
    * combobox
    * expander
    * tooltip
    * grid headings
    * images
    * progress
    * textEntryNumber
    * ColorPickerWidget
    * GridWidget
    * PanedWidget
    * PlotWidget
    * TabsWidget
    * TreeWidget
- Improved support for
     * slider - set orientation
     * DropDownWidget - list items are now labeled correctly.
     * TextEntryWidget - Implement replace for text event.
- Fixes
    * Create accesskit node earlier in `register()` so that ak_node is available in `data_out`
    * Fix panic when focused node was not visible.
- Other
    * Add basic widget radio buttons to a radio group

@david-vanderson If I remember correctly, you weren't in favour of adding 'replace' to the text event, but I can't see a better way to do it. That's the only change I've made to DVUI core in this PR outside of moving the label options to a named union.

Only lightly tested. Needs a full run-through of the demo app with accessibility enabled. Especially test things that have focus becoming invisible (e.g. scrolling off the screen).

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,10 +328,13 @@ DVUI has varying support for different kinds of accessibility infrastructure.  C
 * Language Input
   * IME (Input Method Editor) works in SDL and web backends
 
-* Audio Output
-  * accesskit integration for screen readers and alternate input events
-  * add `-Daccesskit` to `zig build`
-  * see `Options.role` and `Options.label`
-  * this is still new, please report problems and missing widget support
+* High Contrast Themes
+  * dvui's themes can support this
+  * no current OS integration
 
+* Screen Reading and Alternate Input
+  * [AccessKit](https://accesskit.dev/) integration
+  * add `-Daccesskit` to `zig build`
+  * uses `Options.role` and `Options.label`
+  * see [readme-accessibility](readme-accessibility.md) for more information
 
--- a/readme-accessibility.md
+++ b/readme-accessibility.md
@@ -1,0 +1,225 @@
+# Accessibility in DVUI
+
+## What is accessibility?
+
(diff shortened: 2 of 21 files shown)


phatchman:
@datatriny Is the `paned` role equivalent to the DVUI pane widget? The DVUI widget provides a splitter to resize two panels/boxes either horizontally or vertically. 

I was able to set a numeric value on the pane node to indicate the percentage split of the pane, but similar to the scrollbar, this is read-only, so the user can't change it.

Would you prefer me to raise these sorts of issues on the accesskit github?

david-vanderson:
This is excellent!

I spent an hour testing the changes.  I'm not great at this, but it's something.

> @david-vanderson If I remember correctly, you weren't in favour of adding 'replace' to the text event, but I can't see a better way to do it. That's the only change I've made to DVUI core in this PR outside of moving the label options to a named union.

No problem that is good.  It's on me to figure out if trying to do "select all" first will work.

Thanks very much for the readme!  I'll do an edit pass on that later today and link it to the main readme.

david-vanderson:
@phatchman I've done an editing pass.  Please let me know if you don't agree with any of the changes.  Otherwise merge this when you are ready.  Thank you very much!!

david-vanderson:
I fixed up the expander, but we will have to wait on https://github.com/AccessKit/accesskit/issues/551

PR #612, opened 2025-10-24
AccessKit: Linux support

phatchman:
Currently working for -Daccesskit=shared on x86_64 linux. 

The static library has unresolved symbol issues. I think because the static library is compiled against libc++ instead of libc. All of the unresolved symbols appear to be related to c++ exception unwinding. Example at end of message. I'll look into this further, to see if we can change the build options.

@DataTriny I'm having an issue in that the initial tree update is always being called, even on a fresh login session, where I've not previously run orca. Which means we're creating accesskit nodes all the time. I don't think it is anything I am doing that would cause this? 

@david-vanderson I wasn't sure whether to move all of the logic for handing the SDL focus and window shown events into AccessKit.zig instead? I initially split it so the SDL logic is in sdl.zig and created some wrappers for the accesskit functions in AccessKit.zig, but that means passing all of the 8 co-ordinate parameters for SDL_EVENT_WINDOW_SHOWN to a function which just passes them directly on to an access kit function. So, I think it all needs to live either in sdl.zig or AccessKit.zig and since it is mostly sdl logic, I left it in sdl.zig. 

```
error: undefined symbol: _Unwind_Resume
    note: referenced by /home/phatc/.cache/zig/p/N-V-__8AAJ37FAuEZXG593uEuYETrkoxDz78QLL32k0GkaYG/lib/linux/x86_64/static/libaccesskit.a(accesskit.accesskit.ed425a1bcbc715c-cgu.0.rcgu.o/):.text._ZN50_$LT$T$u20$as$u20$core..convert..Into$LT$U$GT$$GT$4into17h38dbd7b2c029bd76E
    note: referenced by /home/phatc/.cache/zig/p/N-V-__8AAJ37FAuEZXG593uEuYETrkoxDz78QLL32k0GkaYG/lib/linux/x86_64/static/libaccesskit.a(accesskit.accesskit.ed425a1bcbc715c-cgu.0.rcgu.o/):.text._ZN8blocking7unblock28_$u7b$$u7b$closure$u7d$$u7d$17h06c4e8c728c32277E
    note: referenced by /home/phatc/.cache/zig/p/N-V-__8AAJ37FAuEZXG593uEuYETrkoxDz78QLL32k0GkaYG/lib/linux/x86_64/static/libaccesskit.a(accesskit.accesskit.ed425a1bcbc715c-cgu.0.rcgu.o/):.text._ZN4zbus10connection7builder7Builder6build_28_$u7b$$u7b$closure$u7d$$u7d$17h352ae292f637f83eE
    note: referenced by /home/phatc/.cache/zig/p/N-V-__8AAJ37FAuEZXG593uEuYETrkoxDz78QLL32k0GkaYG/lib/linux/x86_64/static/libaccesskit.a(accesskit.accesskit.ed425a1bcbc715c-cgu.0.rcgu.o/):.text._ZN42_$LT$$RF$T$u20$as$u20$core..fmt..Debug$GT$3fmt17h19b85d05745acdf4E
    note: referenced 114 more times
```

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -583,7 +583,6 @@ fn accessKitPath(b: *std.Build, target: std.Build.ResolvedTarget, ak_dep: *std.B
     }
 }
 
-
 fn accessKitLibName(target: std.Build.ResolvedTarget) []const u8 {
     return if (target.result.os.tag == .windows) "accesskit.dll" //
     else if (target.result.os.tag.isDarwin()) "libaccesskit.dylib" //
@@ -621,6 +620,10 @@ fn addDvuiModule(
         dvui_mod.addIncludePath(ak_dep.path("include"));
         dvui_mod.addLibraryPath(accessKitPath(b, target, ak_dep, opts.accesskit, false));
         dvui_mod.linkSystemLibrary("accesskit", .{});
+
+        if (target.result.os.tag == .linux) {
+            dvui_mod.linkSystemLibrary("unwind", .{});
+        }
     }
 
     const stb_source = "vendor/stb/";
--- a/build.zig
+++ b/build.zig
@@ -548,14 +548,21 @@ const DvuiModuleOptions = struct {
 };
 
 fn accessKitSupported(target: std.Build.ResolvedTarget) bool {
-    if (target.result.os.tag != .windows and !target.result.os.tag.isDarwin()) return false;
-    if (!target.result.cpu.arch.isAARCH64() and !target.result.cpu.arch.isX86()) return false;
-    return true;
+    if (target.result.os.tag == .windows) {
+        return target.result.cpu.arch.isAARCH64() or target.result.cpu.arch.isX86();
+    } else if (target.result.os.tag.isDarwin()) {
+        return target.result.cpu.arch.isAARCH64() or target.result.cpu.arch == .x86_64;
+    } else if (target.result.os.tag == .linux) {
+        return target.result.cpu.arch.isX86();
+    } else {
+        return false;
+    }
 }
 
 fn accessKitPath(b: *std.Build, target: std.Build.ResolvedTarget, ak_dep: *std.Build.Dependency, opt: AccesskitOptions, full_lib_path: bool) std.Build.LazyPath {
     const os_path = if (target.result.os.tag == .windows) "windows" //
         else if (target.result.os.tag.isDarwin()) "macos" //
+        else if (target.result.os.tag == .linux) "linux" //
         else "unsupported";
     const arch_path = if (target.result.cpu.arch.isAARCH64()) "arm64" //
         else if (target.result.cpu.arch == .x86) "x86" //
@@ -564,6 +571,7 @@ fn accessKitPath(b: *std.Build, target: std.Build.ResolvedTarget, ak_dep: *std.B
 
     const abi_path = if (target.result.os.tag == .windows) "msvc" //
         else if (target.result.os.tag.isDarwin()) "" //
+        else if (target.result.os.tag == .linux) "" //
         else "";
 
     const linkage_path = @tagName(opt);
@@ -575,9 +583,11 @@ fn accessKitPath(b: *std.Build, target: std.Build.ResolvedTarget, ak_dep: *std.B
     }
 }
 
+
 fn accessKitLibName(target: std.Build.ResolvedTarget) []const u8 {
     return if (target.result.os.tag == .windows) "accesskit.dll" //
     else if (target.result.os.tag.isDarwin()) "libaccesskit.dylib" //
+    else if (target.result.os.tag == .linux) "libaccesskit.so" //
     else "unsupported";
 }
 
--- a/build.zig
+++ b/build.zig
@@ -615,10 +615,16 @@ fn addDvuiModule(
         dvui_mod.linkSystemLibrary("ole32", .{});
     }
 
+    // the system integration option check has to always occur even if accesskit is disabled for
+    // it to be displayed in the build help
+    const accesskit_from_system = b.systemIntegrationOption("accesskit", .{});
     if (opts.accesskit.enabled()) {
         const ak_dep = b.dependency("accesskit", .{});
         dvui_mod.addIncludePath(ak_dep.path("include"));
-        dvui_mod.addLibraryPath(accessKitPath(b, target, ak_dep, opts.accesskit, false));
+
+        if (!accesskit_from_system) {
+            dvui_mod.addLibraryPath(accessKitPath(b, target, ak_dep, opts.accesskit, false));
+        }
         dvui_mod.linkSystemLibrary("accesskit", .{});
 
         if (target.result.os.tag == .linux) {
@@ -735,6 +741,8 @@ fn addExample(
             opts.accesskit,
             true,
         ), .bin, accessKitLibName(opts.target)).step);
+        // Converts shared lib path to be relative to exe install directory.
+        exe.root_module.addRPathSpecial("$ORIGIN");
     }
 
     const run_step = b.step(name, "Run " ++ name);
--- a/readme-accessibility.md
+++ b/readme-accessibility.md
@@ -30,9 +30,9 @@ The overhead of creating AccessKit nodes is only incurred when the operating sys
 AccessKit can be included as either a static or dynamic link library. Use `-Daccesskit=static` or `-Daccesskit=shared` to enable AccessKit support. 
 
 Not all combinations of backends and operating systems are currently supported. 
-- Linux - Coming soon.
+- Linux - Supports all backends in for shared and static libraries. SDL2, SDL3 and Raylib.
 - Windows - Support for all backends, except raylib static due to symbol clashes.
-- MacOS - Support for SDL2, SDL3 and Raylib.
+- MacOS - Supports all backends in for shared and static libraries. SDL2, SDL3 and Raylib.
 - Web - Not currently supported by AccessKit but it is planned. 
 
 When you compile you application with AccessKit enabled, most of the accessibility work is taken care of for you and your users will gain most of the benefits of accessibility support. However, DVUI doesn't know your application's intent or any semantic details about the UX, so it is not possible to automate everything. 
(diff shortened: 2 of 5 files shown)


DataTriny:
> I'm having an issue in that the initial tree update is always being called, even on a fresh login session, where I've not previously run orca.

On that fresh login session, could you please execute the following commands which should both return `false` in that case?

```
dbus-send --dest=org.a11y.Bus --print-reply /org/a11y/bus org.freedesktop.DBus.Properties.Get string:org.a11y.Status string:IsEnabled
dbus-send --dest=org.a11y.Bus --print-reply /org/a11y/bus org.freedesktop.DBus.Properties.Get string:org.a11y.Status string:ScreenReaderEnabled
```

david-vanderson:
> @david-vanderson I wasn't sure whether to move all of the logic for handing the SDL focus and window shown events into AccessKit.zig instead? I initially split it so the SDL logic is in sdl.zig and created some wrappers for the accesskit functions in AccessKit.zig, but that means passing all of the 8 co-ordinate parameters for SDL_EVENT_WINDOW_SHOWN to a function which just passes them directly on to an access kit function. So, I think it all needs to live either in sdl.zig or AccessKit.zig and since it is mostly sdl logic, I left it in sdl.zig.

Looks good to me.  One additional good thing to come out of the AccessKit effort is showing where our event handling is missing functionality.  Maybe in this case we need an entry in `dvui.Event` that is something like `.window_shown`, which is handled in `dvui.Window` by passing it to accesskit (probably regardless of whether it was already handled elsewhere).

But until I figure that stuff out, any division that works works for me!

phatchman:
I should have mentioned how I am testing this. It is a physical x86 machine running Ubuntu 24.04.3 LTS and X11.
I'm then remoting into the Linux machine from Windows via remote desktop. 

Edit: I hooked up a monitor and kb and tested directly from the machine. Same behaviour.
> 
> On that fresh login session, could you please execute the following commands which should both return `false` in that case?
> 

Results below:
```
dbus-send --dest=org.a11y.Bus --print-reply /org/a11y/bus org.freedesktop.DBus.Properties.Get string:org.a11y.Status string:IsEnabled
dbus-send --dest=org.a11y.Bus --print-reply /org/a11y/bus org.freedesktop.DBus.Properties.Get string:org.a11y.Status string:ScreenReaderEnabled
method return time=1761340594.757885 sender=:1.19 -> destination=:1.114 serial=34 reply_serial=2
   variant       boolean true
method return time=1761340594.762619 sender=:1.19 -> destination=:1.115 serial=35 reply_serial=2
   variant       boolean false
```
I get the same results, whether orca is running or not.

Then when I run the sdl3-standalone, I get the debug output below, showing the adapter in an active state.

```
setting adapter
info(AccessKit): initialized successfully
init tree update
creating node
creating node
push updates = on
updating if active
frame tree update
creating node
```

When I run Orca, I do get this warning: ~~This went away after a reboot.~~ Appears to be intermittent, but not affecting anything.
```
(orca:98579): dbind-WARNING **: 08:24:17.767: AT-SPI: Error in GetItems, sender=:1.35, error=Unknown object '/org/a11y/atspi/cache'
```

phatchman:
I found [this](https://gitlab.gnome.org/GNOME/at-spi2-core/-/tree/main/bus) in the gnome docs.
```
However, a screen reader is special: it needs to start up automatically during login, and
immediatelly tell the accessibility registry (... via the accessibility bus) that it
is running.  If you need a screen reader to use your computer, you cannot easily launch it
by hand if there is no screen reader already running!

That is, if a screen reader is turned on — and we assume it will start up turned on for
future sessions — we need to launch the accessibility bus unconditionally, not on
demand, at session startup.  This is why at-spi-dbus-bus.desktop, described below, is an
[XDG autostart](https://specifications.freedesktop.org/autostart-spec/autostart-spec-latest.html) file which runs at-spi-bus-launcher --launch-immediately,
but only if a certain GSettings key is turned on.
```

So I understand it to mean that running orca the first time, changed the accessibility bus to start unconditionally?

Will explore this further later.

Edit: OK I understand this now

It is as above. Running orca sets Gnome's `toolkit-accessibility` to true and, at least in my setup, leaves it like that until someone turns it off. To do that you can use:

```
gsettings set org.gnome.desktop.interface toolkit-accessibility false
```
I also had to turn off the auto-start of the screen reader at login, which I probably accidentally enabled somehow.
```
gsettings set org.gnome.desktop.a11y.applications screen-reader-enabled false
```
So now if I run sdl3-standalone, it doesn't produce nodes, until I run orca.
After that it will continue creating nodes, even if orca is not running, until I do `gsettings set org.gnome.desktop.interface toolkit-accessibility false`
At that point sdl3-standalone immediately stops creating nodes!

So, looking good for shared library! I think I've figured out static library as well. It just needs libunwind added.

phatchman:
OK. So that should get both static and dynamic linking working on linux. As far as I can tell it works? 

I couldn't get orca to read the whole screen, but what it did read was correct, so I assume that is just a skill issue.

I snuck in a fix to the sliderEntry in this PR. You could enter a value that was out of bounds in the text_entry mode. I also took out the text handling from sliderEntry in text_mode = true case. I only put that in there to support accesskit actions on the slider entry if someone had clicked enter on it to type in a value. I don't think we need that and the text event in the text_mode = false case should be enough.

phatchman:
@david-vanderson 
Hmm. One question about the shared library version:

ldd shows:
```
raylib-app:
	linux-vdso.so.1 (0x00007fff8d3cc000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007c8f58534000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007c8f58200000)
	/lib64/ld-linux-x86-64.so.2 (0x00007c8f5863a000)
	libaccesskit.so => /home/phatc/.cache/zig/p/N-V-__8AAJ37FAuEZXG593uEuYET
rkoxDz78QLL32k0GkaYG/lib/linux/x86_64/shared/libaccesskit.so (0x00007c8f58049000
)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007c8f58506000
)
```
Does it matter that the dll path is pointing to the zig cache directory for someone distributing an application? I know it will still search the PATH for the shared library, so it will still work. 

I suppose the solution is to give the user an option to use a different dll path for accesskit=shared so that they can point it to /usr/local/lib/libaccesskit.so or wherever they have installed accesskit? Edit: Ahh. I can use the -fsys= for this right?

The other thing I noticed is the vast difference in the number of shared libs required for sdl2 vs sdl3. Is there a way to statically link the sdl2 dependencies on Linux? I couldn't see a way to do it via the current build options. I'd like to distribute the Linux versions of my applications as fully statically linked so they aren't distro-specific and don't require the user to add any more dependencies.

david-vanderson:
> I snuck in a fix to the sliderEntry in this PR. You could enter a value that was out of bounds in the text_entry mode.

I don't know what behavior we want here.  I can imagine both wanting the value to never be outside the bounds, but also scenarios where you want to override the bounds when entering manually.

david-vanderson:
> Does it matter that the dll path is pointing to the zig cache directory for someone distributing an application? I know it will still search the PATH for the shared library, so it will still work.
> 
> I suppose the solution is to give the user an option to use a different dll path for accesskit=shared so that they can point it to /usr/local/lib/libaccesskit.so or wherever they have installed accesskit? Edit: Ahh. I can use the -fsys= for this right?

I think the solution here would be to copy the shared library to the zig-out dir and link that one?  I'm not sure that would work, but maybe the linker would always search the directory that the exe is in?

Yes -fsys= also makes sense, and in some cases is probably preferred (for example the user might have certain stuff compiled in SDL that we don't, like IME stuff).

> 
> The other thing I noticed is the vast difference in the number of shared libs required for sdl2 vs sdl3. Is there a way to statically link the sdl2 dependencies on Linux? I couldn't see a way to do it via the current build options. I'd like to distribute the Linux versions of my applications as fully statically linked so they aren't distro-specific and don't require the user to add any more dependencies.

I'll look into this.  There is generally a limit you hit here where graphics stuff always has to be dynamically linked, but hopefully that's all.

david-vanderson:
> > I snuck in a fix to the sliderEntry in this PR. You could enter a value that was out of bounds in the text_entry mode.
> 
> I don't know what behavior we want here. I can imagine both wanting the value to never be outside the bounds, but also scenarios where you want to override the bounds when entering manually.

Sorry - I should have said - since I don't know what to do, let's go with your fix and see how it feels.

phatchman:
This should fix up those last few items.

Added and tested -fsys=accesskit. The 'shared' or 'static' is ignored when you do this. It will just pick up the first one it finds. The shared lib path points to the system library in this case as expected.

Added `exe.root_module.addRPathSpecial("$ORIGIN");` which I believe is the correct solution to the cache path issue. This makes the shared library paths relative to the exe installation directory.  But does not impact the other system dlls.

I think this is good to merge now if you are happy with it. I've tested building on Windows and Linux in the various configurations, but not MacOS.

- Adds support for shared and static accesskit on Linux.
- Fixes issue with lock being released too early in createNode
- Implements a min/max check on sliderEntry when entering a value in text mode
- Adds accesskit as a system integration option.
- Panic if accesskit adapter can't be created. Used to just panic later when the null adapter was unwrapped.


Edit: I think we want addRPathSpecial("$ORIGIN") for MacOS as well, where it does something similar.

phatchman:
Finished testing MacOS and fixed an issue. 
RPATH works as expected on MacOS, which I think is correct, but that is just based on things I've read recently.
```
otool -L ./zig-out/bin/sdl3-standalone | grep access
        @rpath/libaccesskit.dylib (compatibility version 0.0.0, current version 0.0.0)
```

Should be good to merge.

david-vanderson:
Awesome!!  Will do some testing and look at it, hopefully merge tomorrow.  Thank you!

david-vanderson:
Beautiful - thank you so much!

DataTriny:
I'll change the conditions required for AccessKit to initialize on Linux so that users who don't need accessibility support don't have to pay the price. I'll let you know when it's available.

david-vanderson:
> I'll change the conditions required for AccessKit to initialize on Linux so that users who don't need accessibility support don't have to pay the price. I'll let you know when it's available.

Thank you!

PR #617, opened 2025-10-27
Various fixes for the `data{Get,Set}[...]` functions

jjkh:
`dataGetDefault` now compiles - wasn't caught by the `refAllDecls(@This())` test due to the `anytype` parameter.

While trying to test this change, I noticed a few issues for slices with sentinels (e.g. string literals):

* Slices with a sentinel would be stored with an extra byte
* Slices for a type larger than 1 byte would not allocate enough space for the sentinel values
* When `setSliceCopies` called with `num_copies` > 1, the sentinel would be written after the first copy (overwriting the second copy) rather than at the end
* `getSlice` would return a slice including the sentinel values
* No debug error when slice stored with sentinel but retrieved without (and vice-versa)

I've also added a test to `dvui.zig` that cover some of these cases, but I'm not sure that it's in the right place.

### DIFF
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -1388,6 +1388,29 @@ pub fn dataRemove(win: ?*Window, id: Id, key: []const u8) void {
     };
 }
 
+test "data get/set/remove basic" {
+    var t = try dvui.testing.init(.{});
+    defer t.deinit();
+
+    dataSet(null, .zero, "data", {});
+    try std.testing.expectEqual({}, dataGet(null, .zero, "data", void));
+
+    dataSetSlice(null, .zero, "dataSlice", @as([]const u8, "ab"));
+    try std.testing.expectEqualSlices(u8, "ab", dataGetSlice(null, .zero, "dataSlice", []u8).?);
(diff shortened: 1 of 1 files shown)


RedPhoenixQ:
Thank you so much for taking the time. I guess there just aren't any uses of `dataGetDefault` within the dvui examples which I trusted blindly in my "testing" (if the examples compiled and opened).

Apologies for the many off by one errors with sentinels... I guess I had a bad grasp of when the sentinel byte was included in the length and not when refactoring. For next time, there should probably be more tests before the refactor to verify that the behaviour remains, thank you again.

> I've also added a test to `dvui.zig` that cover some of these cases, but I'm not sure that it's in the right place.

Since it is a test for the `dvui.XXX` apis I think that is a fine place for it. Could potentially be split into decl-tests for each api, but that not important.

> `dataGetDefault` now compiles - wasn't caught by the `refAllDecls(@This())` test due to the `anytype` parameter.

Also very good to know. Should probably add some more dedicated test in other places in the codebase then

david-vanderson:
Thank you very much!

PR #619, opened 2025-10-28
CellStyle: fix compile errors in unreferenced functions

jjkh:
Also updates test to use `refAllDeclsRecursive` to catch these errors in future.


david-vanderson:
Looks good - merge when ready (or let me know if you have trouble merging or would rather I do it).  Thank you!

PR #621, opened 2025-11-02
PanedWidget: Support auto-fitting the first pane when direction is `.horizontal`

jjkh:
The issue (which also affected the vertical auto-fit behaviour) was that `split_ratio` was handled differently between the `getFirstFittedRatio()`, `rectFor()`, and the drawing/event handling code.  
`split_ratio` now refers to the ratio of the first pane's **content** to the second pane's **content**, which I believe was the original intent.

As well as this, fixed a few other issues:

- Auto-fitting when the first pane was collapsed would only half-expand the pane 
    - Now expands before auto-fitting (if required)
- The pane is auto-fit on the first frame, but at that point we don't know the `PanedWidget`'s content rect
    - Now also attempts to auto-fit on the second frame
- On MacOS, I was unable to grab the handle when a `PanedWidget` is positioned against the edge of the window
    - Now leaves `handle_margin` gap before drawing the handle.
    Before/After:
    image image
- The "Options overrides" window in the debug window would show a duplicate widget ID outline when multiple widgets' styles were overridden

### DIFF
--- a/src/Debug.zig
+++ b/src/Debug.zig
@@ -312,13 +312,12 @@ pub fn show(self: *Debug) void {
         defer menu.deinit();
 
         var it = self.options_override.iterator();
-        var i: usize = 0;
         var remove_override_id: ?dvui.Id = null;
-        while (it.next()) |entry| : (i += 1) {
+        while (it.next()) |entry| {
             const id = entry.key_ptr.*;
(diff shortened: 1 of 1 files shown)


david-vanderson:

PR #622, opened 2025-11-08
Plot: add log scale, tick formatting, etc

leventev:
This PR adds the following:
- logarithmic scale for both axes
- ability to set the number of ticks per axis, the ticks are evenly spaced
- toggleable short tick lines at each tick
- ability to set the formatting of ticks per axis (normal, scientific or custom)
- ability to set spine color
- toggleable grid lines

Everything has a default value so there should be no API breakage. I set the default number of ticks to 10 and grid lines to be enabled by default, this might have to be changed to retain the current user experience, I'm not sure how big of a priority that is. One thing that's a more invasive change is that I added more padding to the label of both axes to prevent the axis label from overlapping with tick labels if the ticks are dense. Also usually in plots the label for the Y axis is rotated by 90 degrees but I haven't found a way to rotate text.

I also added another example to showcase a semi-log plot. I chose to visualize an RC low-pass filter. It might be a good idea to replace this with an example that non-engineers could more easily understand. For example a function like f(x)=C*x^n. Note that my example does not use all the features I have added but I have tested all of them individually.

I took some inspiration from the [lilaq project](https://github.com/lilaq-project/lilaq). I think in the future the following features would be worth implementing:
- ticks passed in an array, ticks located by a function passed by the user
- string tick labels instead of numbers like [A B C D E]
- subticks, they would be pretty nice when using log scales
- logarithmic tick formatting, for example 10^5, the problem with this is rendering the superscript and I'm not sure how much work it would be to implement that
- formatting the data displayed when hovering over the line to avoid displaying 60 digits, IMO that's pretty ugly
- marks for the data points(X, hollow circle, full circle)
- legend
- disable line drawing(draw only marks), dotted or dashed lines
- error bars?
- ability to set the axis labels and tick labels to be displayed on the right or on the top
- fill between two plot-lines?
- set margin/padding between the line and the borders of the data box. lilaq does this and IMO it looks better, makes the extreme values easier to read
- stem plots

Please let me know what you think. I hope my code is up to standard, I tried to follow the idiomatic way of doing things but I'm sure some things could have been done better. If there is an issue with anything I will fix it.

And if the features I mentioned above are desired then I will gladly work on implementing them.

### DIFF
--- a/src/widgets/PlotWidget.zig
+++ b/src/widgets/PlotWidget.zig
@@ -46,7 +46,7 @@ pub const Axis = struct {
     scale: union(enum) {
         linear,
         log: struct {
-            base: f64 = 10,
+            base: f64 = 3,
         },
     } = .linear,
 
@@ -326,12 +326,14 @@ pub fn install(self: *PlotWidget) void {
     yaxis.deinit();
 
     // x axis padding
-    var xaxis_padding = dvui.box(@src(), .{ .dir = .horizontal }, .{
-        .gravity_x = 1.0,
-        .expand = .vertical,
-        .min_size_content = .{ .w = x_axis_last_tick_width / 2 },
-    });
-    xaxis_padding.deinit();
+    if (self.x_axis.name) |_| {
+        var xaxis_padding = dvui.box(@src(), .{ .dir = .horizontal }, .{
+            .gravity_x = 1.0,
+            .expand = .vertical,
+            .min_size_content = .{ .w = x_axis_last_tick_width / 2 },
+        });
+        xaxis_padding.deinit();
+    }
 
     // data area
     var data_box = dvui.box(@src(), .{ .dir = .horizontal }, .{
@@ -365,9 +367,6 @@ pub fn install(self: *PlotWidget) void {
 
     const bt: f32 = self.init_options.border_thick orelse 0.0;
     const bc: dvui.Color = self.init_options.spine_color orelse self.box.data().options.color(.text);
-    if (bt > 0) {
-        self.data_rs.r.stroke(.{}, .{ .thickness = bt * self.data_rs.s, .color = bc });
-    }
 
     const pad = 2 * self.data_rs.s;
 
@@ -397,6 +396,11 @@ pub fn install(self: *PlotWidget) void {
         }
     }
     xaxis.deinit();
+
+    _ = dvui.spacer(@src(), .{
+        .min_size_content = .width(x_axis_last_tick_width / 2),
+        .expand = .vertical,
+    });
 
     hbox2.deinit();
 
@@ -556,6 +560,10 @@ pub fn install(self: *PlotWidget) void {
                 dvui.logError(@src(), err, "x axis tick text for {d}", .{xtick});
             };
         }
+    }
+
+    if (bt > 0) {
+        self.data_rs.r.stroke(.{}, .{ .thickness = bt * self.data_rs.s, .color = bc });
     }
 
     self.old_clip = dvui.clip(self.data_rs.r);


jjkh:
I think this is really cool, and I like the RC low-pass filter demo! (but I come from an electronics background, so I'm also probably the target audience).

I have some minor feedback from playing around with the demo.
_Note that I'm not a regular contributor here, so feel free to ignore this._ 😄

- The small sparkline-like plots no longer seem to render correctly on the Plots example:  
Before / After:
    image image

- It seems like the plot line now displays on top of the plot border, whereas before it was (mostly) underneath _(using SDL3 on Windows)_:  
Before / After:  
    image

- The tick lines in the Grid plot demo look good, but I think it would be nice if the ticks could be round numbers (at least with the default values) -- for example, the x axis could go 0 to 50 in steps of 10 rather than 11.11:  
    image

As far as future features, I wanted to note a few from your list that I would have already found useful:

- Rotating the y-axis text
- Changing the hover text (not just reducing the amount of significant figures displayed, but having a custom hover box with additional information)
- Marks for the data points
- Legend
- Custom line drawing (none/dotted/dashed)
- Setting margin between the line and the border of the plot (I've had to do this manually, by setting the limits 10% higher than the data displayed)

leventev:
Thank you for your feedback!

> The small sparkline-like plots no longer seem to render correctly on the Plots example

I somehow missed this while testing. The problem was that padding for the X axis was added even for plots without an X axis. It renders correctly now.

> It seems like the plot line now displays on top of the plot border, whereas before it was (mostly) underneath (using SDL3 on Windows)

I'm not sure why it renders differently now, but the root cause is that the border is drawn in `PlotWidget.install`. Only after this are the lines drawn. I tried putting the border drawing code in `PlotWidget.deinit` (which is after the lines have been drawn) and it worked as expected, the plot border is above the plot line. However I don't think the `deinit` function should have any rendering code inside it so I haven't pushed this change.

> The tick lines in the Grid plot demo look good, but I think it would be nice if the ticks could be round numbers (at least with the default values) -- for example, the x axis could go 0 to 50 in steps of 10 rather than 11.11

I agree, the reason why the step size is 11.11 is because the min and max values are always included and the default number of ticks is 10, so it has to split 100 into 9 evenly spaced parts. 100/9=11.1, if we specified the number of ticks to be 11 the step size would be 10. I think only having the min and max ticks is not enough but more usually results in uneven spacing, so I set the default number of ticks to be 3 and the user can adjust it if it's not enough. In the future a way to automatically determine the number of ticks and spacing between them could be added.

Also I'm glad I'm not the only one who wants these features :smiley:. I think most of them wouldn't be too hard to implement except rotating the Y axis label. I think that would require some changes to the text rendering API and I rather leave that to someone who is more familiar with the codebase. 

Also I moved the plot border rendering code after the tick line and grid line rendering code because previously the first and last grid lines were being rendered over the border.

david-vanderson:
Amazing thank you very much!  Great revisions, and thanks @jjkh for reviewing!

I agree with your whole list, and I will tackle rotated text (hopefully soon).

As for whether the plot data draws above or below the border, I don't know what the right answer is.  I've been in situations where people wanted the data above the border (so you know that it hits zero and stays there), but maybe visually it's better below?  Not a big deal either way and can deal with it in the future.

For the mouse hover display, I agree it's ugly right now.  I'm hoping we can do something "immediate-mode" where the calling code can ask if the mouse is hovered over a point, and then draw whatever it wants.

> set margin/padding between the line and the borders of the data box. lilaq does this and IMO it looks better, makes the extreme values easier to read

I don't understand this - if it's easy can you say more to point to a screenshot?

leventev:
Thank you for merging!

I meant something like this:

lilaq_rc_lp

You can see that there is some distance between the plot and the borders.
> I've been in situations where people wanted the data above the border (so you know that it hits zero and stays there)

This would also make this question irrelevant since the maximums and minimums are clearly visible(except where we want to clip parts of the line off). As @jjkh mentioned this can be sort of achieved by setting the limits higher than the data but that would mess with the tick locations. In the screenshot I sent you can see that the first and last ticks of both axes are not aligned with the edges of the box.

leventev:
I just noticed I accidentally set the default logarithmic base to 3 instead of the default number of ticks to be 3 :facepalm:. Really sorry about that. I'm not really sure how I could fix that since I already deleted my branch. Can someone change `PlotWidget.Axis.scale.log.base` to 10 and the number of ticks to 3 in `PlotWidget.Axis.ticks.locations`'s default value?

david-vanderson:
No problem - done.

jjkh:
> You can see that there is some distance between the plot and the borders.
> 
> > I've been in situations where people wanted the data above the border (so you know that it hits zero and stays there)
> 
> This would also make this question irrelevant since the maximums and minimums are clearly visible(except where we want to clip parts of the line off).

I do think there's some nuance to this - usually I'll have a graph with the y-axis starting at 0, and in those cases I don't want any padding on the y-axis minimum as it can throw off the relative scale of the chart.  
In my opinion, if added this should be configurable for each extent of both axes (e.g. something like `.axis_padding = .{ .y = .{ .max = 0.05, .min = 0 } }`).

david-vanderson:
@leventev Thanks for the screenshot and explanation - makes sense to me now.

@jjkh Yes that makes sense - it could be an option per-axis.  I guess the padding value would be in logical pixels (not graph data space) so that as the graph is resized the margin stays the same?

jjkh:
> I guess the padding value would be in logical pixels (not graph data space) so that as the graph is resized the margin stays the same?

Right, that does make sense - my brain was still thinking about manipulating axis limits.

Issue #618, opened 2025-10-27
FocusGroup: fixing tabbing for grouped widgets


david-vanderson:
Split from #616 

@ethindp showed that dvui is missing a crucial bit of focus movement behavior.  Example is radio buttons.

Right now dvui treats each radio button as a separate widget that participates in tab order.  But this is not how UI toolkits normally do it.  They will have the entire radio group participate in tab order, but each radio button focus is controlled by the arrow keys (as long as focus is within the radio group).

This kind of thing also happens for menus, tabs, tree entries, grid entries, and probably a few other places I haven't thought of yet.

Proposal:
* add a `FocusGroup` widget
* does not do any layout
* records tab_index of children internally, but prevents those from being added to the subwindow's tab ordering
* in deinit() use unhandled arrow key events to move focus internally

I don't think nesting FocusGroup makes any sense?  If you are on the last radio button and press "down", it cycles back to the first in the FocusGroup, so you could never break out of a nested group.

But how would we handle radio buttons inside a grid cell?  Maybe nesting does make sense as long as a nested FocusGroup doesn't cycle back?  I need to find some OS native examples to play around with.

david-vanderson:
@phatchman any insights here regarding widgets inside grid cells?  How does grid handle these currently?

ethindp:
The only example I could provide is a native win32 example using the win32 API in C++ for tree views anyway. And the win32 API is awful. That said, I can tinker with it and add a nested node structure and post it here if you like (even if it would be quite long).

david-vanderson:
> The only example I could provide is a native win32 example using the win32 API in C++ for tree views anyway. And the win32 API is awful. That said, I can tinker with it and add a nested node structure and post it here if you like (even if it would be quite long).

I don't want to impose too much.  The primary reason I'm looking for examples is just to find out what the behavior is.  I'm looking for answers to stuff like:
* how does keyboard navigation work (or how should it work) when there is a "normal" widget inside a group?
  * example is a group like a grid, but inside some of the grid cells are checkboxes
  * can you get to the checkboxes by tabbing?  Or do you have to tab to the grid, and then use arrows to move to that cell?  Does it auto-focus the first checkbox in the cell?  If there are multiple checkboxes, can you move between them inside the cell via arrow keys?

I'm guessing this situation doesn't come up that often, but knowing what existing applications do here would help.  dvui also is more flexible than most UIs - for example we can put any kind of widget inside a menu, so we at least need some plan for how to handle that.

[...]

PR #623, opened 2025-11-09
TreeWidget: collapse and/or expand branches with left/right keys

jjkh:
Makes the left/right keys collapse or expand the currently focused tree widget branch, following the behaviour described here: https://github.com/david-vanderson/dvui/issues/618#issuecomment-3457871555  
The way I'm finding the first parent on left key down feels a bit fragile, so I'd love to hear if there's a better way!

I've also included a change to optionally restrict key navigation within a group to a specified direction.  
Some examples of where this makes a difference in the examples are the icon browser, and the grid "Layouts and Data" tab with "Horizontal scrolling" enabled:
- previously left/right keys would select the next item
- now left/right keys horizontally scroll the window

However, I'm not sure if this affects the accessibility use-case at all -- I'm happy to drop this change if it causes issues for anyone.

### DIFF
--- a/src/widgets/TreeWidget.zig
+++ b/src/widgets/TreeWidget.zig
@@ -17,6 +17,7 @@ id_branch: ?usize = null, // matches Reorderable.reorder_id
 drag_point: ?dvui.Point.Physical = null,
 drag_ending: bool = false,
 branch_size: Size = .{},
+current_branch_focus_id: ?dvui.Id = null,
 init_options: InitOptions = undefined,
 /// SAFETY: Set in `install`
 group: dvui.FocusGroupWidget = undefined,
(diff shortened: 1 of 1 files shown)


david-vanderson:
Looks great to me - merge when ready.  Thank you!

>  The way I'm finding the first parent on left key down feels a bit fragile, so I'd love to hear if there's a better way!

I think the way I would approach this would be to have a "current parent branch" variable inside Tree (or Branch), and each branch updates it when it becomes the new parent, and sets it back in deinit().  Then the code for "left" could look up that value directly.  Hope that makes sense - totally up to you whether to change it or not.

jjkh:
Thanks, that seems a better approach to me.

I'm now storing the current branch's button ID on the tree, and the parent branch's button ID on the branch.  
I'd have prefered to store the branch, but I wasn't able to get the button ID from the parent branch (as the button is already `deinit`ed) so opted to just store the ID to focus instead.

david-vanderson:
Looks great - merge it!

PR #628, opened 2025-11-14
build: add options for tiny-file-dialogs, freetype, and libc

### LINKED ISSUE #627: a `custom` backend always relies on `tinyfiledialogs` and `freetype`
Hi! I've been wanting to port dvui to the 3ds.

However after starting work on the backend I've seen that It won't compile as it currently hardcodes its dependencies on `freetype` and `tinyfiledialogs` even with the `custom` backend, so I can't do much without forking.

Is it possible to add a toggle to *not* include those libs? I see that you do so but only for `wasm`, I'm in a similar situation where i don't really have a libc.

david-vanderson:
Fixes #627 

@GasInfinity needed more control over what we were building with than the current simple "wasm" flag.  Thanks very much for starting this, I'm stealing a lot from https://github.com/GasInfinity-Forks/dvui/commit/b2c551d6b8a31b18460290c1d740b6bbbfb34646

I'm trying to add it as a build-time option, but if not provided, then the backend provides a default.  Kind of wonky, but seems to be okay so far.

- [X] tiny file dialogs
- [x] freetype
- [x] libc

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -38,6 +38,8 @@ pub fn build(b: *std.Build) !void {
         back_to_build = .sdl2;
     }
 
+    const libc_option = b.option(bool, "libc", "Use libc (default is backend specific)");
+    const freetype_option = b.option(bool, "freetype", "Freetype (or stb_truetype if false) for font rendering (default is backend specific)");
     const tiny_file_dialogs_option = b.option(bool, "tiny-file-dialogs", "OS-native file dialogs (default is backend specific)");
 
     const build_options = b.addOptions();
@@ -83,6 +85,8 @@ pub fn build(b: *std.Build) !void {
         .use_lld = use_lld,
         .accesskit = accesskit,
         .build_options = build_options,
+        .libc = libc_option,
+        .freetype = freetype_option,
         .tiny_file_dialogs = tiny_file_dialogs_option,
     };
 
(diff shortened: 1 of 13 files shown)


RedPhoenixQ (changes requested):
The asserts and min_size are needed so that we can guarantee that we can read back the added header and perform a `png_crc32` on it. If there is no buffer it will fail, possibly silently with a corrupted file. The JPNEncoder similarly calls "undo" a certain number of bytes to replace them with the correct dpi values.

I think the fix is to init the Allocating writer with the capacity of `PNGWriter.min_buffer_size` so that the asserts can pass.

david-vanderson:
> The asserts and min_size are needed so that we can guarantee that we can read back the added header and perform a `png_crc32` on it. If there is no buffer it will fail, possibly silently with a corrupted file. The JPNEncoder similarly calls "undo" a certain number of bytes to replace them with the correct dpi values.
> 
> I think the fix is to init the Allocating writer with the capacity of `PNGWriter.min_buffer_size` so that the asserts can pass.

Done - thanks!

david-vanderson:
@GasInfinity How does this look?

GasInfinity:
> @GasInfinity How does this look?

Looks like it works perfectly, thanks!

PR #631, opened 2025-11-15
feat: simple bar plots

leroycep:
This commit adds a function `bar` to draw a rectangle on a plot using the plot's coordinate system.

Bar plots will display "x to x+w, y to y+h" if `mouse_hover` was set when creating the `PlotWidget`.

The name `bar` (for "bar graph") was chosen because `box` conflicts with a field of `PlotWidget`. `rect` is another name that could be used.

`bar` immediately draws a rectangle, unlike the `line` API which returns a builder style object. This was partly because Paths only have a `fillConvex` function, however it is also the case that from one bar to the next there isn't any shared state, so a builder object isn't necessary.

---

This implementation feels like it could be improved, happy to hear any feedback.

### DIFF
--- a/src/Examples/plots.zig
+++ b/src/Examples/plots.zig
@@ -227,6 +227,102 @@ pub fn plots() void {
         }
         s1.stroke(1, if (valid) dvui.themeGet().focus else dvui.Color.red);
     }
+
+    {
+        const S = struct {
+            var stddev: f64 = 1.0;
+            var mean: f64 = 0;
+            var prng_seed: u64 = 2807233815221062137;
+            var npoints: u32 = 64;
+        };
+
+        const Static = struct {
+            var xaxis: dvui.PlotWidget.Axis = .{
+                .name = "Value",
+                .ticks = .{
+                    .locations = .{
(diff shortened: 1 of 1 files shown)


src/widgets/PlotWidget.zig
```diff
@@ -577,6 +587,56 @@ pub fn line(self: *PlotWidget) Line {
     };
 }
 
+pub const BarOptions = struct {
+    x: f64,
+    y: f64,
+    w: f64,
+    h: f64,
+    color: dvui.Color,
```
david-vanderson: Can color be an optional and if it's not given we use dvui.themeGet().focus?

leroycep: Went ahead and made the color optional and made it use `dvui.themeGet().focus` by default.

david-vanderson:
Thank you!

david-vanderson:
@leroycep In my dvui monthly devlog I'm crediting you as "geemili".  Let me know if you'd like me to use a different name.  Thanks!

leroycep:
@david-vanderson geemili works fine, I prefer it over leroycep online

PR #632, opened 2025-11-15
fix tab roles

leroycep:
This doesn't fix all instances where a `tabpanel` role needs to be added to the content, and I'm only going off of the ARIA documentation. I have no idea what effects it has for people actually using accessibility APIs.


david-vanderson:
Good catch.  Thank you!

PR #633, opened 2025-11-16
add opt-in for builtin fonts except the default/fallback

### LINKED ISSUE #630: Binary size skyrocketed from 175KB to 8.2MB with DVUI (ReleaseFast)
After somewhat finishing the rendering and input parts of a 3DS DVUI backend I saw that the binary size of the executable is now *x47 higher* than before. I expected it to be higher (obviously) but 47x is **too much**, something much more reasonable would be 1.8-2.0MB.

RedPhoenixQ:
Closes #630

This is mostly for the binary size savings and flexibility of the library. The default font is still relatively large (~65KB) and could be replaced by an optimized font that only includes ascii and a fallback character. I've quickly tested this and we can easily get to 10-20KB.

For now you must manually opt-in to the fonts of the builtin-in themes, but it might be valuable to have a compile flag to opt back in to automatically including the fonts for the default theme (or any/all builtin themes).

With this flexibility I've also re-added all the fonts available in the repo currently.

The Examples demo adds all its needed fonts when calling the `demo()` function. In the future we might once again want some flag for this. I'm imagining the web-demo perhaps wanting to not include any fonts and instead load them all in at runtime to improve the initial load and total wasm file size.

### DIFF
--- a/examples/web-test.zig
+++ b/examples/web-test.zig
@@ -46,7 +46,9 @@ export fn dvui_init(platform_ptr: [*]const u8, platform_len: usize) i32 {
         return 2;
     };
 
-    try WebBackend.win.fonts.addBuiltinFontsForTheme(gpa, dvui.Theme.builtin.adwaita_light);
+    WebBackend.win.fonts.addBuiltinFontsForTheme(gpa, dvui.Theme.builtin.adwaita_light) catch {
+        return 3;
+    };
 
     WebBackend.win_ok = true;
 


david-vanderson:
This looks good - do you know why the web test is failing?  I'll do some manual testing but might be a day or two.

david-vanderson:
I've done some testing, the only thing is we lost the automatic bundling of the Noto font in the non-web examples.  Not sure whether that's bad or not.  Will think about it.

In any case, this is good to merge when you are ready.  Thank you!

david-vanderson:
I just noticed that `zig build test` is now showing a lot of these:

```
[dvui] (warn): Font VeraId 0xe4eabf0819042c19 not in dvui database, using default
[dvui] (warn): Font VeraBdId 0x7dca5e830db608bf not in dvui database, using default
[dvui] (warn): Font VeraId 0xe4eabf0819042c19 not in dvui database, using default
[dvui] (warn): Font VeraBdId 0x7dca5e830db608bf not in dvui database, using default
```

Do we need to do something extra to add the fonts to the tests?

RedPhoenixQ:
Ah, yes. Just need to add the fonts in the testing init function. Will push a commit in a moment

david-vanderson:
Thank you!

Issue #421, opened 2025-06-29
Raylib Backend


jptrzy:
Is there to not compile raylib inside dvui.

To give raylib dep as a parameter for dvui.

EG.

**build.zig:**
```
...
const raylib_dep = b.dependency("raylib", .{
    .target = target,
    .optimize = optimize,
    .linux_display_backend = .Wayland,
});
const raylib_lib = raylib_dep.artifact("raylib");
exe.linkLibrary(raylib_lib);
...
```

**build.zig.zon:**
```
...
.raylib = .{
    .url = "git+https://github.com/raysan5/raylib#27a4fe885164b315a90b67682f981a1e03d6079c",
    .hash = "raylib-5.5.0-whq8uFqtNARw6t1tTakBDSVYgUjlBqnDppOyNfE_yfCa",
},
...
```

Can I reuse `raylib_dep` for dvui purpose.

PS. I don't understand the whole building process, so this question might sound stupid ;D

RedPhoenixQ:
I'm not sure there is an easy way to allow external build scripts to change the raylib import. What you could do is access dvuis raylib import, so you don't have to depend on raylib yourself. If you build dvui with `.backend = .raylib` you can do:
 
```zig
dvui.backend.c.RAYLIB_FUNCTIONS_AVAILABLE_HERE();
```

david-vanderson:
I agree with @RedPhoenixQ.  Can you say more about what situation you are in where you want to use your own raylib dependency?

jptrzy:
> dvui.backend.c.RAYLIB_FUNCTIONS_AVAILABLE_HERE();

I knew that I was just interested, if that was possible.

Say for eg that I would like to use custom version of raylib or use a newer one which dvui don't support or use a custom binding like https://github.com/Not-Nik/raylib-zig.

david-vanderson:
Got it.  The only way I know of currently to do that is to use a `.custom` backend and then copy-paste our existing raylib backend.  Not great.

I don't know if/when this situation will get better, but I'll leave this issue open for a while.  Thanks!

david-vanderson:
Will reopen this if any more information or interest comes up.  Thanks!

Logickin-Lambda:
Just faced the same situation and I have found this post.

[...]

PR #636, opened 2025-11-17
Raylib-zig Backend

Logickin-Lambda:
## Overview
This pr is related to [this issue](https://github.com/david-vanderson/dvui/issues/421), providing a more idiomatic experience for [raylib-zig](https://github.com/raylib-zig/raylib-zig) user to use dvui with custom raylib rendering without worrying the problem between C and zig type conversion especially for the string part. To double this as a guide, I am going to explain how users could use these functions as well.

## How to use the raylib-zig backend
Instead of using `.raylib` as the backend, `.raylib_zig` is used:

``` zig
	.target = target,
	.optimize = optimize,
	.backend = .raylib_zig,
});

exe.root_module.addImport("dvui", dvui_dep.module("dvui_raylib_zig"));
exe.root_module.addImport("backend", dvui_dep.module("raylib_zig"));
```

The setting above will let you use the raylib-zig based backend on the fly without injecting the raylib dependency, and you can access the raylib and raygui functions within the backend as shown:

``` zig
// It is imported from `backend` because of the code `exe.root_module.addImport("backend", dvui_dep.module("raylib"));`
const backend = @import("backend");
const raylib = backend.raylib; 
const raygui = backend.raygui;
```

Nevertheless, just in case if you wish to use a specific release of raylib-zig release and load raylib from your own, you may override the raylib dependency as shown:

``` zig
const dvui_dep = b.dependency("dvui", .{...});

// Inject the raylib modules you have imported from raylib-zig
const backend_mod = dvui_dep.module("raylib_zig");
backend_mod.addImport("raylib", raylib);
backend_mod.addImport("raygui", raygui);

exe.root_module.addImport("dvui", ...
```

With the set up above, you could also access raylib in the usual way `@import("raylib")`, and [you could mix and match both of them](https://github.com/Logickin-Lambda/dvui_raylib_zig_backend_test/blob/main/src/main.zig#L9) without causing issues; however, mixing both approaches is a bad practice and you should pick either approach based on your preference.

### What is the impact to the existing users:
It should be none to minimal because the current update will still default to c raylib backend because the raylib-zig backend is new and it requires user testing to ensure it work properly before it can become the default implementation, and for the existing C based backend project should still compile and run as usual without modifying build.zig (at least in the current update).

### Why c and zig backends and examples are written in separated files?
[There was a discussion](https://github.com/david-vanderson/dvui/issues/421#issuecomment-3510330290) and I was not satisfied with such implementation too because despite being logically identical, there are great enough difference between c and zig based design leading to two separated files, including minor function naming difference and how enums are accessed. It is possible to merge two backends into one so are the example, but I have realized that raylib backend might be more preferred by beginners who are just into the zig gui applications, and they need clear and easy to understanding examples. If I lump both of the examples and the backends into one, the example files will contain so many conditional compilations that is not necessary for the user end, resulting in confusing and frustrating beginners, and any future contributors into the raylib backend, and the ZLS server. Thus, based on these considerations, I decided to leave the current design as it is.

### Why can't I access glfw in the raylib-zig backend?
Based on my observation, the original reason for using glfw could be to offer [variable framerate](https://github.com/david-vanderson/dvui/blob/main/src/backends/raylib.zig#L876), possibly for improving the performance of the library, and there is a todo to replace the current design with `SUPPORT_CUSTOM_FRAME_CONTROL`, so there might be a chance that glfw will be removed for other solutions. It doesn't seem like `glfw` is intended to be opened for users, so I decided to not expose the `zglfw` dependency for the users unless there is a demand for that.

Alright, Let me know if any of you have any questions to the pr.

### DIFF
--- a/src/backends/raylib-zig.zig
+++ b/src/backends/raylib-zig.zig
@@ -121,7 +121,6 @@ pub fn begin(self: *RaylibBackend, arena: std.mem.Allocator) void {
 pub fn end(_: *RaylibBackend) void {}
 
 pub fn clear(_: *RaylibBackend) void {
-    // c.ClearBackground(c.BLANK);
     raylib.clearBackground(raylib.Color.blank);
 }
 
@@ -137,7 +136,6 @@ pub fn initWindow(options: InitOptions) !RaylibBackend {
 }
 
 pub fn init(gpa: std.mem.Allocator) RaylibBackend {
-    // if (!c.IsWindowReady()) {
     if (!raylib.isWindowReady()) {
         @panic(
             \\OS Window must be created before initializing dvui raylib backend.
@@ -174,7 +172,7 @@ pub fn accessKitShouldInitialize(self: *RaylibBackend) bool {
 
 pub fn accessKitInitInBegin(self: *RaylibBackend) !void {
     std.debug.assert(self.ak_should_initialized);
-    dvui.backend.c.ClearWindowState(dvui.backend.c.FLAG_WINDOW_HIDDEN);
+    raylib.clearWindowState(.{ .window_hidden = true });
     self.ak_should_initialized = false;
 }
 
@@ -211,7 +209,6 @@ pub fn drawClippedTriangles(self: *RaylibBackend, texture: ?dvui.Texture, vtx: [
 
     //make sure all raylib draw calls are rendered
     //before rendering dvui elements
-    // c.rlDrawRenderBatchActive();
     raylib.gl.rlDrawRenderBatchActive();
 
     if (clipr_in) |clip_rect| {
@@ -497,10 +494,8 @@ pub fn cursorShow(_: *RaylibBackend, value: ?bool) bool {
     const prev = !raylib.isCursorOnScreen();
     if (value) |val| {
         if (val) {
-            // c.ShowCursor();
             raylib.showCursor();
         } else {
-            // c.HideCursor();
             raylib.showCursor();
         }
     }
@@ -864,14 +859,11 @@ fn shiftAscii(ascii: u8) u8 {
     };
 }
 
-// pub fn raylibColorToDvui(color: c.Color) dvui.Color {
 pub fn raylibColorToDvui(color: raylib.Color) dvui.Color {
     return dvui.Color{ .r = @intCast(color.r), .b = @intCast(color.b), .g = @intCast(color.g), .a = @intCast(color.a) };
 }
 
-// pub fn dvuiColorToRaylib(color: dvui.Color) c.Color {
 pub fn dvuiColorToRaylib(color: dvui.Color) raylib.Color {
-    // return c.Color{ .r = @intCast(color.r), .b = @intCast(color.b), .g = @intCast(color.g), .a = @intCast(color.a) };
     return raylib.Color{ .r = @intCast(color.r), .b = @intCast(color.b), .g = @intCast(color.g), .a = @intCast(color.a) };
 }
 
@@ -885,7 +877,6 @@ pub fn EndDrawingWaitEventTimeout(_: *RaylibBackend, timeout_micros: u32) void {
     }
 
     if (timeout_micros > 0) {
-        // c.EndDrawing();
         raylib.endDrawing();
 
         // TODO: investigate raylib with SUPPORT_CUSTOM_FRAME_CONTROL that
@@ -914,8 +905,6 @@ extern "c" fn vsnprintf(
 
 extern "c" fn SetTraceLogCallback(callback: *anyopaque) void;
 
-// TODO: Raylib Library doesn't support the SetTraceLogCallback, I am going to skip this
-// part of the code for now until I have understood this callback and found a solution.
 fn raylibLogCallback(
     msgType: c_int,
     text: [*c]const u8,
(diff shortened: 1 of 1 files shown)


david-vanderson:
Code generally looks good!  After thinking about it some more, probably instead of using the `libc` flag we should add a new backend enum `raylib_zig`.

But first, I'm getting this error (on Mac):
```
dvanderson@Davids-Air dvui % zig build raylib-standalone -Dlibc=true 
raylib-standalone
└─ run exe raylib-standalone
   └─ compile exe raylib-standalone Debug native 2 errors
error: undefined symbol: _GuiIconText
    note: referenced by .zig-cache/o/75258c88d8668c4568607e77bfd57cd0/raylib-standalone_zcu.o:_GuiWindowBox
    note: referenced by .zig-cache/o/75258c88d8668c4568607e77bfd57cd0/raylib-standalone_zcu.o:_GuiTabBar
    note: referenced by .zig-cache/o/75258c88d8668c4568607e77bfd57cd0/raylib-standalone_zcu.o:_GuiTabBar
    note: referenced by .zig-cache/o/75258c88d8668c4568607e77bfd57cd0/raylib-standalone_zcu.o:_GuiTabBar
    note: referenced 2 more times
error: undefined symbol: _GuiValueBox
    note: referenced by .zig-cache/o/75258c88d8668c4568607e77bfd57cd0/raylib-standalone_zcu.o:_GuiSpinner
error: the following command failed with 2 compilation errors:
```

Are you seeing this?

david-vanderson:
Hmm I am NOT getting that error on Linux.  Will investigate.

david-vanderson:
It works for me if I revert the translateC changes.  I don't know what is going on there, but sounds like we should use the old way for now so this can move forward?

Logickin-Lambda:
> But first, I'm getting this error (on Mac):

>It works for me if I revert the translateC changes. I don't know what is going on there, but sounds like we should use the old way for now so this can move forward?

Thanks! I am glad you have tested Mac, and macOS is always my biggest concern since I have no mac on hand to test the update, so usually I can't tell what is the result from my end until other have tested the code. Perhaps in some days, I need to buy the cheapest mac mini just to test things due to the kind of application I aimed for. Usually, Windows and Linux should work fine because I can test these os with my laptop.

If that bug doesn't cause any issues using the original cimport, I will switch it back for now.

> Code generally looks good! After thinking about it some more, probably instead of using the libc flag we should add a new backend enum raylib_zig.

Sure, I did thought and even attempted that once before, but I had gotten some issues about the option `linux_display_backend` which zig doesn't allow me to have multiple options with the same name, this means that the whole block to declare the `linux_display_backend `must be relocated at the top of the `buildBackend()` function, it needs to explicitly be declared only if raylib backend are used:

```
pub fn buildBackend(backend: enums_backend.Backend, test_dvui_and_app: bool, dvui_opts_in: DvuiModuleOptions) void {
    var dvui_opts = dvui_opts_in;
    const b = dvui_opts.b;
    const target = dvui_opts.target;
    const optimize = dvui_opts.optimize;

    var linux_display_backend: LinuxDisplayBackend = undefined;
    if (backend == .raylib or backend == .raylib_zig){
        linux_display_backend = b.option(LinuxDisplayBackend, "linux_display_backend", "If using raylib, which linux display?") orelse blk: { ... };
    }
    switch (backend) { ...
}
```

If you don't mind about `linux_display_backend` flying across the switch statements, I will use this approach to separate raylib-c and raylib-zig.

david-vanderson:
> If that bug doesn't cause any issues using the original cimport, I will switch it back for now.

Good - and don't worry - I can test on Mac when needed.

> If you don't mind about `linux_display_backend` flying across the switch statements, I will use this approach to separate raylib-c and raylib-zig.

Sounds good - thanks!

Logickin-Lambda:
All the suggested changes have been done. TranslateC has been reverted to Cimport, and both zig and c raylib backend are now separated.

Nonetheless, the `linux_display_backend` situation was more complicated than I thought since I have [missed out a loop](https://github.com/Logickin-Lambda/dvui/blob/zig-based-backend-prototype/build.zig#L115) for building all the backend at once, and if there are more than one raylib backend, more than one backend will require the option `linux_display_backend`. Thus, my original proposed solution didn't work and I ended up putting the option next to other options in the [function `build()` instead](https://github.com/Logickin-Lambda/dvui/blob/zig-based-backend-prototype/build.zig#L47) which it will be triggered once only if the build process involves raylib.

Since the name of the zig based backend has been changed, I will update the guide above and in the readme later.

Logickin-Lambda:
Done, the readme and the guide on this pr has been updated.

david-vanderson:
I'm getting this compile error when running `zig build`:

```
src/backends/raylib-zig.zig:177:18: error: root source file struct 'raylib-zig' has no member named 'c'
    dvui.backend.c.ClearWindowState(dvui.backend.c.FLAG_WINDOW_HIDDEN);
```

Do you know what the equivalent raylib-zig call would be?

Logickin-Lambda:
Oh, Seems I have missed that function, my bad. Zig version of the ClearWindowState is the following:

``` zig
raylib.clearWindowState(.{ .window_hidden = true });
```

If you have encountered other raylib-zig functions similar to this one, you may use the following rules of thumb for conversion:

- `raylib` instead of `dvui.backend.c`
- The first letter of the function is a small letter instead of capital as the original
- Flags are replaced by struct fields
- Enum values are replaced by zig enums
- if you see c based function started with rl, these can be accessed in `raylib.gl` for the zig version
- `SetTraceLogCallback` is the only exception that is not included in raylib-zig, so this must be loaded externally.

I might missed out some details for the conversion, but ZLS should covers the remaining missing stuff which I have ensured it works before submitting the pr.

Let me know if you have other questions.

Edit: I am going to change it since I also found that there are quite a lot of uncleaned commented code during the prototyping phase, so I am going to clean the file as well.

david-vanderson:
Excellent.  Thank you very much!

david-vanderson:
@Logickin-Lambda I'm writing this up in my monthly dvui devlog, and will credit you as "Accipiter Nova".  Let me know if you'd like me to use a different name.  Thanks!

Logickin-Lambda:
Sure! Please use "Accipiter Nova".

PR #640, opened 2025-11-17
refactor: ziggify accessibility and make it more portable

GasInfinity:
Basically refactors the current `accessibility` support in DVUI and makes it as platform-independent as possible so no checks have to be done in separate places, everything is inside the specific `accessibility.Instance`.

Also adds a `Null` accessibility instance, which is the default when `AccessKit` is not linked and is literally zero-overhead.

I've tested this on linux on all the possible backends available (and with `orca`) and looks like it works great! I've also compiled successfully all windows backends but haven't tested them.

As an example, here's what it looks like now to add actions to an accessibility node:
```
if (b.data().a11y_node) |node| {
    node.addAction(.focus);
    node.addAction(.click);
}
```

which will compile to nothing with a `Null` accessibility and will set the `Role`s in the `AccessKit` one

---
cc @phatchman (as you appear quite a lot in the `git blame`)


david-vanderson:
I like how this looks (although I'm not fond of the "a11y" contraction):
```
if (b.data().a11y_node) |node| {
    node.addAction(.focus);
    node.addAction(.click);
}
```

I'm unclear on other aspects of the change:
* how does it work now that `accessKitInitInBegin()` is gone?
* are there really other accessibility libraries we might use?  I don't like adding extra indirection unless we are gaining another implementation
* why the changes to build.zig and renaming cacheBuster?

But worryingly, when I compile on windows `zig build sdl3-standalone -Daccesskit=shared` and run the accessibility highlights it no longer is showing anything inside the window.

Do you have access to a windows machine for testing?

GasInfinity:
> although I'm not fond of the "a11y" contraction

Bikeshed it please, It's because I don't know other contraction for accessibility that is both short and you know it's "accessibility" when reading it so I used a standard contraction.

> how does it work now that accessKitInitInBegin() is gone?

We assert that the user of the library won't create a window in a shown state so all adapters can work. After initializing the DVUI Window the user can show the window and do as they please. DVUI handles that automatically in `dvui.App`.

> are there really other accessibility libraries we might use? I don't like adding extra indirection unless we are gaining another implementation.

It's not really for other implementations (even though they may be welcome if someone wants to do it), it's to remove all `if (dvui.accesskit_enabled) <> else <>`, the `accessibility.Instance` should handle all platform-dependent behaviour. Also previously even if someone didn't enable `AccessKit` (I don't think we have it in a 3ds :p) it would still have some overhead (the acessibility node still existed in every widget, even if it was a `?*struct{}`), now it doesn't as every node is an `?enum{}`.

> why the changes to build.zig and renaming cacheBuster?

The commit explains the changes "link 'userenv' instead of 'Userenv' to allow cross-compiling to windows from linux with AccessKit enabled". I moved `cacheBuster` as I saw it wasn't being used in the tree directly (and a `tools/` directory existed)

> But worryingly, when I compile on windows zig build sdl3-standalone -Daccesskit=shared and run the accessibility highlights it no longer is showing anything inside the window.
Do you have access to a windows machine for testing?

I will investigate it then. I don't have any way to test it on macos though.

phatchman:
@GasInfinity Thanks so much for looking at this. Apologies, I don't have a lot of time to go through this in detail, but my comments are below.

I think there are really 3 changes here and have different comments on each.
1) There's a Zig interface to AccessKkit that wraps the C API. (AccessKit.zig)
2) It makes the logic in the backends more consistent.
3) It creates a new generic 'accessibility layer'. (accessibility.zig and the Null.Zig 'implementation' which mirrors AccessKit.zig's interface.)

For 1)
Whether it makes sense to zigify the API is a good question, but I deliberately chose not to do that as it was a layer I did not want to maintain while AccessKit evolves. I created the code generator in the tools directory to deal with that problem instead. To me it comes down to the maintenance cost of having c-like code 'polluting' the zig code vs having a layer where we have to do some sort of diff between the old and new AccessKit header files to work out what changed. Then manually keep the zig mapping layer up to date.

Edit: Though to be fair we also need to consider the cost of keeping the AccessKit code generator up to date. I don't think that is a big task, but I'm not the best to judge given I wrote it. 

Also, there is an update coming for AccessKit that will accept slices for strings. That will get rid of some of the worst awkwardness of the current API by eliminating the string null termination step.

For 2) and 3)
My personal opinion is that it makes sense to try and be as generic as possible across backends. But it doesn't make sense (at this time at least) to make it generic across accessibility libraries.

My reasoning is that:
1) There is still more work to do both on the AccessKit side and on the DVUI side to finish integration support. i.e. the APIs being used on either side are not stable and this seems like another layer that needs to be kept up to date every time a new AccessKit library is released.
2) Related to above, it adds a lot of code that does not really do anything, e.g. every time I want to use a new AccessKit API, I'll need to add that function to AccessKit.zig and Null.zig and create new enum definitions.
3) If we are going to make a more generic zig-like layer, I'd prefer to either a) improve the code generator that transforms the AccessKit header file into zig functions or b) Just create those functions directly in AccessKit.zig, without making them generic.
4) I just generally prefer the conditional compilation that we have today over moving it behind a comptime interface. My guess is there are simpler ways to handle the conditional code. e.g. similar to the way the current accesskit_node() function makes sure no AccessKit calls are compiled in if it is not enabled. 
5) I don't really see a clear benefit to that added code vs the current system where I can directly use any AccessKit function I want. That is pretty valuable during what is still an experimental integration stage as I don't always know what API's will work or how they work to use them until I actually use them. Edit: This could be mitigated somewhat by just calling the raw C APIs until it is understood what is needed, but it still is just extra steps.
5) Does this take into account that each top-level window (i.e. Operating System Window) could have its own AccessKit node tree? I'm not sure we handled that case currently, but we don't want to preclude it in the future. e.g. it looks like it's implemented as a Singleton?
6) If we really did need to support another accessibility library, I don't really know what that interface would look like. It seems better to split it out into an interface once we have a clearly understood usecase.

But for the backends, we definitely want them as similar as possible. I'm a little short on time at the moment and have only looked at the diffs, but that part looks like a good improvement. @RedPhoenixQ might be interested in looking at the backend changes.

If we do want to merge the more zig-like wrapper and/or the accessibility layer, my preference is just to have AccessKit.zig and not have the Null version. Mostly because this is an area of code likely to change once I can turn my attention back to the accessibility integration and is just more code that I'd have to add each time I use a new AccessKit feature. 

As an example, the new code I have for implementing text runs reaches into the rendering layer, will need to implement 2 new events and use ~6 more AccessKit APIs. 

Paul.

GasInfinity:
This fixes also one bug, `AccessKit` previously relied on a stable pointer, *but it is not guaranteed have one*. If someone moved the `Window` and had `AccessKit` enabled, the userdata pointer would point to invalid stack memory (unless `dvui.Window` is in the heap), now the state is a separate allocation, fixing that.

That above fixes the problems with windows. Windows and Linux have been tested and they both compile and work in all backends. I can't test MacOS so please if someone can, it would be really great.

---
re: @phatchman
 
> while AccessKit evolves

While working on this I saw [accesskit-c](https://github.com/AccessKit/accesskit-c/tree/main) because obviously working on somewhat generalizing this is useless if it has too many breaking changes and it doesn't look like it changes too much? 

> To me it comes down to the maintenance cost of having c-like code 'polluting' the zig code

I mean, before if some API changed you had to change *all* callsites, now everything goes through one function. Obviously I'm biased (as I wrote this) but I don't see how the code generator resolves the problem of c polluting the code as it's still the C API but renamed.

> it doesn't make sense (at this time at least) to make it generic across accessibility libraries

I'm not trying to make it generic across accessibility libraries (currently), it's an AccessKit abstraction that allows having a `Null` backend (more below on this).

> Also, there is an update coming for AccessKit that will accept slices for strings

Great, then the only change would be dropping `:0` in the `setValue` function and calling `dupe` instead of `dupeZ`. If the API changes, you'd only need to change the wrapped function and each callsite (only to not allocate the sentinel, it would still compile though).


> the APIs being used on either side are not stable

I understand but they don't change too much? (from my perspective).

> it adds a lot of code that does not really do anything

That's the purpose of a `Null` backend, to not do anything. Even though if you look at the diff *I removed more lines* than added.

> Just create those functions directly in AccessKit.zig

You can do that, just check `accessibility.Instance == accessibility.AccessKit` (making it `pub`) and you can still do conditional compilation when experimenting, then you'd add what you've worked on in the `Null` one so you can remove the checks

> similar to the way the current accesskit_node() function makes sure no AccessKit calls are compiled in if it is not enabled.

That's great! However every node was still stored in memory and the state of the `AccessKit` impl was still also in memory so it would be wasted storage when it isn't even enabled.

> Does this take into account that each top-level window (i.e. Operating System Window) could have its own AccessKit node tree? I'm not sure we handled that case currently, but we don't want to preclude it in the future. e.g. it looks like it's implemented as a Singleton?

This PR does not change that in any way, if you look at the dx11 standalone demo, creating a new window automatically makes it accessible so its *not* a singleton, it's even decoupled somewhat from `dvui`.

> If we really did need to support another accessibility library, I don't really know what that interface would look like. It seems better to split it out into an interface once we have a clearly understood usecase.

And that makes a lot of sense, this is more of a way to not have `AccessKit` use memory even when disabled, if you know of other way to do so please discuss it.

> my preference is just to have AccessKit.zig and not have the Null version.

Literally the `Null` version is the whole reason of the PR as e.g a 3DS doesn't have `AccessKit` but it still uses memory.

> As an example, the new code I have for implementing text runs reaches into the rendering layer, will need to implement 2 new events and use ~6 more AccessKit APIs.

You can still prototype it in any way you like, this doesn't make you having to do an API as you do so.


Overall you can bikeshed the APIs but *please* allow a `Null` accessibility backend of some sort as currently people who don't (or can't!) use `AccessKit` have to pay for it (not in code, only in memory usage).

phatchman:
What extra memory usage do you mean? 

There is no ak_node in non-accesskit builds and no additional code as far as I know? If that is the ultimare goal, perhaps we can look at options for that that don't require replicating every acesskit function?

GasInfinity:
> There is no ak_node in non-accesskit builds and no additional code as far as I know? If that is the ultimare goal, perhaps we can look at options for that that don't require replicating every acesskit function?

While there's no `?*c.accesskit_node`, there's still a `?*struct{}` which is still an extra pointer not used for anything. Also all the state in `AccessKit` was still in-memory even if not used. With this PR `@sizeOf(Node)` and `@sizeOf(Instance)` are basically `0` when `AccessKit` is not enabled.

You can either add lots of checks in different parts of the codebase (why? isn't that just tech debt?) or have it everything centralized in one place so things don't have to be changed everytime something external to DVUI changes.

david-vanderson:
> > There is no ak_node in non-accesskit builds and no additional code as far as I know? If that is the ultimare goal, perhaps we can look at options for that that don't require replicating every acesskit function?
> 
> While there's no `?*c.accesskit_node`, there's still a `?*struct{}` which is still an extra pointer not used for anything. Also all the state in `AccessKit` was still in-memory even if not used. With this PR `@sizeOf(Node)` and `@sizeOf(Instance)` are basically `0` when `AccessKit` is not enabled.
> 
> You can either add lots of checks in different parts of the codebase (why? isn't that just tech debt?) or have it everything centralized in one place so things don't have to be changed everytime something external to DVUI changes.

Minimizing memory cost when not using accesskit is a worthy goal - let's get some numbers so we don't miss the target.

@GasInfinity what are we looking at here?  Does it impact binary size at all?  Or is it just the size of `WidgetData`?  How would we know if we did this correctly?

GasInfinity:
> Does it impact binary size at all?

Binary size is *not* affected (at least AFAIK), only runtime/allocated memory.

e.g: check `@sizeOf(Null)` vs `@sizeOf(AccessKit.State)`. Same for a `Node`

david-vanderson:
> > Does it impact binary size at all?
> 
> Binary size is _not_ affected (at least AFAIK), only runtime/allocated memory.
> 
> e.g: check `@sizeOf(Null)` vs `@sizeOf(AccessKit.State)`. Same for a `Node`

Ok now we are getting somewhere.  We can easily put in a conditional to turn `ak_node` into a zero sized thing when not using accesskit.  And the size of the accesskit state is currently 112 bytes per dvui.Window (usually only one).

That's not enough to warrant putting in a whole layer of indirection.

Sorry, but the abstraction layer is out for the moment.  I very much appreciate the consideration here.  If there are other parts of this change you'd like me to look specifically at, let me know.

GasInfinity:
I'm *really* sorry for wasting your time reviewing and looking at this.

After giving it some thought I think I'll just hard fork and maintain a separate branch focused on what I need. You're completely welcome and 100% free to cherry-pick any commit, ping or ask me anything if you'd like or need.

david-vanderson:
> I'm _really_ sorry for wasting your time reviewing and looking at this.
> 
> After giving it some thought I think I'll just hard fork and maintain a separate branch focused on what I need. You're completely welcome and 100% free to cherry-pick any commit, ping or ask me anything if you'd like or need.

It's no problem at all, and certainly not a waste.  I'm sorry we couldn't figure out a solution for now.  And as always, I could be wrong here.  I know maintaining a fork is annoying, hopefully at some point we can figure out how to sync up.

Thank you very much for the effort and patiently answering my questions!

phatchman:
To confirm: `std.debug.print("{}\n", .{ @sizeOf(dvui.WidgetData)});`

As expected, this gives me 504 bytes with the -Daccesskit=off and the ak_node excluded and 512 with ak_node included.

It's not the prettiest line of code in the world, but this change will conditionally compile out the node when -Daccesskit=off

```
ak_node: if (dvui.accesskit_enabled) ?*dvui.AccessKit.Node else void = if (dvui.accesskit_enabled) null else {},
```
I looked at removing the default value and instead set the node to null in init(), but there are places that construct a WidgetData struct directly (e.g. Window).

@david-vanderson OK to apply this?

david-vanderson:
> @david-vanderson OK to apply this?

Yes - thank you!

PR #641, opened 2025-11-17
Plot: fix plots with constant data and add manually specified tick locations

leventev:
I recently noticed that when we provide constant data the max and min values for an Axis are equal. Meaning that when we try to divide by (max - min) we divide by 0. I fixed this by increasing the max by 1 and decreasing the min by 1 when they are equal in `PlotWidget.deinit()`. I tested this and it works but I think it's worth considering adding a relative error field in `PlotWidget.Axis`. Because if the user wants to plot data that is the result of some float arithmetic then it's unlikely the floats are going to be exactly equal when they "should be equal". There will likely be some very small round-off error. For example if the max is 5.00000012 and the min is 4.9999999 then the user likely doesnt want the bounds of the plot to be these values but rather wants to see a straight line at ~5 while the bounds are 4 and 6.

Also another problem I found is that when using logarithmic scales and providing incorrect values(x <= 0) the plot silently fails and no values/ticks are displayed(since there are no min and max values). I think in this case printing warning messages would be helpful since this is unlikely to be what the user wants. A warning would help them realize something is wrong on their end. But I was hesitant to add this because it would flood the user's terminal.

Also I added the ability to manually specify the tick locations. Which is one of the features from the list discussed in #622.
On this note I think it would be worth it to create an issue with a TODO list of planned features and slowly tick them off as they are implemented. It would also make design discussions easier.

### DIFF
--- a/src/Examples/plots.zig
+++ b/src/Examples/plots.zig
@@ -146,9 +146,34 @@ pub fn plots() void {
     }
 
     {
+        const freqs = [_]f64{ 1, 1e1, 1e2, 1e3, 1e4, 1e5, 1e6, 1e7, 1e8 };
+
         const S = struct {
             var resistance: f64 = 159;
             var capacitance: f64 = 1e-6;
+
+            var xaxis: dvui.PlotWidget.Axis = .{
+                .name = "Frequency",
+                .scale = .{ .log = .{} },
+                .ticks = .{
+                    .locations = .{
+                        .custom = &freqs,
+                    },
+                    .format = .{
+                        .custom = formatFrequency,
+                    },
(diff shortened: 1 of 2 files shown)


david-vanderson:
Looks good, and I agree with everything you said:
* great catch on the constant data
* can figure out how to deal with very small ranges in the future (probably some option like you said)
* for invalid data, one thing we could do is put a label "invalid data" on top of the graph itself?
  * this helps end users, who might not be seeing any logs, but would know something went wrong
* adding a plot TODO issue sounds good - go for it!

So this is good - merge when ready.  Thank you!

Issue #104, opened 2024-08-14
Support for inset/outset border styles


VisenDev:
Follow up to #81

Inset/Outset border style, as well as square corners from #81 will open the door to creatings cool 90s user interface themes with dvui. See https://www.w3schools.com/css/css_border.asp for examples of inset/outset border styles

VisenDev:
Example of desired user interface style
![windowmaker](https://github.com/user-attachments/assets/5962c4ef-2a5c-41b3-8cbe-484091870b4b)

david-vanderson:
#643 and #645 have finally given us a way to do this.  There is a new builtin theme "win98" which does some of this.  It probably needs more work, but I think this issue can be closed.  Thank you!

PR #643, opened 2025-11-17
Implement Ninepatch Rendering

desttinghim:
This pull request implements ninepatch rendering and adds a function for rendering one to the dvui API. It does not integrate ninepatches into the theming system in any way.

### DIFF
--- a/src/render.zig
+++ b/src/render.zig
@@ -529,6 +529,7 @@ pub fn renderNinepatch(ninepatch: Ninepatch, rs: RectScale, opts: NinepatchOptio
 
     var rs_center_center = rs;
 
+    //Move rects into position
     rs_top_center.r.w -= rs_top_left.r.w + rs_top_right.r.w;
     rs_top_center.r.x += rs_top_left.r.w;
     rs_top_center.r.h = rs_top_left.r.h;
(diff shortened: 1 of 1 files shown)


david-vanderson:
This is super cool.  I've read through the code once, and it looks pretty good.  I'm also very excited that this will be a general solution to #104 (which I've been stuck on forever).

There are two things I want to do before merging:
* handle a too small size without erroring - we should render something, even if it doesn't look good (and maybe log an error)
* give this more space in the demo with controlling options (maybe could integrate it into the "layout" part where we already do the image stuff, but might be too complicated there, alternately have a separate place in "styling")

If you are willing to do either of these that would be great.  Let me know if/when you want me to run with it.  Thank you!

desttinghim:
> * handle a too small size without erroring - we should render something, even if it doesn't look good (and maybe log an error)

When `error.NinepatchBelowMin` is returned, a black rectangle with a 1px white border is now rendered as a placeholder. Let me know if you want a different solution.

> * give this more space in the demo with controlling options (maybe could integrate it into the "layout" part where we already do the image stuff, but might be too complicated there, alternately have a separate place in "styling")

The example ninepatch code is now in the styling section, and there are sliders to control it's width and height. 

I also fixed a bug where gaps were appearing between the patches. Now the positions are floored and the sizes are ceiled, which seems to have fixed the problem.

foxnne:
I think this is great, just wanted to chime in and say pulled this and tested and it's exciting to have this feature coming! I think this could allow some really interesting UI elements. Well done!

david-vanderson:
This is excellent!  I suspect the api will change a bit as we work to figure out how people generally want to use it, but that's always the case and we can do that in main.

Thank you!

david-vanderson:
@desttinghim When I write this up in the devlog, should I credit you as "Louis Pearson"?

desttinghim:
Louis Pearson is fine.

PR #646, opened 2025-11-18
Merge init and install

### LINKED ISSUE #428: merge widget init and install
The reason dvui has the init/install split was to allow widgets to get a rect and then modify it before install.  It's rarely used, so maybe we can find a separate mechanism for it.

This issue is to track down all the uses in the dvui codebase of a widget doing something between init and install to understand what's needed.

RedPhoenixQ:
Closes #428

I decided to test this out ("it" being a `self: *Self` in the `init` function) and refactored `BoxWidget` to take a mutable self pointer (that should usually be `undefined`). If you add the argument as a normal self parameter you can do: 

```zig
var box: BoxWidget = undefined; // OR `dvui.widgetAlloc(BoxWidget);` here instead
box.init(@src(), .{}, .{});
```

There are some additional issues with nested widgets where the ordering of the `install` calls determined the parent chain, which now forces the previous `install` ordering on the `init` calls. Basically you can't really call `init` on the nested widgets when initializing the main widget. 

I think it would also be valuable to add a `initAllocated` (or similar) to each widget that would handle the `dvui.widgetAlloc` and setting any "is allocated" boolean needed. It should also call init for you (only negative is needing to repeat the init arguments), so it would work pretty much like the previous init without needing any `undefined`, but returning a pointer (importantly without calling all the other widget methods you might not want, like `drawBackground`).


david-vanderson:
I had the same thought to try it on box - you beat me to it!

A quick look over the code and it looks pretty good to me, especially:
> ```zig
> var box: BoxWidget = undefined; // OR `dvui.widgetAlloc(BoxWidget);` here instead
> box.init(@src(), .{}, .{});
> ```

> There are some additional issues with nested widgets where the ordering of the `install` calls determined the parent chain, which now forces the previous `install` ordering on the `init` calls. Basically you can't really call `init` on the nested widgets when initializing the main widget.

Good point, we'll have to take this slowly - no need to convert everything at the same time.

> I think it would also be valuable to add a `initAllocated` (or similar) to each widget that would handle the `dvui.widgetAlloc` and setting any "is allocated" boolean needed. It should also call init for you (only negative is needing to repeat the init arguments), so it would work pretty much like the previous init without needing any `undefined`, but returning a pointer (importantly without calling all the other widget methods you might not want, like `drawBackground`).

I'm unsure if this is needed, but we will figure that out as we go.

I'll take a close look at this tomorrow (hopefully).  Thank you!

david-vanderson:
Looks great!  I'm ready to move forward with this (one widget at a time), so merge when ready.  Thank you!

RedPhoenixQ:
For consistency, I was thinking of just adding more commits to this PR, mostly during this weekend I think. That way the api stays consistent and we can update the README with an up-to-date recommended way of handling widgets

david-vanderson:
> For consistency, I was thinking of just adding more commits to this PR, mostly during this weekend I think. That way the api stays consistent and we can update the README with an up-to-date recommended way of handling widgets

That makes sense, good idea.

RedPhoenixQ:
@david-vanderson Now most of the widgets are done. The missing ones are:

- `Scrol[Area/Container]Widget`
  - I don't fully understand how/why you'd use the current `installScrollBars` and the significance of orderings here
  - Also can't figure out a clean api to represent the issue above
- Most `FloatingXXXWidget`
  - All the floating widgets allow for modifying some part of their placement between `init` and `install`
  - I've made an attempt at an api on `FloatingMenuWidget` [6596e28](https://github.com/david-vanderson/dvui/pull/646/commits/6596e284305ffd0294f525f9bd61815e1a4ff111) (that might work for the others, not sure)
- `ReorderWidget`
  - because I don't fully understand it and its re-registering
- `TreeWidget`
  -  because of `TreeWidget.Branch` that has a function that explicitly requires to be called between `init` and `install` (could maybe be an init_opt?)
  
  Might have missed some nested "widget-like" things (like `TreeWidget.Branch` but for other widgets).

david-vanderson:
Thank you!

> @david-vanderson Now most of the widgets are done. The missing ones are:

No problem, I will look into them.  If I can't figure it out quickly, we can merge most of this anyway.

Here's my list of things to look into:
* Scroll Area/Container
  * yeah that installScrollbars thing is weird, it was to enable user-controlled Container inside ScrollArea, will investigate
* All Floating widgets
  * possibly these just have a more complicated api if I can't figure it out
  * will try to do Dropdown better and see if I can avoid the preInit/postInit
* ReorderWidget
* TreeWidget
* ensure that the demo wiggle button still works (I made a commit to main prior to this branch that shows a way to do it)
* CacheWidget.invalidate()
  * I think we remove this in favor of an init_option
* TextLayoutWidget.init()
  * calling self.data().id during the creation of self seems fishy, does zig guarantee that self.wd is okay at that point?

RedPhoenixQ:
> TextLayoutWidget.init()
> * calling self.data().id during the creation of self seems fishy, does zig guarantee that self.wd is okay at that point?

Your probably right, that's an oversight on my part. Will fix right away

david-vanderson:
> > * calling self.data().id during the creation of self seems fishy, does zig guarantee that self.wd is okay at that point?
> 
> Your probably right, that's an oversight on my part. Will fix right away

Thanks!  And thanks for cleaning up that selection_store thing.

david-vanderson:
I think I've addressed all the things I wanted to.  I'm happy leaving Reorder.Reorderable and Tree.Branch with the init/install split (similar to Tooltip).  Maybe we'll figure out some way of dealing with them in the future, but they all have the idea of "ask for a rect, then decide whether to use it or not".

So I'm ready to merge this.  Anything else you wanted to do or look at before we merge?

david-vanderson:
@RedPhoenixQ I'm assuming you are okay with this, so I plan to merge it later today.

RedPhoenixQ:
@david-vanderson Yeah, sorry didn't notice your question about if there is anything else. All good to merge on my part

Just leaving a note here about [5ac465e](https://github.com/david-vanderson/dvui/pull/646/commits/5ac465e93c29bde156e70dc224f98fbf68ddc997) that removed comments above `undefined`, which I think should be removed as a default value and moved in as a part of the `init` functions. I don't think it should be "easy" to manually assigning a widget (where default fields make it easy like `box = .{};` or close to it) as not calling `init` would be invalid either way. Can easily be done in a followup PR/commit on main

david-vanderson:
> @david-vanderson Yeah, sorry didn't notice your question about if there is anything else. All good to merge on my part

Thanks!

> 
> Just leaving a note here about [5ac465e](https://github.com/david-vanderson/dvui/pull/646/commits/5ac465e93c29bde156e70dc224f98fbf68ddc997) that removed comments above `undefined`, which I think should be removed as a default value and moved in as a part of the `init` functions. I don't think it should be "easy" to manually assigning a widget (where default fields make it easy like `box = .{};` or close to it) as not calling `init` would be invalid either way. Can easily be done in a followup PR/commit on main

Good point - I agree.

david-vanderson:
Thank you!

PR #647, opened 2025-11-27
zig 0.16 updates

david-vanderson:
DON'T USE THIS YET.

only sdl2 and sdl3 are working so far


src/dvui.zig
```diff
@@ -415,30 +415,42 @@ pub fn logError(src: std.builtin.SourceLocation, err: anyerror, comptime fmt: []
...
+            const stderr, const tty_config = std.debug.lockStderrWriter(&.{});
+            defer std.debug.unlockStderrWriter();
+            stderr.writeAll("error return context:\n") catch {};
+            std.debug.writeStackTrace(t, stderr, tty_config) catch {};
+        }
+    }
+    if (!builtin.strip_debug_info) {
+        std.debug.print("\nstack trace:\n", .{});
+        std.debug.dumpCurrentStackTrace(.{});
+    }
```
RedPhoenixQ: On the web backend I believe `std.debug.print` doesn't work, or at least the the `dumpCurrentStackTrace`. I think it was because of their direct use of `stderr`. The previous solution uses the `log` api to avoid this issue.

The if conditions might always disable the traces on web, but that might be something we want to support.

david-vanderson:
Now (with zig 0.16.0-dev.1484+d0ba6642b) I'm able to run `zig build sdl3-app`, sdl3-ontop, and sdl3-standalone.

For some reason the demo lags heavily when rendering the demo buttons (both before and just after), haven't looked into it yet.

PR #648, opened 2025-11-28
sdl3gpu: fix for flickering issue and moving swapchain control to the sample app

peterino2:
After catching the flicker in renderdoc i realize dvui must not be expecting the backend to clear on every single render. I'm now doing the clear as part of the app instead of the backend implicitly.

I also fixed up a few vulkan validation errors involving situations where the swapchain format differs from the rendertarget formats.


david-vanderson:

Issue #642, opened 2025-11-17
Plot improvements


leventev:
A tracker for a TODO list of the features mentioned in #622 and #641.

- [X] manually specify ticks
- [ ] named tick labels instead of numbers, for example 10 -> A, 20 -> B, 30 -> C
- [x] subticks
- [x] automatically locating "pretty" ticks
- [ ] logarithmic tick formatting, for example 10^5
- [ ] formatting hovered data
- [ ] marks for the data points (X, hollow circle, full circle, square, ...)
- [ ] legend
- [ ] different line drawing options (none, line, dotted or dashed)
- [ ] error bars
- [ ] ability to set the axis labels and tick labels to be displayed on the right or on the top
- [ ] fill between two plot-lines
- [ ] set padding between the plot line and the borders of the data box
- [ ] stem plots
- [ ] identify when the data is constant possibly by checking the relative error of (max - min)
- [x] rotated Y axis label
- [ ] unit based axes, for example pi is 1 unit

david-vanderson:
Let me know if you see any weirdness with rotated text.  One limitation that I know of is that if the text is vertical (so the start is lower on the screen than the end), then if the start of the text goes outside the clipping region the whole text disappears.

This is a limitation of how renderText interprets the rectScale argument.  It's on my list.

leventev:
Looks really good! Thanks!

PR #649, opened 2025-11-30
Plot: find nice numbered tick locations, subticks, flexible tick side configuration

leventev:
Implementation of some features listed in #642.

Previously tick locations were assigned in a very naive way, basically dividing the axis into a given number of ticks, resulting in ugly tick numbers. Now instead of providing a fixed tick count a suggestion is provided. The number of ticks and their locations are computed based on the data range (max - min) and the suggested number of ticks. It finds a step size between ticks that is a multiple of a nice number `(1, 2, 5, 10)`, some examples: `(20, 30, 40), (65, 70, 75, 80), (1000, 2000, 3000)`. This is similar to Heckbert's algorithm. In the future a more complex algorithm like Wilkinson's or Modified Wilkinson's could be implemented but I wanted to minimize complexity and instead have a simple solution that would cover most general use cases. If tick locating in the Axis options is set to auto both linear and logarithmic scales now find tick locations like this.

I also added subticks, they are located very simply:
- linear scales: 3 subticks are added at 25% 50% and 75% between 2 ticks
- logarithmic scales: [2, base) subticks are added between 2 ticks, for example in the case of `base=10` the subticks are `(2,3,4,5,6,7,8,9)` which can be seen in the RC filter example

I don't think subticks need to have as detailed configuration as normal ticks. A possibility could be to have as an option the number of subticks for linear scales and the multiples of the tick for logarithmic scales. But for now it's simply a bool switch.

Now it's possible to specify which side of an axis the tick lines are displayed on(none, left/top, right/bottom, both).

Also now it's possible to specify the color of gridlines(and gridlines for subticks) making it possible to integrat with themes. If the colors are `null` then the gridlines are not drawn(this is the default).

### DIFF
--- a/src/widgets/PlotWidget.zig
+++ b/src/widgets/PlotWidget.zig
@@ -651,6 +651,8 @@ fn drawTickline(
     tick_line_color: dvui.Color,
     gridline_color: ?dvui.Color,
 ) void {
+    if (tick_p.x < self.data_rs.r.x or tick_p.x > self.data_rs.r.x + self.data_rs.r.w) return;
+    if (tick_p.y < self.data_rs.r.y or tick_p.y > self.data_rs.r.y + self.data_rs.r.h) return;
     // these are the positions for ticks on the left or top
     const line_start, const line_end, const gridline_start, const gridline_end = switch (dir) {
         .horizontal => blk: {


david-vanderson:

david-vanderson:
@leventev I'm putting these plotting improvements in my monthly dvui devlog, and will credit you as "leventev".  Let me know if you'd like me to use a different name.  Thanks!

leventev:
@david-vanderson Thank you! You can either use leventev or Levente Vámos. As I've seen you used the real names of others so I'm fine with either.

PR #653, opened 2025-12-06
struct_ui: Change missing default value to runtime error.

phatchman:
- Changes missing default values for optionals from a comptime error to a runtime debug log message.
- The optional will remain null if it is attempted to be set through struct_ui.

### DIFF
--- a/src/struct_ui.zig
+++ b/src/struct_ui.zig
@@ -903,7 +903,7 @@ fn canDisplayPtr(ptr: std.builtin.Type.Pointer) bool {
 /// NOTE:
 /// Any modified text fields are dynamically allocated. These are cleaned up during Window.deinit()
 /// If a string should not be automatically cleaned up (i.e will be cleaned up by a struct's deinit() method),
-/// removed the string from struct_ui.string map prior to the Window.deinit() being called.
+/// remove the string from struct_ui.string map prior to the Window.deinit() being called.
 ///
 /// The displayStringBuf() function can be used as an alternative to display strings with a user-supplied buffer.
 pub fn displayStruct(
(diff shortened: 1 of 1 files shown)


phatchman:
This pr makes the initial change made by @david-vanderson in #645 for the struct_ui examples work without compilation errors. In fact no code would be required to just display the ninepatch fields. 

Any optional struct without either 1) a default initialization or 2) a default value supplied through StructOptions, can now be displayed without error, but if that field is ever set to null, it cannot then be set to a non-null via the struct_ui's default display functions. (e.g. `displayStruct()`)

If any attempt is made to do so, a debug message is logged:
```
debug(dvui): struct_ui: Optional field ninepatch_fill cannot be selected as no default value is provided. Use struct_ui.StructOptions(Ninepatch) to provide a default.
debug(dvui): struct_ui: Optional field ninepatch_hover cannot be selected as no default value is provided. Use struct_ui.StructOptions(Ninepatch) to provide a default.
debug(dvui): struct_ui: Optional field ninepatch_press cannot be selected as no default value is provided. Use struct_ui.StructOptions(Ninepatch) to provide a default.
```
With this pr applied, the following lines can be removed from #645 in src/Examples/struct_ui.zig
```
// Set dummy default values for the Ninepatch and the ImageSource union as they can't be default initialized.
StructOptions(dvui.Ninepatch).initWithDefaults(.{}, null),
StructOptions(dvui.Texture).init(.{}, null),
StructOptions(@FieldType(dvui.Texture.ImageSource, "pixelsPMA")).init(.{}, null),
StructOptions(@FieldType(dvui.Texture.ImageSource, "pixels")).init(.{}, null),
StructOptions(dvui.ImageSource.ImageFile).init(.{}, null),
```
This should lessen the need for an "exclude fields" API. So we'll see if another use-case for exclusion comes up.

david-vanderson:
Awesome thank you!

PR #656, opened 2025-12-09
TreeWidget: Allow animating folder closing

foxnne:
Hi!

https://github.com/user-attachments/assets/a0cadc36-962e-47cc-8982-4404d7277a9e

Hopefully the video shows the intent, that we can also use the `startEnd` animation to reverse the animation on close. 

I tested this briefly in my own program and in the demo, both with and without an animation duration and things seem to work. One minor thing to note is that the close animation will not use the passed in easing fn, because any animations with overshoot, will delay the actual close until the animation is complete, which looks funny. I instead opted to just animate the close linearly instead, and allow the open animation to have a fancy easing fn. 

Please let me know if you'd like me to make any changes!

Thanks as always

### DIFF
--- a/src/widgets/TreeWidget.zig
+++ b/src/widgets/TreeWidget.zig
@@ -435,12 +435,8 @@ pub const Branch = struct {
             }
         }
 
-        if (self.anim) |a| {
-            if (a.end()) {
-                if (self.expanded) {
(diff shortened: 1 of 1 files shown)


david-vanderson:
Great!  Thank you!

PR #657, opened 2025-12-10
fonts: start towards font fallback and system fonts

david-vanderson:
This is some steps towards a better font subsystem.  The direction I think this is headed:
* separate Font.Source (ttf files) from Font (which is now parameters to search for a Font.Source)
* one dvui internal bucket of Font.Source
  * used for apps that embed their own fonts
  * also used for apps that load their own fonts at runtime, but they know where the fonts are coming from
* one "system" bucket of Font.Source
  * used for apps that want to let the user select any font on their system

The goals for this pull request are:
* more flexible font choice (this is not implemented yet)
  * like font.bold() giving you a bold version of whatever font you had
  * selecting a font based on family name and parameters rather than specifying a given ttf
  * reducing the number of fonts in a Theme, since now you can get a bold version of the theme font
* baby steps towards font fallback


david-vanderson:
@MichaelBelousov @foxnne @RedPhoenixQ @phatchman and anyone else reading: Would appreciate your input here, specifically about removing `Options.font_style` and only having one font in `Theme`.

For removing `Options.font_style` in favor of directly specifying `Options.font`, it means that most places that did:
`.font_style = .title_4` or `.font_style = .heading`
would now do:
`.font = dvui.themeGet().font.larger(2)` or `.font = dvui.themeGet().font.withWeight(.bold)`

An application could put these into functions, like:
```
fn fontHeader() dvui.Font {
    return dvui.themeGet().font.withWeight(.bold);
}
```

dvui itself only used `font_style` in a few places where it wanted a bold version of the body font, and a slightly larger/smaller version.

Slimming down `Theme` to only have a single font (representing body text) makes some sense:
* the other fonts there had "intent names" (title, caption) but they only affected the visual aspect of the text
* all the builtin themes did the same thing with all of them
* but if you want multiple different fonts in the dvui.Theme it's hard

And I notice that in Pixi you do specify a few different fonts in your theme.

(Also the Theme now has an `embedded_fonts` field where you can list all the fonts and dvui will load them for you)

So I'm thinking of adding a few more fonts back to Theme, like:
* `.font_app1`, `.font_app2`, `.font_app3`
  * accessed like `.font = .app1()`
  * this would address needing multiple fonts together in a theme
  * leaves meaning up to app
  * is 3 enough?  (For Pixi you'd have Vera (and Vera Bold) in Theme.font, so you'd need 2 - one for Vera Mono and one for CozetteVector)
  * is wanting a monospace font common enough to warrant a `Theme.font_mono`?
* maybe a few helpers like `.font = .bold()` or `.font = .larger(2)`
  * makes it obvious these are only affecting the font, and don't imply intent

How does this sound?  Any opinions would be helpful!

RedPhoenixQ:
I really like these changes over all. Removing the `.title_X` is a good change in my opinion but I definitely think a theme should be able to specify more than one font, maybe 2-4. One should clearly be the default and another a clear secondary, that way the places the used the `.title_X` fonts would use "secondary" to indicate that they would _prefer_ to be visually distinct from the "body"/default font. The additional fonts could be used by the application to set a `font_mono` (or we could name it that and it would probably be fine). I know we didn't go with "primary" and "secondary" for color names in `Theme`, but I still prefer that naming as it doesn't specify it's intended use, but you can in your code clearly indicate that you intend for some text to be visually distinct (which should translate between different Themes, so that one Theme doesn't use `app2` as a mono space font and another users `app2` for titles. That would mean any title blindly using `.app2` might have very different visual effects for different themes.)

As for monospace, I think that should be a property of the Font, like italic (but separate from it). That way the font family can have a monospace version and we could even force rendering of non-monospace fonts in fake monospace (mainly thinking for the bundled fallback font).

In this rework it would be nice to consider [variable fonts](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Fonts/Variable_fonts). Think would mainly mean that weight can specify an arbitrary number and that a single font file could be a valid choice for multiple weights (as it can become any weight) and other parameters like italic etc. For weight a 400 might be a `const normal: Weight = 400` and when looking for a source you could maybe use any font within 100 (so 365 would accept a non variable font with weight 400, or "normal"). I know FreeType supports variable fonts, but it would be a bit of work setting it up. There is also "named" presets included in some variable font files, like a preset for "Bold" that sets the weight and maybe changes letter spacing etc, would should probably be exposed in some way. Don't have a good solution here.

For `Options` I would prefer not to have to do `dvui.getTheme()...` whenever I want to modify a font. Not a deal breaker or blocker for the PR, but an unfortunate inconvenience. Specifying some modifiers where you could set `.font_opts = .{ .weight = .bold, .add_size = 8 }` would feel much nicer, still keeping the `.font` field to override the font completely. That way when you override options and want a bold version of whatever the passed in options had, it would look the same as when modifying the theme font. Perhaps the `.font` field should be a `*const Font` not that it's 60+ bytes on it's own, at least when there is a `.font_opts` field available. 

Sorry for the big text dump but wanted to get all my ideas down. I'm heading to bed now

MichaelBelousov:
Sounds good to me.

I actually personally really don't mind the explicit `dvui.themeGet().font` but some `.bold` helpers could indeed also nice.

I do think mono/title font attributes of a theme are worth having. I know I will be adding a mono font soon hopefully.

Also I just want to mention I expect you can still override fonts per widget in options without using a font in the theme. I can imagine an app that uses more than 4 non-theme-bound fonts (e.g. font per character or something maybe if it were a game dialog UI.

foxnne:
I think this sounds great! I don't have much to add I don't think, at least for me I think the fonts I'm using are all I'd ever need. I like the idea of helpers for modifying a current font, instead of needing different fonts for different styles.

david-vanderson:
> I really like these changes over all. Removing the `.title_X` is a good change in my opinion but I definitely think a theme should be able to specify more than one font, maybe 2-4. One should clearly be the default and another a clear secondary, that way the places the used the `.title_X` fonts would use "secondary" to indicate that they would _prefer_ to be visually distinct from the "body"/default font. The additional fonts could be used by the application to set a `font_mono` (or we could name it that and it would probably be fine). I know we didn't go with "primary" and "secondary" for color names in `Theme`, but I still prefer that naming as it doesn't specify it's intended use, but you can in your code clearly indicate that you intend for some text to be visually distinct (which should translate between different Themes, so that one Theme doesn't use `app2` as a mono space font and another users `app2` for titles. That would mean any title blindly using `.app2` might have very different visual effects for different themes.)

This is hard.  I don't expect different apps would be able to use "primary"/"secondary" more usefully than "app1"/"app2", because nobody would agree on what each of those means.

I think the most commonly used font_styles were .heading (bold) and .title_4 (2 pixels larger).

I'll try a few meaningful names, like:
* "font_body" - used for normal text content
* "font_mono" - same situations as font_body but monospace
* "font_header" - bolder than font_body, used for subwindow titles, selected tabs
* "font_title" - larger than font_body

Maybe also a few "app1"/"app2"/"app3" ones purely for applications that have multiple themes (but don't need themes to work across apps, but I'd wait for it to come up.  Likely an app that is doing more sophisticated theming is already doing something on top of dvui Theme.

> Also I just want to mention I expect you can still override fonts per widget in options without using a font in the theme.

Yes this will always be supported.  After this PR lands, you would use `dvui.addFont()` to load the extra fonts, and then reference them by name using `.font = .find(...)`.  Perhaps in the future we'll be able to reference fonts that aren't loaded into the dvui internal bucket (either from the system or one-off).

> 
> As for monospace, I think that should be a property of the Font, like italic (but separate from it). That way the font family can have a monospace version and we could even force rendering of non-monospace fonts in fake monospace (mainly thinking for the bundled fallback font).

I also thought this would work, but it appears this is not how fonts currently work.  For example, "Times New Roman" has no monospace version, and "Courier New" has no variable-width version.  CSS treats mono/serif/sans as totally separate.  I thought a bit about trying to do something here, but I think it would involve shoving multiple fonts with different family names together, and so much of font lookup is based on that.

> 
> In this rework it would be nice to consider [variable fonts](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Fonts/Variable_fonts).

That would be cool, but too much for this effort.  I'm trying to leave the door open with the Font.Weight enum, we could extend it for 100/200/300 etc. weights.

> 
> For `Options` I would prefer not to have to do `dvui.getTheme()...` whenever I want to modify a font. Not a deal breaker or blocker for the PR, but an unfortunate inconvenience. Specifying some modifiers where you could set `.font_opts = .{ .weight = .bold, .add_size = 8 }` would feel much nicer, still keeping the `.font` field to override the font completely. That way when you override options and want a bold version of whatever the passed in options had, it would look the same as when modifying the theme font. 

Let me play around with some other options here.

> Perhaps the `.font` field should be a `*const Font` not that it's 60+ bytes on it's own, at least when there is a `.font_opts` field available.

I agree having Font be relatively large is annoying, but I'd like to try this option for a while:
* it answers the question of memory ownership for the family name
* using a pointer I think forces you to create the Font not inline in the options

> I actually personally really don't mind the explicit dvui.themeGet().font but some .bold helpers could indeed also nice.

Good feedback!

> I think this sounds great! I don't have much to add I don't think, at least for me I think the fonts I'm using are all I'd ever need. I like the idea of helpers for modifying a current font, instead of needing different fonts for different styles.

Great!  Thank you!

david-vanderson:
> Let me play around with some other options here.

Theme now has:
* font_body
* font_heading
* font_title
* font_mono

We have these options for getting a `Font`:
* `.font = .theme(.heading)`
* `.font = themeGet().font_heading`
* `.font = Font.theme(.heading).larger(2)`

Any of the options immediately produces a `Font`, which means they resolve the theme (and pull the font from the theme) right there.

This could result in a corner case where you specify both a font `.theme(.heading)` and a theme pointer `.theme = &my_theme` in the same Options.  The font resolution by default doesn't look through at `Options.theme`.

I tried a few ways to work around this, but it got very complicated.  I suspect this will be a very niche issue, so I'm not going to worry about it right now.

It does mean that widgets internally should take care to source theme fonts through Options.themeGet() rather than using Font.theme directly.

PR #659, opened 2025-12-16
Make AccesskitOptions public

Southporter:
This allows apps that build off of DVUI to expose the same option to the users with out having to juggle enum definitions.


david-vanderson:
Thank you!

Issue #539, opened 2025-09-03
basic editor with code/syntax highlighting


david-vanderson:
In #531 @phatchman showed a cool demo that used the zig tokenizer to syntax highlight some code in a textLayout.

This issue is a place to brainstorm what we would need for a basic highlighting editor.

While textLayout supports adding text chunk by chunk (with separate colors), textEntry doesn't have an api for that.

Questions:
* How do we communicate highlighting to textEntry?
  * zig function? 
  * pile of regexes? 
  * text piece by piece?
* How much text can be in a textEntry before we run into performance issues?
* What kind of different highlighting do people need?

MichaelBelousov:
so just an FYI, if code highlighting is a big use case here (and I suppose all other use cases are subsets technically), I am already aware of:
- [textmate](https://macromates.com/manual/en/language_grammars) grammars (used by VS Code)
- [treesitter](https://tree-sitter.github.io/tree-sitter/) grammars (used by github, has a C-API and the ability to redo highlighting without reparsing all the text)

I'm not sure how heavyweight those are (not that it matters much I guess, I just am thinking about the web case, and thanks to lazy zig it's all opt-in anyway). Being able to swap in your own highlighter (e.g. the zig tokenizer) would be nice, but probably negligible savings for most use cases

br07h3rj:
I am very interested in this use case. I'll be happy to find its limits when the time comes.

> What kind of different highlighting do people need?

basic:
- font and font size (it would be fine to force the same line-height for now)
- normal, bold, italic, underline
- text color, text background color

wishlist:
- boxes around text (`focusBorder()` could probably do this)
- a way to attach tooltips or other functionality to particular points in the text (please?)
- don't force the same line-height (or assume the bigger line-height and use gravity to position the smaller text, like subscript and superscript)
- strike-through (I'm sure someone will eventually want it)

> How much text can be in a textEntry before we run into performance issues?

If we could do it without rendering or applying attributes to any more text than is on the screen (or maybe just a little more), it should perform pretty well. I'll plan on testing for that.

MichaelBelousov:
I'll add that I think there's a high chance I'll need this at some point in the future for minor JavaScript and Python editing, so at that point I'll be contributing, I'm guessing using tree sitter if there isn't any progress yet

br07h3rj:
Found something interesting.
I just checked the `textEntry` I've already got (from Aug 20th DVUI and Zig 0.14.1).

[...]

PR #660, opened 2025-12-19
syntax highlighting

david-vanderson:
Towards implementing #539

It looks like tree-sitter is a good default direction for syntax highlighting.  So one goal is to make it easy to use a compiled tree-sitter grammar.

A second goal is to make it possible to use other syntax highlighting options by giving calling code control over the part of textEntry where it feeds the text to textLayout.

As of now this branch modifies the app example to show some editable c code with the identifiers colored in red.

Stuff to be done:
- [x] do full syntax highlighting using tree-sitters query stuff
- [x] figure out how to match tree-sitter node types (symbols) to dvui.Options
- [x] make it work incrementally
- [x] make it work with large documents (only reparse and rematch the displayed text
- [x] figure out how to make tree-sitter integration nice but optional (how large is integrating the tree-sitter library? how large is each compiled grammar?)
- [x] decide who owns the tree-sitter memory (does dvui figure out a way to clean it up?  do we punt this to client code?)

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -850,6 +850,27 @@ pub fn addDvuiModule(
         }
     }
 
+    const tree_sitter_dep = b.lazyDependency("tree_sitter", .{
+        .target = target,
+        .optimize = optimize,
+    });
+    if (tree_sitter_dep) |tsd| {
+        dvui_mod.linkLibrary(tsd.artifact("tree-sitter"));
+    }
+
+    const tree_sitter_c_dep = b.lazyDependency("tree_sitter_c", .{
+        .target = target,
+        .optimize = optimize,
+    });
+    if (tree_sitter_c_dep) |tsc| {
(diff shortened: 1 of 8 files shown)


david-vanderson:
I think this is in a good shape to push and get some feedback on:
* textEntry has some "tree_sitter" init opts to make the integration easy (hopefully)
* simple (maybe too simple) mapping of capture names to dvui.Options (linear scan so you can put more specific ones earlier)
* works incrementally with "cache_layout" to only reparse the edits, plus only render the visible text with highlights
* dvui always bundles the tree-sitter library
  * doesn't (yet) work on web (I'm hopeful we can make that happen but might need to fork and strip the library)
  * the library itself is tiny, so negligible impact on binary size
  * the large thing is the grammars, but those are brought by the using code

PR #661, opened 2025-12-21
AccessKit: Add text_run support

phatchman:
Initial text run support for testing.

- TextEntryWidget creates the TextLayoutWidget with a role of .none, and passes as the accessibility parent via Options.ak_node_parent.
- Each line of text must be in its own text_run.
- TextLayoutWidget creates a VirtualParentWidget using `self.bytes_seen` as the `extra_id` to generate a node_id for the text_run for that line
- In renderText() records the position and dimensions of each character in the text_info multi array list.
- Must be careful as renderText can delay rendering so text_info might not be populated when rederText() returns.
- In cases where text won't be rendered by renderrText, a text_run should not be created.
- AccessKits textRunPopulate() fills in the text run details for the node based on text_info from renderText(),

Complete:
- text_run for single and multi-line input boxes
- text_run for read-only text layout widgets.
- text selection action

TODO:
- [ ] Better support for text that has scrolled off screen. 
    - From testing, believe I should use the last visible word as the bounds for all non-visible text.

@ethindp Would appreciate if you could try this for the various text inputs and see if it behaves as you expect. 
Does it make sense to implement this for read-only text as well? 
Or is it better to implement this for any selectable text, whether read-only or not?

@david-vanderson Any ideas on what I could use as id_extra for nodeCreate when processing each line of text in the TextLayoutWidget? Alternatives I thought about:
1) The rounded y position, which would be more reliable, but that doesn't seem 100% safe? In theory someone could display 2 different lines of text in y position 10.1 and 10.2 with a very zoomed-out view. Also as above, there are issues if you don't finish an addText with a newline as the next text_run will have the same y position and thus same id_extra.
2) A hash of the string. But that doesn't work if someone puts the same string on multiple lines (e.g. single newlines especially)
3) Make textAdd supply a @src()? Then use line numbers vs the passed in @src()? - Actually, I think you'd need to add an id_extra as well, in case textAdd is called in a loop with the same @src() without any newlines.

If a duplicate id is generated, the application will panic, so I'd like to have something we can 100% rely on if possible.

Edit: I think just keeping a count of how many lines the TextLayoutWidget has processed, combined with how many times textAdd has been called will work. I will try that and see if it fixes things.

Thanks.

### DIFF
--- a/examples/app.zig
+++ b/examples/app.zig
@@ -96,9 +96,8 @@ pub fn frame() !dvui.App.Result {
     defer scroll.deinit();
 
     var tl = dvui.textLayout(@src(), .{}, .{ .expand = .horizontal, .font = .theme(.title) });
-    const lorem = "This is a dvui.App example that can compile on multiple backends.";
+    const lorem = "This is a dvui.App example that can compile on multiple backends.\n";
     tl.addText(lorem, .{});
-    tl.addText("\n", .{});
     tl.format("Current backend: {s}", .{@tagName(dvui.backend.kind)}, .{});
     if (dvui.backend.kind == .web) {
         tl.format(" : {s}", .{if (dvui.backend.wasm.wasm_about_webgl2() == 1) "webgl2" else "webgl (no mipmaps)"}, .{});
@@ -111,14 +110,12 @@ pub fn frame() !dvui.App.Result {
         \\- paints the entire window
         \\- can show floating windows and dialogs
         \\- rest of the window is a scroll area
+        \\
+        \\
     , .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Framerate is variable and adjusts as needed for input events and animations.", .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Framerate is capped by vsync.", .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Cursor is always being set by dvui.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("Framerate is variable and adjusts as needed for input events and animations.\n\n", .{});
+    tl2.addText("Framerate is capped by vsync.\n\n", .{});
+    tl2.addText("Cursor is always being set by dvui.\n\n", .{});
     if (dvui.useFreeType) {
         tl2.addText("Fonts are being rendered by FreeType 2.", .{});
     } else {
--- a/examples/dx11-ontop.zig
+++ b/examples/dx11-ontop.zig
@@ -239,12 +239,9 @@ fn dvui_floating_stuff() void {
     tl.deinit();
 
     var tl2 = dvui.textLayout(@src(), .{}, .{ .expand = .horizontal });
-    tl2.addText("The dvui is painting only floating windows and dialogs.", .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Framerate is managed by the application (in this demo capped at vsync).", .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Cursor is only being set by dvui for floating windows.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("The dvui is painting only floating windows and dialogs.\n\n", .{});
+    tl2.addText("Framerate is managed by the application (in this demo capped at vsync).\n\n", .{});
+    tl2.addText("Cursor is only being set by dvui for floating windows.\n\n", .{});
     if (dvui.useFreeType) {
         tl2.addText("Fonts are being rendered by FreeType 2.", .{});
     } else {
--- a/examples/dx11-standalone.zig
+++ b/examples/dx11-standalone.zig
@@ -177,18 +177,17 @@ fn gui_frame() !void {
         \\- can show floating windows and dialogs
         \\- example menu at the top of the window
         \\- rest of the window is a scroll area
+        \\
+        \\
     , .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Framerate is variable and adjusts as needed for input events and animations.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("Framerate is variable and adjusts as needed for input events and animations.\n\n", .{});
     if (vsync) {
-        tl2.addText("Framerate is capped by vsync.", .{});
+        tl2.addText("Framerate is capped by vsync.\n", .{});
     } else {
-        tl2.addText("Framerate is uncapped.", .{});
+        tl2.addText("Framerate is uncapped.\n", .{});
     }
-    tl2.addText("\n\n", .{});
-    tl2.addText("Cursor is always being set by dvui.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("\n", .{});
+    tl2.addText("Cursor is always being set by dvui.\n\n", .{});
     if (dvui.useFreeType) {
         tl2.addText("Fonts are being rendered by FreeType 2.", .{});
     } else {
--- a/examples/raylib-ontop.zig
+++ b/examples/raylib-ontop.zig
@@ -150,12 +150,9 @@ fn dvuiStuff() void {
     tl.deinit();
 
     var tl2 = dvui.textLayout(@src(), .{}, .{ .expand = .horizontal });
-    tl2.addText("The dvui is painting only floating windows and dialogs.", .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Framerate is managed by the application (in this demo capped at vsync).", .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Cursor is only being set by dvui for floating windows.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("The dvui is painting only floating windows and dialogs.\n\n", .{});
+    tl2.addText("Framerate is managed by the application (in this demo capped at vsync).\n\n", .{});
+    tl2.addText("Cursor is only being set by dvui for floating windows.\n\n", .{});
     if (dvui.useFreeType) {
         tl2.addText("Fonts are being rendered by FreeType 2.", .{});
     } else {
--- a/examples/raylib-standalone.zig
+++ b/examples/raylib-standalone.zig
@@ -139,18 +139,17 @@ fn dvui_frame() bool {
         \\- can show floating windows and dialogs
         \\- example menu at the top of the window
         \\- rest of the window is a scroll area
+        \\
+        \\
     , .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Framerate is set by Raylib (C api).", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("Framerate is set by Raylib (C api).\n\n", .{});
     if (vsync) {
-        tl2.addText("Framerate is capped by vsync.", .{});
+        tl2.addText("Framerate is capped by vsync.\n", .{});
     } else {
-        tl2.addText("Framerate is uncapped.", .{});
+        tl2.addText("Framerate is uncapped.\n", .{});
     }
-    tl2.addText("\n\n", .{});
-    tl2.addText("Cursor is always being set by dvui.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("\n", .{});
+    tl2.addText("Cursor is always being set by dvui.\n\n", .{});
     if (dvui.useFreeType) {
         tl2.addText("Fonts are being rendered by FreeType 2.", .{});
     } else {
--- a/examples/raylib-zig-ontop.zig
+++ b/examples/raylib-zig-ontop.zig
@@ -154,12 +154,9 @@ fn dvuiStuff() void {
     tl.deinit();
 
     var tl2 = dvui.textLayout(@src(), .{}, .{ .expand = .horizontal });
-    tl2.addText("The dvui is painting only floating windows and dialogs.", .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Framerate is managed by the application (in this demo capped at vsync).", .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Cursor is only being set by dvui for floating windows.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("The dvui is painting only floating windows and dialogs.\n\n", .{});
+    tl2.addText("Framerate is managed by the application (in this demo capped at vsync)\n\n.", .{});
+    tl2.addText("Cursor is only being set by dvui for floating windows.\n\n", .{});
     if (dvui.useFreeType) {
         tl2.addText("Fonts are being rendered by FreeType 2.", .{});
     } else {
--- a/examples/raylib-zig-standalone.zig
+++ b/examples/raylib-zig-standalone.zig
@@ -138,18 +138,17 @@ fn dvui_frame() bool {
         \\- can show floating windows and dialogs
         \\- example menu at the top of the window
         \\- rest of the window is a scroll area
+        \\
+        \\
     , .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Framerate is set by Raylib (raylib-zig).", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("Framerate is set by Raylib (raylib-zig).\n\n", .{});
     if (vsync) {
-        tl2.addText("Framerate is capped by vsync.", .{});
+        tl2.addText("Framerate is capped by vsync.\n", .{});
     } else {
-        tl2.addText("Framerate is uncapped.", .{});
+        tl2.addText("Framerate is uncapped.\n", .{});
     }
-    tl2.addText("\n\n", .{});
-    tl2.addText("Cursor is always being set by dvui.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("\n", .{});
+    tl2.addText("Cursor is always being set by dvui.\n\n", .{});
     if (dvui.useFreeType) {
         tl2.addText("Fonts are being rendered by FreeType 2.", .{});
     } else {
--- a/examples/sdl-ontop.zig
+++ b/examples/sdl-ontop.zig
@@ -132,17 +132,15 @@ fn dvui_floating_stuff() void {
     tl.deinit();
 
     var tl2 = dvui.textLayout(@src(), .{}, .{ .expand = .horizontal });
-    tl2.addText("The dvui is painting only floating windows and dialogs.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("The dvui is painting only floating windows and dialogs.\n\n", .{});
     tl2.addText("Framerate is managed by the application", .{});
     if (vsync) {
-        tl2.addText(" (capped at vsync)", .{});
+        tl2.addText(" (capped at vsync)\n", .{});
     } else {
-        tl2.addText(" (uncapped - no vsync)", .{});
+        tl2.addText(" (uncapped - no vsync)\n", .{});
     }
-    tl2.addText("\n\n", .{});
-    tl2.addText("Cursor is only being set by dvui for floating windows.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("\n", .{});
+    tl2.addText("Cursor is only being set by dvui for floating windows.\n\n", .{});
     if (dvui.useFreeType) {
         tl2.addText("Fonts are being rendered by FreeType 2.", .{});
     } else {
--- a/examples/sdl-standalone.zig
+++ b/examples/sdl-standalone.zig
@@ -152,18 +152,17 @@ fn gui_frame() bool {
         \\- can show floating windows and dialogs
         \\- example menu at the top of the window
         \\- rest of the window is a scroll area
+        \\
+        \\
     , .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Framerate is variable and adjusts as needed for input events and animations.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("Framerate is variable and adjusts as needed for input events and animations.\n\n", .{});
     if (vsync) {
-        tl2.addText("Framerate is capped by vsync.", .{});
+        tl2.addText("Framerate is capped by vsync.\n", .{});
     } else {
-        tl2.addText("Framerate is uncapped.", .{});
+        tl2.addText("Framerate is uncapped.\n", .{});
     }
-    tl2.addText("\n\n", .{});
-    tl2.addText("Cursor is always being set by dvui.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("\n", .{});
+    tl2.addText("Cursor is always being set by dvui.\n\n", .{});
     if (dvui.useFreeType) {
         tl2.addText("Fonts are being rendered by FreeType 2.", .{});
     } else {
--- a/examples/sdl3gpu-ontop.zig
+++ b/examples/sdl3gpu-ontop.zig
@@ -154,17 +154,15 @@ fn dvui_floating_stuff() void {
     tl.deinit();
 
     var tl2 = dvui.textLayout(@src(), .{}, .{ .expand = .horizontal });
-    tl2.addText("The dvui is painting only floating windows and dialogs.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("The dvui is painting only floating windows and dialogs.\n\n", .{});
     tl2.addText("Framerate is managed by the application", .{});
     if (vsync) {
-        tl2.addText(" (capped at vsync)", .{});
+        tl2.addText(" (capped at vsync)\n", .{});
     } else {
-        tl2.addText(" (uncapped - no vsync)", .{});
+        tl2.addText(" (uncapped - no vsync)\n", .{});
     }
-    tl2.addText("\n\n", .{});
-    tl2.addText("Cursor is only being set by dvui for floating windows.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("\n", .{});
+    tl2.addText("Cursor is only being set by dvui for floating windows.\n\n", .{});
     if (dvui.useFreeType) {
         tl2.addText("Fonts are being rendered by FreeType 2.", .{});
     } else {
--- a/examples/sdl3gpu-standalone.zig
+++ b/examples/sdl3gpu-standalone.zig
@@ -152,18 +152,17 @@ fn gui_frame() bool {
         \\- Efficient texture uploads via transfer buffers
         \\- Render passes for structured GPU work
         \\- Supports multiple shader formats (SPIR-V, DXIL, MSL)
+        \\
+        \\
     , .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Framerate is variable and adjusts as needed for input events and animations.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("Framerate is variable and adjusts as needed for input events and animations.\n\n", .{});
     if (vsync) {
-        tl2.addText("Framerate is capped by vsync (PRESENTMODE_VSYNC).", .{});
+        tl2.addText("Framerate is capped by vsync (PRESENTMODE_VSYNC).\n", .{});
     } else {
-        tl2.addText("Framerate is uncapped (PRESENTMODE_IMMEDIATE).", .{});
+        tl2.addText("Framerate is uncapped (PRESENTMODE_IMMEDIATE).\n", .{});
     }
-    tl2.addText("\n\n", .{});
-    tl2.addText("Cursor is always being set by dvui.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("\n", .{});
+    tl2.addText("Cursor is always being set by dvui.\n\n", .{});
     if (dvui.useFreeType) {
         tl2.addText("Fonts are being rendered by FreeType 2.", .{});
     } else {
--- a/examples/web-test.zig
+++ b/examples/web-test.zig
@@ -140,18 +140,17 @@ fn dvui_frame() !void {
         \\- can show floating windows and dialogs
         \\- example menu at the top of the window
         \\- rest of the window is a scroll area
+        \\
+        \\
     , .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Framerate is variable and adjusts as needed for input events and animations.", .{});
-    tl2.addText("\n\n", .{});
-    tl2.addText("Cursor is always being set by dvui.", .{});
-    tl2.addText("\n\n", .{});
+    tl2.addText("Framerate is variable and adjusts as needed for input events and animations.\n\n", .{});
+    tl2.addText("Cursor is always being set by dvui.\n\n", .{});
     if (dvui.useFreeType) {
-        tl2.addText("Fonts are being rendered by FreeType 2.", .{});
+        tl2.addText("Fonts are being rendered by FreeType 2.\n", .{});
     } else {
-        tl2.addText("Fonts are being rendered by stb_truetype.", .{});
+        tl2.addText("Fonts are being rendered by stb_truetype.\n", .{});
     }
-    tl2.addText("\n\n", .{});
+    tl2.addText("\n", .{});
     tl2.format("Scale: {d:0.2} (try pinch-zoom)", .{dvui.windowNaturalScale()}, .{});
     tl2.deinit();
 
--- a/readme-accessibility.md
+++ b/readme-accessibility.md
@@ -196,6 +196,18 @@ DVUI offers the following labeling options:
 
 Alternatively, you can also use the `AccessKit.nodeSetLabel` and `AccessKit.nodeSetLabeledBy` functions.
 
+#### Text
+Text created through the TextLayoutWidget will make the text accessibile to the screen reader through a text_run role. This is true for any widget that uses TextLayoutWidget to create text, e.g. TextEntryWidget.
+
+Each call to add_text creates a new text_run for each line or partial line of text added. 
+
+To make your text compatible with text_runs, you should:
+1) If there are no font/style changes, use addText() to add one or more full lines, including any trailing newline.
+2) If there are fonts/styles changes, use a different addText() for each font/style change.
+3) Do not call addText() just to add a newline, unless it is for a blank line, only containing a newline.
+
+Ignoring these rules will not break anything, but will make it harder for accessiblilty users to correctly navigate such text in their screen reader.
+
 ## Current State of Accessibility in DVUI
 
 * A role of `null` means the widget will not be added to the accessibility tree, unless the user passes a `role` for that widget via Options.
@@ -226,8 +238,8 @@ Alternatively, you can also use the `AccessKit.nodeSetLabel` and `AccessKit.node
 | ScrollBarWidget | scroll_bar | Y *| N* | Y | Due to bug in AccessKit, is set to read-only and cannot interact. Actual values displayed may not be accurate | 
 | SuggestionsWidget | suggestion, list_item | Y | Y | N | 
 | TabsWidget | .tab_panel, .tab | Y | Y | N | Sets active tabs and allows tab selection |
-| TextEntryWidget | text_input, multiline_text_input | Y* | N* | Y | Text entry should work for "reasonable" amounts of text. SetValue will replace all text. Needs to implement the .text_run role to properly allow users to fully perform text entry and editing. |
-| TextLayoutWidget | label | Y* | N/A | Y | Currently displays only visible text (TBC). Works OK for "reasonable" amounts of text. Does not support sending of formatting information | 
+| TextEntryWidget | text_input, multiline_text_input | Y* | N* | Y | TextRun support implemented for text. Allows selection and navigation. Scrolling for off-screen text requires improvement. Cached layouts needs more investigation. |
+| TextLayoutWidget | label | Y* | N/A | Y | Most features supported. Needs to implement the text_replace action. | 
 | TreeWidget | tree, tree_item | Y* | Y | Y | Adds tree and nodes. Implement expand / collapse when supported by AccessKit | 
 | windowHeader | label, button | Y | Y | N | |
 | dialogs | window | Y | Y | Y | All dialogs are displayed as windows, rather than dialogs and modal state is not displayed in accessibility insights. AccessKit currently has limted support for dialogs, so leaving as window until the situation changes and can revisit. |
@@ -241,6 +253,6 @@ Alternatively, you can also use the `AccessKit.nodeSetLabel` and `AccessKit.node
 | slider | slider | Y | Y | N | Fully supported |
 | progress | progress_indicator | Y | N/A  | N | Fully supported |
 | checkbox | check_box | Y | Y | N | Fully supported |
-| radio | radio_button | Y* | Y | N | Best practice: Create a radio group with a surrounding box|
+| radio | radio_button | Y* | Y | N | Best practice: Create a radio group with a surrounding radio group widget|
 | textEntryNumber | number_input | Y | Y | N | Fully supported. Supports min, max and valid/invalid. |
 
--- a/src/AccessKit.zig
+++ b/src/AccessKit.zig
@@ -94,7 +94,7 @@ pub fn initialize(self: *AccessKit) void {
 
 fn windowsHWND(window: *dvui.Window) c.HWND {
     switch (dvui.backend.kind) {
-        .sdl3 => {
+        .sdl3, .sdl3gpu => {
             const SDLBackend = dvui.backend;
             const properties: SDLBackend.c.SDL_PropertiesID = SDLBackend.c.SDL_GetWindowProperties(window.backend.impl.window);
             const hwnd = SDLBackend.c.SDL_GetPointerProperty(
--- a/src/AccessKit.zig
+++ b/src/AccessKit.zig
@@ -23,6 +23,8 @@ nodes: std.AutoArrayHashMapUnmanaged(dvui.Id, *Node) = .empty,
 // Note: Any access to `action_requests` must be protected by `mutex`.
 action_requests: std.ArrayList(ActionRequest) = .empty,
 
+// null means don't add text runs
+text_run_parent: ?dvui.Id = null,
 text_runs: std.ArrayList(TextRunOptions) = .empty,
 
 // The last seen node with `role = .label`
@@ -162,6 +164,11 @@ pub fn nodeCreateReal(self: *AccessKit, wd: *dvui.WidgetData, role: Role) ?*Node
     // text runs are created even if not visible.
     if (!wd.visible() and wd.id != dvui.focusedWidgetId() and role != .text_run) return null;
 
+    if (role == .text_run and self.text_run_parent == null) {
+        log.debug("skipping ak text run for widget {x}, text_run_parent null", .{wd.id});
+        return null;
+    }
+
     self.mutex.lock();
     defer self.mutex.unlock();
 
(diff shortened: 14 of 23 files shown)


ethindp:
This needs to be implemented for all text inputs, read only or not. It is nonsensical for labels and such -- these will appear in object navigation as labels, so no need to help the screen reader with those.

For IDs, could do a PCG RNG maybe?

For words off-screen, my best advice would be to pass all the contents in the edit field to the screen reader, onscreen or not, as text runs. I think. I'm not entirely sure how actual GUIs do this, so I could be very wrong and this could turn out to be a very bad idea.

ethindp:
So just tried it and it doesn't seem to be working yet? At least so far? The edit fields read as they have originally but I still can't navigate them. Not entirely sure what's going on? I try mainly with the SDL3 test app, since the others (mainly the DX11 one) are really laggy for some reason. That could probably be because it's constantly printing things like:

```
>dx11-app
debug(dvui): adding embedded font Vera Sans
debug(dvui): adding embedded font Vera Sans Bold
debug(dvui): adding embedded font Vera Sans Italic
debug(dvui): adding embedded font Vera Sans Bold Italic
debug(dvui): adding embedded font Vera Sans Mono
debug(dvui): adding embedded font Vera Sans Mono Bold
debug(dvui): adding embedded font Vera Sans Mono Italic
debug(dvui): adding embedded font Vera Sans Mono Bold Italic
debug(dvui): adding embedded font Vera Serif
debug(dvui): adding embedded font Vera Serif Bold
info(dvui): window logical Size.Natural{ 945.3333 453.33334 } pixels Size.Physical{ 1418 680 } natural scale 1.5 initial content scale 1 snap_to_pixels true accesskit true

info(AccessKit): initialized successfully
debug(dvui): adding embedded font Pixelify Sans
debug(dvui): adding embedded font Pixelify Sans Bold
debug(dvui): adding embedded font Aleo
debug(dvui): adding embedded font Aleo Bold
debug(dvui): adding embedded font Aleo Italic
debug(dvui): adding embedded font Aleo Bold Italic
debug(dvui): adding embedded font OpenDyslexic
debug(dvui): adding embedded font OpenDyslexic Bold
debug(dvui): adding embedded font OpenDyslexic Italic
debug(dvui): adding embedded font OpenDyslexic Bold Italic
Segmentation fault at address 0xffffffffffffffff
dvui\src\backends\dx11.zig:1353:94: 0x7ff670d1b9a3 in wndProc (dx11-app_zcu.obj)
                const result = dvui.AccessKit.c.accesskit_windows_adapter_handle_wm_getobject(
                                                                                             ^
???:?:?: 0x7ff8fe01ef5b in ??? (user32.dll)
???:?:?: 0x7ff8fe01e8cb in ??? (user32.dll)
???:?:?: 0x7ff8fe0310e2 in ??? (user32.dll)
???:?:?: 0x7ff8fe3f1463 in ??? (ntdll.dll)
???:?:?: 0x7ff8fb9e1063 in ??? (win32u.dll)
???:?:?: 0x7ff8fe01a332 in ??? (user32.dll)
???:?:?: 0x7ff8fe01a292 in ??? (user32.dll)
???:?:?: 0x7ff8fd7c8c23 in ??? (combase.dll)
???:?:?: 0x7ff8fd7c8ace in ??? (combase.dll)
???:?:?: 0x7ff8fd7c8a75 in ??? (combase.dll)
???:?:?: 0x7ff8fd791908 in ??? (combase.dll)
???:?:?: 0x7ff8fd791846 in ??? (combase.dll)
???:?:?: 0x7ff8fd7914a5 in ??? (combase.dll)
???:?:?: 0x7ff8fd764f0c in ??? (combase.dll)
???:?:?: 0x7ff8fd764383 in ??? (combase.dll)
???:?:?: 0x7ff8fc506975 in ??? (IMM32.DLL)
???:?:?: 0x7ff8fdb2be6d in ??? (MSCTF.dll)
???:?:?: 0x7ff8fe03676f in ??? (user32.dll)
???:?:?: 0x7ff8fe02e858 in ??? (user32.dll)
???:?:?: 0x7ff8fe02e74d in ??? (user32.dll)
???:?:?: 0x7ff8fe3f1463 in ??? (ntdll.dll)
???:?:?: 0x7ff8fb9e2383 in ??? (win32u.dll)
???:?:?: 0x7ff8c901e2d5 in ??? (nvdaHelperRemote.dll)
dvui\src\backends\dx11.zig:358:33: 0x7ff670d182be in deinit (dx11-app_zcu.obj)
    if (0 == win32.DestroyWindow(hwndFromContext(self))) win32.panicWin32("DestroyWindow", win32.GetLastError());
                                ^
dvui\src\backends\dx11.zig:1631:19: 0x7ff670d169d6 in main (dx11-app_zcu.obj)
    defer b.deinit();
                  ^
...\zvm\data\0.15.2\lib\std\start.zig:602:28: 0x7ff670d18cad in main (dx11-app_zcu.obj)
    return callMainWithArgs(@as(usize, @intCast(c_argc)), @as([*][*:0]u8, @ptrCast(c_argv)), envp);
                           ^
...\zvm\data\0.15.2\lib\libc\mingw\crt\crtexe.c:259:0: 0x7ff67131e68b in __tmainCRTStartup (crt2.obj)
    mainret = _tmain (argc, argv, envp);

...\zvm\data\0.15.2\lib\libc\mingw\crt\crtexe.c:148:0: 0x7ff67131e4ab in WinMainCRTStartup (crt2.obj)
  ret = __tmainCRTStartup ();

???:?:?: 0x7ff8fe257373 in ??? (KERNEL32.DLL)
???:?:?: 0x7ff8fe39cc90 in ??? (ntdll.dll)
```

phatchman:
I'm only testing SDL3 as well.

The main test I've been doing is to go to the Text Entry demo and using the Windows narrator to navigate the entry fields there.  When a text entry has focus, I can use caps-ctrl-right/left to navigate forward and backwards a single word and when I use the read screen option (caps-shift-r), it will highlight the individual words as it reads them. So, I can see at least some parts are working.

Could you give me some steps to recreate what is not working for you? I have nvda installed as well.

phatchman:
@ethindp I've also tried this in NVDA and using the numpad keys seems to work to navigate the text lines as I expect in the multi-line text.

I think that when I implement the events, the cursor will update to match the current word / line selection to match the reader.

ethindp:
Hi there,

I think the events is the problem. Yes, I can navigate the edit fields with object navigation (on laptop, caps-left/right and the like), but I can't navigate by word with control-left/right or by character with left/right, or read lines in a single-line or multi-line edit field. And I'm using NVDA. I don't know dvui in too much detail so I don't know if the missing events is the issue or not or if something else is missing though.

david-vanderson:
This is awesome!  It'll take me a few days to get to it, sorry.

For id_extra, another option is to use the total count of bytes we've gotten so far.  I haven't thought very hard about how stable those ids have to be, but my guess is that if the text changes at all it's probably okay for all the ids to change?

ethindp:
I don't really have an answer for that, sadly. So I guess we'll just have to find out?

phatchman:
> This is awesome! It'll take me a few days to get to it, sorry.

All good. I'm expecting more changes, so its prob worth waiting.
> 
> For id_extra, another option is to use the total count of bytes we've gotten so far. I haven't thought very hard about how stable those ids have to be, but my guess is that if the text changes at all it's probably okay for all the ids to change?

Yeah, I think that works.

phatchman:
> I think the events is the problem. 

Ok. I'll add event support next and let u know.

phatchman:
This latest commit adds reporting of text selection. I would not bother testing this yet.

phatchman:
@david-vanderson Any thoughts on implementing the text selection event? After adding some more support, I'm able to convert the selection action from accesskit into the correct DVUI TextEntry widget id with an absolute start and end character position.

In theory I could turn that into start/end mouse clicks by looking at the bounds of the text run and the width of each character to generate a screen position for the start selection click and the end selection click. But the issue is that clicking doesn't just move the selection, it does word or line select on the second and 3rd clicks. And it would interact with users holding down the shift key.

So, I think I either need to:
1)  Introduce a text_selection event (widget id, start_char, end_char)
2) Treat this as a special-case and have textEntryWidget query AccessKit directly for the selection information each frame. e.g. AccessKit.textSelectionGet(node_id) which returns the start/end selection or null if no selection received this frame.

Any preference or other ideas?

P.S. Happy new year! Hope 2026 is a good one for you and DVUI.

phatchman:
@ethindp This adds support for text selection. The next step is to support text replacement.

I can't recreate the issue you were having. With NVDA if I use the next/prev word (caps-ctrl-left/right or numpad 4/6) it reads the words as expected. Same with navigating to next / previous lines.

The only issue I've found is that ctrl->left/right (i.e. without holding the NVDA key) sometimes reads the word and sometimes doesn't. I'm not sure what is causing that. If you can check the SDL3 demo and see if the issue is still present?

Edit: just to check you are running the correct version, you should see some debug text like the below:
```
Setting selection to: .{ .has_value = true, .value = .{ .anchor = .{ .node = 15614982536272827525, .character_index = 18 }, .focus = .{ .node = 15614982536272827525, .character_index = 18 } } }
Setting selection to: .{ .has_value = true, .value = .{ .anchor = .{ .node = 3699348162616609027, .character_index = 0 }, .focus = .{ .node = 3699348162616609027, .character_index = 0 } } }
```

@david-vanderson For this commit, I special-cased the processing of the text selection in textEntryWidget's processEvents(). It's not pretty, but works. I'll convert this to an event and see what it looks like. Probably I can extend the existing text event.

I need to remove single quote from TextLayoutWidget.word_break because it makes words with apostrophes into two words and the reader won't read them properly. 

~~Note to self: Need to validate bounds on the selection action indexes. They values from the reader are untrusted and could be past the end of text.~~ TextLayoutWidget already caters for this. Can set end to maxInt.

DataTriny:
Hi there,

There seem to be an issue with caret and text selection once they reach the end of the content. For instance using the right arrow key to get to the end of a single line input, you're supposed to land on a blank spot past the last character, but here it does not happen  and I can see that the caret is placed at the start. Using an AT action to move the caret to the end also places the caret at the start. Text changes seem to also be affected by this as text input content does not get updated if you remove text at the end.

Empty text inputs don't seem to report a caret position either so they appear as read only, this might be linked to the issue above.

There might be an issue with placeholder text too as it behave as if it was regular text that you just can't select.

This is already a very good start. Text input is one of the most difficult kinds of widgets to represent with AccessKit.

Happy new year everyone! Oh and the latest version of accesskit-c allows you to pass string slices everywhere which will save you the trouble of copying zig strings.

phatchman:
Thanks for testing. I'll look into these.

> Empty text inputs don't seem to report a caret position either so they appear as read only, this might be linked to the issue above.

This is an issue I've not tracked down yet. I'm currently skipping creating empty text runs as they were triggering an assert in text.rs when unwapping the text run's value. I.e. calling `accesskit_node_set_value(node, "");` seems to set a null value, rather than an empty value.

Should setting a text runs value to "" work? 

> Happy new year everyone! Oh and the latest version of accesskit-c allows you to pass string slices everywhere which will save you the trouble of copying zig strings.

Great. That will simplify things a lot.

ethindp:
> @ethindp This adds support for text selection. The next step is to support text replacement.
> 
> I can't recreate the issue you were having. With NVDA if I use the next/prev word (caps-ctrl-left/right or numpad 4/6) it reads the words as expected. Same with navigating to next / previous lines.
> 
> The only issue I've found is that ctrl->left/right (i.e. without holding the NVDA key) sometimes reads the word and sometimes doesn't. I'm not sure what is causing that. If you can check the SDL3 demo and see if the issue is still present?
> 
> Edit: just to check you are running the correct version, you should see some debug text like the below:
> 
> ```
> Setting selection to: .{ .has_value = true, .value = .{ .anchor = .{ .node = 15614982536272827525, .character_index = 18 }, .focus = .{ .node = 15614982536272827525, .character_index = 18 } } }
> Setting selection to: .{ .has_value = true, .value = .{ .anchor = .{ .node = 3699348162616609027, .character_index = 0 }, .focus = .{ .node = 3699348162616609027, .character_index = 0 } } }
> ```

So, the bug is with the arrow keys. Control-caps-left/right and the NumPad use object/screen review. Left'right/up/down/control-left/right and such are the keys where it broke last time I tested it, but will try with the latest.

david-vanderson:
> @david-vanderson For this commit, I special-cased the processing of the text selection in textEntryWidget's processEvents(). It's not pretty, but works. I'll convert this to an event and see what it looks like. Probably I can extend the existing text event.
> 
> I need to remove single quote from TextLayoutWidget.word_break because it makes words with apostrophes into two words and the reader won't read them properly.

Sorry I haven't gotten to this yet, but sounds good, both adding an event (unsure if we need a new class of events or not) and removing single quotes from word_break.  Thank you!

david-vanderson:
> (unsure if we need a new class of events or not)

Probably change the .text event to have an action field like the rest of the event kinds - just like you said.  This kind of selection event will also be useful in the future for menu items like "Edit > Select All".

Happy new year everyone - looking forward to a great 2026!

phatchman:
> > Empty text inputs don't seem to report a caret position either so they appear as read only, this might be linked to the issue above.
> 
> This is an issue I've not tracked down yet. I'm currently skipping creating empty text runs as they were triggering an assert in text.rs when unwapping the text run's value. I.e. calling `accesskit_node_set_value(node, "");` seems to set a null value, rather than an empty value.
> 
Actually, I take that back. I found a few cases I'm creating text runs and are not being populated with anything. e.g. where text is not being rendered. I'll look into it further, but pretty sure this is a me issue, not an accesskit issue.

phatchman:
> > (unsure if we need a new class of events or not)
> 
> Probably change the .text event to have an action field like the rest of the event kinds - just like you said. This kind of selection event will also be useful in the future for menu items like "Edit > Select All".
> 
For first attempt, I've replaced the "selected" in the Text event with a start/end selection and added convenience `all` and `none` values. This will also work with the replace_text event as it takes the same values, just with "replace" = true, and selection start/end to the range ot be replaced.

What are your thoughts on changing the TextLayoutWidget.textAdd to take a TextLayoutWidget.Selection? And have `all and `none` helpers? We could then potentially pull `Selection` out into a common struct and share it between TextEntryWidget and the Text event? Then anything that can be done with addText can also be done via the event?

The other option is to split the setting of text and the selection of text into separate events. So, we'd lose 'select' from the Text event and have a union of `select` and `set`. In that case I'd be tempted to remove 'selected' from textAdd as well? All it does is call selectAll() anyway.

The only thing I saw using selection for the text event was the sdl backend. So, changing the event will have little impact either way.

david-vanderson:
> What are your thoughts on changing the TextLayoutWidget.textAdd to take a TextLayoutWidget.Selection? And have `all and `none`helpers? We could then potentially pull`Selection` out into a common struct and share it between TextEntryWidget and the Text event? Then anything that can be done with addText can also be done via the event?
> 
> The other option is to split the setting of text and the selection of text into separate events. So, we'd lose 'select' from the Text event and have a union of `select` and `set`. In that case I'd be tempted to remove 'selected' from textAdd as well? All it does is call selectAll() anyway.

I'm getting up to speed here, still a bit confused.  I think when you say doing stuff with "addText" you must mean TextEntryWidget.textTyped()?

Currently in my mind there are distinct events we need to support:
* here's some new selected text (insert text at cursor and change selection to cover new text) - this is important for IME input
* no change to text, but change the selection - this is "Select All" or "Select None", but should also include "Select this range"

In theory we could insert selected text by sending two events, a "new text" event, and then an event that selected exactly that text, but then the code generating the events would need insight into the byte positions the text is being inserted into, which seems hard.

One thing we do have is a "replace" flag, and it seems like that could probably be replaced by two events, a "select all", and then a "new text" event.

So my guess is we want the separate events option you laid out.  I don't understand how the first option would work.

Is this making any sense?

phatchman:
After looking further agree. The new text event will get handled by TextEntryWidget while selection would be handled by TextLayoutWidget.

Edit: The latest commit implements the event, but introduces an issue that can cause the application to hang. Don't bother testing this version.

phatchman:
I believe this latest commit should resolve the issues mentioned above and cursor/selection position is now being reported correctly. Everything seems to work fine with NVDA at least, including the arrow key navigation, but would appreciate others testing as well.

There is some race condition I need to work out that can result in the intentional `@panic("TODO: Missing node");`. I've not worked out any good way to reproduce it yet, so still need to track it down. 

@david-vanderson I still want to clean up a few things before it is ready to review. But should be close. 

Edit: Found an issue, application will hang if you select a blank line of read-only text. So don't do that!

phatchman:
@david-vanderson what triggers the `if (!cw.render_target.rendering)` in render.zig?

I'd be able to simplify things quite a bit if I didn't have to handle that case. The issue is that I have to create the text_run while the TextEntryWidget exists, but the details of the text_run don't get populated until the rendering is done. If nothing actually ends up being rendered, I have to go and remove those text_runs  nodes.

Also if I need to create text runs for off-screen text, do you think I need to do anything else outside of skipping the return on `if (dvui.clipGet().intersect(opts.rs.r).empty())`. It seems to work as expected. 

Edit: Ahh, so it doesn't work for the cache_layout case, but that should be OK as long as I have a few lines before and after to be able to scroll to. 

@DataTrinty Do you have any insight in how to handle off-screen text? From what I can see in other applications the text should scroll to keep the screen in sync with the reader. But I'm not sure how this works? I don't receive anything while the reader is running telling what position it is up to. 

Edit: Ahh. I think I managed to work it out. I must fixed something because I'm getting scroll_into_view actions now! 

As I understand, the "hint" value for that action will be "top edge" to scroll to make the "previous" text run visible and "bottom edge" to scroll to make the "next" text run visible. Or any of the other values to similarly scroll. (where "next" and "previous" are given by the target node id for the action).

phatchman:
@david-vanderson Hmm. We don't have scroll events anymore. So, just want to run past you the approach I plan to take to scroll text into view.

When I get the scroll_into_view, I have the following:

1) The widget id of the text entry widget to be scrolled. Let's assume I can get the widget id of the scroll area if needed,
2) The bounds of the part of the text entry to be made visible (in screen coordinates)
3) The text position (in bytes / characters) of the start and end of the section to be made visible.
3) How that should be aligned (top-left, bottom-right, top-edge, bottom-edge, left-edge, right-edge)

I can add a ScrollIntoView event that has an alignment and a rect. We could just limit it to the text event (i.e. make it a Text Event action), but I expect we are going to need this for  the grid scrolling.
But maybe there is a better way similar to the "scroll to focused" change? I can't really think of anything.

david-vanderson:
> @david-vanderson what triggers the `if (!cw.render_target.rendering)` in render.zig?

Usually that's when you are in a floating subwindow, and the rendering stuff is being deferred until `Window.end()`.  I'm not sure what to do here, maybe we need to attach more information to the render command?

> Also if I need to create text runs for off-screen text, do you think I need to do anything else outside of skipping the return on `if (dvui.clipGet().intersect(opts.rs.r).empty())`. It seems to work as expected.

Phew!

> 
> Edit: Ahh, so it doesn't work for the cache_layout case, but that should be OK as long as I have a few lines before and after to be able to scroll to.

That is also the strategy I'm using for when the user does text motion (like holding down the up key).  When cache_layout is true we render a bit more on the top and bottom than what is visible, and we can enlarge that if needed.

> @david-vanderson Hmm. We don't have scroll events anymore. So, just want to run past you the approach I plan to take to scroll text into view.

Sorry - that got replaced with `dvui.scrollTo()` which you should be able to call any time during the widget lifetime, and it'll find the nearest scroll container and do the right thing (hopefully).  You can see it being used in `WidgetData.register()` to scroll a newly focused widget into view.

For this case usually the right thing is for textLayout to record that information when it gets the event, and then when text is being laid out react once you find the matching text.  Does that make sense?

DataTriny:
Gave this PR another try. Happy to report that placing the caret at the end no longer triggers a bug.

Empty text inputs still don't get the caret position though, even if you type text into them. You have to move the focus to a different widget and then focus the text input again for it to get the caret. This also causes an issue when removing the very last character. Because the input becomes empty, it seem that no tree update is pushed so my screen reader still display that input with the last character.

Placing the caret on a blank line in a multi-line input seem to not report a blank line in the accessibility tree but the content of the first non-blank line for some reason. Inserting a blank line at the start also displays the first non blank line after it.

In a multi-line input, when a line ends with a hard line break (e.g. LF character), the last position for the caret should be on that LF character. Here it seem to land on the first character of the next line.

There must be another bug in how lines are broken down because when I hover the mouse over a multi-line input, my screen reader always read the whole content. It should just read the paragraph that the mouse is pointing at instead.

Using my Braille display, I should be able to select a word by double-clicking a routing key: this should send a text selection action request with character bounds for that word. This does not work yet, unlike double clicking with the mouse.

> The only issue I've found is that ctrl->left/right (i.e. without holding the NVDA key) sometimes reads the word and sometimes doesn't.

This may be linked to the fact that word navigation works differently depending on the direction in DVUI. Jumping to the next word places the caret at the end of the current word while jumping to the previous word places the caret at the start of the previous word. Typically the caret is always placed at the start of the word, no matter the direction. I know this is a small detail, but that's the sort of thing that make or break UX.

phatchman:
> that got replaced with `dvui.scrollTo()` which you should be able to call any time during the widget lifetime, and it'll find the nearest scroll container and do the right thing (hopefully). You can see it being used in `WidgetData.register()` to scroll a newly focused widget into view.

As I understand scollTo() requires a scroll container to be instantiated. But these actions are processed at end of frame when all widgets are deinit-ed. I still need some way to pass the action to the widget to process next frame, which either needs to be a scoll event or similar?

phatchman:
> In a multi-line input, when a line ends with a hard line break (e.g. LF character), the last position for the caret should be on that LF character. Here it seem to land on the first character of the next line.

OK. In the DVUI rendering all LF are stripped. That might be causing the issue.

I'm still unclear on how to handle empty text entry / text runs.

If I create a text run without a value, AcessKit panics when trying to unwrap the value.

If I set value to empty string, and set all of the character length, word positions etc to empty arrays, I get infinite loops for multi-line text inputs when AccessKit is trying to find the start of the line on those "blank" lines. 

Perhaps setting them to newline for blank lines will fix both issues.

david-vanderson:
> Gave this PR another try. Happy to report that placing the caret at the end no longer triggers a bug.

Thank you so much for the testing!  Once we figure out the event stuff I will go through this and see what can be done.

david-vanderson:
> As I understand scollTo() requires a scroll container to be instantiated. But these actions are processed at end of frame when all widgets are deinit-ed. I still need some way to pass the action to the widget to process next frame, which either needs to be a scoll event or similar?

Oh yes, sorry.  You are right - we'll have to have an event like "scroll_to_id" that can pass the info to the next frame.  Don't worry too much about the naming or where it goes, we can figure that out later.

david-vanderson:
> > In a multi-line input, when a line ends with a hard line break (e.g. LF character), the last position for the caret should be on that LF character. Here it seem to land on the first character of the next line.
> 
> OK. In the DVUI rendering all LF are stripped. That might be causing the issue.

That could be the problem.  Another possibility is that we have to set `Selection.affinity` differently.  If the caret is logically going to the "end of the line" then Selection.affinity must be set to ".before".  @phatchman - I can deal with this once you get to place where I won't be stepping on your toes.

phatchman:
Adding the newlines fixed up the paragraph issue.
The scrolling is getting a little complicated. So, I'm going to make that a separate PR.

I *think* there might be a simple thing we can do to just scroll as the cursor moves to the next word (i.e. the same behaviour as ctrl-right/left would do, moving word by word). We just won't get the nice effect of scrolling as the reader reads the entire text since that doesn't move the cursor.

david-vanderson:
> Adding the newlines fixed up the paragraph issue.
> The scrolling is getting a little complicated. So, I'm going to make that a separate PR.

Good plan.

ethindp:
I was wondering why trying this a few days ago kept carshing it in the text entry demos. Am very hopeful that this gets resolved! I'm not sure what to do about blank fields; I don't really know what Godot did to solve this, sadly.

DataTriny:
@phatchman FYI single-line text inputs or the last line of a multi-line text input does not have to end with a newline character. Empty text inputs can be represented with a text run node on which you should set value to an empty string and at least character_lengths to an empty slice. I just built a quick Rust app to double check that the current version of AccessKit doesn't contain a bug here. If this is not the case, please share any crash you may hit.

Still, congratulations on getting it so far. DVUI already has decent accessibility support, which is quite amazing given the project size.

phatchman:
> 
> Still, congratulations on getting it so far. DVUI already has decent accessibility support, which is quite amazing given the project size.

Thanks. And appreciate you taking the time to assist us! Each time I thought I understood text runs, I found another thing I didn't know. But I think I finally have a handle on them now.

What I was seeing previously, when I was creating empty text_runs, was that backward_to_paragraph_start called backward_to_line_start, which was hanging. 

I think that will now be fixed by including the `\n` for the paragraph ends on the multi-line text runs and empty line's will be set as value "\n". (assuming they are newline terminated).

Because the blank lines in the multi-line text runs were causing hangs, I have not been creating text runs for any blank lines, including single line text entries. So, I'm expecting the single line text entries work again once I start creating the empty text runs again. Edit: Yeah, all seems working now!!! Thanks for the pointers. I'll clean up a few other things you mentioned and we should be good.

Maybe there is an assert you can put in to detect when there is no start/end paragraph and prevent the infinite loop?

Example stack trace of the hang below:
```
- frame #0: 0x00007ff69d55a580 sdl3-standalone.execore::iter::adapters::map::impl$2::next<accesskit_consumer::node::Node,accesskit_consumer::iterators::PrecedingSiblings,accesskit_consumer::node::impl$0::preceding_siblings::closure_env$0>(self=0x000000095edff5c0) at 1159e78c4747b02ef996e55082b704c09b970588\library\core\src\iter\adapters\map.rs:108frame #1: 0x00007ff69d55a6c0 
- sdl3-standalone.exeaccesskit_consumer::iterators::previous_filtered_sibling<accesskit_consumer::text::impl$11::preceding_text_runs::closure_env$0>(filter=<unavailable>) at iterators.rs:234 frame #2: 0x00007ff69d55b2ac 
- sdl3-standalone.exeaccesskit_consumer::iterators::impl$15::next<accesskit_consumer::text::impl$11::preceding_text_runs::closure_env$0>(self=<unavailable>) at iterators.rs:367frame #3: 0x00007ff69d55afc8 
- sdl3-standalone.exeaccesskit_consumer::text::InnerPosition::biased_to_end(self=0x000000095edff778, root_node=<unavailable>) at text.rs:103 frame #4: 0x00007ff69d55bf64 
- sdl3-standalone.exeaccesskit_consumer::text::Position::is_paragraph_start(self=0x000000095edff760) at text.rs:223frame #5: 0x00007ff69d55c410 
- sdl3-standalone.exeaccesskit_consumer::text::Position::backward_to_paragraph_start(self=<unavailable>) at text.rs:450 frame #6: 0x00007ff69d55e530 
- sdl3-standalone.exewindows::Win32::UI::Accessibility::impl$184::new::ExpandToEnclosingUnit<accesskit_windows::text::PlatformRange_Impl,-1> [inlined] accesskit_windows::text::impl$2::ExpandToEnclosingUnit::closure$1(range=<unavailable>) at text.rs:395    frame #7: 0x00007ff69d55e504 
- sdl3-standalone.exewindows::Win32::UI::Accessibility::impl$184::new::ExpandToEnclosingUnit<accesskit_windows::text::PlatformRange_Impl,-1> [inlined] accesskit_windows::text::impl$0::write::closure$0(tree_state=<unavailable>) at text.rs:286 frame #8: 0x00007ff69d55e49c 
- sdl3-standalone.exewindows::Win32::UI::Accessibility::impl$184::new::ExpandToEnclosingUnit<accesskit_windows::text::PlatformRange_Impl,-1> [inlined] accesskit_windows::text::impl$0::with_tree_state::closure$0(state=<unavailable>, =<unavailable>)frame #9: 0x00007ff69d55e49c 
- sdl3-standalone.exewindows::Win32::UI::Accessibility::impl$184::new::ExpandToEnclosingUnit<accesskit_windows::text::PlatformRange_Impl,-1> [inlined] accesskit_windows::text::PlatformRange::with_tree_state_and_context(self=<unavailable>) frame #10: 0x00007ff69d55e46c 
- sdl3-standalone.exewindows::Win32::UI::Accessibility::impl$184::new::ExpandToEnclosingUnit<accesskit_windows::text::PlatformRange_Impl,-1> [inlined] accesskit_windows::text::PlatformRange::with_tree_state(self=<unavailable>) at text.rs:239frame #11: 0x00007ff69d55e46c 
- sdl3-standalone.exewindows::Win32::UI::Accessibility::impl$184::new::ExpandToEnclosingUnit<accesskit_windows::text::PlatformRange_Impl,-1> [inlined] accesskit_windows::text::PlatformRange::write(self=<unavailable>, f=<unavailable>) at text.rs:283
frame #12: 0x00007ff69d55e46c 
sdl3-standalone.exewindows::Win32::UI::Accessibility::impl$184::new::ExpandToEnclosingUnit<accesskit_windows::text::PlatformRange_Impl,-1> [inlined] accesskit_windows::text::impl$2::ExpandToEnclosingUnit(self=<unavailable>, unit=<unavailable>) at text.rs:386frame #13: 0x00007ff69d55e39c 
sdl3-standalone.exewindows::Win32::UI::Accessibility::impl$184::new::ExpandToEnclosingUnit<accesskit_windows::text::PlatformRange_Impl,-1>(this=0x0000011f0da166a8, unit=<unavailable>) at mod.rs:6289 frame #14: 0x00007ff883af8640 UIAutomationCore.dll`UiaRaiseTextEditTextChangedEvent + 6816
```

phatchman:
I need to do more testing on this, but I think we are mostly there. A few issues I noticed:
1) When moving the reader cursor, the whole of the next word isn't displayed when horizontal scrolling. That's because the cursor is moved to the start of the word (rather than the end as DVUI expects). We probably need to enhance the text entry widget to scroll to display the whole word if the cursor is at the start of one.
2) No width is calculated for spaces, so when moving character by character, hitting a space makes the whole text_run selected instead of just the space. This should be easy to fix.
3) When I have 2 or more newlines at the bottom of a multiline, navigating backwards doesn't move the cursor. Could be an issue with how I'm setting the last blank line?
4) When scrolling down or upwards, the bounds of the text run is set before the scrolling occurs. Until the next navigation, the highlighting is in the wrong spot. I think we're just going to live with that until we get the full scrolling solution in place.

I'll fix those last few issues before I ask you to test for (hopefully) one last time.  

@david-vanderson the code should be good to take a look at now. 

The one thing I'd point out is the reason for adding of the ak_text_parent to the text layout widget's options and the corresponding ak_node_parent in dvui.Options.
This is because we need to parent the text_runs to the TextEntryWidget, but the TextLayoutWidget is the one creating them. I didn't want to rely on the widget parent or the accessibility node parent to determine this as TextLayoutWidget isn't the direct child of the TextEntryWidget (e.g. it's in a ScrollContainerWidget).

Also, I don't need opts.newline_terminated in TextRunOptions anymore. I'll remove that as TextRunOptions.Text now contains the newline.

david-vanderson:
> @david-vanderson the code should be good to take a look at now.
> 
> The one thing I'd point out is the reason for adding of the ak_text_parent to the text layout widget's options and the corresponding ak_node_parent in dvui.Options. This is because we need to parent the text_runs to the TextEntryWidget, but the TextLayoutWidget is the one creating them. I didn't want to rely on the widget parent or the accessibility node parent to determine this as TextLayoutWidget isn't the direct child of the TextEntryWidget (e.g. it's in a ScrollContainerWidget).

Awesome!  Will look at it soon, and thank you for this explanation.

phatchman:
OK. I *think* this should be fairly good.

Outside of the text scrolling issues mentioned previously, there are two things I noticed:
1) Something still not 100% with the last line of a multi-line text if it only contains newlines. (i.e. cursor is 1 character past the last newline and on the first position of the next line). I've checked the text_run info for these cases and it looks OK. So not sure. 
2) When you navigate the static text on the main window word by word (or character by character), the newline at the end of each line is in a separate textrun.

That's because the newlines get added via separate textAdd()s:
```
   tl2.addText("Framerate is capped by vsync.", .{});
   tl2.addText("\n\n", .{});
```
My brain is too fried right now to work out the implications of merging 2 different addText's into a single text_run. Unless someone thinks this would cause major issues, I think we'll live with that until I can come up with a better solution.

You should be able to move the text selection focus / anchor using a screen reader and then select text with it. (this works with  Accessibility Insights at least).

Let me know if you find any other issues.

david-vanderson:
> That's because the newlines get added via separate textAdd()s:

This is totally reasonable, separate textAdd() calls also cause problems for ligatures and text shaping in general.  I will think about how we can communicate that people should only do this if they need it (like changing font/color properties).

ethindp:
Man, I love how this is coming along! Happy to do more testing and to play with it some more. Super excited for when all of this is done! It'll be good to have another reliable UI framework to use instead of needing to recommend the usual ones.

phatchman:
@david-vanderson Sounds good. That lines up with how text_runs are supposed to work. i.e. Any font/style changes should start a new text_run.

I'll update the examples where needed and add them to the PR.

@ethindp Great to hear. Thanks so much for helping out.

david-vanderson:
I've looked over the code, and it looks pretty good to me!  Maybe we can change the Options.ak_node_parent for a more specialized "swap the ak parent node for this one for this function" kind of thing - I will figure that out later.

I'm starting to do testing now.

phatchman:
I did create a nodeReparent() previously. I can't remember now what caused me to abandon it. 

I'll revisit and see if that is viable as it's much nicer. We still need to tell the TextEntryWidget who to parent the text_runs to, but at least that would move it out of Options.

Edit: One thing I'm wondering is if we need to disable the text_runs for some text layouts. e.g. should a tooltip cotnain a text_run? Not sure, I'll investigate further.

```
 if (tt.shown()) {
        var tl2 = dvui.textLayout(@src(), .{}, .{ .background = false });
        tl2.format(fmt, fmt_args, .{});
        tl2.deinit();
        if (tt.data().accesskit_node()) |ak_node| {
            var str_builder: std.Io.Writer.Allocating = .init(currentWindow().arena());
            str_builder.writer.print(fmt, fmt_args) catch {};
            AccessKit.nodeSetLabel(ak_node, str_builder.toOwnedSliceSentinel(0) catch "");
        }
    }

```

david-vanderson:
> I did create a nodeReparent() previously. I can't remember now what caused me to abandon it.
> 
> I'll revisit and see if that is viable as it's much nicer. We still need to tell the TextEntryWidget who to parent the text_runs to, but at least that would move it out of Options.

Looking more we could have TextEntryWidget set some kind of "parent override" and reset it once all the text has been sent to TextLayoutWidget, unsure though.

phatchman:
Some backends are not working. Also need to add support for raylib-zig backend.

This is to be resolved in a separate PR, but documenting here until created. Note results are valid for Windows (aarch64)

| Backend / Example| Working | Issue? |
|--------------------|----------|-------|
| dx11-app | Y | |
| dx11-ontop | N |Crashes on startup |
| dx11-standalone | Y |  Reports of slowness when using a screen reader.  |
| raylib-app | Y | Only shared library supported |
| raylig-ontop | N | Crashes on startup |
| raylib-standalone | Y | Only shared library supported |
| sdl2-app| Y | | 
| sdl2-ontop| Y | |
| sdl2-standalone| Y | |
| sdl3-app| Y | | 
| sdl3-ontop| Y | |
| sdl3-standalone| Y | |
| sdl3gpu-ontop| N | Crashes on startup |
| sdl3gpu-standalone| Y |  |
| raylibzig-* | N | No support added yet. |

One of the downsides of not enabling by default is that people add/change backends without making them compatible with accesskit. It would be good to at least ask any new backend implementer to test changes against accesskit before merging.

phatchman:
Updated examples.

phatchman:
@david-vanderson  Not sure how far into the weeds you want to get with testing. 

But if you want to play with the text selection, I was using accessibility insights. The "..." menu at the same level as the "Text Range(s)" dropdown, you can use the Move-> By Unit (moves anchor and focus) or Move Endpoint By Unit to move anchor and focus independently.

Then you can use "Select" from the same menu to select the text.

Changing the TextRanges from Document to Selection will highlight show you just the selected text.

For any screen readers you need to find their text navigation key. Some of them, will ask DVUI to move the cursor when you navigate with them, so the text will follow along. But some (e.g. NVDA) don't. So, text won't until I implement the "scroll_into_view" action.

david-vanderson:
> One of the downsides of not enabling by default is that people add/change backends without making them compatible with accesskit. It would be good to at least ask any new backend implementer to test changes against accesskit before merging.

That is a good idea, I will try to do that for new backends.

> @david-vanderson Not sure how far into the weeds you want to get with testing.

Thanks very much for the pointers!  I've done one round of high level testing with NVDA, and it seemed okay (of course I'm super new to this).  I'm going to do another round with accessiblity insights, and go back through all the comments here.

phatchman:
@david-vanderson The reParent() version is here: https://github.com/phatchman/dvui-dev/tree/accesskit_textrun_reparent

It's a single commit on top of this PR.

After thinking about it more and how reparenting requires allocation and iterating and copying the current parent's children, I think it might be simpler just to special-case nodeCreate for text-runs and not create a parent for them. The parent can be set in textRunPopulate instead. If you create a text_run without populating it, AccessKit is going to have a bad time anyway.

As to getting rid of the `ak_text_parent` in TextEntryWidget I can't really see a nice way around it. The text layout has no knowledge of the text entry widget, unless it walks up the widget chain. But what would it look for? How would it know it found a TextEntry widget (or whichever widget) that is expecting to be the text_run parent?

If I added some extra infra to keep track of parents in the accessibility node tree, a widget could mark itself there as a "text parent" or something and the text_run would walk the accessibility parents to the closest "text parent". Acesskit only gives me a way to find children, not parents, so we'd need to keep this ourselves.

The only thing we keep now is a map of dvui.Id to *Node. So, it would require extra book-keeping just for that.

Edit: The delayed parent assignment version is here: https://github.com/david-vanderson/dvui/compare/main...phatchman:dvui-dev:accesskit_textrun_alt

I think that is cleaner, even though it does create special treatment just for text runs.

Edit: Ahh this introduces a bug for empty lines in multiline text edits. I need to work that out.

DataTriny:
Unsure if it is in scope for this pull request: the placeholder text is not communicated properly to AccessKit. My understanding is that right now it goes through the normal text layouting system, thus a text run is created for it. AccessKit expects you to not create a text run node for placeholder text, but rather set the dedicated node property using `accesskit_node_set_placeholder`. You can always set this property even when the text input is not empty, we will make sure to only expose the placeholder when it is needed.

DataTriny:
There is an issue with soft line breaks where moving the caret at the end of a line will go through the first character of the next line twice. When navigating by characters the caret will visually go past the word breaking character so the issue is probably not just related to accessibility. Paragraphs spanning multiple lines seem to mess up the text run bounds though: if you have a two-line paragraph and then a single line, it seem to me that the single line bounds overlap with the first paragraph.

phatchman:
> Unsure if it is in scope for this pull request: the placeholder text is not communicated properly to AccessKit. 

To confirm, when you say placeholder text, do you mean blocks of read-only text, like on the DVUI demo's main window? 
Are the text_runs only applicable to text inputs / editable text and nothing else?

Right now, everywhere that uses our TextLayoutWidget will get a text_run created. Even if it is something like tooltip text. But if it is limited to TextEntry (and/or a few others) I can sort that out fairly easily.

DataTriny:
> Are the text_runs only applicable to text inputs / editable text and nothing else?

No in theory they can apply to a lot of widgets, web areas being the primary use case.

By placeholder text I mean for instance "enter a password" for the password input of the SDL3 standalone example. Right now it appears as readonly text that cannot be selected, so for instance if I use the right arrow key I can hear "e" because the screen reader sees it as regular text, just that the caret is stuck at the start.

Also, is text selection set for empty text inputs? Right now I'd say no, because using arrow keys don't trigger anything, whereas the NVDA screen reader will typically say "blank" when you hit arrow keys on empty inputs.

david-vanderson:
> This may be linked to the fact that word navigation works differently depending on the direction in DVUI. Jumping to the next word places the caret at the end of the current word while jumping to the previous word places the caret at the start of the previous word. Typically the caret is always placed at the start of the word, no matter the direction. I know this is a small detail, but that's the sort of thing that make or break UX.

I was initially confused by this, because I very carefully copied how browsers work here.  Surprise - it's OS dependent!  On Windows the caret is always at the beginning.  Mac and Linux work the way dvui does - the caret is at the beginning when moving backwards, and the caret is placed at the end of each word when moving forward.  I will make a separate issue for this: #671

david-vanderson:
> Unsure if it is in scope for this pull request: the placeholder text is not communicated properly to AccessKit.

This should be fixed now, thanks!

david-vanderson:
@phatchman I based this on your reparent code, but using a separate kind of stack of values that textEntry overrides.

I did have some trouble with it - during testing it seemed like sometimes on a single frame some widgets would get an `ak_node` and others not.  Does that ring any bells?  I will keep looking into it.

ethindp:
Gotta love OS-dependent stuff.

Re: accessibility testing tools. I've never used many of these (e.g., I never knew of Accessibility Insights until it was mentioned on here) because I don't feel like they provide the same experience as actual screen readers do. Screen readers implement all kinds of hacks and workarounds that make them not really equivalent to an accessibility testing tool which only sees the accessibility tree.

I'm not really sure what input I can provide here regarding this issue, since I'm not too familiar with text runs and AccessKit in that regard.

phatchman:
> I did have some trouble with it - during testing it seemed like sometimes on a single frame some widgets would get an `ak_node` and others not. Does that ring any bells? I will keep looking into it.

I can't think of anything that would cause that. The visible check in nodeCreate was tripping me up, but that should be skipped for text runs. Otherwise maybe the delayed text rendering vs immediate?

david-vanderson:
> I can't think of anything that would cause that. The visible check in nodeCreate was tripping me up, but that should be skipped for text runs. Otherwise maybe the delayed text rendering vs immediate?

Aha thanks, that helped me figure out that I was missing that textEntry tells the embedded textLayout to have a role of "none", so it had no `ak_node`.  Phew.

I'm almost ready to merge this.  I'm aware of 3 remaining issues:
1. As @DataTriny reported, moving the caret through a soft line break reports the first character of the next line twice.  I'm going to look into this before merging, but might punt it to a separate post-merge issue.
2. In the demo text layout section, when I mouse over the text "Body font is Vera Sans" NVDA reads the "caption text" (separate text layout in the upper-right corner).  I think this is because the caption text is in a nested text layout.  Probably will be a post-merge issue.
3. I'm hitting a weird accesskit crash on linux when mouse-drag selecting text from a line, down through an empty line and into a 3rd line with text.  (Not sure all of that is required).  Accesskit reports:
```
thread '<unnamed>' panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/accesskit_consumer-0.30.1/src/text.rs:612:23:
byte index 31 is out of bounds of `Framerate is capped by vsync.
`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Will probably push that into a post-merge issue as well, because I want to try updating accesskit first.

phatchman:
> 3. I'm hitting a weird accesskit crash on linux when mouse-drag selecting text from a line, down through an empty line and into a 3rd line with text.  (Not sure all of that is required).  Accesskit reports:
> 

The index out of bounds usually means that the text set though nodeSetValue is not the same length as the arrays passed to the text run. It can also be that the word length calculations are wrong. In that case I usually set the debug_text_runs to true and confirm all the text run values look good.

It's crashing on
```
                let slice_end = slice_start
                    + character_lengths[start_index..end_index]
                        .iter()
                        .copied()
                        .map(usize::from)
                        .sum::<usize>();
                &value[slice_start..slice_end]
```
Which suggest the character lengths array is shorter than the text.

Edit: When I had a similar issue I was able to debug under lldb and print out the node_id of the text_run causing the issue.

phatchman:
> Will probably push that into a post-merge issue as well, because I want to try updating accesskit first.

I have a [branch ](https://github.com/david-vanderson/dvui/compare/main...phatchman:dvui-dev:accesskit_upgrade) which updates to the latest accesskit-c. Main changes:
1) Changed to use the WithLength versions of any string setting.
2) Text runs changed from using word lengths to word starts.

Good news is that accesskit_gen.zig works without modification to create the new mappings.

david-vanderson:
Found the crash - it was my fault from an old part of textLayout that added a space if there was a selected newline.  The intention is to be able to show that the newline is selected.  But it screwed up accesskit when the selection bridged two text runs and the first one had this extra space that wasn't fully accounted for.

I've changed it so we do pass the newline into renderText, and swap it for a space inside rendering.

david-vanderson:
I've filed issues for the 2 remaining things I'm aware of.  Please let me know if I'm missing anything.

This is a huge step forward for dvui, and I'm super grateful.

@phatchman Thank you so much for putting all the time and effort into making this happen!

@DataTriny and @ethindp Thank you thank you for the testing and helping us figure out the details!

phatchman:
I like that last commit. It really cleans up the newline handling. 

I wasn't really happy with how I implemented that, especially with having to pass in almost identical text twice. I'll give this some more testing later on.

PR #666, opened 2026-01-08
Integrate kb_text_shape for ligatures

### LINKED ISSUE #662: Text shaping/Ligatures
There was some talk about text shaping [here](https://github.com/david-vanderson/dvui/issues/121#issuecomment-3145034239), but I didn't see a dedicated issue for this (since it's a big thing on it's own) so I'm opening one here. Some fonts with heavy ligature (ab)use that are good for testing text shaping are fonts like [linja pona](https://musilili.net/linja-pona/) or [Neptunian Shpre](https://www.laghariportals.com/_files/archives/b5ea1a_1155d3b9bea3438486f517b7711825d4.zip?dn=neptunian_shpre.zip) which are for constructed languages, these use ligatures to emulate an Input Method Editor, letting you type in the script without needing extra software or copy+paste to type bespoke unicode codepoints.

As a temporary workaround I can just move all these ligatures to their own dedicated codepoints in the reserved codepoint section of unicode and manually convert the ligatures to the correct codepoints before giving to dvui, but this is much more difficult, and makes integrating new fonts harder. However it will be enough for my current uses (porting a conlang calendar app from HTML+JS -> dvui+Zig).

linja pona example text: `tokiasoweli[_pilin_e_lili_ijo] nasinpi++ao`
Image

Neptunian Shpre example text: `CAFRERHMWNYE`
Image

david-vanderson:
Fixes #662 

This is really rough so far, but I can run `zig build sdl3-app` and type `tokiasoweli[_pilin_e_lili_ijo] nasinpi++ao` and it almost works.  Something is off at the end with those ++ things.

Got a ways to go, but this is at least promising.

Still a bunch of stuff to be figured out:
- [x] loading Neptunian Shpre font crashes
- [x] pressing a single enter crashes
- [ ] kb_text_shape sizing (and other metrics) don't account for hinting (how do we handle this?)
- [ ] textEntry cursor positioning and movement
- [ ] performance implications?


MichaelBelousov:
I'd be curious about measured binary size increase, especially on the web backend

PR #672, opened 2026-01-12
Initial library support for browser consumers

MichaelBelousov:
Some consumers want to use DVUI in the browser more flexibly, e.g. instantiating on multiple canvas elements easily, bundling (even minifying) the result with "modern/industrial" web bundlers, adding additional wasm imports/exports, or even use DVUI in a JavaScript browser-platform library.
I currently maintain a local patch for this, but some upstream changes could allow me and others to use DVUI directly as a library instead. 

This PR is some initial work to better enable those workflows, with the following changes:
- shipping minimal types (web.d.ts) for typescript users (useful for people wrapping DVUI in a typescript library and also LSP completion when using it)
- allow typechecking (you can now `npm install` in DVUI if you want to set up typing
- allow installing of DVUI as a dependency of a javascript project (add package.json)
- basic bundler support (only `vite` has been tested so far)

I will happily own the newly added code and review new additions to `web.js`, and I have some future planned changes as I work my code around these changes, but I think this is a fine start to get an idea of where I'm thinking of going. I am trying to be minimal in changes while useful externally.

### Notes

- I chose to encourage users calling `dvui` with their own instantiated WASM module to allow for interoperability with the most bundlers, as loading WASM can be done in many ways
- Need to add a vite example project and/or bundler documentation. Doing this may also make me decide to _not_ build the wasm executable when installing the dependency. That makes sense for some bundlers/loading methods, but not all, so I'm not quite sure, but hopefully making some example project for this will tighten my thoughts on this
- the `npm run typecheck` currently yields many, some somewhat real, errors in `web.js`. These will be fixed in future work.

### Current State

Locally I tested setting up a dummy `vite` project with a very simple index.html like so:

```html
<!doctype html>
<html lang="en" style="height: 100%;">
  <head>
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
  </head>
  <body style="height: 100%; margin: 0; padding: 0;">
    <canvas id="dvui-canvas" tabIndex="1" style="display: block; width: 100%; height: 100%; outline: none; caret-color: transparent;"></canvas>
    <script type="module">
      import { dvui } from "dvui";
      dvui("#dvui-canvas", "./node_modules/dvui/zig-out/bin/web-app/web.wasm");
    </script>
  </body>
</html>
```

and I was able to "install" DVUI as a javascript dependency for use with vite, and it ran zig build for me (via `postinstall` script) using the following dependency:

```sh
npm install https://github.com/MichaelBelousov/dvui.git#industrial-js
```

That will of course work with the official DVUI repo once this is merged, and without a `#commit`  (although I'd recommend one)

### DIFF
--- a/src/backends/index.html
+++ b/src/backends/index.html
@@ -18,6 +18,7 @@
     <script type="module">
       // TEMPLATE_HASH replaced by build.zig
       import { dvui } from "./web.js?cache-buster=TEMPLATE_HASH_WITH_PADDING__ITS_64_BYTES_LONG_THE_SAME_AS_SHA256";
+      // when using vite, this can be 'node_modules/dvui/zig-out/bin/web-app/'
       dvui("#dvui-canvas", "web.wasm?cache-buster=TEMPLATE_HASH_WITH_PADDING__ITS_64_BYTES_LONG_THE_SAME_AS_SHA256");
     </script>
   </body>


RedPhoenixQ:
I did the refactor of web.js a while back, adding most of the types there. When this lands I can go and fix the types and eventual bugs that exist.

david-vanderson:
Thank you so much!  I'm a bit behind here, but I'll try to keep up!

david-vanderson:
Sorry this took me so long.  Thank you!

MichaelBelousov:
was integrating this into Graphl. Looks like I'll need I'll need a follow up PR to allow initializing a dvui instance to get wasm imports before I attach the wasm instance.

Should have something today I think

MichaelBelousov:
also I will remove the postinstall script... I can disable it kind of with bun but it's not conventional and not possible to disable in other js package managers.

Mostly, 99.9% of people will be building their own executable with their own UI code so it's pretty useless imo and I don't anyone to accidentally build it lol.

Issue #675, opened 2026-01-13
Empty text inputs must report caret position to AccessKit


DataTriny:
As I said in #661 

> Also, is text selection set for empty text inputs? Right now I'd say no, because using arrow keys don't trigger anything, whereas the NVDA screen reader will typically say "blank" when you hit arrow keys on empty inputs.

#### To reproduce

- Have the NVDA screen reader running
- Open the sdl3-standalone example
- Go to the Text Entry demo
- Focus the first text entry which should be empty
- Press Left/Right arrow keys

#### Current behavior

You won't hear anything, as if the key presses were ignored. I also think using Accessibility Insights for Windows you would be able to see that no selection is reported. This is particularly obvious for Braille display users as the screen reader would not display a blinking cursor.

#### Expected behavior

NVDA should report "blank" when Left/Right arrow keys are pressed on an empty text input, indicating that there is a caret in that field.

david-vanderson:
Sorry I missed it, thank you for the detailed report!

david-vanderson:
I believe this is fixed now - does it work for you?

I noticed a similar issue with the password field, so am going to keep this issue open until I can look into that.

DataTriny:
Yes, strangely regular inputs are fixed but not the password field of the demo. The issue seem not to be with the kind of the input but rather whether a placeholder text is set for it. If I comment out line 134 of src/Examples/text_entry.zig then caret position is reported for that input as expected.

phatchman:
The issue is that this line in addTextDone() does not pass when there is placeholder text:
```
            if (self.bytes_seen == 0 or self.newline) {
```
That line is trying to pick up that the textRun is empty. If I don't see anything else maybe I'll just keep a text_added flag instead? Can look at this with the other fixes I'm working on.  Edit: Ahh I see that won't work, that's correctly not creating a text run for the placeholder text. So, we need to also create and empty text run for that case.

phatchman:
OK. Have a fix for this. coming in #677. Basically just exposed a function in textLayout to create empty text runs and call that from text entry if there is placeholder text. then set the text_run_parent to null to stop the text run being created for the placeholder as we do today.

Issue #674, opened 2026-01-13
accesskit: nested text layouts


david-vanderson:
Deferred from #661

In the demo text layout section, when I mouse over the text "Body font is Vera Sans" NVDA reads the "caption text" (separate text layout in the upper-right corner). I think this is because the caption text is in a nested text layout.

I haven't looked into it deeply yet, but my suspicion is that we need the parent textLayout to not have the "label" role, and instead insert some extra accesskit node in there.  We need to communicate to accesskit that textLayout might have normal widgets in the corners before we start adding text runs.

phatchman:
Looking at this with some web and desktop examples, the expected accessibility tree would be (ignoring the icons etc):
```
document
  + label (for main body text)
      + text_run
      + ...
  + label (for caption text)
      + text_run
      + ...
```
OR
```
document
  + label (for main body text)
      + text_run
      + ...
  + group
      + label (for caption text)
         + text_run
         + ...
```
Optionally, the second label could be in a group. i.e. top-level text entries create a document and nested create a group. My browser example doesn't nest into a group and has all text under document. My desktop example creates a group for any nested text and places the label in that group.

The only roles that support text selection are document, label and terminal. i.e. the text run parent needs to be one of those.


The current tree looks like:
```
  + label (for main body text)
      + text_run
      + ...
      + label (for caption text)
         + text_run
         + ...
```
I think I'll need to implement a "text run container stack" similar to the text_run_parent stack to keep track of the parent of the text run parent. (i.e. the document or the group). The main issue being that the from an accessibility point of view the embedded text layout needs to be parented to the "document", rather than the parent text layout widget.

phatchman:
Believe this the correct tree for the example. It reads as expected.

<img width="540" height="267" alt="Image" src="https://github.com/user-attachments/assets/5196a762-7861-4f50-bd54-222910a3265a" />

<img width="606" height="151" alt="Image" src="https://github.com/user-attachments/assets/55516abf-f635-4bc1-99a9-dc26fce19b1f" />

I tried re-arranging the demo so that all of the "caption" text is inside the second text layout, which would put all of the caption information into the same group as the text run. But that gave me a null dereference at TextLayoutWidget.zig: line 1300
```
// widget on right side, need to add width to min_size below
width_after = self.corners_min_size[i].?.w;
```
Issue appears to be with the cbox being inside the second text layout.

In the short-term I can use a .group role on the "cbox" (caption box) to group everything together.

[...]

Issue #673, opened 2026-01-13
accesskit: moving caret through soft line break reads start character twice


david-vanderson:
Deferred from #661 

* enable screen reader
* open demo, text entry section
* check "Break Lines"
* add some text that breaks across a line
* position the cursor/caret on the first line and move by pressing "right"
* screen reader reads the first character of the second line when the caret moves past the last character on the first line
  * "line end position"
* screen reader reads (again) the first character of the second line when the caret moves to the second line

I think this is a consequence of dvui having two logical positions for the caret that correspond to the same byte position.  `Affinity` controls where we show the caret - either in front of the next byte, or just after the previous byte.

Some quick testing says that some places (Firefox on Windows) solve this by reading "blank" for the "line end position".  Others (Edge on Windows) don't let the caret be in the "line end position".

Here we need more testing of what is done in other applications (both web text area and normal apps), before deciding what kind of behavior we want.

DataTriny:
@david-vanderson As you pointed out, the behavior vary between implementations. Firefox would say "blank" when the caret lands on a soft line break, even though the caret is visually placed after the last character, it is placed on the first character of the next line for people who use a Braille display. I've only tested this on Windows so far so maybe other platforms have different conventions, but right now my opinion is that the caret should never be able to reach a soft line break. At least this is how Chromium and modern Windows UI behave, and I tend to trust them more.

So, to be clear, "soft line breaks" don't appear in the AccessKit schema. You would just have a text run which value ends with a space for instance, and the next text run would represent the line below, with the appropriate bounds. The soft line breaks are not communicated through speech synthesis, but Braille display users would see where the line end. It is of course the responsibility of the UI toolkit to make sure that Home/End keys correctly jump at the start/end of lines with soft breaks, and not at the start/end of a paragraph.

Editable text is hard...

david-vanderson:
> Firefox would say "blank" when the caret lands on a soft line break, even though the caret is visually placed after the last character, it is placed on the first character of the next line for people who use a Braille display.

Do you mean that the behavior changes depending on the display?  Or do you mean we could visually show the caret in one place (end of the first line) while communicating a different location (start of the second line) to accesskit?

> So, to be clear, "soft line breaks" don't appear in the AccessKit schema.

Good - thank you for the information, that helps me understand the data model.

> Editable text is hard...

No kidding!

DataTriny:
[...]

Issue #616, opened 2025-10-27
Some Accessibility bugs


ethindp:
Sorry for the super generic issue title, not really sure what I should title this since these are just issues I encountered while playing with the demos. This is quite rambly, sorry!

First, there's a slight delay/lag in the DirectX 11 demos. Not entirely sure why but this seems go vanish with SDL3. Godot had this same issue; not sure how they fixed it.

Another issue Godot had that your running into is edit boxes not really acting like edit boxes. As in, I can focus them, but I can't navigate them with the arrow keys, or read the line with the up/down arrows.

Another is that the menu bar (the file/edit menus) are focusable. I don't know if you have to do something different to make them act more like menu bars; I wonder if we could potentially simulate OS behavior here? I don't really know how best to do that or if AccessKit would even allow us to do that.

Lastly, the tree view widget is... Kinda broken. If you launch NVDA and run the sdl3-standalone app, you'll notice a few things after going into the demo window and selecting the reordering/tree button:

1. Tree view widgets appear in tab order. Usually you want the tree view itself to be in tab order but not the individual nodes. Ideally, a tree view is navigable with the arrow keys: up and down navigate among either the nodes at the current level or, if the node has children and is expanded, descends to the next level automatically, and so on all the way down. Left and right arrows collapse and expand nodes: if you are on a node and it is collapsed, right arrow expands the node but does not descend into it, and left arrow anywhere in a list of nodes which has a parent node at the top jumps to that parent node first, and another press of the left arrow collapses it. If your already on the parent node, one press of the key collapses the node.
2. Tree view items aren't read right now; NVDA just says "tree view item". (It does announce the level your at though!) I was able to sort-of navigate the tree with object nav, but it felt awkward since I also saw the graphics/images and such.

Tabbing also felt a bit weird. Like, I would tab to the end of the window, and hitting tab again would bring me to the actual window itself and nothing would be focused again. I'm not sure if we want this behavior or not. Part of me wants it to be like this (after all, that could be where you might want to pass all keyboard input through to the application), like a keyboard area of sorts. But part of it doesn't feel very natural and definitely gives "this isn't a native UI" vibes. Which isn't a bad thing, by any means. Sometimes, you *want* that.

I'm not really sure if elements not appearing as part of groups when they are supposed to be grouped (or I would assume they would usually be that way) is just not that being implemented or if that's intentional. Some other thoughts as I go through the app:

[...]

PR #677, opened 2026-01-14
Upgrade accesskit_c to v 0.20.0 plus fixes and enhancements

### LINKED ISSUE #670: Update accesskit-c library to v0.19.0
The latest AccessKit C bindings provide APIs to send text via a ptr and len, rather than null-terminated C-strings.

This will greatly reduce the amount of string copying / duplication required just to null-terminate strings.

https://github.com/AccessKit/accesskit-c/releases/tag/0.19.0

Also fix any backend issues identified on https://github.com/david-vanderson/dvui/pull/661#issuecomment-3733711220

phatchman:
- Regenerated accesskit API mappings.
- Use the "WithLength" functions for strings.
- Change from using word lengths to word starts for text runs

Fixes #670

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -43,8 +43,8 @@
             .hash = "svg2tvg-0.0.0-HpMZuWZ3BgApRrFbd9CxtabLoXEA5odBFUDFrDqno8Ih",
         },
         .accesskit = .{
-            .url = "https://github.com/AccessKit/accesskit-c/releases/download/0.19.0/accesskit-c-0.19.0.zip",
-            .hash = "N-V-__8AAMnMRQu3EtJ1QgonwvPKhD8L4BUHBUDiVfyLoSr4",
+            .url = "https://github.com/AccessKit/accesskit-c/releases/download/0.20.0/accesskit-c-0.20.0.zip",
+            .hash = "N-V-__8AAPwQuguI8Jsun6Omu2HHb9EADaaP8Yn2TxYplJtC",
         },
 
         // Experimental zig bindings, any zig based backend will be located here
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -43,8 +43,8 @@
             .hash = "svg2tvg-0.0.0-HpMZuWZ3BgApRrFbd9CxtabLoXEA5odBFUDFrDqno8Ih",
         },
         .accesskit = .{
-            .url = "https://github.com/AccessKit/accesskit-c/releases/download/0.17.1/accesskit-c-0.17.1.zip",
-            .hash = "N-V-__8AAJ37FAuEZXG593uEuYETrkoxDz78QLL32k0GkaYG",
+            .url = "https://github.com/AccessKit/accesskit-c/releases/download/0.19.0/accesskit-c-0.19.0.zip",
+            .hash = "N-V-__8AAMnMRQu3EtJ1QgonwvPKhD8L4BUHBUDiVfyLoSr4",
         },
 
         // Experimental zig bindings, any zig based backend will be located here
--- a/examples/raylib-ontop.zig
+++ b/examples/raylib-ontop.zig
@@ -26,6 +26,8 @@ pub fn main() !void {
     ray.SetConfigFlags(ray.FLAG_WINDOW_RESIZABLE);
     ray.SetConfigFlags(ray.FLAG_WINDOW_HIGHDPI);
     ray.SetConfigFlags(ray.FLAG_VSYNC_HINT);
+    if (dvui.accesskit_enabled)
+        ray.SetConfigFlags(ray.FLAG_WINDOW_HIDDEN);
     ray.InitWindow(800, 600, "DVUI Raylib Ontop Example");
     defer ray.CloseWindow();
 
--- a/examples/raylib-zig-ontop.zig
+++ b/examples/raylib-zig-ontop.zig
@@ -28,6 +28,7 @@ pub fn main() !void {
     raylib.setConfigFlags(.{
         .window_resizable = true,
         .vsync_hint = true,
+        .window_hidden = if (dvui.accesskit_enabled) true,
     });
     raylib.initWindow(800, 600, "DVUI Raylib Ontop Example");
     defer raylib.closeWindow();
--- a/examples/raylib-zig-ontop.zig
+++ b/examples/raylib-zig-ontop.zig
@@ -28,7 +28,8 @@ pub fn main() !void {
     raylib.setConfigFlags(.{
         .window_resizable = true,
         .vsync_hint = true,
-        .window_hidden = if (dvui.accesskit_enabled) true,
+        // Window must initially be created hidden if accesskit is used.
+        .window_hidden = dvui.accesskit_enabled,
     });
     raylib.initWindow(800, 600, "DVUI Raylib Ontop Example");
     defer raylib.closeWindow();
--- a/readme-accessibility.md
+++ b/readme-accessibility.md
@@ -21,7 +21,7 @@ The primary benefit is making your appliation useable by a wider array of users.
 
 Incorporating accessibility makes your UI application scriptable. Each widget that has a node in the accessibility tree publishes its current value along with the set of actions it can perform. By using your platform's accessibility API, you can write automated tests as well as allowing your users to script common tasks.
 
-Outside of the additional executable size from including the AccessKit library, there is very little performance impact from supporting AccessKit. (Essentially, two additional if statements per widget) 
+Outside of the additional executable size from including the library, enabling AccessKit support has very little impact on runtime performance or memory requirements. 
 
 The overhead of creating AccessKit nodes is only incurred when the operating system's accessibility API is activated by a screen reader or similar application.
 
@@ -35,31 +35,22 @@ Not all combinations of backends and operating systems are currently supported.
 - MacOS - Supports all backends in for shared and static libraries. SDL2, SDL3 and Raylib.
 - Web - Not currently supported by AccessKit but it is planned. 
 
-When you compile you application with AccessKit enabled, most of the accessibility work is taken care of for you and your users will gain most of the benefits of accessibility support. However, DVUI doesn't know your application's intent or any semantic details about the UX, so it is not possible to automate everything. 
+When you compile you application with AccessKit enabled, most of the work is taken care of for you and your users will gain most of the benefits of accessibility support. However, DVUI doesn't know your application's intent or any semantic details about the UX, so it is not possible to automate everything. 
 
-As an example, you may have a set of radio buttons, but DVUI has no way to know if those radio buttons all belong to the same radio group.
-
-For the radio button example, you can give DVUI and AccessKit some additional information by setting the 
-role option on the box surrounding the radio buttons to `.role = .radio_group`.  This will let AccessKit know that all child radio buttons within the box all belong to the same radio group. You might also choose to supply a `.label` to give the user more information about the meaning of the radio group.
+As examples, any images should be given an accessible label by passing the .label option. When placing radio button in a radio group it is also helpful to supply a label to the group.
 
 There are other AccessKit APIs that let you provide more details to the accessibility API. See [Widget users](#widget-users) for more details on how to use the AccessKit API.
 
+Example of setting a label for an image:
 ```
-{
-    var hbox = dvui.box(@src(), 
-        .{ .dir = .horizontal }, 
-        .role = .radio_group, 
-        .label = . { .text = "Make a choice"} },
-    );
-    defer hbox.deinit();
-    
-    if (dvui.radio(@src(), radio_choice == choice1), "Choice 1", .{ .id_extra = i })) {
-        radio_choice = choice1;
-    }
-    if (dvui.radio(@src(), radio_choice == choice2), "Choice 2", .{ .id_extra = i })) {
-        radio_choice = choice2;
-    }
-}
+    const image_source: dvui.ImageSource = .{ .imageFile = .{ .bytes = zig_favicon, .name = "zig favicon" } };
+    const imgsize = dvui.imageSize(image_source) catch dvui.Size.all(50);
+    _ = dvui.image(@src(), .{ .source = image_source }, .{
+        .gravity_y = 0.5,
+        .min_size_content = .{ .w = imgsize.w + icon_image_size_extra, .h = imgsize.h + icon_image_size_extra },
+        .rotation = icon_image_rotation,
+        .label = .{ .text = image_source.imageFile.name },
+    });
 ```
 
 ## Testing for accessibility
@@ -218,13 +209,13 @@ Ignoring these rules will not break anything, but will make it harder for access
 | BoxWidget | null | Basic | N/A | N | User can pass .role and .label via options |
 | ButtonWidget | button | Yes | Yes | N | Focus and Click actions supported |
 | ColorPickerWidget | slider x 4 | Partial | Partial | Y | AccessKit does not currently support 2d-sliders. Other sliders are supported. |
-| DropDownWidget | combo_box / list_item | N | Y | Y | more testing required |
+| DropDownWidget | combo_box / list_item | N | Y | Y | Supported. Add expand/collapse when supported by accesskit |
 | FlexBoxWidget | null | Basic | N/A | N | User can pass .role and .label via options |
-| FloatingMenuWidget | none | N/A | N/A | N | Accessibility Handled by Menu / MenuItem|
+| FloatingMenuWidget | none | N/A | N/A | N | Accessibility Handled by Menu / MenuItem |
 | FloatingToolTipWidget | tooltip | Y* | N | N | Tooltips are added only when shown. No support for showing / hiding tooltips |
 | FloatingWidget | N/A | N/A | N/A | N | |
-| FloatingWindowWidget | window | Y | Y | N | Can close via close button. |
-| GridWidget | grid, header, cell | Y? | N/A | Y | Needs more real-world testing. Setting row and col numbers doesn't appear to do anything |
+| FloatingWindowWidget | dialog | Y | Y | N | Defaults to .dialog. User should override with .window if more appropriate. |
+| GridWidget | grid, header, row, cell | Y | N/A | N? | Needs more real-world testing.  |
 | IconWidget | image |  Y | N/A | N | |
 | LabelWidget | label |  Y | N/A | N | | 
 | MenuItemWidget | menu_item | Y | Y | Y | Add keyboard shortcuts when supported by dvui | 
@@ -235,19 +226,19 @@ Ignoring these rules will not break anything, but will make it harder for access
 | ReorderWidget | null | N | N | Y | Not currently supported. Unlikely that interaction will work? | 
 | ScaleWidget | null | N/A | N/A | N | Not required |
 | ScrollAreaWidget | scroll_view | Y | N | N | Appears as a pane |
-| ScrollBarWidget | scroll_bar | Y *| N* | Y | Due to bug in AccessKit, is set to read-only and cannot interact. Actual values displayed may not be accurate | 
+| ScrollBarWidget | scroll_bar | Y| N* | Y | Due to bug in AccessKit, is set to read-only and cannot interact.  | 
 | SuggestionsWidget | suggestion, list_item | Y | Y | N | 
 | TabsWidget | .tab_panel, .tab | Y | Y | N | Sets active tabs and allows tab selection |
-| TextEntryWidget | text_input, multiline_text_input | Y* | N* | Y | TextRun support implemented for text. Allows selection and navigation. Scrolling for off-screen text requires improvement. Cached layouts needs more investigation. |
-| TextLayoutWidget | label | Y* | N/A | Y | Most features supported. Needs to implement the text_replace action. | 
+| TextEntryWidget | text_input, multiline_text_input | Y* | N* | Y |  Allows selection and navigation. Cached layouts needs more investigation. Needs to implement the text_replace action. Scrolling for off-screen text requires improvement. |
+| TextLayoutWidget | label | Y* | N/A | Y | Most features supported. TextRun support implemented for text. Scrolling for off-screen text requires improvement. Needs to split lines of more than 255 chars into multiple text runs | 
 | TreeWidget | tree, tree_item | Y* | Y | Y | Adds tree and nodes. Implement expand / collapse when supported by AccessKit | 
 | windowHeader | label, button | Y | Y | N | |
-| dialogs | window | Y | Y | Y | All dialogs are displayed as windows, rather than dialogs and modal state is not displayed in accessibility insights. AccessKit currently has limted support for dialogs, so leaving as window until the situation changes and can revisit. |
-| toasts | label | Y? | N/A | ? | These are set as polite annoucements via node_set_live but have not seen this cause anything to be read from the reader. |
-| comboBox | combo_box | Y | Y | N | Displays as combo box and shows list items when dropped |
+| dialogs | window | Y | Y | N | Same as floating window |
+| toasts | label | Y | N/A | N | These are set as polite annoucements. |
+| comboBox | combo_box | Y | Y | Y | Displays as combo box and shows list items when dropped. Add expand / collapse when supported by accesskit |
 | expander | group | Y* | Y* | Y | AccessKit does not currently support expand / collapse |
 | context | ? | ? | ? | ? | Will need to implement "show context menu" action. Which will be mapped to right-click. |
-| gridHeadingSortable | button | Y* | Y | Y | Sets sort state of asc/desc if sorted, but does not come through in accessibility insights. further investigation required |
+| gridHeadingSortable | button | Y* | Y | Y | Sets sort state of asc/desc if sorted, but does not come through in accessibility insights. requires accesskit support. |
 | gridHeadingCheckbox | button | Y | Y | N| Labels checkbox as select all / select none. |
 | image | image | Y* | N/A | N | Requires user to label the image with .label |
 | slider | slider | Y | Y | N | Fully supported |
--- a/src/AccessKit.zig
+++ b/src/AccessKit.zig
@@ -116,6 +116,9 @@ fn windowsHWND(window: *dvui.Window) c.HWND {
         .raylib => {
             return @intFromPtr(dvui.backend.c.GetWindowHandle());
         },
+        .raylib_zig => {
+            return @intFromPtr(dvui.backend.raylib.getWindowHandle());
+        },
         .dx11 => {
             return @intFromPtr(window.backend.impl.hwndFromContext());
         },
@@ -234,9 +237,7 @@ pub fn nodeCreateReal(self: *AccessKit, wd: *dvui.WidgetData, role: Role) ?*Node
                 }
             },
             .text => |txt| {
-                const str = dvui.currentWindow().arena().dupeZ(u8, txt) catch "";
-                defer dvui.currentWindow().arena().free(str);
-                nodeSetLabel(ak_node, str);
+                nodeSetLabelWithLength(ak_node, txt.ptr, txt.len);
             },
         }
     }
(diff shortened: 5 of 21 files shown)


phatchman:
I plan to add another commit to make this failing backends / demos work again.
Because of the change to use "word starts", instead of "word lengths" it is worth doing some quick testing with various combinations of text starting and ending with puctuation, spaces etc.

phatchman:
Both raylib and raylib-zig ontop examples are now working.

For DX11 ontop:
1) The GET_WMOBJECT messages are being sent to be handled by accesskit after the DVUI window has been deinitialized. This causes a crash on exit. 
2) Even though accesskit is now initialized, it does not interact with the screen reader. I need to investigate further, but if anyone has any ideas, please share. I know that WM_GETOBJECT is being received at least once by the DX11 backend because of the crash above. 

For SDL3-GPU ontop:
Running with accesskit enabled crashes on first call to SDL_BeginGPURenderPass I have no idea where to go with this one. Maybe the GPU render doesn't play well with the accesskit hooks? Maybe it's just memory corruption somewhere?

Stack trace below in case it helps.

```
 thread #1, stop reason = Exception 0x80000003 encountered at address 0x7ff72f96feac
  * frame #0: 0x00007ff72f96feb0 sdl3gpu-ontop.exe`tree_sitter_json at posix.zig:689
    frame #1: 0x00007ff72f94edbc sdl3gpu-ontop.exe`tree_sitter_json at debug.zig:722
    frame #2: 0x00007ff72f94edbc sdl3gpu-ontop.exe`tree_sitter_json at debug.zig:722
    frame #3: 0x00007ff72f94dcc4 sdl3gpu-ontop.exe`tree_sitter_json at debug.zig:602
    frame #4: 0x00007ff72f94d878 sdl3gpu-ontop.exe`tree_sitter_json at ubsan_rt.zig:461
    frame #5: 0x00007ff73017f408 sdl3gpu-ontop.exe`SDL_BeginGPURenderPass_REAL(command_buffer=0x00000197f0e3d650, color_target_infos=0x0000006a1b1ff678, num_color_targets=1, depth_stencil_target_info=0x0000000000000000) at SDL_gpu.c:1738
    frame #6: 0x00007ff72ff8fba4 sdl3gpu-ontop.exe`SDL_BeginGPURenderPass(a=0x00000197f0e3d650, b=0x0000006a1b1ff678, c=1, d=0x0000000000000000) at SDL_dynapi_procs.h:68
    frame #7: 0x00007ff72f9db528 sdl3gpu-ontop.exe`main(backend=sdl3gpu @ 0x0000006a1b1fcbf8) at sdl3gpu-ontop.zig:111
    frame #8: 0x00007ff72f9df998 sdl3gpu-ontop.exe`main [inlined] callMain at start.zig:627
    frame #9: 0x00007ff72f9df98c sdl3gpu-ontop.exe`main [inlined] callMainWithArgs(argc=1, argv=0x00000197f013a210, envp=[][*:0]u8 @ 0x0000006a1b1ff830) at start.zig:587
    frame #10: 0x00007ff72f9df908 sdl3gpu-ontop.exe`main(c_argc=1, c_argv=0x00000197f013a210, c_envp=0x00000197f011ec80) at start.zig:602
    frame #11: 0x00007ff72ff8f240 sdl3gpu-ontop.exe`__tmainCRTStartup at crtexe.c:259
    frame #12: 0x00007ff72ff8f054 sdl3gpu-ontop.exe`WinMainCRTStartup at crtexe.c:148
    frame #13: 0x00007ffca58d8740 kernel32.dll`BaseThreadInitThunk + 64
```

phatchman:
@david-vanderson I'm good to merge if you are happy.

dx11-ontop no longer crashes, but still doesn't actually create an accessibility tree. accesskit_windows_adapter_update_if_active never calls frameTreeUpdate(). Not sure if what I've done in dx11.zig is the best way, but I need to stop the backend processing windows events after it has been deinitialized.

The sdl3gpu-ontop still crashes, but I don't know where to even start with that one.

Everything else looks to work fine (at least on Windows).

phatchman:
Ahhh. 

I found an issue. It only reads the first paragraph of text in a multi-paragraph text layout. 

I'll need to look into that later. 

I also have a handful of fixes to various other things, which I think I'll include in this PR. Just to cut down on testing effort.

david-vanderson:
> dx11-ontop no longer crashes, but still doesn't actually create an accessibility tree.
> The sdl3gpu-ontop still crashes, but I don't know where to even start with that one.

I think this is okay.  We need to find a balance between backends and accesskit support.  I don't know exactly what that will look like, but my guess is that we shouldn't put too much effort into accesskit support for a given backend unless/until someone needs it.  We're likely to always have some backends that are in various stages of being brought up to full functionality, and no way to know what people need unless they speak up.

At some point I think we'll switch accesskit from default off to a backend-specific default so we can have it on by default in at least the sdl3 backend.

> I found an issue. It only reads the first paragraph of text in a multi-paragraph text layout.
> 
> I'll need to look into that later.
> 
> I also have a handful of fixes to various other things, which I think I'll include in this PR. Just to cut down on testing effort.

Sounds good thanks!

phatchman:
Agree. I wasn't planning to spend a lot of time getting the on-top ones to work. But not having them crash would be nice :)

Fixed the issue with text runs. With the new API, we need to call `nodeSetPreviousOnLine` and `nodeSetNextOnLine`.
Edit: Well, no that just made it a bit better. Still needs work.

phatchman:
@DataTriny Sorry to bother you again, but I need your help on this one.

Since upgrading to using the 0.19.0 C API, the screen reader stops reading text after the first line of text for any text runs. This happens both in our static text (text layout) and editable text (text entry).

To troubleshoot what was going on, I've recreated the scenario in the test function [main_multiline_tree](https://github.com/AccessKit/accesskit/blob/main/consumer/src/text.rs#L1692) from text.rs. 

See below for example:

image

Under the previous version of accesskit, the screen reader will read all the text in the text layout. 
Under the new version. it only reads the first line. 

The DVUI-side differences since the upgrade are using `node_set_value_with_length` and the change from using word lengths to word starts and using `node_set_prev/next_on_line` for text runs on the same line.

The only difference to the test.rs example is that I'm sticking to the ascii character set.

I have dumped out the text_run values we create for this test below. As far as I can tell we are matching the text.rs version as closely as we can. I can't see any difference to what we do in the test that would be causing this change in behaviour.

If you are able to test, this is the [branch](https://github.com/phatchman/dvui-dev/tree/textrun_testing). Would appreciate any insight here as I just can't see anything wrong.

Edit: If I incorrectly use `node_set_prev/next_on_line` to link all lines in the same paragraph, the reader will read the first paragraph and then stop. If I incorrectly link all lines, regardless of whether they are in the same paragraph, it is the same behaviour, the reader reads the first paragraph and then stops.
If I use word navigation in the reader, it can navigate between lines, word by word fine.

Edit2: To clarify, this is when I am testing with windows narrator, especiailly using CAPS-R to read the entire screen. NVDA seems to work and be able to read all text. .... I've done some more testing with NVDA, and as far as I can tell it is working and reads everything. It's just Windows narrator that is not playing ball.

Edit3: Fixed the @. Thanks DV.

The first [number] in the log message is the node_id of the parent and the second is the node_id of the text_run.

```
---------------------------------
TEXT_RUN [739440b30fd92069:80f9b366d5d732fe]: text = `This paragraph is long enough to wrap `
TEXT_RUN [739440b30fd92069:80f9b366d5d732fe]: bounds = .{ .has_value = true, .value = .{ .x0 = 7.5, .y0 = 52.5, .x1 = 338.75, .y1 = 72.5 } }
TEXT_RUN [739440b30fd92069:80f9b366d5d732fe]: lengths = { 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 }
TEXT_RUN [739440b30fd92069:80f9b366d5d732fe]: positions = { 1.5, 11.5, 22.5, 27.5, 36.5, 41.5, 52.5, 62.5, 69.5, 79.5, 90.5, 97.5, 107.5, 118.5, 129.5, 134.5, 139.5, 148.5, 153.5, 158.5, 168.5, 179.5, 190.5, 195.5, 205.5, 216.5, 226.5, 237.5, 248.5, 259.5, 264.5, 271.5, 281.5, 286.5, 300.5, 307.5, 317.5, 328.5 }
TEXT_RUN [739440b30fd92069:80f9b366d5d732fe]: widths = { 12, 9, 3, 9, 5, 9, 8, 6, 8, 10, 6, 8, 9, 9, 5, 3, 9, 5, 3, 10, 9, 10, 5, 10, 9, 10, 9, 10, 9, 5, 7, 10, 5, 14, 6, 8, 9, 5 }
TEXT_RUN [739440b30fd92069:80f9b366d5d732fe]: word_starts = { 5, 15, 18, 23, 30, 33 }
TEXT_RUN [739440b30fd92069:80f9b366d5d732fe]: prev_on_line = null
TEXT_RUN [739440b30fd92069:80f9b366d5d732fe]: next_on_line = null
---------------------------------
TEXT_RUN [739440b30fd92069:e7c0280f36523218]: text = `to `
TEXT_RUN [739440b30fd92069:e7c0280f36523218]: bounds = .{ .has_value = true, .value = .{ .x0 = 7.5, .y0 = 76.5, .x1 = 29.5, .y1 = 96.5 } }
TEXT_RUN [739440b30fd92069:e7c0280f36523218]: lengths = { 1, 1, 1 }
TEXT_RUN [739440b30fd92069:e7c0280f36523218]: positions = { 0.5, 7.5, 17.5 }
TEXT_RUN [739440b30fd92069:e7c0280f36523218]: widths = { 7, 10, 5 }
TEXT_RUN [739440b30fd92069:e7c0280f36523218]: word_starts = { 0 }
TEXT_RUN [739440b30fd92069:e7c0280f36523218]: prev_on_line = null
TEXT_RUN [739440b30fd92069:e7c0280f36523218]: next_on_line = c8c561062b62e7f7
---------------------------------
TEXT_RUN [739440b30fd92069:c8c561062b62e7f7]: text = `another`
TEXT_RUN [739440b30fd92069:c8c561062b62e7f7]: bounds = .{ .has_value = true, .value = .{ .x0 = 29.5, .y0 = 76.5, .x1 = 95.5, .y1 = 96.5 } }
TEXT_RUN [739440b30fd92069:c8c561062b62e7f7]: lengths = { 1, 1, 1, 1, 1, 1, 1 }
TEXT_RUN [739440b30fd92069:c8c561062b62e7f7]: positions = { 0.5, 10.5, 21.5, 31.5, 38.5, 49.5, 59.5 }
TEXT_RUN [739440b30fd92069:c8c561062b62e7f7]: widths = { 8, 9, 10, 7, 9, 10, 6 }
TEXT_RUN [739440b30fd92069:c8c561062b62e7f7]: word_starts = { 0 }
TEXT_RUN [739440b30fd92069:c8c561062b62e7f7]: prev_on_line = e7c0280f36523218
TEXT_RUN [739440b30fd92069:c8c561062b62e7f7]: next_on_line = b1f461b4c8f94cce
---------------------------------
TEXT_RUN [739440b30fd92069:b1f461b4c8f94cce]: text = ` line.
`
TEXT_RUN [739440b30fd92069:b1f461b4c8f94cce]: bounds = .{ .has_value = true, .value = .{ .x0 = 95.5, .y0 = 76.5, .x1 = 136.5, .y1 = 96.5 } }
TEXT_RUN [739440b30fd92069:b1f461b4c8f94cce]: lengths = { 1, 1, 1, 1, 1, 1, 1 }
TEXT_RUN [739440b30fd92069:b1f461b4c8f94cce]: positions = { 0.5, 5.5, 10.5, 15.5, 26.5, 36.5, 41 }
TEXT_RUN [739440b30fd92069:b1f461b4c8f94cce]: widths = { 5, 3, 3, 9, 10, 3, 5 }
TEXT_RUN [739440b30fd92069:b1f461b4c8f94cce]: word_starts = { 1 }
TEXT_RUN [739440b30fd92069:b1f461b4c8f94cce]: prev_on_line = c8c561062b62e7f7
TEXT_RUN [739440b30fd92069:b1f461b4c8f94cce]: next_on_line = null
---------------------------------
TEXT_RUN [739440b30fd92069:170e7e87924cda29]: text = `Another paragraph.
`
TEXT_RUN [739440b30fd92069:170e7e87924cda29]: bounds = .{ .has_value = true, .value = .{ .x0 = 7.5, .y0 = 100.5, .x1 = 173.5, .y1 = 120.5 } }
TEXT_RUN [739440b30fd92069:170e7e87924cda29]: lengths = { 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 }
TEXT_RUN [739440b30fd92069:170e7e87924cda29]: positions = { 0.5, 12.5, 23.5, 33.5, 40.5, 51.5, 61.5, 68.5, 73.5, 84.5, 94.5, 101.5, 111.5, 122.5, 129.5, 139.5, 150.5, 161.5, 166 }
TEXT_RUN [739440b30fd92069:170e7e87924cda29]: widths = { 12, 9, 10, 7, 9, 10, 6, 5, 9, 8, 6, 8, 10, 6, 8, 9, 9, 3, 5 }
TEXT_RUN [739440b30fd92069:170e7e87924cda29]: word_starts = { 8 }
TEXT_RUN [739440b30fd92069:170e7e87924cda29]: prev_on_line = null
TEXT_RUN [739440b30fd92069:170e7e87924cda29]: next_on_line = null
---------------------------------
TEXT_RUN [739440b30fd92069:9dfaf4f2450d5f4]: text = `
`
TEXT_RUN [739440b30fd92069:9dfaf4f2450d5f4]: bounds = .{ .has_value = true, .value = .{ .x0 = 7.5, .y0 = 124.5, .x1 = 8.5, .y1 = 144.5 } }
TEXT_RUN [739440b30fd92069:9dfaf4f2450d5f4]: lengths = { 1 }
TEXT_RUN [739440b30fd92069:9dfaf4f2450d5f4]: positions = { 0 }
TEXT_RUN [739440b30fd92069:9dfaf4f2450d5f4]: widths = { 5 }
TEXT_RUN [739440b30fd92069:9dfaf4f2450d5f4]: word_starts = {  }
TEXT_RUN [739440b30fd92069:9dfaf4f2450d5f4]: prev_on_line = null
TEXT_RUN [739440b30fd92069:9dfaf4f2450d5f4]: next_on_line = null
---------------------------------
TEXT_RUN [739440b30fd92069:28da76582f402015]: text = `Last non-blank line  
`
TEXT_RUN [739440b30fd92069:28da76582f402015]: bounds = .{ .has_value = true, .value = .{ .x0 = 7.5, .y0 = 148.5, .x1 = 178.5, .y1 = 168.5 } }
TEXT_RUN [739440b30fd92069:28da76582f402015]: lengths = { 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 }
TEXT_RUN [739440b30fd92069:28da76582f402015]: positions = { 0.5, 9.5, 19.5, 28.5, 35.5, 40.5, 51.5, 61.5, 72.5, 78.5, 89.5, 94.5, 104.5, 115.5, 125.5, 130.5, 135.5, 140.5, 151.5, 161.5, 166.5, 171 }
TEXT_RUN [739440b30fd92069:28da76582f402015]: widths = { 9, 8, 9, 7, 5, 9, 10, 9, 6, 9, 3, 8, 9, 9, 5, 3, 3, 9, 10, 5, 5, 5 }
TEXT_RUN [739440b30fd92069:28da76582f402015]: word_starts = { 5, 9, 15 }
TEXT_RUN [739440b30fd92069:28da76582f402015]: prev_on_line = null
TEXT_RUN [739440b30fd92069:28da76582f402015]: next_on_line = null
---------------------------------
TEXT_RUN [739440b30fd92069:98b1e3f8a198a895]: text = ``
TEXT_RUN [739440b30fd92069:98b1e3f8a198a895]: bounds = .{ .has_value = true, .value = .{ .x0 = 7.5, .y0 = 172.5, .x1 = 8.75, .y1 = 192.5 } }
TEXT_RUN [739440b30fd92069:98b1e3f8a198a895]: lengths = { }
TEXT_RUN [739440b30fd92069:98b1e3f8a198a895]: positions = {  }
TEXT_RUN [739440b30fd92069:98b1e3f8a198a895]: widths = {  }
TEXT_RUN [739440b30fd92069:98b1e3f8a198a895]: word_starts = {  }
TEXT_RUN [739440b30fd92069:98b1e3f8a198a895]: prev_on_line = null
TEXT_RUN [739440b30fd92069:98b1e3f8a198a895]: next_on_line = null
```

david-vanderson:
@DataTriny Can you look at @phatchman 's comment above?  I think they typod trying to ping you.

phatchman:
Upgrade accesskit-c to v0.19.0 with fixes
    - Use new withLength functions for setting string values
    - Use word_starts instead of word_lengths for text runs
    - Link text runs that are on the same line
    - Fix issue with text runs not being created when placeholder text is set #675
    - Fix raylib and raylib-zig ontop demos
    - Stop dx11 ontop from crashing on exit
    - Create dialogs with dialog role.
    - Fix tree widget accessibility hierarchy. #686
    - Fix scrollbar position reporting.

@ethindp I've tried to address a few of the issues you raised in #616. Hopefully the tree widget works better now? 

At one point I did have Windows narrator announcing, "1 of 8", "3 of 6" etc as I navigated the tree. But once I got all the widgets in the right hierarchy, it stopped doing that. So not sure why that went away. But hopefully it behaves like a tree now?

Dialogs should now announce themselves correctly. And tab widgets look like they were fixed at some point? At least they look good to me, but if you could check if you are still having issues with those?

Some of the toasts are working. But that is still a work in progress,

@david-vanderson I'll have a go at #674 and see if I can include that as well. I think I'll leave #673 to you.


Fixes #675
Fixes #686

DataTriny:
@phatchman your tree seem fine to me. I can confirm the regression and I can reproduce it in another implementation where Narrator behavior stopped properly working after AccessKit 0.22 update.

david-vanderson:
> @david-vanderson I'll have a go at #674 and see if I can include that as well. I think I'll leave #673 to you.

Sounds good - thank you!

ethindp:
So I've been a bit distracted on my own project, but just got the chance to try this. And it looks pretty great to me!

Some notes:

1. Multi-line edit fields are not reported; they appear as single-line but function as multi-line.
2. Some widget labels aren't being discovered (e.g., in the text box example, the labels for some of the combo boxes and edit fields). This is really hard to get right and as I noted in my original issue there's not really a "standard" so to speak, so we may wish to force it.
3. The tree views example is... A bit interesting. Node state (expanded, collapsed) isn't reported, but the actual node items themselves is. I can, for example, tell what is a folder and file, based purely on the "folder icon" versus "file icon" labelling. I think the reason the state is unreported is because, according to NVDA, these items appear as drop buttons, whatever those are, and not tree view nodes.
4. The simple grouping doesn't appear to have anything in it? Neither does the "advanced" one? Also, these groupings are focusable, so I'm guessing there's supposed to be something within it?

ethindp:
Other than that the text runs work fine for me with paragraphs and stuff. Like, the current selection is read when I un-focus and focus the item, but I think this is expected behavior.

DataTriny:
> Multi-line edit fields are not reported; they appear as single-line but function as multi-line.

If you're testing on Windows, this is a limitation of UI Automation. Not much we can do about it.

> Some widget labels aren't being discovered (e.g., in the text box example, the labels for some of the combo boxes and edit fields). This is really hard to get right and as I noted in my original issue there's not really a "standard" so to speak, so we may wish to force it.

The standard way is to call `accesskit_node_push_labelled_by` on the input widget with the ID of the label.

> Node state (expanded, collapsed) isn't reported, but the actual node items themselves is.

Expanded state is not yet exposed by AccessKit so this is expected.

> Also, these groupings are focusable, so I'm guessing there's supposed to be something within it?

Nodes with `Role::Group` must never be focusable, only their children should be.

DataTriny:
@phatchman we've just released [AccessKit 0.20](https://github.com/AccessKit/accesskit-c/releases/tag/0.20.0) which contain a fix for the Windows Narrator not reading the whole paragraph.

You might want to upgrade directly. There is one breaking change in the API around color and text style properties but I don't think you're using them at the moment. The main highlight of this version though is that we now support subtrees. DVUI may want to take advantage of this in the future to produce its AccessKit subtree and let an embedding application manage it (e.g. DVUI UI inside of a game engine). You might also want to do the opposite and manage a subtree produced by a web view or other complex widgets that DVUI would host.

phatchman:
The advanced drop down doesn't work very well in the demo. The issue is that when there is no selected index, there is no focused widget id. (And the focused subwindow is not in the accessibility tree). This results in the accessibility focus defaulting to the main window and there is no way to move it from the main window to anything else using tab.

I'm proposing to change this so that if nothing is selected when dropped-down, we focus the first item. This is essentially what happens now if you press the down-arrow after opening the drop-down. There is no way to return the drop-down to an unselected state, so the only functionality we lose is the ability to drop-down and close without selecting anything. 

But I'd suggest if we really want a "nothing selected" option for drop-downs we have a blank selectable entry as the first entry to represent the "not selected" state.( Whether that is fully user-implemented or dvui-assisted)

Any issues with that change? Essentially adding the else case below:

```
    if (self.drop_first_frame) {
        if (self.init_options.selected_index) |si| {
            if (si == self.drop_mi_index) {
                dvui.focusWidget(self.drop_mi.data().id, null, null);
                dvui.dataSet(null, self.data().id, "_drop_adjust", self.drop_height);
            }
        } else if (self.drop_mi_index == 0) {
            dvui.focusWidget(self.drop_mi.data().id, null, null);
        }
    }

```

As to other options, I don't think I can add the focused subwindow to the tree as the "drop down pane" itself is not focusable in dvui. (though it could be?)

There is a bigger fix needed as if I tab or shift-tab focus out of the drop-down, the focus is also lost. I can raise an issue for that if you like.

ethindp:
> If you're testing on Windows, this is a limitation of UI Automation. Not much we can do about it.

How do things like WinUI3 apps do it? Or Windows forms? Do they use accessibility infra that we just don't have access to?

ethindp:
> But I'd suggest if we really want a "nothing selected" option for drop-downs we have a blank selectable entry as the first entry to represent the "not selected" state.

I'm also pretty sure this is how web browsers and such do it.

phatchman:
> How do things like WinUI3 apps do it? Or Windows forms? Do they use accessibility infra that we just don't have access to?

My guess is it would use the aria properties? But accesskit (top) is setting multiline=true, just like chrome is.

image

>

image

phatchman:
Quite a few small changes, but added together I think they make a big improvement:

- Upgrade to accesskit-c 0.20.0
- Improve focus tracking to minimize focus being set to the main window
- Grid Widget now creates row nodes in the accessibility tree.
- FIX: Selection actions are now targeted to the correct widget.
- FIX: Word starts algorithm for text runs now handles word breaks correctly.
- Updated text layout example so that the caption is wholly included in the child text layout widget
- Floating windows now default to the .dialog role. This improves screen reader behaviour for dialogs.
- Toasts now announce their content when displayed.
- FIX: labelClick now supports the click action. This also enables clicking on URLs.
- Sliders now set a step and jump value.
- Drop down widgets now focus the first element if nothing has been selected.
- Windows now advertise the focus action. (note: focus action not yet fully implemented).
- Text Layout Widget now has a role of group with text contained in a label within the group.
  This allows other widgets to be embedded in the text area and be read correctly.
- FIX: Nesting of text layouts now produces the correct accessibiltiy tree. Fixes #674

I found the grid changes hard to test because of #690. But on the occasions, I did manage to get the reader's cursor into the grid, it did seemed to be able to navigate the columns and rows as expected. But definitely needs more testing.

Known issues:
1) The dx11 backend is broken with this upgrade. I need to look into that, but probably as a separate change.
2) Accessibility Insights reports the rect for the radio group in basic widgets demo extends from the radio group to the bottom of the window. The correct rect is set on the AccessKit node.

@david-vanderson No rush if you are busy. I'm going to take a break from the accessibility stuff after doing a bit more testing, updating the readme etc. So you aren't holding me up or anything. I'll try and do some Mac and Linux testing over the next couple of days.

Edit: Is there a preference for using overlay widget over virtual parent widget? I've used both here, overlay when I pass a rect and virtual parent when I want to have the parent's content rect.

phatchman:
> How do things like WinUI3 apps do it? Or Windows forms? Do they use accessibility infra that we just don't have access to?

@ethindp do you have an example of an application on Windows where nvda detects a multi-line edit. Or does it always do it?

I've got the win sdk accessibility tools instaled now, so I can dump all the attributes in both applications to see what is different.

david-vanderson:
> Any issues with that change? Essentially adding the else case below:

Nope, sounds good!

> Quite a few small changes, but added together I think they make a big improvement:

No kidding, this is fantastic!  Really appreciate all the detailed work here - thank you!

> No rush if you are busy.

Thanks, I've gotten a bit behind recently.  My plan here is to look at the code, and do some spot testing myself, but not hold this up more than that.

> Edit: Is there a preference for using overlay widget over virtual parent widget? I've used both here, overlay when I pass a rect and virtual parent when I want to have the parent's content rect.

Not really - I haven't thought that much about it.  Overlay is the simplest, so I usually use that unless something else is needed - you can pass `.rect = .{}, .expand = .both` to get the parent's content rect.  VirtualParent excludes itself from layout, which might be exactly what you want.  If it works, go for it.

ethindp:
@phatchman I mean, just look at Notepad, or VS code, or really anything that lets you enter multiple lines of text. Hell a simple WXWidgets app would would it.

phatchman:
> @phatchman I mean, just look at Notepad, or VS code, or really anything that lets you enter multiple lines of text. Hell a simple WXWidgets app would would it.

It tuns out that Notepad, VSCode etc, don't present themselves as multi-line text edits. They are dcouments, which support the text replacement action.  I've been  struggling to find something outside of web pages that actually has a true multi-line text entry. 

Even the multi-line text edit I did with python WX presents itself as a document.

I guess my only other question would be does NVDA detect a web page with a multi-line edit as a multi-line edit on windows? Because browsers do present them as `edit` instead of `multiline_edit` on Windows, same as DVUI.

I'll add that as something to look at in a future update. Maybe that's the reason why native apps don't present them as edits?

Edit: OK I found one. If you have Word, on a blank document do File->Info and the comment box is a true multi-line edit control. No Aria info is set, so my assumption is that NVDA won't detect that as a multi-line edit either?

DataTriny:
Windows Forms probably still use the older MSAA API, which has a dedicated state to represent multiline text inputs. The "recommended" way to expose a multiline text field through UI Automation is, like we do, to set an ARIA property, according to Microsoft's documentation. But NVDA explicitly ignores it, as I found while I was working on this.

As @phatchman found out, the only way to have a widget report as multiline on Windows nowadays is by using `UIA_DocumentControlTypeId` which maps to among other roles, `ACCESSKIT_ROLE_DOCUMENT`. But as I said elsewhere we have not yet implemented this control pattern so I would be surprized if using this role would result in something usable right now. Please just be patient.

ethindp:
Wow, okay, I didn't know that at all. I haven't looked too deeply into the accessibility properties of individual apps but I might (like, seriously) because this is super fascinating.

As for NVDA ignoring the multi-line UIA attribute, sounds like an NVDA bug to me. In theory it should be trivial to fix (since it really is just a speech change and NVDA doesn't actually have a problem interacting with such fields) and the bug is just that NVDA doesn't indicate the status properly. We could probably submit a PR upstream.

ethindp:
Well, in theory it would be easy to fix... Not in practice. Lots of abstraction from what I can see.

DataTriny:
@ethindp when I say "NVDA explicitly ignores it" I mean that I found proof in the source code. This is a deliberate choice:

https://github.com/nvaccess/nvda/blob/aa95ea8e798ef875521d9ea9874d7ff9f66e05b4/source/NVDAObjects/UIA/__init__.py#L608-L612

Feel free to open an issue or submit a pull request though.

ethindp:
@DataTriny I was trying to find that and couldn't, thanks for pointing that out. I will open a PR reverting this change because it certainly creates bug-looking behavior from our side. And I suspect people will get really confused when NVDA just says "edit blank" without a multi-line indicator but the edit field behaves like a multi-line edit box when NVDA claims it's a single-line edit.

ethindp:
@DataTriny I removed all four of those lines and NVDA still does not see the fields as multi-line. Ran it from source too. My PR is PR 19502 if you'd like to see my exact changes. So I'm guessing there's something else I need to change somewhere.

david-vanderson (approved):
Looking great!  I will do some testing, but you can merge this when you are ready.  Thank you!!

phatchman:
> Looking great! I will do some testing, but you can merge this when you are ready. Thank you!!

I think we regressed with raylib on MacOS not compiling? The SDL examples compile OK on mac, but I can't test them.

david-vanderson:
> I think we regressed with raylib on MacOS not compiling? The SDL examples compile OK on mac, but I can't test them.

On MacOS I'm able to compile the raylib examples with accesskit - are you getting an error?

For the raylib and sdl3 examples on Mac, I've never tried to test them before.  I turned on "VoiceOver" in the settings, but it only reads the window description, none of the dvui widget stuff.  I'm not sure if I'm doing something wrong, but I get the same behavior before and after this PR.

PR #679, opened 2026-01-16
Examples: Add view_source button to show each example's source code

phatchman:
I got distracted from fixing accessibility issues and the tree sitter API is far too easy to use not to try this.

The `tree_sitter_zig_queries.scm` should come from the dependency, but I couldn't see any nice way to do that. There was an existing sample of this file in `examples`, so I just moved it to `src\Examples` instead.

- Display source code with treesitter if supported
- Otherwise open in browser

I'm pretty bad at colours, so if someone with better taste wants to improve the highlighting colours, please feel free.
It would be nice to have a read-only mode on the text entry widget that handles selection, copy/paste etc, but not modification. Even if it was a different processEvents that just did a sub-set of the events? But it's probably not generally useful as without the tree sitter requirement, I'd just use a text layout widget instead.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1002,6 +1002,18 @@ pub fn addDvuiModule(
                 .flags = &.{"-std=c11"},
             });
         }
+        const tree_sitter_zig_dep = b.lazyDependency("tree_sitter_zig", .{
+            .target = target,
+            .optimize = optimize,
+        });
+        if (tree_sitter_zig_dep) |tsc| {
+            dvui_mod.addIncludePath(tsc.path("src"));
+            dvui_mod.addCSourceFiles(.{
+                .root = tsc.path(""),
+                .files = &.{"src/parser.c"},
+                .flags = &.{"-std=c11"},
+            });
+        }
     }
 
     return dvui_mod;
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -69,6 +69,10 @@
             .hash = "N-V-__8AAHNaAgD6kjK3HX6lAf89chPqS5zIm0tLpvBQLOtX",
             .lazy = true,
         },
+        .tree_sitter_zig = .{
+            .url = "git+https://github.com/tree-sitter-grammars/tree-sitter-zig.git#6479aa13f32f701c383083d8b28360ebd682fb7d",
+            .hash = "tree_sitter_zig-1.1.2-AAAAAJ57XACsD_XBHC4i6G6rnCsZ1dZj7CLRTC09toEB",
+        },
         .zgl = .{
             .url = "git+https://github.com/ziglibs/zgl#6ee54c287f3c4f176bd05a92746828bb791a1ad0",
             .hash = "zgl-1.1.0-p_NpAJNECgCzUGx0aF5KFsmD5VOwfw8LaOsN0fcoU2bi",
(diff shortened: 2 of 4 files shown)


david-vanderson:
> I got distracted from fixing accessibility issues and the tree sitter API is far too easy to use not to try this.

Hahaha!  This is cool!  Give me a bit to look at the details.

> 
> The `tree_sitter_zig_queries.scm` should come from the dependency, but I couldn't see any nice way to do that. There was an existing sample of this file in `examples`, so I just moved it to `src\Examples` instead.
> 
>     * Display source code with treesitter if supported
> 
>     * Otherwise open in browser

That is a good idea that I had not thought of.  Makes sense at least until we get tree-sitter working on the web.

> I'm pretty bad at colours, so if someone with better taste wants to improve the highlighting colours, please feel free.

Me too.

> It would be nice to have a read-only mode on the text entry widget that handles selection, copy/paste etc, but not modification. Even if it was a different processEvents that just did a sub-set of the events? But it's probably not generally useful as without the tree sitter requirement, I'd just use a text layout widget instead.

This is a good point.  I was initially thinking that I'd add tree-sitter support to TextLayoutWidget for a read-only usecase when it came up.  Maybe the time is already here - I'll try it.

eamonburns:
> The tree_sitter_zig_queries.scm should come from the dependency, but I couldn't see any nice way to do that

@phatchman - You can import arbitrary files using `@embedFile` by creating an anonymous import:

```zig
// build.zig
exe.root_module.addAnonymousImport("dep_file", .{
    .root_source_file = my_dep.path("file/in/dependency.txt"),
});
```
```zig
// main.zig
const dep_file = @embedFile("dep_file");
```

Reference: https://ziglang.org/learn/build-system/#embed-file

phatchman:
@eamonburns From memory I also have to generate the file as it's not just a straight copy. But this helps!

Thanks.

phatchman:
1) I've rebased this to main 
2) I had the very simple idea to call te.textLayout.processEvents() on the text entry, which gives us the same functionality as if we were using a text layout widget that supported tree sitter. So, we get selection, copy, paste, focus etc.

@david-vanderson What did you want to do with this one? Happy to close or modify if it's not what you want for this.

Cheers.

Edit: I was just playing with this and looking at "Basic Widgets". It seems to work quite well for that as it is a lot simpler than some of the other examples. if you expand the window and put them side by side it seems quite easy to follow. A search box would be handy though.

david-vanderson:
> 1. I've rebased this to main
> 
>     2. I had the very simple idea to call te.textLayout.processEvents() on the text entry, which gives us the same functionality as if we were using a text layout widget that supported tree sitter. So, we get selection, copy, paste, focus etc.

That is clever - well done!

> 
> 
> @david-vanderson What did you want to do with this one? Happy to close or modify if it's not what you want for this.

This is awesome.  Sorry I sat on it.  Can you please make the view source conditional on calling `demo()` with `.full`?  

Then merge when ready.  Thank you!

phatchman:
No problem. Changed to only include for full demo.

phatchman:
The grid widget view source seems quite slow? Even scrolling it. 
I'll profile it later.

image

image

david-vanderson:
> The grid widget view source seems quite slow? Even scrolling it. I'll profile it later.

Does it matter what part you are looking at?  For me there seems to be two places (in Debug mode) where it gets slow, and both are literal slices:
* the `all_cars` list (about 1/4 of the way down)
* the `directory_examples` list (close to the bottom)

I wonder if the zig tree sitter grammar has some weird corners?

PR #680, opened 2026-01-16
Zig16 progress

kristoff-it:
Made some progress, most notably I've made forks of:

- raylib
- raylib-zig

I would imagine raylib (and maybe also raylib-zig) has no interest in keeping up with 0.16-dev so I would recommend at some point making your own fork.

This commit still doesn't compile but does reduce the list of compile errors to:

- some places where there's a missing `io` argument. I haven't fixed those because it wasn't immediately obvious what would be the right way to pass the argument around (see the fix to `cacheBuster.zig` for an example of how to use the new style of main function, which is required to access environment variables, among other things)
- a bajillion cImport errors, which seem to point at header files in zig's library. @Vexu, would you be able to diagnose what the problem could be?

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -21,8 +21,8 @@
             .lazy = true,
         },
         .raylib = .{
-            .url = "git+https://github.com/raysan5/raylib?ref=master#720dd22491cdc4a29ec0ea87b572d9380872198c",
-            .hash = "raylib-5.6.0-dev-whq8uMVQCgVTgRxG3REWgBBSi2BR-nf4bq6eqANxImt7",
+            .url = "git+https://github.com/kristoff-it/raylib?ref=zig-0.16#68eff3f612091d9e713d2229b4f840582dbfc28f",
+            .hash = "raylib-5.6.0-dev-whq8uLAJDQXCUoWbI50KR2RX8f1NhyzmklCim1bT2EH2",
             .lazy = true,
         },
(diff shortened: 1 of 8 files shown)


Vexu:
> * a bajillion cImport errors, which seem to point at header files in zig's library. Vexu, would you be able to diagnose what the problem could be?

How do I reproduce the errors?

kristoff-it:
`zig build` on this branch. If you don't see errors on your machine try targeting macOS.

Vexu:
I'm only seeing `std.Io` related errors when targetting Linux. I'm not sure how to make targeting macOS work, simply pointing `--sysroot` at https://github.com/hexops/xcode-frameworks didn't work.

david-vanderson:
I had some weird behavior regarding this - I was trying to disable the raylib backends for the moment.  They are still listed in build.zig.zon, but not referenced in build.zig.

But that wasn't quite enough - I had to blow my zig caches to get it to work.  @kristoff-it does blowing your caches help?

david-vanderson:
Note that I'm only trying to get `zig build sdl3-app` working for now.

kristoff-it:
I can confirm clearing the cache removes all the cImport failures, sorry for the false alarm!

david-vanderson:
> I can confirm clearing the cache removes all the cImport failures, sorry for the false alarm!

Phew good - thanks for the changes, will need them in any case!

PR #692, opened 2026-01-25
export Dvui class and expose types for constructor and methods and remove post install script

MichaelBelousov:
follow up to #672.
This now works in Graphl after exposing a few more things.

- expose `Dvui` class for consumers (like Graphl) that need custom initialization (e.g. custom wasm imports and exports)
- expose more types and methods while we're here
- this allowed graphl to delete 1400 lines of our custom web backend patch!


david-vanderson:
Awesome thank you!

PR #693, opened 2026-01-27
Fix tab ordering on grid demos and improve accessibility

### LINKED ISSUE #690: GridWidget: Focus not working correctly
## Description ##
It is not possible to focus the GridWidget or anything in the grid widget using tab navigation. 
This causes issues with:
1) Navigating the grid and select rows solely with keyboard
2) Screen readers can't easily "enter" the grid to start reading the content.
3) The user can't change the sort order using keyboard as they can't tab into the header.

## Repro ##
1) Open the Grid -> Layouts and Data Demo
2) Attempt to select something using keyboard
3) Attempt to tab into the grid to scroll via keyboard.

## Expected ##

Using windows explorer as an example:
1) Tab should tab into the grid body. Then navigation within the grid is via keyboard navigation (up/down keys).
2) Second tab should tab into the header to allow the user to navigate with arrow keys (left/right) to move withing the headers.

But I think the first step is to understand why the grid is not getting added to the tab order and so never receives focus.

phatchman:
- Pass tab order to internal widget tab groups
- Tab groups no longer include themselves in the tab index if they don't contain any tab-able widgets
- Grid headers are now in a focus group
- Sortable grid headers now display focus rings
- Sortable grid headers now retain focus when sort changes.
- Improve tab order on grid styling demo.
- Grid cells are read correctly.

Known Issues:
- When reading the grid, the reader needs to be moved a second time to the right on the first cell in any row to read the cell label. Moving again to the right reads the label from the next column. It is only when moving to a new row that I see the issue.
- AccessKit appears to make grid cells selectable, even if they are not. The reade rreads "not selected" for each cell in addition to the text.

Fixes #690

I still need to fix the tab order for the remaining grid demos. And back out some speculative changes to them.

### DIFF
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -3430,8 +3430,8 @@ pub fn gridHeadingSortable(
     const sort_changed = switch (g.colSortOrder(col_num)) {
         // Use same src for each button so they get the same id and can retain focus accross frames.
         .unsorted => button(src, heading, .{}, heading_opts),
-        .ascending => buttonLabelAndIcon(src, .{ .label = heading, .tvg_bytes = icon_ascending, .button_opts = .{} }, heading_opts),
-        .descending => buttonLabelAndIcon(src, .{ .label = heading, .tvg_bytes = icon_descending, .button_opts = .{} }, heading_opts),
+        .ascending => buttonLabelAndIcon(src, .{ .label = heading, .icon_label = "sorted ascending", .tvg_bytes = icon_ascending, .button_opts = .{} }, heading_opts),
+        .descending => buttonLabelAndIcon(src, .{ .label = heading, .icon_label = "sorted descending", .tvg_bytes = icon_descending, .button_opts = .{} }, heading_opts),
     };
 
     if (sort_changed) {
@@ -4002,6 +4002,8 @@ pub const ButtonLabelAndIconOptions = struct {
     label: []const u8,
     tvg_bytes: []const u8,
     icon_first: bool = false,
+    // Best practice is to supply an icon label for accessibility.
+    icon_label: ?[]const u8 = null,
 };
 
 pub fn buttonLabelAndIcon(src: std.builtin.SourceLocation, combined_opts: ButtonLabelAndIconOptions, opts: Options) bool {
@@ -4018,7 +4020,7 @@ pub fn buttonLabelAndIcon(src: std.builtin.SourceLocation, combined_opts: Button
     {
         var outer_hbox = box(src, .{ .dir = .horizontal }, .{ .expand = .horizontal });
         defer outer_hbox.deinit();
-        icon(@src(), combined_opts.label, combined_opts.tvg_bytes, .{}, options.strip().override(.{ .gravity_x = if (combined_opts.icon_first) 0.0 else 1.0, .color_text = opts.color_text }));
+        icon(@src(), combined_opts.icon_label orelse combined_opts.label, combined_opts.tvg_bytes, .{}, options.strip().override(.{ .gravity_x = if (combined_opts.icon_first) 0.0 else 1.0, .color_text = opts.color_text }));
         labelEx(@src(), "{s}", .{combined_opts.label}, .{ .align_x = 0.5 }, options.strip().override(.{ .expand = .both }));
     }
 
--- a/src/selection.zig
+++ b/src/selection.zig
@@ -1,7 +1,7 @@
 //! Helpers for Multi and Single selection
 //! Implements an event queue for selections.
 //! Selectable objects can optionally raise selection events which
-//! are the processed by these helpers
+//! are processed by these helpers
 
 const std = @import("std");
 const dvui = @import("dvui.zig");
--- a/src/widgets/GridWidget.zig
+++ b/src/widgets/GridWidget.zig
@@ -170,8 +170,10 @@ pub const default_col_width: f32 = 100;
 
 //Widgets
 vbox: BoxWidget,
-/// SAFETY: Set by `bodyScrollContainerCreate`, is valid when `bscroll` is non-null
-group: dvui.FocusGroupWidget,
+/// SAFETY: Set by `headerScrollAreaCreate`, is valid when `hscroll` is non-null
+header_group: dvui.FocusGroupWidget,
+/// SAFETY: Set by `bodyScrollAreaCreate`, is valid when `bscroll` is non-null
+body_group: dvui.FocusGroupWidget,
 scroll: ScrollAreaWidget, // main scroll area
 hscroll: ?ScrollAreaWidget = null, // header scroll area
 bscroll: ?ScrollContainerWidget = null, // body scroll container
@@ -225,7 +227,8 @@ pub fn init(self: *GridWidget, src: std.builtin.SourceLocation, cols: WidthsOrNu
 
         // SAFETY: Widgets set bellow
         .vbox = undefined,
-        .group = undefined,
+        .header_group = undefined,
+        .body_group = undefined,
         .scroll = undefined,
     };
 
@@ -355,6 +358,7 @@ pub fn deinit(self: *GridWidget) void {
         !std.math.approxEqAbs(f32, self.header_height, self.last_header_height, 0.01);
 
     if (self.hscroll) |*hscroll| {
+        self.header_group.deinit();
         hscroll.deinit();
     }
 
@@ -365,7 +369,7 @@ pub fn deinit(self: *GridWidget) void {
     _ = dvui.spacer(@src(), .{ .min_size_content = this_size, .background = false });
 
     if (self.bscroll) |*bscroll| {
-        self.group.deinit();
+        self.body_group.deinit();
         bscroll.deinit();
     }
     self.scroll.deinit();
@@ -663,12 +667,14 @@ fn headerScrollAreaCreate(self: *GridWidget) void {
         if (!std.math.approxEqAbs(f32, self.header_height, self.last_header_height, 0.01)) {
             self.resizing = true;
         }
+        self.header_group.init(@src(), .{ .nav_key_dir = .horizontal }, .{ .tab_index = self.data().options.tab_index });
     }
 }
 
 fn bodyScrollContainerCreate(self: *GridWidget) void {
     // Finished with headers.
     if (self.hscroll) |*hscroll| {
+        self.header_group.deinit();
         hscroll.deinit();
         self.hscroll = null;
     }
@@ -687,7 +693,7 @@ fn bodyScrollContainerCreate(self: *GridWidget) void {
         self.bscroll.?.processEvents();
         self.bscroll.?.processVelocity();
 
-        self.group.init(@src(), .{ .nav_key_dir = .vertical }, .{});
+        self.body_group.init(@src(), .{ .nav_key_dir = .vertical }, .{ .tab_index = self.data().options.tab_index });
     }
 }
 
--- a/src/widgets/TabsWidget.zig
+++ b/src/widgets/TabsWidget.zig
@@ -44,7 +44,7 @@ pub fn init(self: *TabsWidget, src: std.builtin.SourceLocation, init_opts: InitO
 
     self.scroll.init(src, scroll_opts, defaults.override(opts));
 
-    self.group.init(@src(), .{ .nav_key_dir = self.init_options.dir }, .{});
+    self.group.init(@src(), .{ .nav_key_dir = self.init_options.dir }, .{ .tab_index = opts.tab_index });
 
     const margin: Rect = switch (self.init_options.dir) {
         .horizontal => .{ .y = 2 },


phatchman:
To make the demo work, I needed to:

1) Set the tab index of anything created in the main part of the Example window to 1. (This is only the back to demos button)
2) Set the tab index of the "tabs widget to 2.
3) Set the grid tab index to 3.
4) Everything else is left as null.

The only issue this creates is that the "close" button is the first of the "null" tab indexes, so is focused before the right-hand panel. I'd need to also add tab index of everything in the right-panel as well. Do you think we should also be able to set the tab order of the dialog header controls?

Adding a focus group for the grid header removes the need to specify a tab order for the headers. I did find that when creating a focus group inside a widget we should pass the tab_index from the parent's options to make sure the group gets the correct tab order.

The slightly weird thing is that if you don't set any tab order, the window close button will be first in the tab order. But if you do set tab indexes for everything, it will be last. Maybe it should always be last? I tried a few windows apps and the window/dialog controls don't appear in the tab order at all. But I don't think that is great either.

phatchman:
I need to rework this PR.

The main non-accesskit issue that came up is that the grid demos often render the right-side of a screen before the left side for layout reasons. But this means all the right-side widgets get a tab-order before the left-side, forcing us to set all the tab_index's manually.

Seems like this would be a common issue. 

The only thing I was thinking (and happy to experiment with it) is having a "tab group". Everything inside a tag_group of 0, would get a sequential tab_index, and they all would have a tab_index less than any widgets in tab_group 1. 

I could then put the left-side of the screen's widgets in tab_group 0, the right-side in tab_group 1 and would not have to worry about setting individual tab indexes on each widget.

I doubt we'd use different tab_indexes in the implementation, rather it would probably look at tab_groups first and then tab_indexes within each group or similar. Or maybe we make a TabGroupWidget like you did for the FocusWidget, so it handles the tabbing withing it.

Thought?

david-vanderson:
Yes we are thinking along the same lines.  I've also been thinking of having TabGroupWidget (TabIndexGroupWidget?) similar to FocusGroupWidget that does tab ordering all locally inside.

Seems like a promising solution because it makes it a local change.  If there are two buttons you want to swap the focus order for, you wrap them in a tab group and give them reversed tab_index values.

There are two things I haven't figured out yet:
* this would have to be fully recursive? (FocusGroupWidget only does a single layer)
  * if dvui.Window catches an unhandled tab, then it has to be able to look up the "next" widget
  * focusGroup is okay because it always puts itself into the global list
  * I guess maybe we could say no here and see if anybody actually needs a tabGroup inside another tabGroup
* tabbing backwards?
  * when focus moves to a focusGroup, it knows internally which widget should get focus
  * but for tabGroup it also needs to know if focus is moving forwards or backwards
  * but it doesn't get a chance to run until the next frame, so we'd have to store this somewhere
  * maybe we could store the last tabIndex movement direction (either .prev or .next)?

david-vanderson:
Alternate idea:
* change tab index from u16 to f32
* make tabGroup something that divides and offsets its children's tab indexes (in a way that nests well)

For example, if we have tab indexes:
* 1 - normal widget
* 2 - tab group
  *  4, 5, 6 widgets inside the tab group
* 3 - normal widget

Then the actual tab indexes recorded would be something like 1, 2.4, 2.5, 2.6, 3

I think this could be done with nested tab groups, and it has the simplicity of maintaining a single global tab ordering.

Thoughts?

Also somewhat separate - we should have documented well-known tab index values for standard things like
* floatingWindow close button
* dialog ok/cancel buttons
* this is not needed for this effort but I'm thinking through how much of the problem is trying to insert tab index values into a global order that is currently unspecified

phatchman:
> * this would have to be fully recursive? (FocusGroupWidget only does a single layer)

I think so. Those swapped buttons could easily be in a right-panel, so you have 2 levels just there.
But maybe not "fully". I'm sure there are exceptions, but it is hard to image nesting layouts more than 3 or so levels. So maybe we can get away with having a reasonable maximum if that makes some solutions viable.
   
>   * maybe we could store the last tabIndex movement direction (either .prev or .next)?

Sounds a reasonable to me.

> Alternate idea:
> 
> change tab index from u16 to f32
> make tabGroup something that divides and offsets its children's tab indexes (in a way that nests well)

I'm struggling on the "in a way that nests well". Wouldn't we have to either:
1) Choose a small divisor, which means limiting the number of widgets that can be in a tab index group
2) Choose a large divisor and limit the number of nesting levels

Otherwise, we'll run into fp precision issues.

e.g. for f32 if we had a divisor of 1000, I think we could only support 2-3 levels (depending on if the first level has numbers > ~10) before we run out of unique fp numbers (i.e. f32 epsilon is ~1.19e7)

david-vanderson:
> Otherwise, we'll run into fp precision issues.
> 
> e.g. for f32 if we had a divisor of 1000, I think we could only support 2-3 levels (depending on if the first level has numbers > ~10) before we run out of unique fp numbers (i.e. f32 epsilon is ~1.19e7)

Good point, I was imagining the tabGroup picks a power of two larger than the number of children it has:
* first tab group has 10 children, picks divisor 16
* second nested tab group has 7 children, picks 8 (meaning total divisor is now 16*8 = 128)
* third nested tab group has 3 children, picks 4 (total divisor now 512)

I think this should give us plenty of room?

phatchman:
> Also somewhat separate - we should have documented well-known tab index values for standard things like

I think we need to do this too. I already ran into the issue in this PR that if you provide even 1 tab_index, the close button will get a tab_order after any widgets with a tab index, but before any widgets without a tab index. So, it forces you to index everything if you index 1 thing. 

And even if you do that, the close button ends up at the end of the tab_index, rather than the start.

I've not thought this through yet, but wondering if there is a way that widgets can inherit their tab_index from the parent. Then I could have a right-box created first, but with a tab_index of 2, a left box with a tab index of 1, and then anything inside the box gets the tab_index of the parent box, and so gets a tab order based on creation order within the box.

If I also wanted to have control over the tab index of children in the boxes, I can create the boxes with a tab index of e.g. 10 and 20. Then the buttons in the box would have their tab_index manually set to 22 and 21 to give them a different index to their creation order.

This would allow you to arbitrarily nest tab indexes, with the trade-off that if you want to make the tab order different to creation order, you will need to specify a tab_index for all widgets in that container.

Otherwise instead of allowing inheritance of tab_index, that is what the TabIndexGroupWidget could essentially do, set the tab_index of its children based on its tab index and creation order. (though I can't think how just yet). This would solve the arbitrary nesting issue as we just make that the user's problem to make the gap between the box / TabIndexGroupWidget tab_indexes big enough to accommodate the number of widgets that need a unique tab_index.

phatchman:
My little POC of the above looks promising.

The simplest implementation is that `dvui.tabIndexSet(self.data().id, self.data().options.tab_index)` takes a self.data() instead of self.data().id and if tab_index is null, it gets a tab_index from the first parent with a non-null tab_index.

Then in the grid demo I set the right-hand box to tab_index 20, and the left_hand grid to 10 it all worked as expected.

But I don't think that is the correct implementation of the idea. I think I'd rather the shared tab index is set on a specific widget e.g. the TabIndexGroupWidget, instead of having the 'magic' automatic inheritance via options. 

I think it is basically the same as what you are proposing by making it f32, but instead of trying to work out how to partition the available numbers, we make that the user's problem. And it gives them more flexibility as maybe they want the focus to go from innermost level 4 to 2 to 3 to 1 for some weird reason. And they could do that just by picking the appropriate numbers for the group.


```
    // Needs to not inherit when in a FocusGroupWidget. Need to add some "inherit_focus_from_parent: bool" param or similar.
    var ti = TabIndex{ .windowId = cw.subwindows.current_id, .widgetId = widget_id, .tabIndex = (tab_index orelse parent: {
        var iter = wd.parent.data().iterator();
        while (iter.next()) |pwd| {
            if (pwd.options.tab_index != null) break :parent pwd.options.tab_index;
        }
        break :parent null;
    } orelse math.maxInt(u16)) };
```

It does also solve the window close button issue. We could set the floating window pane to have a tab index of 100 and then everything in the window would have an index of 100 by default. The window header would be able to use any value less than 100.

phatchman:
I think this idea will work? Documenting in case I don't get time to try it out today:
On init, our TabIndexGroupWidget records the length og currentWindow().tabIndex.
On deinit it looks for all entries added since it was created with the default (max u16) value and sets them to the tab index of the tab group widget.

This should work for nested widgets as well as they would have already set the inner values before the outer deinit() runs.

What I fear is that this doesn't set the tab indexes early enough. The normal tab processing is handled in Window.end(), but it is also handled in other event loops like text entry.

david-vanderson:
> What I fear is that this doesn't set the tab indexes early enough. The normal tab processing is handled in Window.end(), but it is also handled in other event loops like text entry.

I'm still digesting all of this, but this specifically shouldn't be a problem - all tab processing works off last frame's data.

We definitely have something here.  Let me write out some usecases that seem useful, maybe we don't end up supporting all of them.
* put widgets in tab order before/between/after common dialog elements (close, ok/cancel)
* have a bunch of widgets in default order, want to locally swap 2 keeping the pair in the same place
* have a bunch of widgets with specified tab indexes, want to wrap the whole group and move it with respect to something else
* have a bunch of widgets with specified tab indexes, want to replace one widget with a set of widgets with tab indexes without having to change all the rest

This suggests that tab indexes are no longer a global thing, but a local ordering within a tab group.  Not totally sure that's the right call.  You are right that would prevent breaking out of the local ordering, and also maybe it's more complicated than necessary and we should push it back to the user code.

If we can have a solution that assigns global tab indexes I think it will work better than having nested tab stuff running inside each tab group (the way focus group does it).  If we have to run the tab group to figure out where the focus moves it prevents focus from moving multiple times per frame.  Also more difficult to test and understand.

One thought is that our tab index range is currently much larger than needed.  Maybe we shrink it to like (1-10) - so at each nesting level you can have 9 independent orderings (of course most of the time you'll have many widgets with the same number, usually the default highest).  Or maybe we should make the default be in the middle?

Then if we bump tab index to u32 we can have a max of 9 levels, which seems plenty?

So concretely:
* tabGroup tracks the nesting level, starting at 1e9
* each nested tabGroup divides the nesting level by 10
* a widget's tab index is added to the current nesting level

Oh wait - separate nested groups can step on each other's toes.  If you have two sets of widgets that are wrapped a the same nesting level, then the focus might jump between the two sets.  Argh.

Will continue to think about it.

phatchman:
This [code ](https://github.com/david-vanderson/dvui/compare/main...phatchman:dvui-dev:focus_groups) doesn't quite work (and I've not really looked into why), but demonstrates what I was thinking.

Find all of the tab indexes that are not set since the tab index group was created and set those tab indexes to the tab index of the tab group. (Note the current code doesn't do this, it goes and sets everything with 65535 to the group's tab index).

This should end up with the left-panel buttons getting a tab-index of 10, the "middle right" buttons 20, the top-right "30" and the "bottom right" 40, with the focus of the 2 buttons in the top-right reversed because I manually specify a tab-index within the groups "range".

It doesn't quite do that, but did when I was using the "inheritance" version (The one where I changes SetTabIndex to find the tab index from any parent). So, I'm not sure if this is viable, but it seems very simple and I think should do everything we need?

There are also many other ways to do this, e.g. keeping a stack of TabIndexGroups and doing it during tabIndexSet. Similar to the FocusGroup, but we'd need a stack.

david-vanderson:
Okay, maybe we directly represent the tab groups in the tab index data structure:
* store both given tab index and Id of current tab group
* placeholder entry for where a tab group is in its parent group
* when moving "down into" a nested tab group, you know which way you are going, can recur properly
* when hitting the "edge" of a tab group, go find the placeholder entry for that group, set the parent, and continue

This makes the logic for moving through the tab index data structure more complicated, but makes reasoning about tab groups easy.  Everything in the tab group is logically "inside" the tab index of the group itself.

It does rule out your example of busting through the group hierarchy.  In that case we'd have to say to remove all the tab groups, because they are trying to do some kind of global tab index thing.

This seems the most promising to me so far.  How does it sound to you?

david-vanderson:
> This [code ](https://github.com/david-vanderson/dvui/compare/main...phatchman:dvui-dev:focus_groups) doesn't quite work (and I've not really looked into why), but demonstrates what I was thinking.

Whoops we wrote stuff at the same time.

It seems like we have a choice between a strategy that pushes managing the tab index values more onto the user code (with some helpers) vs. a strategy where we manage them more (might become a bit opaque as to what's going on).

I honestly am unsure what approach to shoot for.  One thing I think we've both found is that wrapping groups of widgets into functions works really well.  Does that imply something here?  Maybe it means a solution where you don't have to think about the tab index values of widgets inside the function is better?

phatchman:
This [code ](https://github.com/david-vanderson/dvui/compare/main...phatchman:dvui-dev:focus_groups) is a cleaned up, working version of that idea. Pushing the tab_index_group to the window, so that it follows the normal logic for setting a tab group.

Basically the only change is:
```
 var ti = TabIndex{ .windowId = cw.subwindows.current_id, .widgetId = widget_id, .tabIndex = (tab_index orelse cw.tab_group_index orelse math.maxInt(u16)) };
 ```
 
I like this primarily because it is simple and flexible and the tab order stays a linear sequence.

The hierarchy version of this wouldn't look too different. I'd just store the tab_group in the TabIndex struct and search by tab_group prior to tab_index. Happy to make a version of that for comparion if you like.

I guess my question is does dvui need to know the tab structure? What would it help us do vs the current linear implementation? From the user perspective there's a minor difference in which numbers are used, but i think keeping it linear makes it easier to follow? If I see 31 as a tab index, i know that comes after tab index 20. vs seeing a tab index of 1 and then having to check that the parent tab group is 3.

Yeah agree this wrapping approach works for lots of things. Also, the pattern I copied here of "widgets" building their own stacks, by saving and restoring previous values is really effective.

Edit: If you agree this is the way to go, I'm not sure where TabIndexGroup should live? I'm thinking it just goes in dvui.zig? It's not really related to anything else and I don't think it deserves its own file.

Edit2: Now working for focus groups. The two buttons in a hbox are in a focus group.

phatchman:
[This](https://github.com/david-vanderson/dvui/compare/main...phatchman:dvui-dev:tab_groups) shows how it would work for the grid demo, taking into account focus groups.

I arbitrarily picked an index of 5 for the window header tab index, but not convinced that is the best way to do it yet. They could wrap the window header in a focus group, pass the tab index as an option etc. But I think we need to see what dialogs look like first as there are 2 sets of indexes to set there.

This maybe shows the limitation of the single level tab index in that we "arbitrarily" pick 10 for the main examples window and every sub-widget needs to know that and pick a higher number. But you sort of have to do that with the 2-level tab_index solution. You are still picking arbitrary numbers for the tab_groups, it is just that this only matters for ordering between your siblings, not globally, so is a little easier to manage in a complicated layout across lots of files / functions, I suspect). 

So, I think either the single or 2 levels works. Single is just conceptually simpler in that everything has a single number.

As I said, happy to do one with the tab group stored in the TabIndex struct to give us the 2 levels if you want to compare.

david-vanderson:
Good catch on the focus group maybe not having any children.

I'm still unsure.  I like the simplicity of rewriting the default index values.  But I can't see how it addresses these cases:
* in a long list of default index widgets, we want to locally switch the order of 2
* a function wraps a bunch of widgets that have non-default indexes, how to order that whole function beside other widgets
  * this in particular feels important, could be a library function used in many projects, otherwise the tab index stuff would have to be communicated some other way

You're right that a recursive solution means the tab indexes are no longer comparable in source without knowing the tab group.  Maybe that would be okay?

I'll try a stab at the recursive solution and we can see how the code looks.

phatchman:
> * in a long list of default index widgets, we want to locally switch the order of 2

Not sure how this is different to to the recursive case. You put all the widgets before the custom ordering in 1 tab group, all the ones after it in another tab group and specify individual tab indexes for the current one.
I think you'd have to do the same in the recursive case, except the custom ordering would be inside a separate tab group.

> * a function wraps a bunch of widgets that have non-default indexes, how to order that whole function beside other widgets.

Options.tab_index can pass tab-index to that function which can then  be passed to a tab group if needed.
Or wrap the function in it's own tab group. 
>   
>   * this in particular feels important, could be a library function used in many projects, otherwise the tab index stuff would have to be communicated some other way

But, I agree non-recursive is much weaker here. Above doesn't cover if the function needs to create multiple internal tab groups. That would only be addressed by documenting that the function uses a particular range on top of the passed in tab index.
> 
> You're right that a recursive solution means the tab indexes are no longer comparable in source without knowing the tab group. Maybe that would be okay?

I don't think it's an issue.
> 
> I'll try a stab at the recursive solution and we can see how the code looks.

Recursive solution probably works better for things like grid cells where you just set yhe group id of the cell and then can index the widgets in the cell without having to know the parent group index.

Pretty sure recursive wins. We only lose the ability to have an ordering different to the tab group heirachy. But it's hard to think of a realistic use for that.

david-vanderson:
> Pretty sure recursive wins. We only lose the ability to have an ordering different to the tab group heirachy. But it's hard to think of a realistic use for that.

https://github.com/david-vanderson/dvui/pull/786

What do you think of that?  The implementation is a little complicated, but using it seems okay.  The standalone example changes (which I will revert if we move forward) really overuse it to exercise the various scenarios.

We can even nest a tabIndexGroup inside a focusGroup in case someone wants to have a complicated ordering there.

Reordering the grid examples (I only did 2) works pretty well?

phatchman:
Haven't been able to test yet, but looks good. I'll try it out tmr.

phatchman:
After the changes to include tab groups, there was only a handful of things left in this branch.

- Grid headers are now in a focus_group
- Pass tab order to the focus group in the TabsWidget
- Provide accessibility labels for icons in buttonLabelAndIcon()

Believe should be good to merge.

There is only one accessibility item we didn't address, which is navigating into a grid header or body that doesn't have any focusable elements. This is a reminder for me to go and work out what we were supposed to do in that situation and create an issue for it.

david-vanderson:
Great - thank you!

Issue #702, opened 2026-02-01
SDL3 GPU backend crashes with many graphical elements (u16 index buffer overflow)


Hcode00:
Hi! First, thanks for the great library - I've been using dvui for a desktop application and it's been working well.

I ran into a crash when rendering a lot of graphical elements at once. My app is a visualization tool that sometimes needs to display thousands of lines and circles on a canvas. When I have too many elements visible, the app panics with an integer overflow.

The crash happens in `src/backends/sdl3gpu.zig`:

```
thread 27699 panic: integer overflow
/home/hcode/Project/lib/dvui/src/backends/sdl3gpu.zig:297:41: 0x1ae6c4e in push (project)
            self.index.pushAssumeCap(id + vertex_start);
                                        ^
/home/hcode/Project/lib/dvui/src/backends/sdl3gpu.zig:1216:28: 0x1ae729b in drawClippedTriangles (project)
    self.frame_uploads.push(self, backendTexture, vtx, idx, clip) catch {
                           ^
/home/hcode/Project/lib/dvui/src/Backend.zig:69:42: 0x19cd653 in drawClippedTriangles (project)
    return self.impl.drawClippedTriangles(texture, vtx, idx, clipr);
                                         ^
/home/hcode/Project/lib/dvui/src/render.zig:100:40: 0x191e196 in renderTriangles (project)
    try cw.backend.drawClippedTriangles(tex, triangles.vertexes, triangles.indices, clipr);
                                       ^
/home/hcode/Project/lib/dvui/src/Path.zig:327:25: 0x19fe533 in stroke (project)
    dvui.renderTriangles(triangles, null) catch |err| {
                        ^
```

It looks like the vertex buffer length exceeds what a `u16` can hold (65,535). I traced through the code and it seems the index buffer is typed as `u16` in several places.

Is there a recommended way to handle this? Would switching to 32-bit indices be feasible, or is there a better approach?

david-vanderson:
Thank you!  Will comment in your pull request.

david-vanderson:
If you can, I'd love to see a screenshot of your app.  No pressure, just curious!

Hcode00:
> If you can, I'd love to see a screenshot of your app. No pressure, just curious!

just emailed you some screenshots.

PR #703, opened 2026-02-01
feat(sdl3gpu): support 32-bit index buffers

Hcode00:
## Summary
- add a build option for SDL3GPU index buffer bit size (16 or 32, default 16)
- route index buffer types through a backend `IndexType` alias
- bind the appropriate SDL GPU index element size when drawing

## Context
Fixes https://github.com/david-vanderson/dvui/issues/702 by allowing large render batches to use 32-bit indices while keeping 16-bit as the default.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -20,6 +20,11 @@ pub const AccesskitOptions = enum {
     }
 };
 
+pub const VertexIndex = enum {
+    u16,
+    u32,
+};
+
 const CommonSdl = struct {
     mod: *std.Build.Module,
     options: *std.Build.Step.Options,
@@ -116,6 +121,13 @@ pub fn build(b: *std.Build) !void {
         b.option(bool, "log-error-trace", "If error logs should include the error return trace (automatically enabled with log stack traces)"),
     );
 
+    const vertex_index = b.option(VertexIndex, "vertex-index", "Vertex index type (default u16)") orelse .u16;
+    build_options.addOption(
+        VertexIndex,
+        "vertex_index",
+        vertex_index,
+    );
+
     const accesskit = b.option(AccesskitOptions, "accesskit", "Build with AccessKit support") orelse .off;
 
     build_options.addOption(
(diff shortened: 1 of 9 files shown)


david-vanderson:
Thanks very much!  Give me a bit of time, I'll fix the testing backend and merge it.

Is it worth exposing this as a build option?  Or should we make it backend-specific (for backends that still require u16) and move to u32 for backends that can handle it?

Hcode00:
Thanks for the feedback! I think we should keep both approaches.

**Most users won't hit the u16 limit** - it's enough for small/medium apps. So keeping it as a build option makes sense; only someone creating a heavy visualization app would need to think about it.

Right now I'm using my forked dvui (from this PR until it gets merged) like this:

```zig
// in build.zig
    .target = target,
    .optimize = optimize,
    .backend = .sdl3gpu,
    .@"sdl3gpu-index-bits" = 32,
});
```

## Suggested Implementation

1. **Backend-specific defaults** - set sensible defaults per backend:

```zig
const default_index_bits = switch (backend) {
    .sdl3gpu => 32,
    .dx11 => 32,
    else => 16,       // web, sdl2, raylib - keep u16 for wider compatibility
};
```
   
2. **Dependency option for override** - keep `.@"sdl3gpu-index-bits"` (and equivalents for other backends) so users can explicitly set it when needed

3. **Update other backends** to use `IndexType` alias instead of hardcoded `u16`

4. **Add a runtime warning** when vertex_count > 65535 with u16 indices:  
   *"Index buffer overflow. Set .@"sdl3gpu-index-bits" = 32 in your dvui dependency options"*

This way apps with heavy rendering (CAD tools, data visualizations, etc.) work out-of-the-box with SDL3 GPU, but we don't break compatibility for web/mobile targets. Best of both worlds?

david-vanderson:
After playing with this a bit, I realized this is not really a backend-specific option, because it affects dvui+backend together.  So I promoted it to a top-level option, and made the build error out if u32 is given for a backend that doesn't support it.

How does that sound?

Hcode00:
Perfect, thanks for improving it! Top-level is definitely a cleaner approach. Closing in favor of commit [0fffb20](https://github.com/david-vanderson/dvui/pull/703/changes/0fffb20420c76afaefe9b4559ece1ce11fcc9ca4).

david-vanderson:
Thank you!

Issue #669, opened 2026-01-10
WEB Backend Improvements


Didas-git:
As of opening this issue the utility of DVUI on the WEB is fairly limited, while a lot of different apps can be made the situation limits more complex applications that require some for of access to web apis.

With the structure of dvui there are a few paths i'd like to discuss and push forward to make dvui not only the goto for desktop applications but also for dynamic web apps built in zig.

### Import/Export Symbols

The main blocker with dvui is that there is no way for a user to easily expose JS functions to zig and vice-versa, this can be achieved with a couple of ways but in my opinion there are only 2 viable options for this.

The first approach is to allow the user to specify a second javascript file to be included by dvui, this is the way i have been doing on my project, some people might complain about the `window` object "polution" by adding 2 objects to it but given that dvui doesnt heavily depend on JS i dont think this is a concern worth having given all the benefits we gain.

```js
// web.js
// We can use this to export stuff like `stringFromPointer` etc
// which dvui already uses in its js code its just duplicated all over instead of a reusable function
window.dvui_utils = {
    utf8decoder,
    utf8encoder,
};

export async function dvui(canvas_id, wasm_file_path, js_file_path) {
    let js_mod = null;
    if (typeof js_file_path === "string") {
        js_mod = await import(js_file_path);
    }

    const dvui = new Dvui();
    const result = await WebAssembly.instantiateStreaming(fetch(wasm_file_path), {
        dvui: dvui.imports,
        env: js_mod === null ? {} : js_mod.registerModule(),
    });

    window.dvui_exports = result.instance.exports;


    dvui.setInstance(result.instance);
    dvui.setCanvas(canvas_id);
    dvui.run();
}

// module.js
export function registerModule() {
    return {
        print(ptr, len) {
            // for the sake of the example, this function is defined
            // tho if we go with this approach it would be defined under 
            // `window.dvui_utils`
            const message = stringFromPointer(ptr, len);
            console.log(message);
        }
    }
}

// main.zig
extern fn print(message: [*]const u8, len: usize) void;

pub fn AppInit(win: *dvui.Window) !void {
    const msg = "Test Message From zig";
    print(msg.ptr, msg.len);
}
```

[...]

PR #706, opened 2026-02-04
Initial cleanup of `web.js`

Didas-git:
This PR is a starting point towards some of the conversations had #669, there is still a lot to do (opening `web.js` alone fills my IDE with errors) but with this PR at least changing some things will be more trivial as we dont need to deal with transforming pointers manually every time

### DIFF
--- a/src/backends/web.js
+++ b/src/backends/web.js
@@ -131,7 +131,7 @@ const fragmentShaderSource_webgl2 = `# version 300 es
     }
 `;
 
-/** @typedef {string | WebAssembly.WebAssemblyInstantiatedSource | Promise<WebAssembly.WebAssemblyInstantiatedSource} WasmArg */
+/** @typedef {string | WebAssembly.WebAssemblyInstantiatedSource | Promise<WebAssembly.WebAssemblyInstantiatedSource>} WasmArg */
 
 /**
  * @typedef {object} DvuiOptions
@@ -158,6 +158,7 @@ export function dvui(canvas, wasmRef) {
 
 const utf8decoder = new TextDecoder();
 const utf8encoder = new TextEncoder();
+
 
 export class Dvui {
     /** @type {WebGL2RenderingContext | WebGLRenderingContext} */
@@ -270,6 +271,31 @@ export class Dvui {
         return idx;
     }
 
+    stringFromPointer(ptr, length) {
+        return utf8decoder.decode(this.bytesFromPointer(ptr, length));
+    }
+
+    bytesFromPointer(ptr, length) {
+        return new Uint8Array(this.instance.exports.memory.buffer, ptr, Number(length))
+    }
+
+    bytesToPointer(allocator, bytes) {
+        const pointer = allocator(bytes.length + 1);
+        const slice = new Uint8Array(
+            this.instance.exports.memory.buffer,
+            pointer,
+            bytes.length + 1
+        );
+        slice.set(bytes);
+        slice[bytes.length] = 0;
+        return pointer;
+    }
+
+    stringToPointer(allocator, string) {
+        const buffer = utf8encoder.encode(string);
+        return this.bytesToPointer(allocator, buffer);
+    };
+
     constructor() {
         this.hidden_input = document.createElement("input");
         this.hidden_input.setAttribute("autocapitalize", "none");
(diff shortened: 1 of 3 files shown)


david-vanderson:
Looks good - why do we need the +1 (and zero end byte) on bytesToPointer?

Didas-git:
In some cases users might prefer to pass the string like you would in C, no length properly and use `std.mem.span` at a later date (also allows for simply passing strings to C apis without much extra work).

Its not needed for what dvui itself does internally but this api is also intended for users to use and for example, when returning a string from a JS function you sadly dont have much choice unless you opt for the "return" to be calling an intermediary zig function (very overkill in some scenarios)

src/backends/web.js
```diff
@@ -270,6 +271,31 @@ export class Dvui {
         return idx;
     }
 
+    stringFromPointer(ptr, length) {
+        return utf8decoder.decode(this.bytesFromPointer(ptr, length));
+    }
+
+    bytesFromPointer(ptr, length) {
+        return new Uint8Array(this.instance.exports.memory.buffer, ptr, Number(length))
```
RedPhoenixQ: The `Number(length)` feels weird and unnecessary. Ideally we should expect the correct type to be passed in and allow the implicit conversion in the `Uint8Array` constructor handle any convertible type. Anything else should throw an error anyways

src/backends/web.js
```diff
@@ -1304,22 +1241,14 @@ export class Dvui {
...
-                    this.instance.exports.memory.buffer,
-                    ptr,
-                    str.length,
-                );
-                dest.set(str);
+                const ptr = this.bytesToPointer(this.instance.exports.arena_u8, str)
                 this.instance.exports.add_event(
                     5,
                     ptr,
                     str.length,
```
RedPhoenixQ: I know this was here before, but this should also use the returned length from the allocation

david-vanderson:
I sympathize with that argument, but I'm uncomfortable doing it for everything (including the pixels), at least for now.  I suspect it'll cause weird issues to always have an extra byte on there, and already you've got a few "- 1" things to correct for it.

Please revert that part of it.  Everything else looks good!

Didas-git:
Instead of reverting what if we take a more zig-like approach with `bytesToPointer` and `bytesToPointerSentinel`? We could use this to also allow custom sentinels besides just null termination, (thinking like `std.fmt.bufPrintSentinel`)

david-vanderson:
> Instead of reverting what if we take a more zig-like approach with `bytesToPointer` and `bytesToPointerSentinel`? We could use this to also allow custom sentinels besides just null termination, (thinking like `std.fmt.bufPrintSentinel`)

I'm okay with this in general, but I don't like adding a function unless we have a use for it - otherwise it's too easy to get it wrong.  So if you have a use right now, go for it, otherwise we should wait.

RedPhoenixQ:
I still think a simple `alloc` that returns `{ ptr: number, len: number }` and another function for `allocZ` that includes the null terminator if needed. Returning the length should also remove the confusion of adding a +1 manually

david-vanderson:

Didas-git:
While i do want to work on improving `web.js` further it wouldn't fit the cleanup description.
I would consider it ready given the purpose of the PR but i do not have permission to merge

david-vanderson:
> While i do want to work on improving `web.js` further it wouldn't fit the cleanup description. I would consider it ready given the purpose of the PR but i do not have permission to merge

Sorry, try now.

Didas-git:
I got it now.
Just to take note for future PRs, are squash merges acceptable or would you prefer a normal merge? (Personally I prefer squash merges to keep main a little cleaner)

david-vanderson:
I don't have a preference, so either is okay!

MichaelBelousov:
@Didas-git why did this have to break web.d.ts for people wrapping the web backend? It removed the public API types

Didas-git:
That could potentially be discussed in an issue for others who come across it but, the types were not removed, all of them are in jsdoc in the web.js file (web.d.ts was just cleaned to remove all duplicate types and is only needed because doing unions in jsdoc is actual pain), you just need to review how you are importing the types. in all my testing (and in my own app) i didn't face any issues.

MichaelBelousov:
Thanks for clarifying. I do have type check errors on my end but I saw errors even on the original change. I'll have to go plumbing as I don't remember. The intention might have been to fix in a separate PR

PR #715, opened 2026-02-07
update to zig 0.16.0-dev.2534+12cb5b928

andrewrk:
all the lockUncancelable should be audited, almost always lock (cancelable) is correct.


david-vanderson:
Thank you!  Please feel free to merge as needed to the zig16-dev branch.  I will cherry pick the paths part back to main.

I will look into lockUncancelable at some point, but no need to hold this up in the meantime.

andrewrk:
Thanks!

PR #718, opened 2026-02-08
Add Group Box widget

### LINKED ISSUE #685: Add fieldset widget
Coming from discussion https://github.com/david-vanderson/dvui/discussions/678:
> I suggest adding a fieldset widget ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/fieldset), [Dlang UI](https://github.com/buggins/dlangui#dlang-ui)).

Useful for categorising inputs.

This would probably have purely visual effect by default, so not much new functionality-wise.

Image

[Image source](https://www.geeksforgeeks.org/html/how-to-achieve-fieldset-like-effect-without-using-fieldset-tag/)

phatchman:
dvui.groupBox() creates a bordered group with a title. Used for containing related widgets.
Also known as a fieldset in HTML.
Creates an accessibility role of group, labeled with the text of the title.
Adds example usage to basic_widgets.zig.

Fixes #685

image

image

image

image

image

image

image

### DIFF
--- a/src/Examples/basic_widgets.zig
+++ b/src/Examples/basic_widgets.zig
@@ -130,63 +130,66 @@ pub fn basicWidgets() void {
 
         dropdownAdvanced();
     }
-
     {
-        var hbox = dvui.box(@src(), .{ .dir = .horizontal }, .{ .expand = .horizontal, .min_size_content = .{ .h = 40 } });
-        defer hbox.deinit();
+        var gbox = dvui.groupBox(@src(), "Slider examples in a group box", .{ .expand = .horizontal });
+        defer gbox.deinit();
 
-        dvui.label(@src(), "Sliders", .{}, .{ .gravity_y = 0.5 });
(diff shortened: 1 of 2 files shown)


phatchman:
Looking at this again visually, the top padding looks too big around the sliders in the demo. I'll take a look into that later.

Edit: turns out the padding is correct, but there were some other cases where it wasn't. 

@david-vanderson Good for review.

david-vanderson:
Looks great!  I changed it so that when the label width is zero, we don't render it at all.  I was seeing that when I changed the width of the basic widgets pane towards zero the label showed back up at that point.

This is due to a quirk of how passing `Options.rect` works - if the width/height are 0, then it uses the min size width/height from last frame.  This allows using it to position a widget but let the widget figure out its size.  Sorry it's confusing, I will add it to my notes about widget layout for when I get around to writing a blog post about it.

If that looks okay to you, then please merge when ready.  Thank you!

PR #719, opened 2026-02-09
Support display of optional values in drop downs

### LINKED ISSUE #710: dropdown and dropdownEnum: add nullability
@Didas-git on discord was asking for these functions to have:
* the ability to accept a pointer to a nullable variable
* placeholder text if the variable is null

These allow a dropdown that requires the user to select an option but doesn't default to a specific choice.  Seems like a common enough thing to warrant adding to the high level functions.

phatchman:
Support display of optional values in dropdowns.:
- Add placeholder init_options to DropDownWidget
- Add init options to dropdown() and dropdownEnum() to control null display
- Add DropdownChoice(T) to accept optional and non-optional choice parameters.
- Optionally allow display and selection of the blank "null" value.

Fixes #710

### DIFF
--- a/src/Examples/animations.zig
+++ b/src/Examples/animations.zig
@@ -97,7 +97,7 @@ pub fn animations() void {
 
             dvui.labelNoFmt(@src(), "Animate", .{}, .{ .gravity_y = 0.5 });
 
-            _ = dvui.dropdown(@src(), &.{ "alpha", "horizontal", "vertical", "none" }, &global.animation_choice, .{});
+            _ = dvui.dropdown(@src(), &.{ "alpha", "horizontal", "vertical", "none" }, .{ .choice = &global.animation_choice }, .{}, .{});
 
             dvui.labelNoFmt(@src(), "easing", .{}, .{ .gravity_y = 0.5 });
 
@@ -106,7 +106,7 @@ pub fn animations() void {
                 recalc = true;
             }
 
-            if (dvui.dropdown(@src(), &easing_names, &global.easing_choice, .{})) {
+            if (dvui.dropdown(@src(), &easing_names, .{ .choice = &global.easing_choice }, .{}, .{})) {
                 global.easing = easing_fns[global.easing_choice];
                 recalc = true;
             }
--- a/src/Examples/basic_widgets.zig
+++ b/src/Examples/basic_widgets.zig
@@ -18,7 +18,7 @@ const RadioChoice = enum(u8) {
     _,
 };
 var radio_choice: RadioChoice = @enumFromInt(0);
-var dropdown_val: usize = 1;
+var dropdown_val: ?usize = null;
 
 /// ![image](Examples-basic_widgets.png)
 pub fn basicWidgets() void {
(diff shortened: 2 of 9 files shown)


phatchman:
If we're happy with this, I think I need to add placeholder to dropdown() as well?

Placeholder:
image

With null selection allowed:

image

With null selection *not* allowed:

image

non-optional (i.e. current fucntionality):

image

Null, without placeholder text:

image

```
debug(dvui): dropdownEnum 8fad947ca13c0734 displaying an optional enum requires the placeholder init_option to be supplied
```
Here we could default the placeholder text to "-- Select --" instead?

david-vanderson:
Looks great to me - thank you!

> If we're happy with this, I think I need to add placeholder to dropdown() as well?

If you are willing that would be wonderful.

> Here we could default the placeholder text to "-- Select --" instead?

Good idea - I like it.

phatchman:
Changed the dropDown() api to have the same init options as dropdownEnum(). I had to replace choice: *usize with:

```
pub const DropdownChoice = union(enum) {
    choice: *usize,
    choice_nullable: *?usize,
};
```
Which seems to work fine. Just need to add the default placeholder text to both dropdowns and _maybe_ make  dropdownEnum use the DropdownChoice union for consistency.

phatchman:
OK I need to do some testing and review, but happier with this version. I think the change to using DropdownChoice(T) to handle nullable and not-nullable choice values works well for both styles of dropdowns..

` _ = dvui.dropdownEnum(@src(), dvui.Options.Expand, .{ .choice_nullable = &expand }, .{ .placeholder = "Select something", .disable_null_selection = true }, .{});`

Edit: Good to review. 

Default placeholder:

image

Now we've got 3 places doing self.data().options.color(.text).opacity(0.65) for placeholder, seems like we should make that common somewhere.

Didas-git:
I'm still of the opinion that the label for the null option should not be blank, it should be the same as the placeholder or configurable, a blank/empty choice just looks like a bug to anyone using an app.

david-vanderson:
> OK I need to do some testing and review, but happier with this version. I think the change to using DropdownChoice(T) to handle nullable and not-nullable choice values works well for both styles of dropdowns..
> 
> ` _ = dvui.dropdownEnum(@src(), dvui.Options.Expand, .{ .choice_nullable = &expand }, .{ .placeholder = "Select something", .disable_null_selection = true }, .{});`

Yeah that looks better - good idea!

> Now we've got 3 places doing self.data().options.color(.text).opacity(0.65) for placeholder, seems like we should make that common somewhere.

Good point.  I'm not sure whether to have `Options.placeholder()` or `Color.placeholder()`.  Will think about it.

> I'm still of the opinion that the label for the null option should not be blank, it should be the same as the placeholder or configurable, a blank/empty choice just looks like a bug to anyone using an app.

Agree - @phatchman would you mind duplicating the placeholder text for the null option?

david-vanderson (approved):

phatchman:
OK both those changes done:

image

image

@david-vanderson We could change `DropDownWidget.addChoiceLabel()` to take a dvui.Options to allow styling of the text? Instead of having to use `addChoice()` just to change the opacity?

Didas-git:
Im not a fan of a single selection (specially what is essentially the default selected one) to have a different opacity than everything else, it doesnt feel very coherent with everything else.

The styling should be the same for all options, if the user wants to specialize a single option (personally cant think of a reason to do so) i think thats when they should start copy pasting some code...

david-vanderson:
> @david-vanderson We could change `DropDownWidget.addChoiceLabel()` to take a dvui.Options to allow styling of the text? Instead of having to use `addChoice()` just to change the opacity?

Good idea.

> The styling should be the same for all options

I agree, but at this point we are way down in the weeds. @phatchman - whatever you think is best.  I'm good - merge when ready!

phatchman:
> Im not a fan of a single selection (specially what is essentially the default selected one) to have a different opacity than everything else, it doesnt feel very coherent with everything else.
> 
My view is that it is placeholder text and to be consistent it needs to be the same style as the other placeholders. It is different to a "default selection", which would have the same style as everything else.

What did look weird to me was when the opacity of the placeholder changed between the collapsed and dropped down state. Which is why I made them look the same.

I'm included to leave it as-is.

david-vanderson:
Thank you!

PR #725, opened 2026-02-11
textEntryNumber() add init option to display or clear text

### LINKED ISSUE #720: Feature request: Simple way to clear the text in a textEntryNumber field
I'm using `dvui.textEntryNumber` for data entry, where I want to type a value in, hit enter, and have my UI enter a new entry and clear this field so I can type the next number in.

If I were using `dvui.TextEntryWidget` directly, no problem. I could just call `te.textSet("", false);`. But wrappers like `textEntryNumber` provide a lot of boilerplate so I don't have to, so I'd like to use them. The problem is that `textEntryNumber` both instantiates the `TextEntryWidget` and calls `deinit` on it, and doesn't expose it to the caller, so I don't really have access to its functions.

The function signature in `dvui.textEntryNumber` might have some way to drill down and get access to the backing data store or widget data in order to clear it, but it's not at all obvious from parameters.

I'd like either some demo code for drilling down into the state and manually controlling that (i.e. in the text_entry sample), if that's even possible, or I'd like a feature to make it easier to simply clear the text and empty the field.

I was able to copy/paste/modify my own implementation of this function to add it, quite easily, but this seems to me like some fairly basic functionality that might be a common use case, and it would be nice to have it without having to customize things.

I can think of a few different approaches to solve this, and I don't know the library well enough to suggest which one is most DVUI-like.

phatchman:
When text is set via init_option, the textEntryNumber enters "text_mode".
If a key down is received while in text mode, return to numeric mode.

OLD - showing invalid ring in text_mode.
image

NEW 
image

Fixes #720

### DIFF
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -4876,9 +4876,6 @@ pub fn textEntryNumber(src: std.builtin.SourceLocation, comptime T: type, init_o
             AccessKit.nodeSetInvalid(ak_node, AccessKit.Invalid.ak_true);
             AccessKit.nodeSetNumericValue(ak_node, 0);
         }
-        if (false) {
-            AccessKit.nodeSetPlaceholderWithLength(ak_node, te.textGet().ptr, te.textGet().len);
-        }
     }
 
     te.deinit();


phatchman:
Initially I thought I could used te.text_changed to detect the keypress to change from text to number mode. _But_ the first character would get eaten if I set the text to blank after te.processEvent(). 

So, instead I'm doing the key press detection within textEntryNumber and setting the text entry to blank prior to the te.processEvent().

This is only an issue if the text contains a number. For non-numeric text  te.filterIn() would filter the text for us.

text_mode probably shouldn't trigger the red invalid ring. I'll fix that. Edit: Fixed

The only other tweak we might want to make is to force the cursor to be in the 0th char position while in text mode. If you mouse click in you can currently change the cursor position. Which might be fine? I'm not sure. OR maybe we should clear text_mode when the te is focused?

david-vanderson:
Sorry I confused everything and mixed up putting a textual value in there with a placeholder.  We need to separate those, and likely need to change the behavior around it being empty:
* the new `text` init option should work the same as if a person cleared the existing text and typed that value in
* for placeholder text we probably need a new init option
  * can we reuse the TextEntryWidget placeholder internally?
* maybe empty should not be considered an error?

Does this make sense?

phatchman:
> Sorry I confused everything and mixed up putting a textual value in there with a placeholder. We need to separate those, and likely need to change the behavior around it being empty:
> 
> * the new `text` init option should work the same as if a person cleared the existing text and typed that value in

That's what it currently does. (except that you can't type non-numeric characters into a text entry number, as the non-numeric chars are filtered out. So, I have to disable the filtering if someone has set text and persist that disabling of the filter until some event (e.g. keypress). 

> * for placeholder text we probably need a new init option

I don't really understand what the difference between "text" and "placeholder" functionality should be.

Also note that, we currently have existing "Placeholder" text to show min and max values. Not an issue, just noting.

>   
>   * can we reuse the TextEntryWidget placeholder internally?
> * maybe empty should not be considered an error?

That was my original intention to just allow the user to set the placeholder for the text entry widget. I'm pretty sure that would work.

Also, empty isn't an error currently or with this PR. 

Let me know how you see the text option working vs placeholder.

david-vanderson:
Hmmm maybe I'm the only one confused.  Give me some time to play with it.

david-vanderson:
> That's what it currently does. (except that you can't type non-numeric characters into a text entry number, as the non-numeric chars are filtered out. So, I have to disable the filtering if someone has set text and persist that disabling of the filter until some event (e.g. keypress).

I think I misled you, sorry.  My intention is that the given text would be filtered, exactly as if the user had typed it.

> I don't really understand what the difference between "text" and "placeholder" functionality should be.
> 
> Also note that, we currently have existing "Placeholder" text to show min and max values. Not an issue, just noting.

Good point.  I adjusted it to actually use TextEntryWidget's placeholder functionality (I think that code predated that?).  I still need to test Accesskit to make sure the placeholder is coming through, that's why I left that "if (false)", I'm not totally sure which widget it should be associated with.

> That was my original intention to just allow the user to set the placeholder for the text entry widget. I'm pretty sure that would work.

This I think is where I messed up.  I hadn't intended `.text` to be for the placeholder.  My understanding was that it would be used for two purposes:
* clearing the widget (so that it is now empty, possibly showing the min/max placeholder text)
  * I've added a "clear" button to the demo text entry parsing section
* inserting a particular value (as text), like you click a separate button and it inserts "123" into the widget

> 
> Also, empty isn't an error currently or with this PR.

I thought it was?  Like if you delete all the text in a textEntryNumber widget, then it highlights itself in red, right?

> 
> Let me know how you see the text option working vs placeholder.

I've committed the kind of behavior I was thinking of, let's make sure we are on the same page.  Is this making more sense now?

phatchman:
Ahhhh! OK now makes sense. In my mind, text meant setting any text, not "filtered" text.

I can sort out the rest of the testing and any cleanups.

Sorry, you are right about empty. I must have already changed something when I did that testing. Or setting text to "" doesn't trigger empty somehow. 

image

david-vanderson:
Excellent!  I will ask about empty on the issue.  Thank you!

phatchman:
> Excellent! I will ask about empty on the issue. Thank you!

All good. Glad we workout out where the miscommunication was! Everything is easy in hindsight.

I think Empty should remain an error, unless we allow null values.

Assuming you should be able to set every valid state via `init_opts.value`, how would you set a value of blank unless value is allowed to be null? As we don't allow null values, no one could set the valid state of blank/empty using `init_opts.value`.

You could argue blank is an exception, but it makes sense to me that it should remain an error.

phatchman:
I added support for init_opts.placeholder and everything else looks good. I've left it that if you have a placeholder, supply an init_opts.value and the value is empty that it still shows the red error ring.

So you get:
image
or
image

Depending on whether value is passed. But if value is passed, then it normally wouldn't show placeholder text unless the user deleted the text, or the text was set to ""




But I'm looking into an issue with AccessKit support where the reader is saying "newline" when switching from placeholder to non-placeholder text. 

~~That will prob end up being a separate PR, unless it is trivial. I'm pretty sure it's the same id changing issue (i.e. the id of the palceholder text run is not the same as the id of the text entry text run.~~

~~Edit: ugh it's worse. It's not just placeholders. Even when you start typing from an empty box, the cursor doesn't update and the characters aren't read, again presumably because the text run id changes. It's definitely going to be a separate PR for this one.~~
I'm pretty sure I found the issue and it is to do with how word starts are being calculated, but that is a job for tomorrow. If I'm right it's a pretty trivial fix and can include in this PR.

phatchman:
On a related, but different topic. To test these changes, I tend to create these throw-away test harnesses like this:
image

Sometimes with a way to set some of the important layout options for the widget.

Rather that throwing these away and recreating them each time, it would be nice to save them and over time develop a "control panel" for each widget that allows you to test all its functionality when making changes.

I don't have a really good idea on how to implement that. We could have standard function named testWidget() in the widgets, but most widgets get created in dvui.zig, so we'd have to have lots of different test function names. 

Not sure how I feel about how this would mix the test code with the widget creation code in dvui.zig, which is already quite large. 

Maybe it is a positive as there would be an example of how to use the widget immediately following the creation function and people could call that test function from their own code to try out lots of different options for the widget to get it working how they want?

Anyway, thought I'd raise it to see if there is any interest in pursuing that further. I see it as something we'd just build up over time as we worked on each widget, rather than making a big PR adding lots of test functions.

It might also give us a nice use case for struct_ui as it should be perfect for things like setting init_options and whatever sub-set of Options is applicable to that widget.

phatchman:
OK. there were a few changes I had to make to how we handle empty text runs:

1) Don't populate any of the length or width values etc if it is empty.
2) Fix an issue where word starts were not being calculated correctly for the first text run.
3) Keep the first text_run's id consistent whether empty / placeholder or with text.

There is still something weird going on with backspace. The Windows narrator cursor doesn't follow the backspace, but as far as I can see we're setting the cursor position correctly.

Good to go from me.

david-vanderson:
Thank you!

> Rather that throwing these away and recreating them each time, it would be nice to save them and over time develop a "control panel" for each widget that allows you to test all its functionality when making changes.

This makes a lot of sense, and it could take some of the pressure off the normal demo (so we don't feel like the normal demo has to show off all the features).  I'm also not sure where it should go.  We could try a new section of the demo with a tab/dialog/something for every one of these widget-specific control panels?  Sort of like a reference section.

phatchman:
This is what we get out of the box with StructUI. With some tweaking, I think it could work.

image

phatchman:
I've got a prototype using struct_ui (with some enhancements) up here: https://github.com/david-vanderson/dvui/compare/main...phatchman:testing_controls to sdl-standalone.

I'll probably add some way to test common scenarios via drop-down or radio buttons. e.g. to force an error or empty.

image

The checkboxes are "manual reset" boolean display fields. The idea is you can see whether "changed" was set to true since you last reset the checkbox, and also see the current value.

But looks promising to me. I'll think more about how to integrate it, but probably a separate window that you can open from the main screen. A "widget playground" or reference section as you suggest or ... I'll play and see what looks good.

david-vanderson:
> The checkboxes are "manual reset" boolean display fields. The idea is you can see whether "changed" was set to true since you last reset the checkbox, and also see the current value.

I've never thought about this kind of "latched" display.  Will think about it!

> But looks promising to me. I'll think more about how to integrate it, but probably a separate window that you can open from the main screen. A "widget playground" or reference section as you suggest or ... I'll play and see what looks good.

It does look promising - thanks!

phatchman:
I implemented that custom display callback idea for struct_ui and it came together nicely. You can assign a custom display function to any field (or to a container type) and display something like this for Color.

image

I'll probably use the options editor from the debug window for the options that are covered by that. Still thinking about how to best handle them.

david-vanderson:
Looks cool!

phatchman:
I've updated the prototype to work for box and textEntryNumber, which can be selected from the tree. Box is still pretty basic. I imagine box is the sort of widget that needs lots of different pre-set test cases for testing different layouts.

The main limitation I see is that struct ui doesn't allow you to order / group fields. They are displayed in the order of declaration.  I don't have a good solution for that as the order is determined at comptime. Runtime just chooses which ones are shown / hidden.

For common things like Options we can use the custom display function to order things, so there is a workaround at least.

Setting optionals feel a little  clunky. But I don't see a better solution. Maybe for enums and bools we could use a dropdown with a null state. But for other types, I'm not sure how we'd represent an optional int or string any better. The current way works consistently for everything at least.

david-vanderson:
Cool.  I think you are right about box.

The more issues we run into with struct_ui, the more it seems like the way to think about it is "a quick way to get a debug-level editor for a struct" rather than "a good way to display a struct for editing generally".  Not totally sure that's right, but wanted to put that out there for you to think about.

PR #732, opened 2026-02-16
dropdown: Highlight correct selected index when dropped

### LINKED ISSUE #729: dropdowns: fix list offset
I think this broke with #719 - when a dropdown is clicked, the floating list is supposed to be offset so that the selection lines up with where the dropdown widget is (so the selected item is under the mouse).

phatchman:
- The incorrect selected item was highlighted when placeholder text is shown in the dropdown list

There is still an inconsistency when first dropped down with the placeholder vs after a selection has taken place. The placeholder is highlighted on the first dropdown, but in subsequent drops, it is not. But is a minor visual bug compared to this main issue.

Fixes: #729


david-vanderson:
Did my commit fix that issue, or did you mean something else?

If it did, then this looks good - merge when ready.  Thank you!

phatchman:
I didn't notice. Will test and merge.

PR #735, opened 2026-02-18
Add a widget reference to the demo

phatchman:
Implements the start of a "Widgetpedia", which:

1) Allows testing of widgets during development of DVUI
2) Showcases all available init_opts and options that affect a widget.
3) Provides a quick reference for users to see all available widgets.
4) Allows users to interact with widgets to customize their options and test their capabilities.

A small collection of widgets is provided as a start.

Currently only sdl-standalone is modified to add the Widgetpedia.

Let me know thoughts and whether you think it is worth developing further to a cleaner PR.

P.S. I'm not wedded to the name. We can make it whatever.


david-vanderson:
First impressions:
* I really like the idea and layout!
  * widget list down the left is great
  * grouping widgets with convenience functions is great
  * excellent way to research widget options
  * way to take some stuff out of the demo (which is getting overwhelming)
* I'm less happy with the struct_ui approach
  * not obvious which options are widget options vs options that change other stuff you'd pass to the widget (like the group box title)
  * I immediately wanted to hand-craft it

Based on that, my guess is that the way forward here is to add "Widgetpedia" (love the name) to the demo (maybe a top-level button that toggles it as a separate window?).  But approach it similarly to the demo (hand written) - so then we can better balance the demo (showing off all the widgets) with this (visually showing all the widget-specific options).  I wonder if there is a way to add a button in the demo that opens the Widgetpedia with a specific widget already selected?

What do you think?

phatchman:
> I'm less happy with the struct_ui approach

So, my question is can we make improvements to struct_ui to give you what you are looking for? In my mind I want:
1) A standard UX for the configuration of most widgets.
2) An "easy" way to display and set those configuration options, without having to deal with state for optionals etc.

And I think struct_ui gives both of those, but you can use a fully customized approach when it is needed. 

The one I would really like to improve in struct_ui is the handling of optionals. But I don't have a good idea on that at the moment.

> not obvious which options are widget options vs options that change other stuff you'd pass to the widget (like the group box title)

Is this really a struct_ui issue? I separated them out into "test_options", which control what the user would pass, like labels, "init_options", "options" and "results" to try separate that. How would you make that clearer?

> I immediately wanted to hand-craft it

What sorts of things would you do differently e.g. for init_options? 

Can we pick a particular widget , compare the two approaches to see how close we could get with struct_ui?

> I wonder if there is a way to add a button in the demo that opens the Widgetpedia with a specific widget already selected?

Hmm. The technical side is fairly straight forward, but not sure on the UX side. e.g. on Basic Widgets, how would you identify the widget they are after? 

I'll probably also add a seach feature at the top of the tree, so potentially you could open the window with a search string set to the widget you want.

david-vanderson:
> So, my question is can we make improvements to struct_ui to give you what you are looking for? In my mind I want:

I'm not sure, and I fully admit that I don't have a detailed understanding of what I'm looking for, so thanks for the push!  I think there are two parts:
* user-intention can map non-obviously to widget options
  * example is a multiline textEntry where `break_lines` and `scroll_horizontal` are usually coupled
  * or the layout section image example, where if you check "shrink" it shows the shrink enum radio buttons
  * widgetpedia is an opportunity for us to guide people (in addition to showing the widget options)
  * I think this means we don't want (or don't necessarily want) to mechanistically show the widget options?
* the effect of certain widget options (or widget interactions) can be non-obvious
  * example is the large doc demos for textEntry
  * some widget options only make sense in a particular circumstance
  * I think this means we would want the example to (possibly) change (or have multiple examples) based on the options selected

> Can we pick a particular widget , compare the two approaches to see how close we could get with struct_ui?

This is a good idea.  Also not unreasonable to start with a struct_ui quick thing, and then hand craft later.  I will try this.

> I'll probably also add a seach feature at the top of the tree, so potentially you could open the window with a search string set to the widget you want.

That is a great idea.

phatchman:
I largely agree with that. 

Take the entry for box(), which I picked because it didn't really have anything to expose other than an options editor. You only really see behaviour when you put something in the box.  So, I imagine this would have more hand-crafted UI where you select some scenario and that exposes various parameters you can tweak.

But the really basic widgets like check boxes, buttons etc, I don't see much need to exposer anything but init_options, options and results.

In your mulit-line text example. I'd still want to be able to set these unusual combinations, so you can test break_lines with or without scrolling.

But, I agree text entry is another clear case where there are a lot of different scenarios to cover, that will need to be hand crafted.

Where I think we are at cross-purposes is:

> I think this means we don't want (or don't necessarily want) to mechanistically show the widget options?

I actually want this, even if it is not the primary interface that people use. I want to set a options that should not be set together as part of testing the widget. My primary purpose is not having to create a throw-away test harness each time I change a widget, which might also be useful for DVUI library users. I think you are leaning more to extended version of the demo that can show different aspects of each widget.

I think we can have both over time. Start with just showing raw the init_options, but as people test various widgets they can add the pre-configured scenarios and UI as they need. 

Then we can choose to put the raw strucf_ui feature in an "Advanced" section or something once the custom UI is created.

If you strip away the comptime recursively find all the fields in a struct, struct_ui it is literally just a bunch of functions that display a field with a label and a widget.  So, my thought is that we'd likely be creating similar display functions in the widgetpedia to keep a consistent look. If we need to create those anyway, we can put that effort into making struct_ui display things more nicely and both get a benefit. Whether that is practical I guess we will find out.

Just in this short time of dog-fooding struct_ui, it's seen lots of bug fixes and improvements, that would not have  happened otherwise. (e.g. the theme editor now exposes most of ninepatch out of the box)
And I think the more we use it ourselves, the more usable it would become. But I do understand your other comment that it may be a quick-way, but to necessarily a good-way.

david-vanderson:
> I actually want this, even if it is not the primary interface that people use. I want to set a options that should not be set together as part of testing the widget. My primary purpose is not having to create a throw-away test harness each time I change a widget, which might also be useful for DVUI library users. I think you are leaning more to extended version of the demo that can show different aspects of each widget.
> 
> I think we can have both over time. Start with just showing raw the init_options, but as people test various widgets they can add the pre-configured scenarios and UI as they need.

Very good point - I agree.  We can try to have both and see how it goes!

I also agree that dogfooding struct_ui is very useful.  Let's move forward with this - how about as a new demo section?  We can start with just a few widgets and go from there.

phatchman:
One thing I'm trying to think of a way to improve, is to show what dvui.Options are set (not null) for a widget.

e.g. For most of the widgets, options is empty when you first open it, but for a few we need to set gravity or expand etc. It would be nice to show the user which options are set for the widget, so if they can easily replicate what they see.

I think it could just list what options are set, rather than including their value. OR maybe we show simple values, but not colors etc.

david-vanderson:
> One thing I'm trying to think of a way to improve, is to show what dvui.Options are set (not null) for a widget.
> 
> e.g. For most of the widgets, options is empty when you first open it, but for a few we need to set gravity or expand etc. It would be nice to show the user which options are set for the widget, so if they can easily replicate what they see.
> 
> I think it could just list what options are set, rather than including their value. OR maybe we show simple values, but not colors etc.

I'm possibly following here, but could you give some examples?  It sounds like maybe we need a read-only version of struct_ui, but maybe this is too Options-specific.

phatchman:
This is what I've come up with so far as a tooltip over the options editor expander.

image

The idea is that the user can more easily tell what options are set on a widget, without having to Options the options editors.

I don't typically like using tooltips to show state, but maybe it works in this case.

david-vanderson:
Playing around with this:
* I like being able to see which Options are set
* but I want to see them all the time, especially as I change stuff in the Options editor
* maybe unrelated, but checking and unchecking "min size" now says a lot more Options are set?

I wonder if we should separate the Options from the widget-specific init_opts.  Maybe Options are something set across all the widgets in the widgetpedia?  Maybe we put them in a sidebar on the right?
Screenshot 2026-02-20 at 8 58 03 PM

Also possible that there's only a few cases within Options that matter in this context, and we don't need the full editor.  Like if we know the box the widget is going into, maybe for demonstration it's enough to have corner radius on/off, but not really necessary to be able to set it to a particular value on each corner?

phatchman:
> Playing around with this:
> 
> * I like being able to see which Options are set
> * but I want to see them all the time, especially as I change stuff in the Options editor

I've been coming to the same conclusion.

> * maybe unrelated, but checking and unchecking "min size" now says a lot more Options are set?

I've been slowly improving the null handling in the options editor. There are still a few that haven't been done yet, so as soon as you set a value, some of the null options get a default value.

There are also a few other options that need to be added to the editor.
> 
> I wonder if we should separate the Options from the widget-specific init_opts. 

Yep, I'll give this a go.

> Also possible that there's only a few cases within Options that matter in this context, and we don't need the full editor. Like if we know the box the widget is going into, maybe for demonstration it's enough to have corner radius on/off, but not really necessary to be able to set it to a particular value on each corner?

Yeah. I was thinking this sort of stuff goes in the "widget display options". But if someone really wants to make the corner radius 30 to see what it looks like, they still can.

phatchman:
We're really constrained horizontal space. 

Here's the colorPicker, with the tree expanded.

image

The options aren't quite wide enough to show the slider labels and values at the same time. 

You lose the init_options off the bottom of the screen, but we need to make a tradeoff somewhere. It's one of those few times where I'd like to show a window outside the main dvui window, so that it can be larger than the demo window.

Making the demo 850px wide and 650px tall, with the widgetpedia taking up 825px wide looks a lot better:

image

And animate looks like this:
image

I still have space to extend the options down into the "init_options" area like you had. I'll try that out if I have some time tomorrow, but I suspect it will make the things uneven.

Edit: I had a quick attempt at launching a separate OS window for the widgetpedia so we can make it whatever size it needs to be. But I didn't really work out how to do it. It seems SDL needs the backend created on the main thread, so somehow I'd have to manage both main loops on a single thread?

david-vanderson:
> We're really constrained horizontal space.

Maybe we could put some kind of button in the upper-right to toggle showing the Options sidebar?

> Edit: I had a quick attempt at launching a separate OS window for the widgetpedia so we can make it whatever size it needs to be. But I didn't really work out how to do it. It seems SDL needs the backend created on the main thread, so somehow I'd have to manage both main loops on a single thread?

That's a good idea, but you're right.  I think how it has to work is a single main loop that serially runs the dvui loop for each window.  That's what the dx11 backend does right now.  Eventually we'll get there, but that's a longer-term project.

phatchman:
Maybe I can implement this a little more nicely, but what about something like this? I've pushed this if you want to try it.

image

&nbsp;

image

&nbsp;

image

Edit: Fixed the issue with menu not scrolling.

I think it will work better on the right-side of the window. That way it doesnt cover the widget while you are selecting. Now you have to move off the menu to see the widget. But feels a bit weird to put the selection on the right.

Or I could just implement it like a normal menu and have it at the top of the screen.

phatchman:
Here I have the menu on the right-side. Tooltip doesn't seem to understand gravity with position = .absolute? As it was still popping up the floating window on the left-side of the screen, even when the tab was on the right. So, I'm using .horizontal positioning instead.

BTW these are all using the 800x600 resolution with the widgetpedia window taking up a width of 775px,

image

image

Edit: OR another possibility is to change that group-box label into a drop-down and allow people to select widgets from there. It would still cover up the widget, until you selection.

I see the advantage of the tree is that you can quickly flick through multiple widgets until you see something you want, vs the menu / drop-down approach where you have to select widgets one by one. 

The advantage of the menu or drop-down over the floating tree is that it disappears once selected, so you can see the widget right away, even if the menu was previously covering it. I prefer the floating tree, just not how it hides the selected widget.

I thought about other options, like an expander style approach where the menu slides in from the right. But I don't like that it would squash the widget display, which could trigger other layout changes within the widget while the menu is open. I'd prefer that the layout is static unless the user resizes the window. But it is still a viable option.

Edit2: I also tried reversing the layout of the options editor and widget, so the options editor was on the left. But I don't really like how that looks either.

Edit3: I'll try and make better use of the width at the bottom. We can have 2 columns there.

I'm leaning to the right-hand side floating tree, if I can make it work. We could have other tabs like "reference", "sample code" etc in the future? There is something satisfying about being able to click through each widget. So I'm trying to find a way to keep that,  instead of just using a menu or combo box etc.

david-vanderson:
Nope I was totally wrong.  The Options editor is very visually distracting, and my assumption is that it's not going to be used that often (since mostly what this is for is showing the widget-specific options).

I very much appreciate all the different layouts, and it's good to see that the mouse-hover-menu thing works fairly well.

But I think we should go back to a static tree on the left, widget on the right, widget options under the widget plus an expander at the bottom to show the Options editor.  What do you think?

phatchman:
No worries. Agree, that is the tradeoff with the widget editor always being on the right. I'm sure we could make it a little better with some borders and padding (I didn't really try and make it look polished), but it will always be busier than the single widget being displayed.

I'll try one more layout with the display options and init options in the right-panel where the widget editor is now and the options editor below it. Maybe in a tab, so you can put as many different types of options as you want there and see how that compares.

phatchman:
Might be a winner? If a widget has lots of options, they can be put in a tab widget, rather than scrolling.

image

image

Heights and margins etc still need fixing. But here is one with lots of init_opts.

image

Though some of the options get a bit squishy. I still think it's probably the best compromise. And if someone has a big tree of options, they can still put it in the bottom box.

image

I can probably steal a bit of padding / margin around the place to get us a bit more space, and/or make the widget box a bit narrower. Actually, we could just truncate the struct_ui labels if they get too long, and put a tooltip on them if they get truncated. That will solve that issue where one long options name pushes everything to the right.

Edit: I've managed to improve this just by making the widget box 50px narrower and some minor layout changes.

phatchman:
I've pushed a commit for the new layout. The layout's not perfect with expand etc. I'll fix that up if you agree this is the way forward.

There needs to be a couple of tweaks to struct_ui to work better for narrower widths, but looks reasonable to me, even without those changes. Can add horizontal scrolling as an interim if we need.

image

Edit: I've also improved the null handling for bools and enums in struct_ui. It now combines them into a single drop-down with a "null" option. See comboBox widget for good example of this and for how it looks when space is constrained.

Options edit default to open or collapsed? Can't decide?

phatchman:
I've fixed up the layout now and added a few more examples.

I added a vertical splitter to animate() to try it out. ~~(It doesn't expand the top section properly yet).~~ Will test that a bit more, to see if it makes sense.

Edit: I've added the pane to colorPicker() and combobox().

I like it and I think we can do some nifty things with it, like showing the options editor expander as collapsed when the bottom pane is collapsed and then when they expand it, we could animate split to 80% or something and show the options editor expanded.

I reckon this solves most of the issues as if the user is mainly focused on options, they can make the options area larger. IF they don't care about them, they can collapse the options editor expander, which would collapse the bottom pane to just show the collapsed expander.

david-vanderson:
I agree - this is looking pretty good!
* can we have the Options editor start collapsed?
* is there a way to not have the "Widget display options" expander (just have that stuff vertically there?)
* same for the color picker results (maybe "Results" is the name of the group box?)
* for the color picker results, the checkbox is still confusing - what do you think about doing a fade (like in the demo > "text entry" > singleline where you press enter and it flashes "Enter!" (maybe that fade should use a different easing)?

phatchman:
> * can we have the Options editor start collapsed?

Yes. I'm planning to make the bottom pane collapsed when the Options editor's expander is collapsed. Then open the bottom pane to some minimum size when the options editor is expanded. I'm pretty sure the pane should be able to support that.

> * is there a way to not have the "Widget display options" expander (just have that stuff vertically there?)

Often, we're showing multiple things there though. Usually just the widget display options and init_options.

> * same for the color picker results (maybe "Results" is the name of the group box?)

Sure, will try this. 

> * for the color picker results, the checkbox is still confusing - what do you think about doing a fade (like in the demo > "text entry" > singleline where you press enter and it flashes "Enter!" (maybe that fade should use a different easing)?

That sounds nicer. Will try. Something like an option of `triggered_when: ?bool`.

I'm also going to look at adding an optional expander above the options editor to put any more complicated options that can't fit in the right-panel. 

I'll put something together around the text entry widget to see if we might need it or not. e.g. to implement "Large Doc" and all the other scenarios on text entry demo. And to what extent we need to hand-craft those etc.

david-vanderson:
Awesome thank you!

phatchman:
Is there an option to set a "min_size_content" for the first pane in paned? i.e. I want to stop them making the top pane smaller than 300px tall.

Clamping the split_ratio init_opt doesn't work.

Ahh it does. I needed to do it after paned.init(), instead of before.

phatchman:
Weird thing I noticed. There is a difference between how `animate()` and `animation()` works for easing functions.

With `animation()`, `a.value()` is actually 1 - the value returned from the easing function whereas for `animate()` the value returned from the easing function is used directly.

In otherwords, if I use the same easing function for both `animate()` and `animation()` for alpha. they give the opposite results . I'd actually have to pass `1 - a.value()` to `dvui.alpha()` to make the easing function work as expected. i.e. I have to undo the `1 - easing function value` and turn it back into the actual easing function value.

Is that intentional?

Also, I had to add this to get the "false" value re-displayed after the animation is complete. 
```
if (a.value() == a.end_val) {
    dvui.refresh(null, @src(), null);
}
```

phatchman:
OK.
I've updated Animate, Button and ColorPicker to the new layouts. 
I've implemented the new `trigger` display for bools. I tried a few different easings. Not sure I'm 100% happy, but we can tweak it later.

I think I got everything from your last, except what to do with the right-hand box containing the widget controls I could turn that into a group box, but I want to see a few more examples where we have multiple controls and multiple options to see how that looks.

Let me know what you think. I like how the options expander turned out. (There's a bug with resizing the window horizontally, which I'll look at later.)

Edit: I've updated all of the existing demos to the new layout. It is much easier to change the layout for everything now, so  should be much easier to tweak. 

Similar to the issue with the "Controls" box, the results box sometimes gets a "results" label twice. I think there just needs ot be an option to struct_ui not to display the top-level container and that should fix it.

I didn't work out how to get the options editor expander to stay the right size during resizing the window. Not sure how to fix that, so if you have any ideas, appreciated. It is in `displayWidgetTemplate()`

So still some more tweaking, but should be easier now, because I only have to do it once for everything.

david-vanderson:
> Weird thing I noticed. There is a difference between how `animate()` and `animation()` works for easing functions.
> 
> With `animation()`, `a.value()` is actually 1 - the value returned from the easing function whereas for `animate()` the value returned from the easing function is used directly.
> 
> In otherwords, if I use the same easing function for both `animate()` and `animation()` for alpha. they give the opposite results . I'd actually have to pass `1 - a.value()` to `dvui.alpha()` to make the easing function work as expected. i.e. I have to undo the `1 - easing function value` and turn it back into the actual easing function value.
> 
> Is that intentional?

I'm having trouble understanding what you mean.  I am using this modification to examples/app.zig:

```zig
    {
        var animator = dvui.animate(@src(), .{ .kind = .horizontal, .duration = 500_000, .easing = dvui.easing.linear }, .{});
        if (dvui.firstFrame(animator.data().id)) {
            dvui.animation(animator.data().id, "blah", .{.end_time = 500_000});
        }
        if (dvui.animationGet(animator.data().id, "blah")) |a| {
            std.debug.print("animation value {d} animate value {d}\n", .{a.value(), animator.val orelse 0});
        }
        defer animator.deinit();
        if (dvui.button(@src(), "Debug Window", .{}, .{})) {
            dvui.toggleDebugWindow();
        }
    }
```

Can you modify this example to show the problem?

> 
> Also, I had to add this to get the "false" value re-displayed after the animation is complete.
> 
> ```
> if (a.value() == a.end_val) {
>     dvui.refresh(null, @src(), null);
> }
> ```

I can't find this, where did you have to add it?  But yes - the current way animations trigger refreshes is implicit, so you get a frame where Animation.done() is true, but to get a frame after that you need an explicit refresh.  I think that makes sense?

david-vanderson:
> OK. I've updated Animate, Button and ColorPicker to the new layouts. I've implemented the new `trigger` display for bools. I tried a few different easings. Not sure I'm 100% happy, but we can tweak it later.
> 
> I think I got everything from your last, except what to do with the right-hand box containing the widget controls I could turn that into a group box, but I want to see a few more examples where we have multiple controls and multiple options to see how that looks.
> 
> Let me know what you think. I like how the options expander turned out. (There's a bug with resizing the window horizontally, which I'll look at later.)

I like the layout a lot - and the options expander is quite nice!

> Similar to the issue with the "Controls" box, the results box sometimes gets a "results" label twice. I think there just needs ot be an option to struct_ui not to display the top-level container and that should fix it.

I didn't see this, is there a specific way to trigger it?

> I didn't work out how to get the options editor expander to stay the right size during resizing the window. Not sure how to fix that, so if you have any ideas, appreciated. It is in `displayWidgetTemplate()`

Will look into it.

phatchman:
OK. I've fixed the paned issue. Just changed how I calculate the "closed" split ratio.

If you look in the color picker prior to this commit, the group box would say "Results" and then it as displaying the "results" struct, so you got results twice. But have stopped it doing that.

It's interesting the sorts of things you find when playing around with the widgets. One example. 

image

david-vanderson:
> OK. I've fixed the paned issue. Just changed how I calculate the "closed" split ratio.
> 
> If you look in the color picker prior to this commit, the group box would say "Results" and then it as displaying the "results" struct, so you got results twice. But have stopped it doing that.
> 

Excellent!

> It's interesting the sorts of things you find when playing around with the widgets. One example.

Haha!  I guess I thought icons would all be square enough that it wouldn't matter.  At some point I'll have to do the same thing that LabelWidget does to recalculate the bounds.

Also shows the benefit of a widgetpedia for bug finding!

phatchman:
With the same easing function, the top label is not show (as I'd expect), but the bottom label is shown.
> I'm having trouble understanding what you mean. I am using this modification to examples/app.zig:
> 

Ahhh!!!! Don't worry. Had the start and end values swapped around and didn't realise. Sorry about that! I was using the same easing function for animate() and animation() but getting opposite alpha fades. But had start set to 1 instead of 0.

> > Also, I had to add this to get the "false" value re-displayed after the animation is complete.
> > ```
> > if (a.value() == a.end_val) {
> >     dvui.refresh(null, @src(), null);
> > }
> > ```
> 
> I can't find this, where did you have to add it? But yes - the current way animations trigger refreshes is implicit, so you get a frame where Animation.done() is true, but to get a frame after that you need an explicit refresh. I think that makes sense?

Yep. Makes sense. It's in struct_ui.boolFieldWidget. I had issue with using animate.done() previously, which is why I didn't use it here (I don't recall the exact circumstance), but I've changed to if (a.done()) and it is working fine.

david-vanderson:
Phew good!

phatchman:
> Also shows the benefit of a widgetpedia for bug finding!

That brings up one minor concern I have that people will create a lot of new issues from playing with it. 

There are lots of things you just shouldn't do, like enabling multi-line input on a combobox.
There are others that are minor visual bugs where the user uses large values for border and corner rounding and displays a background.
And others like the one I showed, where unless a real user actually needs it, then there are lots of other things that could be done instead.

But I guess we will find out.

I played around with a hacked-up version of a couple of new concepts this morning that look promising. But something to look at in the future. 
Just like we can copy options to the clipboard on the options editor. I can also do the same with struct_ui, where it will create the Zig code to populate a struct with the values from struct_ui.

But the nice bit is we can actually take that a step further and add a "copy" to the widget itself and generate zig code to create the widget with the init options and options from the widgetpedia.

For more complicated widgets it might be nice, you could generate a whole menu structure in widgetpedia and then generate the Zig code for it. Might not really be worth fully implementing, but as a quick hard-coded POC, it looks really cool.

phatchman:
Actually, I thought about it some more. 

The below demonstrates the issue I found. 

Would you expect these to have opposite alphas if the easing function only returns 0? I'd assumed that start_val through end_val would be passed to the easing function and the value from it returned as-is? This seems to be doing something else.

```
{
    var anim = dvui.animate(@src(), .{ .duration = 5_000_000, .kind = .alpha, .easing = struct {
        fn easing(_: f32) f32 {
            return 0;
        }
    }.easing }, .{});
    defer anim.deinit();
    dvui.labelNoFmt(@src(), "Alpha Zero", .{}, .{});
}
{
    var box = dvui.box(@src(), .{}, .{});
    defer box.deinit();
    if (dvui.firstFrame(box.data().id)) {
        dvui.animation(box.data().id, "trigger", .{ .start_val = 1, .end_val = 0, .start_time = 0, .end_time = 5_000_000, .easing = struct {
            fn easing(_: f32) f32 {
                return 0;
            }
        }.easing });
    }
    if (dvui.animationGet(box.data().id, "trigger")) |a| {
        const prev_alpha = dvui.alpha(a.value());
        defer dvui.alphaSet(prev_alpha);
        dvui.label(@src(), "Alpha Zero", .{}, .{});
    }
}
```

phatchman:
I've pushed a rough example for text entry. It's not complete, it's just to demonstrate the concept.

I basically just copied the text entries from the demo and turned them into different "Configurations". For the multiline one, I'd want to either hide "text" or have to default to collapsed.

Booleans can now be shown as checkboxes.

I think this where the limitation with struct_ui on field ordering starts to bite. For the multi-line example, we'd want the multi-line and break lines options close together etc.

image

image

So there's two options; 
1) Use the underlying struct_ui functions to display the individual fields or each scenario in the order we want. Which would keep the look consistent.
2) Bypass struct_ui and lay out with normal dvui widgets. 

So, I'll do some more experimenting.  I'm not really sure what the interface should look like. But I'm sure something will emerge!

phatchman:
I've customized the multi-line display of init options,to control the display order. "large", which is not implemented yet shows the original layout.

image

So my current theory is we keep improving struct_ui to get the type of display we need. Underneath it is just a set of functions that display fields in a standard way. 

```
.multiline => {
    struct_ui.displayBool(@src(), "multiline", &field_value_ptr.multiline, .{ .boolean = .{ .checkbox = true } }, &al);
    struct_ui.displayBool(@src(), "break_lines", &field_value_ptr.break_lines, .{ .boolean = .{ .checkbox = true } }, &al);
    struct_ui.displayOptional(@src(), T, "scroll_vertical", &field_value_ptr.scroll_vertical, 0, .defaultBool, .{}, &al, false);
    struct_ui.displayOptional(@src(), T, "scroll_vertical_bar", &field_value_ptr.scroll_vertical_bar, 0, .default, .{}, &al, .auto);
    struct_ui.displayOptional(@src(), T, "scroll_horizontal", &field_value_ptr.scroll_horizontal, 0, .defaultBool, .{}, &al, false);
    struct_ui.displayOptional(@src(), T, "scroll_horizontal_bar", &field_value_ptr.scroll_horizontal_bar, 0, .default, .{}, &al, .auto);
},
```

I've made a few more improvements to how stuct_ui handles nulls and read-only values. It's definitely getting nicer to use.

david-vanderson:
Wow all of this looks great!

> Would you expect these to have opposite alphas if the easing function only returns 0? I'd assumed that start_val through end_val would be passed to the easing function and the value from it returned as-is? This seems to be doing something else.

The idea is that the easing functions all take 0-1 and output 0-1 (sometimes a bit beyond) and then the output is lerped between the start_val and end_val.  I suppose we could change it but then the easing functions would have to take 3 args (start, end, t).  Why did you want to override the easing function output?

phatchman:
> The idea is that the easing functions all take 0-1 and output 0-1 (sometimes a bit beyond) and then the output is lerped between the start_val and end_val. I suppose we could change it but then the easing functions would have to take 3 args (start, end, t). Why did you want to override the easing function output?

I think this is just a documentation issue then. My mental model was that over the period of start_time to end_time, the normalized value beween start_val and end_val would be passed to the easing function. And given my function just returns 0, it seemed to me that it should not matter what start_val or end_val are, as value should always be the return value of the easing function, i.e. 0.

But how it works now makes absolute sense, as of course the easing function should only take time as a parameter. But unless you have that background knowledge, I don't think it is obvious how the parameters map onto the easing function. At least it wasn't for me. Probably what tripped me up is that start and end value were between 0 and 1, but start and end time were not, so I didn't make the connection.

And I did what I think most people do in that I copied the code form the examples and it has the non-intuitive configuration of start = 1, end = 0.

I can add some doc comments as part of this PR based on your description above.

david-vanderson:
> And I did what I think most people do in that I copied the code form the examples and it has the non-intuitive configuration of start = 1, end = 0.

Which example, and is there a way we could improve it?

> 
> I can add some doc comments as part of this PR based on your description above.

That would be helpful thank you!

phatchman:
> Which example, and is there a way we could improve it?
> 
Both the "Enter Pressed" and "Changed" animations do this:
```
dvui.animation(hbox.data().id, "value_changed", .{ .start_val = 1.0, .end_val = 0, .start_time = 0, .end_time = 500_000 });
```
So, they are doing what they are supposed to. Can't think of anything you'd do outside of a comment.

phatchman:
Was thinking I'll finish off the text entry example, then turn what we have now into a cleaned-up PR ready for merging.

Then can fill out the rest of the widgetpedia a few widgets at a time in separate PRs?

Edit: Just pushed out "finished" text entry. Not sure what to do about syntax highlighting, whether to include it here or separately. But I'll look into that later.

The options editor has had a few additions to the "Style" section to change font and the box_shadow. They need some polishing, but work well enough I think.

Also, see what you think of labelClick().

david-vanderson:
Looking good!

> Then can fill out the rest of the widgetpedia a few widgets at a time in separate PRs?

Yeah that makes sense, no reason to front-load all the effort.  Let's put this in as a separate demo button section?

Feedback:
* in animate() selecting the very last easing function in the dropdown crashes?
* I like the textEntry section where the options shown are specific to the configuration
* I noticed in the textEntryNumber the min/max switch to null automatically when you delete all the text - very nice
* in labelClick() I like showing the click event, but it's a bit jarring when it first appears, maybe we could have it start there with "null"?

phatchman:
Were you thinking of adding it like this? (as per latest commit). I'm concerned about the available width and height.

image

david-vanderson:
> Were you thinking of adding it like this? (as per latest commit). I'm concerned about the available width and height.

Yes, but you're right - that doesn't look good.  And probably a separate window makes sense if we ever figure out how to put links/buttons next to stuff in the demo that changes what the widgetpedia is showing.

How about a new button along the top of the demo window?  We could change the "Zoom In" and "Zoom Out" to be buttonIcons to give more horizontal room for the new button?

Screenshot 2026-03-01 at 9 35 23 PM

phatchman:
I did a hacky version where just the widgetpedia (latest commit) that has a bigger window (doesn't work with the expanded pane, but could). But I think your suggestion works better.

david-vanderson:
We could also bump up the default size of the app window and the demo window, but probably we want the widgetpedia in a separate window anyway (like the icon browser).

phatchman:
I think this works:
image

david-vanderson:
I'm good with this being merged - looks great!

phatchman:
> * in animate() selecting the very last easing function in the dropdown crashes?

This was a weird one. If there are exactly enough entries in the enum to fill the backing integer, dropDownEnum fails with an integer overflow when adding 1 to accommodate the null.

Fix is below, but something to be mindful of if you ever manipulate the result of @intFromEnum.
```
if (init_opts.null_selectable) @as(usize, @intFromEnum(ch.*.?)) + 1 else @intFromEnum(ch.*.?)
```

phatchman:
I've cleaned up the PR as #753.

I don't have a good solution the transition between no click event and having a click event yet for labelClick(). There isn't really an "empty" event that can be created there. I'm sure something will come to me.

david-vanderson:
> Fix is below, but something to be mindful of if you ever manipulate the result of @intFromEnum.

Good lesson.

> I don't have a good solution the transition between no click event and having a click event yet for labelClick(). There isn't really an "empty" event that can be created there. I'm sure something will come to me.

Cool no problem.  Thank you!

PR #738, opened 2026-02-20
Add a delay init_option to FloatingTooltipWidget

### LINKED ISSUE #731: tooltip: Add an option to delay before showing
This enhancement would add a `delay` init_option to provide s small delay before the tooltip is shown.

Currently, tooltips display as soon as the mouse enters the active_rect.

tooltips usually require the user to have the mouse stationary for a small period of time before the tooltip shows to avoid tooltips being displayed unnecessarily.

phatchman:
- Add a delay option to init_options
- Hide the tooltip until 80% of delay microseconds has passed
- Linearly fade in the tooltip over the remaining time.

Fixes #731

NOTE: The second commit is not intended to be merged. It is only to demonstrate the functionality of the first commit and will be dropped prior to merging.

### DIFF
--- a/src/widgets/FloatingTooltipWidget.zig
+++ b/src/widgets/FloatingTooltipWidget.zig
@@ -47,6 +47,9 @@ pub const InitOptions = struct {
 
     /// Is true if the user should be able to hover the tooltips content without it disappearing
     interactive: bool = false,
+
+    // delay the tooltip display for `delay` microseconds. Tooltip will fade in after 80% of `delay`
+    delay: ?i32 = null,
 };
 
 parent_tooltip: ?*FloatingTooltipWidget = null,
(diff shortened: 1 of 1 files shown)


phatchman:
I added a delay 1 second to the existing button widget with tooltip example and tooltips with a 1 sec delay on the 2 checkboxes under the "gray" button, just to try out the change. (another use for widgetpedia in future!)

I think this works OK at 1 second. 0.5 seconds and the tooltip was on the verge of showing with a normal interaction and at 2 seconds it felt like it was never going to show up. Somewhere between 1-1.5 seconds seems to work. 

You do occasionally get the effect of it popping up just as you are leaving the window and then disappearing really quickly. I'll do some testing with different fade-in periods to see if that helps, but see what you think.

This would work well enough for my purposes and keeping it out of the demo should discourage use unless someone really needs it.


Edit: After some testing, I think easing in after 80% of the time has elapsed works better. It gets rid of the "flicker" when you are unlucky with the timing. Also increased the delay to 1.2 seconds, so the fade-in should start at 1 second and last 0.2 seconds. I've updated the commits to implement this.

Regardless of whether we merge this, it is pretty cool that this works just by combining existing widgets!

david-vanderson:
That looks better than I expected!  Merge when ready.  Thank you!

PR #742, opened 2026-02-21
Added backend for OpenGL and GLFW

Carboxide68:
Added a backend for OpenGL and GLFW. There are a few things missing, such as saving of OpenGL state and then restoring it after rendering as well as content scale. Opening urls is also not supported currently. I have tested the backend on a personal project, as well as on the two created examples, but no heavy testing has been done.

### DIFF
--- a/src/backends/glfw-opengl.zig
+++ b/src/backends/glfw-opengl.zig
@@ -256,7 +256,9 @@ pub fn windowSize(ctx: *@This()) dvui.Size.Natural {
 }
 
 pub fn contentScale(ctx: *@This()) f32 {
-    return ctx.window.getContentScale()[0];
+    _ = ctx;
+    // Figure out what to do here
+    return 1;
 }
 
 pub fn drawClippedTriangles(


david-vanderson:
Very cool!  I'm looking into this.

david-vanderson:
I'm getting an error about linking GL on both Mac and Windows:
```
error: error: unable to find dynamic system library 'GL' using strategy 'paths_first'. searched paths:
```

Does that ring any bells?

I'm getting a self-hosted error on Linux saying `compiler backend 'stage2_x86_64' does not support tail calls'`.  I'll look into forcing llvm here.

The -standalone example uses dvui.App, so that should be an "-app" example.  Is there a reason you can't use `examples/app.zig`?

Should the backend be named "glfw_opengl"?  Is there any chance we'd have a different opengl version that forces us to make a whole new backend (as opposed to an option for this backend)?

I'm totally fine merging before things are totally complete and working on it in main, but I'd like to at least sort out the 'GL' thing first.

Carboxide68:
I forgot to add linking of OpenGL differently depending on Linux and Windows, don't know the specifics for MacOS though. The library should only use OpenGL3.3 functions, which is usually the lowest supported version so there should be no reason to make multiple versions of the library for different opengl versions since they're all backwards compatible.

The tail-call problem is an upstream problem with zgl, which would require us to either recompile a custom version of it for ourselves if we wanted, but might create problems when people try to bring their own zgl. Maybe not though...

I must've misunderstood what `standalone` was, but you're correct in that we could simply use `App.zig`.

I'll come with a commit that fixes these hopefully tomorrow.

Carboxide68:
Latest commit on my branch hopefully fixes all issues but the tail_call one, with some additional polishing and a little refactoring added.

Currently, frame-scheduling doesn't seem to work and I haven't had time to look into it. Relevant code should be at
https://github.com/Carboxide68/dvui/blob/0e2dd547d639bd6635602b0325dcfc3b4f2b66bf/src/backends/glfw-opengl.zig#L476-L479

david-vanderson:
I'm able to run this on Linux by adding .use_llvm = true to the addExecutable in build.zig.

Mac: I've updated the build.zig to use linkFramework, but I'm still getting errors about not having an opengl context.  I'll look into this a bit, but I know opengl is deprecated on Mac.

Windows: I'm getting errors finding only a few opengl functions: glGetnTexImage and glGetnUniformdv.  Some searching suggests these aren't always availble?  (My test laptop has integrated Intel graphics)

Carboxide68:
Thank you! This is really helpful, I had a difficult time figuring out which zgl-abstracted functions matched to 3.3, and which mapped to higher, since I didn't find anything that validated the version. I'll come with a commit fixing these instances hopefully later today.

I thought MacOS would work from the [Khronos](https://wikis.khronos.org/opengl/Programming_OpenGL_on_macOS) docs, maybe the linking is incorrect?

david-vanderson:
This is now working for me on Mac - thanks!  I'll do some more testing on Windows and Linux.

Is there anything else you want to do before merging?

Carboxide68:
I've got it working on all platforms myself, so I'm happy merging it if you are. I will try and fix the constant mouse event issue I described on Discord later but I wouldn't consider it blocking.

david-vanderson:
On my windows machine I get empty green (the fill color) windows, and no widgets show up.  Will continue investigating, but no reason to block merging.

Thank you very much!

david-vanderson:
I'm writing this up in the dvui devlog, and I'm going to credit you as "Jakob Ingvast".  Let me know if you want something else.

PR #744, opened 2026-02-23
build: switch SDL3 dependency

brodo:
This PR switches the SDL3 package to [this one](https://codeberg.org/brodo/SDL). We need this for [the Awebo CI process](https://codeberg.org/awebo-chat/awebo/issues/69). I've opened a [PR](https://github.com/castholm/SDL/pull/37) in the original repo, so we might be able to switch back in the future.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -11,8 +11,8 @@
             .lazy = true,
         },
         .sdl3 = .{
-            .url = "git+https://github.com/castholm/SDL#887fa805e40d5c7bb3955ae69b89ed566c6a0471",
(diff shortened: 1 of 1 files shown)


david-vanderson:
I've pushed this commit to the zig16-dev branch, which is what Awebo is using.  Does that work for you?

brodo:
Yes that works. Currently Awebo is using Andrews fork of DVUI, but I'll do a PR tomorrow that uses the branch.

david-vanderson:
This will come in when we merge the zig16-dev branch after zig16 is released.

PR #745, opened 2026-02-25
Allow transparent windows, and clear to transparent color to allow windows to use transparency

foxnne:
This should hopefully not affect anyone that isn't attempting to use a transparent window, but when nothing is drawn, should use the window transparency to draw behind the window. I'm not certain yet if this is supported on all platforms (titlebar transparency), but it seems to work well on macOS.


david-vanderson:
Those CI errors are my fault, I'll fix them sorry.

david-vanderson:
Works for me on Windows, Mac, and Linux (x11).  Thank you!

PR #746, opened 2026-02-26
add an example for drawing and erasing strokes with dvui

amoghyermalkar123:
Added an example for sdl2 backend for creating and erasing strokes via mouse drags.


david-vanderson:
I like this - since it doesn't use any backend-specific stuff I'd like to put it into the demo window somewhere (maybe sharing space with the calculator somehow).  If you are okay with it I'll leave this open until I'm able to do that.

Thank you!

amoghyermalkar123:
@david-vanderson ofcourse, Thanks!

david-vanderson:
I'm merging this into #749, hopefully I've kept the git history intact.

Thank you!

PR #747, opened 2026-02-26
App: Add option to start with transparent flag, currently only affect…

foxnne:
…s SDL backend

Sorry about breaking things a bit! I tested this on Windows and MacOS and it seems SDL3's behavior here is inconsistent. I haven't found any way with SDL3 to make a transparent titlebar/border on Windows, and on macOS the titlebar transparency covers the entire window. It seems it's impossible to currently support matching the titlebar and app body transparency, at least without a hacky solution using the native window pointers in SDL3. 

Anyway, I've just made the transparency flag optional so I can play with this in the future but it should return things to previous behavior.

Thanks!

### DIFF
--- a/src/App.zig
+++ b/src/App.zig
@@ -85,6 +85,8 @@ pub const StartOptions = struct {
     icon: ?[]const u8 = null,
     /// use when running tests
     hidden: bool = false,
+    /// Set the window to be transparent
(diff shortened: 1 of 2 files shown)


david-vanderson:
Great - thanks!

PR #749, opened 2026-02-27
texture target: clearing and explicit destroying

### LINKED ISSUE #748: renderTarget: function to clear
@foxnne asked for a way to clear a render target so it can be reused.  I think the general strategy is to make that target the current renderTarget, then fill a rect that matches the target size.

This issue is to make that a function on renderTarget so it's easy.

david-vanderson:
fixes #748

The idea is to move towards a place where an app could reuse a texture target over time.

This adds:
* Texture.Target.clear() for resetting a texture target back to full transparency
* Texture.Target.destroyLater() for lazily destroying a texture target

Still need:
- [x] update other backends (besides sdl)
- [x] move my hacked testing code from app.zig into the demo somewhere
- [x] figure out what to do about `textureFromTarget`


david-vanderson:
The problem with `textureFromTarget()` is that on most backends it returns the same texture, but on dx11 it returns a newly-created texture.  This makes ownership/lifetime of the returned value tricky.

After some thought, I think there are 2 categories of use for this:
* "convert this target to a normal texture for the future" (this is what CacheWidget wants)
* "just want to draw this target" (this is what dvui.Picture wants)

Proposal: split into 2 functions:
* Target.textureConvert() - this logically destroys the target, usage code owns and must destroy the returned texture
* Target.textureTemp() - the returned texture has already been marked for destruction via `destroyLater()`, so it's only good for rendering in the current frame

@foxnne Does this sound reasonable to you?

foxnne:
> @foxnne Does this sound reasonable to you?

Yes it does! Thank you so much for looking into this!

david-vanderson:
I kept `textureFromTarget` with the existing (destroy target) semantics, and added `Texture.fromTargetTemp` with the new (temporary texture, target not destroyed) semantics.

The demo now has 3 applets where the calculator used to be:
* calculator
* drawing (from @amoghyermalkar123 that I modified a bit)
* texture (showing accumulating a long-lived target texture)

david-vanderson:
The sdl3gpu and dx11 backends are not hanging on to target texture contents like I expect.  That will have to be dealt with later.

foxnne:
> The sdl3gpu and dx11 backends are not hanging on to target texture contents like I expect. That will have to be dealt with later.

I should have tested this sooner, but it seems SDL3 uses the same backend for windows? i.e. the target texture becomes blank for me on windows but works on macOS. I think for now I'll have to have a separate strategy for windows.

PR #753, opened 2026-03-02
Introduce Widgetpedia - a widget reference and playground

phatchman:
Add a widget reference to showcase all dvui widgets and allow users to interact with them.
Also provides a place for developers to test changes to widgets under various conditions.

This PR replaces #735


david-vanderson:
Thank you!

PR #758, opened 2026-03-03
make enums_backend.Backend public to build.zig consumers

MichaelBelousov:
I think this is necessary if you use an `std.Build.option` to choose your DVUI backend, since it becomes a non-comptime resolution, e.g.:

```zig

const test_use_gui = b.option(bool, "test_use_gui", "use a graphical backend, no screenshot tests if disabled") orelse true;

const dvui_testing_dep = b.dependency("dvui", .{
    .target = target,
    .optimize = optimize,
    .backend = if (test_use_gui) dvui.Backend.sdl3 else dvui.Backend.testing,
    .@"snapshot-images" = if (test_use_gui) "before" else null,
});
``` 

Sorry for not making an issue first, I feel like this description with code is OK, but lmk if you'd have preferred a less code-forward discussion first.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,5 +1,6 @@
 const std = @import("std");
 const enums_backend = @import("src/enums_backend.zig");
+pub const Backend = enums_backend.Backend;
 const Pkg = std.Build.Pkg;
 const Compile = std.Build.Step.Compile;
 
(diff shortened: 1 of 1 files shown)


david-vanderson:
Looks great - thanks!

PR #759, opened 2026-03-04
add dvui build utility `svgPathToTvgPath` with inline documentation for using custom SVGs more easily in build scripts

MichaelBelousov:
Putting this first pass up for discussion.

If this looks good, I intend to add a test in the DVUI code base, possibly usage in an example, to confirm it works.
I have confirmed it works in my own DVUI application, and replaces the local utility I was using.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1111,3 +1111,50 @@ fn addWebExample(
 
     b.getInstallStep().dependOn(compile_step);
 }
+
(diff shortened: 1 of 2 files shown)


MichaelBelousov:
I forgot this is branched off of #758 so there is a small changeset included there

david-vanderson:

PR #761, opened 2026-03-04
Pass options when using `addTabLabel`

Didas-git:
The user needs to be able to pass options to customize the tab, for example remove the rounded corners or the margin


david-vanderson:
Good idea - thank you!

PR #762, opened 2026-03-04
Widgetpedia2

phatchman:
Additional widgets and improvements.


phatchman:
This implements all of the "trivial" widgets and a few others.

I can't get the toast example to work with sub_window_id and I don't know why.

I've checked that the currentWindow().subwindows.current_id is the same as the floating window id.
I've also tried setting it to a box containing the button / toast and it still doesn't display.

Any ideas?

foxnne:
Just wanted to say I finally got to check out the new Widgetpedia stuff and its great! I really like this and how easily it is to explore the widgets and different capabilities. Great job!

david-vanderson:
> This implements all of the "trivial" widgets and a few others.
> 
> I can't get the toast example to work with sub_window_id and I don't know why.
> 
> I've checked that the currentWindow().subwindows.current_id is the same as the floating window id. I've also tried setting it to a box containing the button / toast and it still doesn't display.
> 
> Any ideas?

This is my fault.  The way I did it was to require a subwindow that wanted to display toasts to call a line like this:
`dvui.toastsShow(float.data().id, .cast(float.data().rect));`

This makes sense to be able to target toasts to things other than subwindows, but it clearly fails the expectations test.

I think the solution is that `FloatingWindowWidget` should automatically call this at some point, so that the code you wrote (nicely using the current subwindow id) would work.  That would clean up the other demo code as well.

Does that sound right?

phatchman:
> Does that sound right?

Yes. That makes sense to me. Then you'd only have to do the call manually if you are targeting something other than a subwindow.

david-vanderson:
Love it - I especially appreciate the attention to detail, like having the border option in the spacer so you can see what is going on.

I've added a commit to fix the toasts issue.  If that looks good to you, then merge when you are ready.  Thank you!

david-vanderson:
I merged main because there was a failed ci.  But looking at it closer, somehow the ci ran the wrong branch?  Not sure what happened there, but I think this is sorted now.

PR #765, opened 2026-03-06
widgetpedia: add menu builder

phatchman:
If you have some time would appreciate your thoughts on the menu builder in the `menu()` entry.

Is it obvious what each of the + and - do?
Is it weird that - deletes a top-level menu, but the + on the same level adds a sub-menu item?

Something just doesn't feel quite right to me about it.

I'm still trying to refine how this will work, but I'll prob move this to menuItemLabel() and have a similar one for the icons etc. And then options would be applied to all the menu items created.

### DIFF
--- a/src/Examples/widgetpedia.zig
+++ b/src/Examples/widgetpedia.zig
@@ -110,8 +110,7 @@ pub fn widgetpedia() void {
                         defer branch_child.deinit();
                         dvui.labelNoFmt(@src(), child.name, .{}, .{ .expand = .horizontal });
                         if (branch_child.button.clicked()) {
-                            current_widget = child;
-                            reset_widget = true;
+                            setCurrentWidget(child);
                         }
                     }
                 }
@@ -126,6 +125,11 @@ pub fn widgetpedia() void {
     }
 }
 
+fn setCurrentWidget(widget: WidgetHierarchy) void {
+    current_widget = widget;
+    reset_widget = true;
+}
+
 // Only supports 2 levels, parent and children.
 const WidgetHierarchy = struct {
     name: []const u8,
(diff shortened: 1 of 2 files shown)


phatchman:
I've added the other menu item widgets.

I'll probably turn menu() into something simpler that has one of every menu item type, with recursive submenus or something like that. And keep the menu builder in menuItemLabel().

The submenu init_opt is a little tricky to represent because it needs to be calculated per menu item. At the moment it just gets the init_opt the user sets gets anded with whether the menu item has children to work out the real value.

david-vanderson:
Menu builder is awesome.

> Is it obvious what each of the + and - do?

I wouldn't worry about making it more obvious.  I only had to click on them once to understand.

> Is it weird that - deletes a top-level menu, but the + on the same level adds a sub-menu item?

Could you put all the + below the menu (like how the top-level + is below on its own row)?

Other random stuff:
* maybe start the menu builder off with a few prepopulated things?
* menuItemIcon() has "tooltip()" for the group box label
* can we make the menu background the "normal" gray?

phatchman:
> * can we make the menu background the "normal" gray?

What are you seeing as different here? I'm not setting any options by default.

image

image

Below are for the light theme.

image
image

david-vanderson:
Sorry should have been more specific.  This is what I'm seeing when using keyboard navigation with "focus as outline" and the light theme:
Screenshot 2026-03-06 at 7 57 22 PM

This is because the icon is drawn white, but the background is also white.  My knee-jerk reaction was to ask for the background (of a box inside the groupbox maybe?) to have the `.window` style, since that is usually what widgets get put on top of.

I'm not totally sure that's the right answer, maybe menuItem with "focus as outline" is buggy where it's not drawing the right background?  Will look at it.

david-vanderson:
> I'm not totally sure that's the right answer, maybe menuItem with "focus as outline" is buggy where it's not drawing the right background? Will look at it.

This is a bit of a mess right now.  "focus as outline" was added to make dropdowns work.  Ignore this for now, and I'll come back to it at some point.  Sorry for the noise!

phatchman:
All good. I've pushed a commit to address the other issues.

david-vanderson:
Looks great - merge when ready.  Let me know if you want me to look at anything specific.  Thank you!

PR #769, opened 2026-03-08
Widgetpedia4

phatchman:
widgetpedia:
- Add a splitter between the widget box and widget controls
- floatingWindow()
- windowHeader()
- paned()
- scrollArea (in progress)

Options editor:
- Default color chooser to the theme color for each null color on the Style tab, previously the color picker defaulted to white for all null Style values.

### DIFF
--- a/src/Debug.zig
+++ b/src/Debug.zig
@@ -834,10 +834,13 @@ fn stylePage(self: *Options, id: dvui.Id) bool {
             var vbox = dvui.box(@src(), .{}, .{});
             defer vbox.deinit();
 
-            const field: *?dvui.Color = switch (active_color.*) {
-                inline else => |c| &@field(self, "color_" ++ @tagName(c)),
+            const field: *?dvui.Color, const default: dvui.Color = switch (active_color.*) {
+                inline else => |c| .{
+                    &@field(self, "color_" ++ @tagName(c)),
+                    dvui.themeGet().color(.control, std.meta.stringToEnum(dvui.Options.ColorAsk, @tagName(c)) orelse unreachable),
+                },
             };
-            var hsv = dvui.Color.HSV.fromColor(field.* orelse .white);
+            var hsv = dvui.Color.HSV.fromColor(field.* orelse default);
             if (dvui.colorPicker(@src(), .{ .hsv = &hsv, .dir = .horizontal }, .{})) {
                 changed = true;
                 field.* = hsv.toColor();
--- a/src/Debug.zig
+++ b/src/Debug.zig
@@ -493,11 +493,11 @@ fn copyOptionsToClipboard(id: dvui.Id, options: Options) void {
     dvui.clipboardTextSet(aw.written());
 }
 
-fn sliderRectOptional(src: std.builtin.SourceLocation, comptime label: []const u8, rect: *?Rect, comptime field: std.meta.FieldEnum(dvui.Rect), link_all: bool, default: dvui.Rect) bool {
-    return sliderRectOptionalWithInitOpts(src, label, rect, field, link_all, default, null);
+fn sliderRectOptional(src: std.builtin.SourceLocation, comptime label: []const u8, comptime fmt: []const u8, rect: *?Rect, comptime field: std.meta.FieldEnum(dvui.Rect), link_all: bool, default: dvui.Rect) bool {
+    return sliderRectOptionalWithInitOpts(src, label, fmt, rect, field, link_all, default, null);
 }
 
-fn sliderRectOptionalWithInitOpts(src: std.builtin.SourceLocation, comptime label: []const u8, rect: *?Rect, comptime field: std.meta.FieldEnum(dvui.Rect), link_all: bool, default: dvui.Rect, init_opts: ?dvui.SliderEntryInitOptions) bool {
+fn sliderRectOptionalWithInitOpts(src: std.builtin.SourceLocation, comptime label: []const u8, comptime fmt: []const u8, rect: *?Rect, comptime field: std.meta.FieldEnum(dvui.Rect), link_all: bool, default: dvui.Rect, init_opts: ?dvui.SliderEntryInitOptions) bool {
     var changed: bool = false;
     var hbox = dvui.box(src, .{ .dir = .horizontal }, .{ .min_size_content = .{ .w = 160 }, .max_size_content = .width(160) });
     defer hbox.deinit();
@@ -525,7 +525,7 @@ fn sliderRectOptionalWithInitOpts(src: std.builtin.SourceLocation, comptime labe
 
         if (dvui.sliderEntry(
             @src(),
-            label ++ ": {d:0.1}",
+            label ++ ": " ++ fmt,
             slider_init_opts,
             .{ .margin = .{ .x = 0, .y = 4, .w = 4, .h = 4 }, .gravity_y = 0.5 },
         )) {
@@ -571,6 +571,7 @@ fn layoutPage(self: *Options, id: dvui.Id, wd: *const dvui.WidgetData) bool {
         changed = sliderRectOptionalWithInitOpts(
             @src(),
             "rotation",
+            "{d:0.2}",
             &rot_rect,
             .x,
             false,
@@ -595,7 +596,7 @@ fn layoutPage(self: *Options, id: dvui.Id, wd: *const dvui.WidgetData) bool {
         defer row.deinit();
 
         var has_min_size = self.min_size_content != null;
-        if (dvui.checkbox(@src(), &has_min_size, "min size", .{ .gravity_y = 0.5, .min_size_content = .{ .w = 90 } })) {
+        if (dvui.checkbox(@src(), &has_min_size, "min_size_content", .{ .gravity_y = 0.5, .min_size_content = .{ .w = 90 } })) {
             if (self.min_size_content) |_| {
                 self.min_size_content = null;
             } else {
@@ -619,7 +620,7 @@ fn layoutPage(self: *Options, id: dvui.Id, wd: *const dvui.WidgetData) bool {
         defer row.deinit();
 
         var has_max_size = self.max_size_content != null;
-        if (dvui.checkbox(@src(), &has_max_size, "max size", .{ .gravity_y = 0.5, .min_size_content = .{ .w = 90 } })) {
+        if (dvui.checkbox(@src(), &has_max_size, "max_size_content", .{ .gravity_y = 0.5, .min_size_content = .{ .w = 90 } })) {
             if (self.max_size_content) |_| {
                 self.max_size_content = null;
             } else {
@@ -644,20 +645,20 @@ fn layoutPage(self: *Options, id: dvui.Id, wd: *const dvui.WidgetData) bool {
         { // Top Left
             var col = dvui.box(@src(), .{}, .{ .border = .all(1), .expand = .vertical });
             defer col.deinit();
-            changed = sliderRectOptional(@src(), "radius", &self.corner_radius, .x, link_radius.*, wd.options.corner_radiusGet()) or changed;
+            changed = sliderRectOptional(@src(), "corner_radius", "{d}", &self.corner_radius, .x, link_radius.*, wd.options.corner_radiusGet()) or changed;
         }
         { // Top Center
             var col = dvui.box(@src(), .{}, .{ .border = .all(1) });
             defer col.deinit();
-            changed = sliderRectOptional(@src(), "margin", &self.margin, .y, link_margin.*, wd.options.marginGet()) or changed;
-            changed = sliderRectOptional(@src(), "border", &self.border, .y, link_border.*, wd.options.borderGet()) or changed;
-            changed = sliderRectOptional(@src(), "padding", &self.padding, .y, link_padding.*, wd.options.paddingGet()) or changed;
+            changed = sliderRectOptional(@src(), "margin", "{d:0.0}", &self.margin, .y, link_margin.*, wd.options.marginGet()) or changed;
+            changed = sliderRectOptional(@src(), "border", "{d:0.0}", &self.border, .y, link_border.*, wd.options.borderGet()) or changed;
+            changed = sliderRectOptional(@src(), "padding", "{d:0.0}", &self.padding, .y, link_padding.*, wd.options.paddingGet()) or changed;
         }
         { // Top Right
             var col = dvui.box(@src(), .{}, .{ .border = .all(1), .expand = .vertical });
             defer col.deinit();
 
-            changed = sliderRectOptional(@src(), "radius", &self.corner_radius, .y, link_radius.*, wd.options.corner_radiusGet()) or changed;
+            changed = sliderRectOptional(@src(), "corner_radius", "{d:0.2}", &self.corner_radius, .y, link_radius.*, wd.options.corner_radiusGet()) or changed;
         }
     }
 
@@ -667,9 +668,9 @@ fn layoutPage(self: *Options, id: dvui.Id, wd: *const dvui.WidgetData) bool {
         { // Middle Left
             var col = dvui.box(@src(), .{}, .{ .border = .all(1) });
             defer col.deinit();
-            changed = sliderRectOptional(@src(), "margin", &self.margin, .x, link_margin.*, wd.options.marginGet()) or changed;
-            changed = sliderRectOptional(@src(), "border", &self.border, .x, link_border.*, wd.options.borderGet()) or changed;
-            changed = sliderRectOptional(@src(), "padding", &self.padding, .x, link_padding.*, wd.options.paddingGet()) or changed;
+            changed = sliderRectOptional(@src(), "margin", "{d:0.0}", &self.margin, .x, link_margin.*, wd.options.marginGet()) or changed;
+            changed = sliderRectOptional(@src(), "border", "{d:0.0}", &self.border, .x, link_border.*, wd.options.borderGet()) or changed;
+            changed = sliderRectOptional(@src(), "padding", "{d:0.0}", &self.padding, .x, link_padding.*, wd.options.paddingGet()) or changed;
         }
         { // Middle Center
             var col = dvui.box(@src(), .{ .dir = .horizontal }, .{ .border = .all(1), .expand = .both });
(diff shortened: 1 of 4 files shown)


phatchman:
@david-vanderson For the scrollArea() widget, I need to restrict some of the inputs in scroll_info depending on the scroll options. e.g. only allow the user to enter a value in scroll info if that axis is .given.

I'm confused about if .given is a valid value for the ScrollMode in init_opts.horizontal_bar and .vertical_bar. 
ScrollInfo also has a ScrollMode for .horizontal and .vertical, which are mandatory, and dvui complains if a scroll_info is provided when .horizontal/vertical_bar is also set.

So, I don't think it is ever valid to set .horizontal/vertical_bar to .given as it will always conflict with the scroll_info, even if the scroll_info values are also set to .given. 

If you want to see what I mean use the scrollArea() demo, set scroll_info to not null, set the axis to .given and also set .horizontal_bar and .vertical_bar to .given. Then see the debug log messages.

Not saying any of the above needs to change, just want to make sure I understand so I can work out the right UI.

Edit: I've done a bit more work on the scrollArea options. See what you think of the new widget for "Velocity" in scroll_info. I need to make it more discoverable that you have to press "enter" to apply the velocity number, unlike the other number entries that apply immediately. Probably all the struct ui number entries should only apply a new value on "enter".

Also I've done something custom for setting the x and y viewport values when you set the scroll info horizontal / vertical to .given. 

Edit 2: I've pushed some more changed. It still needs some work.
For the focused_id, should it auto-scroll to the focused widget? Or does it just set the focus?
Ahh I think I see. It only focuses the widget if you click on the scroll bar? Hmm. Need to think more about how to present that. What is that used for?

hpcfzl:
Great progress so far! Here's my experience using this as of yet, jotting down as you're making changes from `0aa22c9`:
* As the top-left pane in showcases is usually quite oversized for what its contents are, the default starting view could be changed to have the options expanded, like so:

  widgetpedia-0aa22c9-view
  Although I did remember the https://github.com/david-vanderson/dvui/pull/735#issuecomment-3955191471, I thought I would mention this anyway.
* When I enlarge the floating window, the size of icon widgets in the `paned` showcase causes frame drops. Only **while resizing** to a larger size though, so might not be that important:
  <video src="https://github.com/user-attachments/assets/c847aafe-859a-4a55-83e5-3721669d6c2c"/>
* The `Style` tab of options reminds me of the debugger's. So same as there, I'm missing `.style` and `.theme`, wouldn't they be useful to have? In fact, more of the general options seem to be missing, is there a reason behind that?
* Again in the `Style` tab, I think `font` could have its settings named more after the actual options as done elsewhere, Like what's the option name for changing the font, its size, and line height here?
  
  image

* Each button could use a brief description, for example as a `tooltip` in this area when `slider` is hovered over:

  image

  Or there could even be a dedicated pane containing description automatically taken from docs? That way, we could also improve docs for each widget. Though I don't know whether I can handle a fourth pane when we already have three.

* Why are these circles **green**? That tells me there's a way to turn one like `background` off, becoming **red**, but `background` obviously just disappears along with the dot:
  image

* Value names and `groupBox` titles could be using the `.mono` font. I'm aware not all themes come with one, so this could be something for the future.

* Any ETA when Widgetpedia might come to Web demo? I'm guessing the site just needs updating?

Also I believe you forgot to give `scrollArea` button a lighter grey WIP background other buttons currently have.

phatchman:
> Great progress so far! Here's my experience using this so far, jotting down as you're making changes from `0aa22c9`:
> 

Thanks very much for the feedback.

> * As the top-left pane in showcases is usually quite oversized for what its contents are, the default starting view could be changed to have the options expanded, like so:

It's tricky, but I *think* it is a good balance at the moment. There are some widgets that needs the extra space (like splitter, scroll area, text entry), where others don't. Open to what others think as well, but it seems fairly balanced to me at the moment.

> * The `Style` tab of options reminds me of the debugger's. So same as there, I'm missing `.style` and `.theme`, wouldn't they be useful to have? In fact, more of the general options seem to be missing, is there a reason behind that?

It is calling into the debugger's editor. So it is exactly the same. 

No reason. They just haven't been added yet. The only options I known that are missing are ninepatch, theme and style. Do you think there are others? For theme, I figure there is already the theme selector in the Demo, so adding it isn't urgent. Style is only really useful if someone has set up application-specific styles. And ninepatch, well that's just tricky.

> * Again in the `Style` tab, I think `font` could have its settings named more after the actual options as done elsewhere, Like what's the option name for changing the font, its size, and line height here?

Not sure what you mean? It is displaying the font struct you would pass to the .font option. Agree the font selector could be improved. It is just a modified version from the demo.

Edit: it's not showing weight or style. I can add those.

> * Each button could use a brief description, for example as a `tooltip` in this area when `slider` is hovered over:

Yes, there are various things that can be done here with documentation. But not really thinking that far ahead until we have all the base widgets done. 

> * Why are these circles **green**? That tells me there's a way to turn one like `background` off, becoming **red**, but `background` obviously just disappears along with the dot:

It just looked nicer to me than a block of white text. If people really find it confusing, I'll look at it, but I prefer it.

> * Value names and `groupBox` titles could be in the `.mono` font. I'm aware not all themes come with one, so this could be something for the future.

Is there a reason? We're very width limited and monspace will make that worse.

> * Any ETA when Widgetpedia might come to Web demo? I'm guessing the site just needs updating?

Yeah. I think DV does that manually?

> 
> Also I believe you forgot to give `scrollArea` button a lighter grey WIP background other buttons currently have.

There is a demo page for scrollArea, so it will have an active color. The colors are automatically generated based on whether it has the "under construction" display page or not.

Thanks again!

Edit: apologies if my responses come off a bit short. I was just about to head off to bed when you replied.

hpcfzl:
> No reason. They just haven't been added yet. The only options I known that are missing are ninepatch, theme and style. Do you think there are others?

You're right, it's just that there are multiple options for `ninepatch` alone which probably threw me off. Correction: while preparing the screenshot about `fonts`, I noticed `weight` is missing, so indeed there might be more missing? Or not. Though `.label` which is under `Info` has more fields; are they missing because they're not settable like everything else under this tab? I haven't tried those much yet so I wouldn't know.

> Style is only really useful if someone has set up application-specific styles.

Fair enough, but I often find myself having difficulty telling apart even the built-in themes' styles:
theme-styles

> Not sure what you mean? It is displaying the font struct you would pass to the .font option.

Specifically I was aiming for something like this:
image

> Is there a reason? We're very width limited and monspace will make that worse.

It's something more associated with code, at least compared to sans fonts. Also signifies that something is the actual option name:

option-naming

For example, seeing _min size_ and _max size_, then looking those names up, when in fact I believe it's `min_size_content` and `max_size_content`. This might not be so obvious to everyone. In this example, I would leave both names as-is or have their exact names in a `.mono` font.

While I understand the spacing argument, I couldn't find many instances of text where that's true:

image

Still mostly holds even after corresponding sliders and so on are revealed. Is that depictive enough?

> Edit: apologies if my responses come off a bit short.

No worries, timezones happen.

phatchman:
> 
> Specifically I was aiming for something like this: 

Ah I see. I didn't do it that way because the rest of the options editor doesn't look like that, and I'm deliberately not using struct_ui (which is formatting amost of the widgetpedia  options) to not make dvui.Window dependent on struct_ui. So, it is going to behave differently to the other options for that reason.

I think this also ties in with your other point that the options editor shows `min size` instead of `min_size_content`. I agree the labels should reflect the actual option name everywhere. As for the rest of the font section, I plan to come back to it at some point. I'm also not really happy with the way the expander works to set null / not null. But if anyone has the inclination to polish up the options editor, it would be welcome. I don't have any immediately plans outside of adding those fields we aren't displaying.

> > Is there a reason? We're very width limited and monspace will make that worse.
> 
> While I understand the spacing argument, I couldn't find many instances of text where that's true:

I think you just have to trust me on this one. I spent a lot of time tinkering to squeeze as much space as I could to fit in the demos and controls side by side, without making it look cramped. There's also been a lot of tinkering with struct_ui to make it suitable for narrower widths.

> Still mostly holds even after corresponding sliders and so on are revealed. Is that depictive enough?

Widgets like progress aren't the issue. Think menu, grids, flexbox, scrollarea etc. Even with things like comboBox we need enough horizontal space for gravity to be useful. With the box demo we need enough space for things like the equal_space option to show their effect. With the existing menu() we'd have to restructure the demo to use less horizontal space.

Also, I think you'd have to do a lot to convince me to move away from the symmetrical view we have now, especially now there is a pane to allow the user to adjust the size to their preference. 

Cheers.

david-vanderson:
> @david-vanderson For the scrollArea() widget, I need to restrict some of the inputs in scroll_info depending on the scroll options. e.g. only allow the user to enter a value in scroll info if that axis is .given.
> 
> I'm confused about if .given is a valid value for the ScrollMode in init_opts.horizontal_bar and .vertical_bar.
> ScrollInfo also has a ScrollMode for .horizontal and .vertical, which are mandatory, and dvui complains if a scroll_info is provided when .horizontal/vertical_bar is also set.
> 
> So, I don't think it is ever valid to set .horizontal/vertical_bar to .given as it will always conflict with the scroll_info, even if the scroll_info values are also set to .given.
> 
> If you want to see what I mean use the scrollArea() demo, set scroll_info to not null, set the axis to .given and also set .horizontal_bar and .vertical_bar to .given. Then see the debug log messages.

Yeah what's going on here is that I couldn't figure out a better way to pass the scroll mode.  If ScrollInfo is provided, we want to use that, but if ScrollInfo is not provided, then you need a way to control those.  So `.horizontal` and `.vertical` are duplicating those fields of ScrollInfo.

.horizontal_bar and .vertical_bar are different - they control if the scrollbar is shown.

Does that help?

> Edit: I've done a bit more work on the scrollArea options. See what you think of the new widget for "Velocity" in scroll_info. I need to make it more discoverable that you have to press "enter" to apply the velocity number, unlike the other number entries that apply immediately. Probably all the struct ui number entries should only apply a new value on "enter".

I like the (tooltip?) saying to press enter to change, seems natural!

> Also I've done something custom for setting the x and y viewport values when you set the scroll info horizontal / vertical to .given.

Very nice!

> Edit 2: I've pushed some more changed. It still needs some work. For the focused_id, should it auto-scroll to the focused widget? Or does it just set the focus? Ahh I think I see. It only focuses the widget if you click on the scroll bar? Hmm. Need to think more about how to present that. What is that used for?

ScrollArea's focus_id is only to pass to the Scrollbars, and it's only used by TextEntryWidget currently.  TextEntryWidget is complicated, because it can be focused, but it has a whole ScrollArea with scrollbars inside of it, and you don't want any of those components to get focused, only the top TextEntryWidget.  Maybe there's a better way to do it, but that's why it exists.

david-vanderson:
> > * Any ETA when Widgetpedia might come to Web demo? I'm guessing the site just needs updating?
> 
> Yeah. I think DV does that manually?

Yes - I always do it at the start of the month when writing the devlog.  Sometimes additionally for other reasons.  It can be useful to quickly check the web demo against whatever I'm currently hacking on.

Probably eventually it'll be automated, but it's working out ok so far.

@hpcfzl thanks for the detailed feedback!

@phatchman I'm good with these - merge when ready!  Thank you!

phatchman:
I've made changes to the options editor to align the field names with the options struct.

phatchman:
Added plot(), textEntryColor()

Various struct_ui enhancements:
1) You can now specify different field options for an optional vs it's value. i.e. this allows you to have a read_write optional field, but the value is read-only. Useful for const string optionals.
2) Can now restrict StructOptions to a field name. So if you have multiple fields with Point structs, you can now chose to show some of them as sliders, some of them as number_entry etc based on their field name in the parent struct.
3) Fixed up some union default value initialization bugs.

I'll finish off the next plot one and then prepare it for commit.

phatchman:
This adds:
- Rebase to main
- plotXY()
- structui: For container fields (arrays, structs, unions) control whether the container defaults to expanded or collapsed via FieldOptions. Used in PlotXY to keep the xs and rs collapsed. Also used in labelClick to stop the "last click event" display from being so jarring.

I'll give it another look over, but should be good to merge.

Yep looks good. Will leave open in case you want to test.

phatchman:
Widgetpedia addtions, Debug and struct_ui enhancements

Debug
- Allow configurable number formats for Style sliders
- Make label names consistent with underlying dvui options names

struct_ui
- Add `constant` display options which behaves like `read_only` but recursively propagates to child fields.
- Add `default_expanded` FieldOption to control whether container fields (structs, arrays, unions) are displayed expanded or collapsed.
- Add OptionalFieldOptions to allow different FieldOptions to be specified for an optional vs its value.
- Add `entry_on_enter` widget_type for NumberFieldOptions. This widget type only sets the field value when the enter key is pressed.
- Allow StructOptions to apply per field_name. This allows multiple StructOptions for the same type to be used when displaying a struct. e.g. to display some Points as Text Entries and others as sliders. 

Widgetpedia
- Added a horizontal splitter between the widget display box and the widget controls box
- Remove some unused variables.
- flexbox()
- floatingWindow()
- windowHeader()
- paned()
- plot()
- plotXY()
- scrollArea()
- sliderVector()
- textEntryColor()
- LabelClick() uses the collapsed struct_ui to display the `last_click_event`

hpcfzl:
Nicely progressing. Since this is essentially a demo inside demo, I have been pondering about documenting the addition of Widgetpedia in the README: *Find the **Widgetpedia** button in the demo for a condensed overview of all widgets.*

Also documenting the new `struct_ui`-related flag which was introduced along the way, as well as other existing flags. Coincidentally, I see [the latest commit](https://github.com/david-vanderson/dvui/commit/39e87cc932dbfa3c399bcbc98d9245388ed789d3) to DVUI is very much relevant.

By the way, what's the interaction with these checkboxes in `scrollArea`? I have seen them before, but I though that was part of WIP:
image

phatchman:
@hpcfzl yeah, I'm not sure whether to just remove those? There is a focus_widget init_option you can set to either of those checkbox id's. If you click on the scrollbar, it will focus that widget.

I think DV said it is only used to focus the the textlayout in a text entry. No one is really going to discover that just by clicking around. 

Any thoughts? Maybe hide that option and the checkboxes?

Re readme updates. Good ideas. Will add to next PR.

hpcfzl:
> I think DV said it is only used to focus the the textlayout in a text entry.

Yes, I noticed that tabbing through Widgetpedia overall feels quite intuitive. Not being able to move around inside `scrollArea` would be slightly unexpected I think? So now that you have explained what their purpose is, I would say we could be better conveying that to the user, rather than removing anything. Here's _an example_ adding a couple of labels:
```
[ ] sample top-left focus_id

                                                                                sample bottom-right focus_id [ ]
```

The horizontal scrolling is also disabled by default, so when I tab to the bottom-right checkbox, I lose sight of its position. Though I think that's simply the consequence of `horizontal?` being `null` by default.

phatchman:
I've pushed some changes that I think addresses these issues. I wish the focus ring stood out a bit more, but I think it still works.

> Yes, I noticed that tabbing through Widgetpedia overall feels quite intuitive. 

I found this as well. I think that is mostly because DVUI options just being very well named and operating intuitively and a little bit me adding some UX details to make it easier where I found an option difficult to understand. Plus, even if you don't know what something does, you can just try it and see.

> The horizontal scrolling is also disabled by default,

Yeah, this is a tough one. I generally don't want to mess too much with the default init_options or options, because I want the user to experience what they would see if they passed blank init_options. But there are exceptions in a few places.

I've made the boxes 1px wider/taller so they don't neatly end on the scrollbar in the default view. Hopefully that conveys there are more boxes to the right?

I was working on Flexbox in another branch, so that got added as well.

hpcfzl:
> Hopefully that conveys there are more boxes to the right?

That's more apparent now, yes.

The `entypo.hair_cross` icon might be more appropriate than `entypo.box`, but either way I would say it's better than a checkbox without explanation.

phatchman:
> Since this is essentially a demo inside demo

This made me think. Does it have to be? Instead of including it the demo we could just have:
```
Examples\widgetpedia.zig
```
Which just launches widgetpedia, separate from the demo, and build it for all the backends, the same way app.zig can build for all the backends.

It would fix out compile time issues and also allow us to use the full 800x600 screen size (assuming we want to limit to that).

I'm, not sure how I feel about it. But there are a few positives. Would mean there are 2 different web demos though. 

It's also something we could do later.

hpcfzl:
> Does it have to be?

I had a similar thought when you started making Widgetpedia, because it's quite involved on its own, yet it's only visible through a button inside a window which is behind the `Show Demo Window` button.

I can recall that you like to depend on the theme selector from the demo, so at least that would have to be added in case it's decided Widgetpedia is more beneficial as its own completely separate demo. I think moving the existing Widgetpedia button to the main demo window as a `Show Widgetpedia Window` button below the demo's would be more reasonable, at least for now.

phatchman:
@david-vanderson Is this expected? That the text flows around text icons on left and right, but not at top or bottom? Am I doing something wrong?
```
var tl = dvui.textLayout(@src(), init_opts, options.override(.{ .data_out = &wd }));
defer tl.deinit();
dvui.icon(@src(), "align_top", dvui.entypo.align_top, .{}, .{ .min_size_content = .all(35), .gravity_y = 1, .gravity_x = 0.5 });
dvui.icon(@src(), "align_bottom", dvui.entypo.align_bottom, .{}, .{ .min_size_content = .all(35), .gravity_y = 0, .gravity_x = 0.5 });
dvui.icon(@src(), "align_left", dvui.entypo.align_left, .{}, .{ .min_size_content = .all(35), .gravity_y = 0.5, .gravity_x = 1 });
dvui.icon(@src(), "align_right", dvui.entypo.align_right, .{}, .{ .min_size_content = .all(35), .gravity_y = 0.5, .gravity_x = 0 });
tl.addText(lorem[0], text_options);
```
image

david-vanderson:
Sorry I fell behind, but thank you for pushing this forward!

> @david-vanderson Is this expected? That the text flows around text icons on left and right, but not at top or bottom? Am I doing something wrong?

Sorry this is not well documented.  The idea is to support widgets in the corners, so the logic inside textLayout is pretty simple:
* use the gravity values (less than 0.5 or not) to figure out which corner
* last widget in that corner wins

I think what you are seeing is:
* "align_top" got kicked out by "align_left"
* "align_bottom" is assumed to be in the top-right corner, and the text is actually flowing around it, but the 0.5 gravity_x meant it rendered in the middle
* I'm nicely surprised that the text flowed around the left and right ones.  I never tested that case.

This logic is not amazing, it's just what I could figure out back when I did it.  Flowing around widgets in arbitrary places sounds pretty hard and more of a niche usecase than stuff in the corners.

One thing I've learned is that flowing around stuff at the top is pretty easy, but trying to size correctly with text flowing around stuff at the bottom is tricky.  It's relatively easy to get into an infinite loop:
* based on last frame's height, we position the bottom-right widget, the text is long enough to break onto a newline after it (meaning we need more space)
* next frame the bottom-right widget is lower, so now the text doesn't break the same and doesn't take as much vertical space
* next frame the bottom-right widget is back higher, the text breaks again
* repeat forever

I think stuff in the bottom corners still works as long as the height of the textLayout is not being determined by the text itself.

Given all that, what do you think?  Do we just need better documentation?  Does the api here need to change?

phatchman:
> Sorry I fell behind, but thank you for pushing this forward!

No issues. These is all fairly low impact changes, so no need for you to test / review, unless you have time. @hpcfzl has been helping out with the testing, and giving feedback, so that's been quite helpful.
 
> Given all that, what do you think? Do we just need better documentation? Does the api here need to change?

That all makes sense. So, I think it is just a documentation thing. 

If the main intention is to support corner icons, I'll change the widgetpedia entry to reflect that.

hpcfzl:
A couple more things I noticed now that Widgetpedia is live in the Web demo:
* Regarding the `focus_id` option in `scrollArea`, its value can't be reverted to `null` once it's already been changed to `top left` or `bottom right`. Could be that this has been present all this time.
* In `flexbox`, I can't observe any changes when I change the `expand` controls option.

In the case that I find more, I will attempt to save that for future PRs, as this one is already closed.

phatchman:
Thanks very much, will look into those.

> Could be that this has been present all this time.

You are in fact **not** going mad. I broke it when trying to fix a different issue. When you changed the number of boxes, the bottom-left widget id wasn't being reset, so I totally changed how it worked.

PR #776, opened 2026-03-12
Adding multi-threading

KernelState:
So, remember my pull request? This is a solution to it, it is simple enough comply with the zig approach of doing things, yet so useful, like you can have one state struct, give it to the update thread and have it do the thing, want many threads, sure just copy over the queue from the old thread so there is no corruption of data and set queue_initialized to true. I like how it went out to be, there is an example for it which looks very useless but has the code that shows how helpful it can be. Also, I added use_llvm because for some reason the native compiler screamed about single-threaded mode and some libc conflicts, I just made it like that for better experience you can remove that for your personal setup for speed.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1099,7 +1099,6 @@ fn addWebExample(
             .link_libc = false,
             .strip = if (opts.optimize == .ReleaseFast or opts.optimize == .ReleaseSmall) true else false,
         }),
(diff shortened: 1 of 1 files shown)


david-vanderson:
Looks like this is missing a few new files?  update.zig and update_thread.zig?

PR #777, opened 2026-03-12
Multi threading

KernelState:
forgot to add the files, @david-vanderson knows the pull request


david-vanderson:
I can at least understand the idea you are going for.  But especially for tricky stuff like this, we need applications that are trying to do this kind of thing to understand what the details of the api and interactions should be:
* Do they want a message queue kind of interface like you have here?  Or direct manipulation protected by mutexes?
* How should the details of thread creation/destruction work?  Are they tied to a widget lifetime?  Do we need a thread pool?
* Does the update function run on the background thread or on the ui thread?  Both?

These are questions that we won't get answers to by just trying to imagine what people will need.

KernelState:
This is not an imagination, this is what I needed when creating cherry. Also, this is just like the dvui.animation function, it's there as a simple way to abstract the need of creating threads and having to call them, you can make your own if it doesn't fit you but this seems like the more general thing you can make, the way it works is very open too as you can see in the update_thread example it contains the thread in the state and the window in the state so the update thread can do stuff with these. This is more like the dvui.App.main or renderers, you can make your own if your case is special but you shouldn't need more than what's provided 80% of the time.

david-vanderson:
I agree that a thread-safe queue is generally useful, but I don't see how it's tied into the rest of dvui.  As it stands this is more suited to a utility library.

In the future we might revisit some kind of thing like this, but for now it will have to live outside of dvui.  Thank you.

PR #778, opened 2026-03-14
wio backend

ypsvlq:
Tested on Linux (Wayland+X11), macOS, and Windows.

Known issues:

- DPI scaling does not work on platforms where natural size is the same as physical size (Win32, X11)
- Touch events are not handled
- AccessKit support is not implemented

(to be fixed in wio):

- ~~`refresh` does nothing~~
- ~~`waitEventTimeout` does not wait on Linux~~
- Under macOS, there is a "bad cursor rect event" warning (but everything seems to work)

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -26,8 +26,8 @@
             .lazy = true,
         },
         .wio = .{
-            .url = "https://github.com/ypsvlq/wio/archive/65b65d00acc65925bbff27ba0426417dc867ca81.tar.gz",
-            .hash = "wio-0.0.0-8xHrr6kmBQB_BvMsoONvrQPjzyj0xCSM-JJNmyZmqmcK",
+            .url = "https://github.com/ypsvlq/wio/archive/665305435019fce313c55c6c3cfb1f9e9f89058a.tar.gz",
+            .hash = "wio-0.0.0-8xHrr-cvBQA6pUTUXtUwKvoKRNKk3QboY8Vw7XSg8-JI",
             .lazy = true,
         },
         .opengl = .{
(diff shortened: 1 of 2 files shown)


ypsvlq:
For the DPI issue, wio exposes:

- Window size (i.e. coordinate space for mouse and touch events)
- Framebuffer size
- "Scale" (in DVUI terms, this is either `pixelSize() / windowSize()` or `contentScale()` depending on platform)

but I'm not sure this can map nicely to what DVUI expects.

The closest solution I can think of is something like
```
pub fn contentScale(self: *@This()) f32 {
    return if (self.size_natural.w == self.size_physical.w)
        1
    else
        self.scale;
}
```
but content scale isn't dynamic and the value isn't known until events are processed.

david-vanderson:
I'm running a bit behind but I will get to this soon!

For the dpi stuff (you probably already know this) "contentScale" is separate from the window and framebuffer size - it's an extra factor set by the user/windowing system.  My windows machine has "display scaling 125%" for example.  In dvui we (currently) only try to get that information at startup, which is why `contentScale()` works the way it currently does.

ypsvlq:
No worries!

I see how it should work in dvui, wio just only exposes one value across all systems.

I think maybe content scale should be dynamic in dvui, since different monitors can have different scales configured.

It could also be worked around in the wio backend by scaling the natural size on those platforms.

david-vanderson:
This is fantastic!

I noticed the mouse wheel scroll direction seems reversed, but that also made me see that the dvui logic for touchpad horizontal scrolling was messed up.  I've pushed a fix to main for that, so I don't want to touch this until it's merged and I can retest.

I'm will merge and we can continue to work on it in main.  Thank you!

david-vanderson:
In the devlog, I will credit you as "Elaine Gibson".  Let me know if you would like something else.

PR #781, opened 2026-03-16
DRAFT: Win98 Theming

desttinghim:
This PR has hackily implemented more of the Win98 style. The dvui Win98 theme does not match actual Win98 styles very closely. Part of the reason for this is that dvui's theming system is not currently featureful enough to enable it. The code in this branch is something of an exploration to figure out how theming will need to be expanded on. That is to say it may be useful for reference, but it will probably be replaced by deeper changes to the theming system before being merged.


david-vanderson:
Thank you very much!  I've gone through once looking at the things you had to do, will have to go through again more seriously trying to list here the kinds of things required.

vlappa:
Ha, this is pretty cool actually. :)

PR #785, opened 2026-03-18
defer rendering for overlapping widgets

david-vanderson:
fixes #783

Problem:
* load the scroll canvas demo
* move the boxes on top of each other
* try to click-drag the boxes back off each other
* a click on both boxes drags the *bottom* one

When rendering multiple overlapping widgets, the first one you run gets events first, but also draws first.

This PR adds `dvui.DeferRender` which records all drawing commands, and then in `deinit()` inserts them at the beginning of the "after" render command queue.  This way it inverts the draw order of the widgets (while maintaining of the order within the DeferRender extent).

Todo

- [x] integrate with floating windows/widgets (since they also defer their rendering - make sure a floating window inside a DeferRender works, and also the other way around)
- [x] @foxnne needed `renderingSet(false)` because they were drawing in the main window (not a subwindow), make sure that works without needing that workaround

### DIFF
--- a/src/Examples/scroll_canvas.zig
+++ b/src/Examples/scroll_canvas.zig
@@ -97,9 +97,10 @@ pub fn scrollCanvas() void {
     }
 
     for (boxes, 0..) |*b, i| {
-        const pt = dataRectScale.rectToPhysical(.{ .x = b.x, .y = b.y });
+        const phys = dataRectScale.rectToPhysical(.{ .x = b.x, .y = b.y });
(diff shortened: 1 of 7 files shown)


david-vanderson:
@foxnne I'm ready to push this to main, just wanted to check on the naming.  Does `RenderFrontToBack` feel okay?

foxnne:
> @foxnne I'm ready to push this to main, just wanted to check on the naming. Does `RenderFrontToBack` feel okay?

That works for me! I think that's probably clearer than defer, and makes the most common use case more obvious 👍

PR #786, opened 2026-03-20
TabIndexGroup

david-vanderson:
Makes tab_index a local order within the current TabIndexGroup, which can have its own tab_index.

This allows things like locally reordering the tab indexes of some widgets without globally changing anything.

Or a set of widgets with weird tab orders wrapped in a function that can be given a single tab_index.

### DIFF
--- a/examples/sdl-standalone.zig
+++ b/examples/sdl-standalone.zig
@@ -103,75 +103,85 @@ pub fn main() !void {
     }
 }
 
+// pretending to be some library widget
+fn libWidget(tab_index: ?u16) void {
+    var tig = dvui.tabIndexGroup(@src(), .{ .tab_index = tab_index });
+    defer tig.deinit();
+
+    _ = dvui.button(@src(), "libWidget Focus 2", .{}, .{ .tab_index = 2 });
+    _ = dvui.button(@src(), "libWidget Focus 1", .{}, .{ .tab_index = 1 });
+}
+
 // both dvui and SDL drawing
 // return false if user wants to exit the app
 fn gui_frame() bool {
     const backend = g_backend orelse return false;
+    if (false) {
+        {
+            var hbox = dvui.box(@src(), .{ .dir = .horizontal }, .{ .style = .window, .background = true, .expand = .horizontal, .name = "main" });
+            defer hbox.deinit();
 
-    {
-        var hbox = dvui.box(@src(), .{ .dir = .horizontal }, .{ .style = .window, .background = true, .expand = .horizontal, .name = "main" });
-        defer hbox.deinit();
+            var m = dvui.menu(@src(), .horizontal, .{});
+            defer m.deinit();
 
-        var m = dvui.menu(@src(), .horizontal, .{});
-        defer m.deinit();
+            if (dvui.menuItemLabel(@src(), "File", .{ .submenu = true }, .{})) |r| {
+                var fw = dvui.floatingMenu(@src(), .{ .from = r }, .{});
+                defer fw.deinit();
 
-        if (dvui.menuItemLabel(@src(), "File", .{ .submenu = true }, .{})) |r| {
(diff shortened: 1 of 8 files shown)


hpcfzl:
This seems like the right place to mention my observation with focus:
1. I tab around or click to focus on a widget.
2. I click somewhere empty.
3. The widget loses focus, the blue border.
4. I tab again.
5. I find that I'm focused at the very beginning.

So what goes against the web browser behaviour and, I suppose for that reason, my instincts, is that the focus isn't placed where _I clicked somewhere empty_ in the second step. Is there anyone else with opinions on this?

To be clear, here's a hypothetical example: that _somewhere empty_ could have been the invisible bounding box of a text layout. Even though that area might not be focusable, the click would update the _focus position_ anyway. The new position would be the nearest focusable widget, but no focus effects like the blue border would be visible. Essentially behaving like web browsers.

So far, I haven't really been thinking how that could be resolved, or whether it's even desirable for DVUI. In case it's not really, this PR's `tabIndexGroup` widget could still behave like a *checkpoint* for focus; when I click on any otherwise unfocusable widget which is present inside a certain `tabIndexGroup`, my focus would now default to the first widget contained in there.

I imagine there are probably three approaches:
* Entirely copy the web browser behaviour
* Implement only as a *side-effect* of placing widgets inside `tabIndexGroup`
* Combine both approaches, where `tabIndexGroup` is itself an unfocusable widget?

---

A different, but still relatable to focus, specifically that of tabbing, is the topic of **skip links**.

Open for example Reddit, then tab once. A button appears which prompts you to skip over the navigation. I understand that we already have menus for this in DVUI, because you can tab to skip across, or use arrow keys to remain in the menu, but that's inapplicable to interfaces which don't use the menu widgets. I suppose existing `groupFocus` is useful then, unless navigating via arrow keys isn't ideal for a particular use case.

---

These were a couple of thoughts of mine on focusing widgets. Was that relevant enough and have you had similar considerations surface? I'm beginning to notice that the majority of my takes stems from web browsers, nevertheless I attempt to be mindful of what really would be reasonable for native applications.

david-vanderson:
> So what goes against the web browser behaviour and, I suppose for that reason, my instincts, is that the focus isn't placed where _I clicked somewhere empty_ in the second step. Is there anyone else with opinions on this?

Good feedback!  The current behavior does not have that much thought behind it.  I was generally following the idea that clicking outside of any focusable element should make nothing focused.

I'm not sure TabIndexGroup is part of the solution, but I agree we should at least attempt something better here.

Testing in web browsers suggests that they are doing some kind of tree containment test.  Sometimes tabbing gets you to a "nearby" element, but sometimes it restarts at the beginning.  My guess is that you get the first element within the subtree you clicked in.

I tried testing on my Mac, but either I don't know what I'm doing or mac apps are really bad at keyboard navigation.  Even with "Full Keyboard Access" turned on.  Best guess here is that clicking outside any widget doesn't change the focus at all.

I will test on Windows and Linux.  Anybody's experience here would be useful!

My only decent idea would be to keep track of the "closest" focusable widget to the mouse, and if a mouse click would defocus widgets, we remember the closest and restart tab navigation from that widget.

> A different, but still relatable to focus, specifically that of tabbing, is the topic of **skip links**.

This is the first time I've heard of these.  I'd want to see an example outside of web browsers, or at least someone in a situation where this would be useful before trying to figure out what if anything to do.  My guess is the same as yours - because web pages don't have "normal" menus, you have to tab through all the "menu stuff" to get to the content.  But in a dvui app one tab skips the whole menu.

phatchman:
@david-vanderson The recursive model was definitely the way to go.

I've added the minimal set of commits to get those grid examples working. Edit: and a widgetpedia entry.

david-vanderson:
> I've added the minimal set of commits to get those grid examples working. Edit: and a widgetpedia entry.

Awesome - thank you!

PR #788, opened 2026-03-21
demo: add simple search box to view source

phatchman:
Adds a search box. I found this handy when I wanted to look or a particular widget in the source.

I put the prev and next buttons in the search box, because I thought it looked nicer. But there isn't really a way to stop the buttons scrolling off to the right, as far as I can tell? Most people won't see it, so maybe it's fine for now?

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -574,7 +574,8 @@ fn displayZigSourceCode(filename: []const u8, source: []const u8, showing: *bool
                     var hbox_inner = dvui.box(@src(), .{ .dir = .horizontal }, .{
                         .gravity_x = 1.0,
                         .background = true,
-                        .margin = .{ .x = 0, .y = 4, .w = 4, .h = 4 },
+                        .margin = .{ .x = -1, .y = 4, .w = 4, .h = 4 },
+                        .padding = .{ .x = 4 },
                         .corner_radius = .{ .x = 0, .y = 5, .w = 5, .h = 0 },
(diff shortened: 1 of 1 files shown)


david-vanderson:
> Adds a search box. I found this handy when I wanted to look or a particular widget in the source.

This is awesome!

> I put the prev and next buttons in the search box, because I thought it looked nicer. But there isn't really a way to stop the buttons scrolling off to the right, as far as I can tell? Most people won't see it, so maybe it's fine for now?

I tried passing `.scroll_horizontal = false` to the textEntry.  That keeps the buttons from scrolling off, but now the text runs on top of the buttons.  Maybe there's a way to put the buttons outside the textEntry but visually have them in the same background area (so they kind of look like they are inside).  But I'm totally fine with it as it is.

Super cool!  Merge when ready.  Thank you!

phatchman:
I think this works. There is a very tiny gap at the top where the two borders join, if you look closely.

image

focused:

image

Edit: Adding a margin of .x = -1 seems to have fixed it. But my latest commit 66d23bcecac44787f4bc0c3ccef546aeb3ff361d is not coming up yet. Github has been stuck on "processing updates" for 10 minutes.

david-vanderson:
Looks great!  I guess merge whenever github comes back.  I'll put on my list to audit what happens with a negative margin, but no reason to hold this up.

david-vanderson:
Looked into negative margins, and they seem to operate as expected:
* size of the element for layout is reduced
* size of borderRect and contentRect unaffected

So negative margins seem like a reasonable tool for situations like this.  Great!

PR #791, opened 2026-03-23
Support separate render backends

ypsvlq:
With these changes:

- `-Drender-backend=default` will use the render code in the main backend (status quo)
- `-Drender-backend=opengl` will use the OpenGL renderer if supported by the main backend, currently implemented for GLFW and wio

This enables separating the SDL backends in the future, and implementing other shared renderers such as Vulkan.

CC @Carboxide68 for the GLFW backend changes, I can revert them if you would prefer.

### DIFF
--- a/src/Backend.zig
+++ b/src/Backend.zig
@@ -16,9 +16,16 @@ pub const TextureError = GenericError || error{ TextureCreate, TextureRead, Text
 pub const kind = Implementation.kind;
 
 impl: *Implementation,
+render_impl: if (dvui.render_backend.kind == .default) void else *dvui.render_backend,
 
-pub fn init(impl: *Implementation) Backend {
-    return .{ .impl = impl };
+pub const init = if (dvui.render_backend.kind == .default) initDefault else initRenderer;
+
+fn initDefault(impl: *Implementation) Backend {
+    return .{ .impl = impl, .render_impl = {} };
+}
+
+fn initRenderer(impl: *Implementation, render_impl: *dvui.render_backend) Backend {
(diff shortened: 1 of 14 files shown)


src/Backend.zig
```diff
@@ -34,11 +41,13 @@ pub fn sleep(self: Backend, ns: u64) void {
 /// arg is cleared before `dvui.Window.begin` is called next, useful for any
 /// temporary allocations needed only for this frame.
 pub fn begin(self: Backend, arena: std.mem.Allocator) GenericError!void {
+    if (dvui.render_backend.kind != .default) try self.render_impl.begin(arena);
```
david-vanderson: I think renderer.begin() should happen after backend.begin()?  That way the backend begin/end wraps the renderer begin/end.

david-vanderson (approved):
Excellent!  Please address my one question and then merge when ready.  Thank you!

src/Backend.zig
ypsvlq: @david-vanderson

Sorry for the delay, I didn't get an email for some reason.

I wrote it that way originally, but the Raylib backend doesn't return an error union from `begin` so this was a less intrusive workaround.

I can change the Raylib backend, or add some `@typeInfo` to the call site?

ypsvlq: Actually it's not as intrusive as I thought, assuming `RaylibBackend.begin` should only be called from `Backend.begin`. I'll change it there.

david-vanderson (approved):
Looks great - merge it!  Thank you!

PR #796, opened 2026-03-27
Experiment with only updating portions of textures

foxnne:
@david-vanderson I've been looking into increasing Pixi's performance greatly, and through my experiments, I've made good progress on this [branch](https://github.com/foxnne/pixi/tree/perf), which took me from ~11 fps with a large file (4096x4096, 12 layers, 64x64 tiles) even in release mode, up to 55-60 fps in debug mode.

The changes I needed from DVUI are included in this PR. Please let me know if this is something possible for DVUI to implement, or if specialized things (I'm not sure if other backends can do this?) are allowed for certain backends. 

Anyway, thanks for any help!

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,6 +1,7 @@
 const std = @import("std");
 const enums_backend = @import("src/enums_backend.zig");
 pub const Backend = enums_backend.Backend;
+const RenderBackend = enums_backend.RenderBackend;
 const Pkg = std.Build.Pkg;
 const Compile = std.Build.Step.Compile;
 
@@ -60,6 +61,7 @@ pub fn build(b: *std.Build) !void {
     const optimize = b.standardOptimizeOption(.{});
 
     var back_to_build = b.option(Backend, "backend", "Backend to build");
+    const render_backend = b.option(RenderBackend, "render-backend", "Render backend to build (default: implied by backend)") orelse .default;
 
     const test_step = b.step("test", "Test the dvui codebase");
     const check_step = b.step("check", "Check that the entire dvui codebase has no syntax errors");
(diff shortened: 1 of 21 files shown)


foxnne:
I added what I thought we needed here, but please review and let me know what all needs to change if anything for the other backends, if this is something we could potentially accommodate?

david-vanderson:
This is awesome, will definitely work to integrate this.  I've only taken a quick look so far - my initial reaction is to make the update rects optional somehow, to consolidate full and partial texture update into one function.  (I need to rework those apis anyway to support YUV texture updates).  For some backends we can either have them error or fall back to a full update.

Less sure if doing the same for the clearing, but there must be a decent way of combining `renderClearRect` with `textureClearTarget`.  Is is meaningful to run `renderClearRect` against the output framebuffer, or does it only make sense against a texture target?

foxnne:
> Less sure if doing the same for the clearing, but there must be a decent way of combining `renderClearRect` with `textureClearTarget`. Is is meaningful to run `renderClearRect` against the output framebuffer, or does it only make sense against a texture target?

Just pushed removing all that, I no longer need it, and would like to keep this as minimal as possible :) thanks for asking about that. Now this should just be adding enough to update rects of textures only. Feel free to modify or ask for changes or anything, I just did this to get a proof of concept for performance fixes and make sure it works

david-vanderson:
I'm not sure how fast I'll be able to get to this and also mix in the yuv stuff.  Let's merge this now and I'll help update pixi when I get to it later.

If that sounds good, go ahead and merge it.  Thank you!

foxnne:
I suppose this shouldn't have any negative effect and I'll likely be the only one using it before then :) merging, thank you!

PR #798, opened 2026-03-30
add `modal_fade_alpha`

ZackeryRSmith:


david-vanderson:
This is on the right track, but changing from `.text` to `.fill` for the color means a light theme makes the background lighter, not darker.

ZackeryRSmith:
True, although I do feel it's a bit odd to use `.text` as then when you want to change the color of the modal fade you have to set `.color-text`. It's not all that important though, this was just an issue with my local changes. I'll revert it back to `.text`!

david-vanderson:
I've tweaked it to keep the old behavior by default and shorten the name.  Let me know if you need anything else.  Thank you!

PR #799, opened 2026-03-31
Reduce text jitter when smoothly scaling text with snap to pixels off…

foxnne:
…, for cases when this is priority.

@david-vanderson 

Hi! I know we had talked about this before for graphl, I've been trying to find a way to smooth out text on nodes which determine node sizes, and generally reduce text jitter when scaling.

I found that these changes produced smooth and nice looking text for me, though I know this should sacrifice text clarity as we are no longer snapping to pixels. I think this is the only way to maintain smooth zooming, which I believe is a larger priority for us since the user can zoom in on the text. 

I'm a newb to text rendering though, and I know you've studied and worked on this type of thing a lot, so I wanted your input. If you'd like to see a before and after comparison of these changes for us, please let me know. I'm hoping since this is gated on snap to pixels that it would affect almost no one unless the goal is smoothly scaling text, which makes sense to me that you'd turn off snap to pixels if you want that type of smoothness. 

I was also thinking about potentially just temporarily turning off snap to pixels during a zoom, and returning it to on after the zoom is complete, which I think would solve all cases. Anyway. Please let me know what you think!

### DIFF
--- a/src/Examples/scroll_canvas.zig
+++ b/src/Examples/scroll_canvas.zig
@@ -3,6 +3,21 @@ pub fn scrollCanvas() void {
     var vbox = dvui.box(@src(), .{}, .{ .expand = .both });
     defer vbox.deinit();
 
+    var fractional_font_cache_ref_scale: bool = dvui.fractionalFontCacheRefScale() != 1.0;
+    var snap_to_pixels: bool = dvui.snapToPixels();
+
+    if (dvui.checkbox(
+        @src(),
+        &fractional_font_cache_ref_scale,
+        "Fractional Font Cache Ref Scale",
+        .{},
+    )) {
+        _ = dvui.fractionalFontCacheRefScaleSet(if (fractional_font_cache_ref_scale) 32.0 else 1.0);
+    }
+    if (dvui.checkbox(@src(), &snap_to_pixels, "Snap to Pixels", .{})) {
+        _ = dvui.snapToPixelsSet(snap_to_pixels);
+    }
+
     const scroll_info = dvui.dataGetPtrDefault(null, vbox.data().id, "scroll_info", ScrollInfo, .{ .vertical = .given, .horizontal = .given });
     const origin = dvui.dataGetPtrDefault(null, vbox.data().id, "origin", Point, .{});
     const scale = dvui.dataGetPtrDefault(null, vbox.data().id, "scale", f32, 1.0);
@@ -322,6 +337,8 @@ pub fn scrollCanvas() void {
                         zoom *= zs;
                         zoomP = me.p;
                     }
+                } else if (me.action == .wheel_x) {
+                    e.handle(@src(), scrollContainer.data());
                 }
             },
             else => {},
--- a/src/Font.zig
+++ b/src/Font.zig
@@ -295,11 +295,11 @@ pub fn textSizeEx(self: Font, text: []const u8, opts: TextSizeOptions) Size {
     // might give us a slightly smaller font
     const fce = dvui.fontCacheGet(sized_font) catch return .{ .w = 10, .h = 10 };
 
-    // this must be synced with dvui.renderText()
+    // this must be synced with dvui.renderText() (`target_size / fce.em_height` when snap is off)
     const target_fraction = if (cw.snap_to_pixels)
         1.0 / ss
     else
-        self.size / fce.em_height;
+        (self.size * ss) / fce.em_height;
 
     var options = opts;
     if (opts.max_width) |mwidth| {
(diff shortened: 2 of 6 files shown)


foxnne:
Thoughts here: Text quality is visually degraded and blurry, but I am essentially caching a single font glyph atlas and retrieving the same one each time. I believe I could mitigate the issues by generating the "full res" or full zoom font atlas and maintaining that atlas rather than the one when the app starts? Or choosing a scale at which we are most "likely" to see the text, where its most clear at that zoom level?

david-vanderson:
> I was also thinking about potentially just temporarily turning off snap to pixels during a zoom, and returning it to on after the zoom is complete, which I think would solve all cases. Anyway. Please let me know what you think!

This is quite interesting.  I was hoping that "simply" turning off snap to pixels for this (or just during a zoom like you say) would be enough.

My own experience was that rasterizing slightly smaller (at a pixel size) and scaling up to the (fractional) ask size looked better than the opposite (rasterizing at a pixel size slightly larger).  But it looks like you are finding the opposite, so maybe my experience was out-dated by other changes.

> Thoughts here: Text quality is visually degraded and blurry, but I am essentially caching a single font glyph atlas and retrieving the same one each time.

The current behavior (when snap to pixels is off) should be to rasterize at the pixel size smaller and then scale it up.  Is that not working because it looks terrible?  Or are there performance problems rasterizing the font at different pixel sizes (as the ask size changes quickly)?

I want to make this work - can I reproduce the problem by turning off snap to pixels in the scroll canvas demo (do you see it if you do that)?  Or do I need to do something else to see it?

foxnne:
> The current behavior (when snap to pixels is off) should be to rasterize at the pixel size smaller and then scale it up. Is that not working because it looks terrible? Or are there performance problems rasterizing the font at different pixel sizes (as the ask size changes quickly)?

> I want to make this work - can I reproduce the problem by turning off snap to pixels in the scroll canvas demo (do you see it if you do that)? Or do I need to do something else to see it?

For some reason when I test `main` with just setting snap to pixels off for the scroll_canvas demo (and fixing the events so we don't pan and zoom at the same time) I get this: (shows text jitter still, not indicating scaling? or we are still changing font spacing per zoom)

https://github.com/user-attachments/assets/832a7ea3-d16b-4695-9809-2cae621ddadc

This next video is with snap to pixel on, and theres very little difference between snap on and off.

https://github.com/user-attachments/assets/a95e39c3-0e5e-4d37-94b0-159ae32fd1af

Compare that with just scaling the cached texture and we get perfectly smooth zooming, but of course the texture becomes blurry as we zoom in. Thats expected, though I was hoping there would be some way to increase the resolution of the font glyphs while just scaling the distance per-glyph of the 1.0 scale text? It seems that if I just simply keep the cached texture and scale up, that I lose kerning correctly between scaled up glyphs. This is a tricky issue.

I guess when we boil down the issue, we have snap to pixels which potentially increases the width of a character by a pixel or more, which when stacked with 10+ characters can shift by a large amount. We can get around this by not calculating new texture and positions as we zoom, but we lose precision. Im pushing to this branch now with some checkboxes that let you try out the scroll_canvas demo with/without pixel snapping and with/without scaling the texture for higher resolution, but its clearly broken

This branch is just for experimenting for now, I don't intend on merging any of this as its a big mess now. I was hoping that some of this could be helpful.

Below is trying to use a font scale of 32 for the atlas and then allowing zooming on it which looks sharp, but Idk what other issues this causes.

https://github.com/user-attachments/assets/08ee6b28-9d43-40db-8f1d-4c34fef969a6

david-vanderson:
Thanks for the excellent information!  Good catch on the kerning thing.  It might be a few days before I can really dig into this, but we'll figure this out!

foxnne:
> Thanks for the excellent information! Good catch on the kerning thing. It might be a few days before I can really dig into this, but we'll figure this out!

Thanks so much for looking into it, no rush! I really appreciate the help!

david-vanderson:
Current thinking here - we are making sure to align each glyph to the physical pixel grid, even when snap_to_pixels is off.  I think that is the root cause.

We need to keep our glyph measurements fractional until using them.  One wrinkle is that freetype will always report aligned-to-pixel values.

Are you okay with using stb_truetype all the time?  (We already always use it for the web, and you can test with -Dfreetype=false)

foxnne:
> Are you okay with using stb_truetype all the time? (We already always use it for the web, and you can test with -Dfreetype=false)

I've tested with freetype=false and everything seems fine, and we have a web version so I see no issue with us being locked to stb_truetype, but I also don't honestly know what tradeoffs we are making. If we are already using it for web I don't see any reason not to use it all the time, if it can help solve this issue.

PR #803, opened 2026-04-06
fonts: fixes to render smoothly zooming text

david-vanderson:
fixes #799

Reported and initial work by @foxnne 

* must use stb_truetype (-Dfreetype=false)
* must turn off snapToPixels for the text you want to smoothly zoom

Normally you want glyphs to be aligned to the physical pixel grid.  This means not only starting on a pixel boundary, but also ensuring that glyph advance values are all integers.

The problem here was that we were doing that despite the snapToPixels setting. Now when snapToPixels is false, we keep the fractional advance values (and the font height value).

Using this means fonts will zoom nicely, but at the cost of glyphs rendering blurry due to not aligning to the screen pixels.

### DIFF
--- a/src/Examples/scroll_canvas.zig
+++ b/src/Examples/scroll_canvas.zig
@@ -105,7 +105,7 @@ pub fn scrollCanvas() void {
 
         // This makes the text smoothly zoom, as long as you are using
         // stb_truetype (-Dfreetype=false)
-        const snap = dvui.snapToPixelsSet(false);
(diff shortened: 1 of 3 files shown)


david-vanderson:
@foxnne Does this fix it for you?

foxnne:
> @foxnne Does this fix it for you?

In my brief testing this seems to look perfect! Thank you so much for this work for this fix, this is a great help!

PR #806, opened 2026-04-10
wio fixes

ypsvlq:
Touch support is currently untested as I don't have a touch device on hand.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -26,8 +26,8 @@
             .lazy = true,
         },
         .wio = .{
-            .url = "https://github.com/ypsvlq/wio/archive/8aecae4e3b63dbdcdecde6fcc0c474902e4e23d8.tar.gz",
-            .hash = "wio-0.0.0-8xHrr01dBQDLwISVPogLZjCYuQsuVECb8CO5Jq5uTCIU",
+            .url = "https://github.com/ypsvlq/wio/archive/8f874372e6644769f6a964912cd819b02fcb8c6b.tar.gz",
+            .hash = "wio-0.0.0-8xHrr3sEBgDdNTJi4NMsC_SED4XiK8M96IUFfCyVlVRS",
             .lazy = true,
         },
         .opengl = .{
--- a/src/backends/wio.zig
+++ b/src/backends/wio.zig
@@ -211,8 +211,8 @@ pub fn addEvent(self: *@This(), win: *dvui.Window, event: wio.Event) !bool {
         .scroll_horizontal => |ticks| return try win.addEventMouseWheel(-ticks * dvui.scroll_speed, .horizontal),
         .touch => |touch| {
             const button = touchIdToDvuiButton(touch.id) orelse return false;
-            const xnorm = @as(f32, @floatFromInt(touch.x)) / self.size_physical.w;
-            const ynorm = @as(f32, @floatFromInt(touch.y)) / self.size_physical.h;
+            const xnorm = @as(f32, @floatFromInt(touch.x)) / self.size_natural.w;
+            const ynorm = @as(f32, @floatFromInt(touch.y)) / self.size_natural.h;
             const old = &self.touch[touch.id];
             if (std.math.isInf(old.x)) {
                 old.* = .{ .x = xnorm, .y = ynorm };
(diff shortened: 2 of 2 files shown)


david-vanderson:
I have a https://furilabs.com/ FLX1 which is a linux aarch64 phone.  I've been trying to run `zig build wio-app` on it.

First bump:
```
error:  `wayland-scanner` may not be installed on the system.
 You can switch to X11 in your `build.zig` by changing `Options.linux_display_backend`
thread 10073 panic: `wayland-scanner` not found
```

Solved by `apt install libwayland-dev`

Second bump:
```
Couldn't open plugin directory: No such file or directory
No plugins found, falling back on no decorations
Segmentation fault at address 0x40
```

Solved by `apt install libdecor-0-plugin-1-gtk` (from https://github.com/Ferdi265/wl-mirror/issues/74)

Third bump:
```
error(wio): eglCreateWindowSurface failed, error 0x3005
error: Unexpected
/home/furios/.cache/zig/p/wio-0.0.0-8xHrr5z7BQDMoPDfxAX02WK9NT8-XA5-YVXbUjqfLKyR/src/unix/wayland.zig:748:5: 0x15c6993 in logEglError (wio-app)
    return error.Unexpected;
    ^
/home/furios/.cache/zig/p/wio-0.0.0-8xHrr5z7BQDMoPDfxAX02WK9NT8-XA5-YVXbUjqfLKyR/src/unix/wayland.zig:353:108: 0x15cd16f in createWindow (wio-app)
            self.egl.surface = c.eglCreateWindowSurface(egl_display, config, self.egl.window, null) orelse return logEglError("eglCreateWindowSurface");
```

Unsure what to do there.

Instead tried `zig build wio-app -Dwio_unix_backends=x11`.  That runs, but a touch on the screen comes through wio as a mouse move followed by a mouse left down.

That's where I've gotten.  No idea if this is worth digging into right now.

Go ahead and merge this when you are ready, and let me know if there's something worth testing/investigating more here.  Thanks!

ypsvlq:
```
error:  `wayland-scanner` may not be installed on the system.
 You can switch to X11 in your `build.zig` by changing `Options.linux_display_backend`
thread 10073 panic: `wayland-scanner` not found
```

This shouldn't happen, but it seems like an issue in the dvui build script. Unless you specify `-fsys=wio`, wio has no system dependencies.

```
Couldn't open plugin directory: No such file or directory
No plugins found, falling back on no decorations
Segmentation fault at address 0x40
```

Will look into this.

```
error(wio): eglCreateWindowSurface failed, error 0x3005
```

This is probably because it needs to request OpenGL ES instead of OpenGL, I meant to change `wio.CreateContextOptions` to support that for a while now.

> That runs, but a touch on the screen comes through wio as a mouse move followed by a mouse left down.

X11 doesn't have touch handling in wio, so that's probably to be expected. Thank you for the reports!

ypsvlq:
@david-vanderson Could you retry the Wayland build with [this patch](https://github.com/user-attachments/files/26635203/gles.patch) applied?

david-vanderson:
Good point about wayland-scanner, that is probably from the raylib build.zig

> @david-vanderson Could you retry the Wayland build with [this patch](https://github.com/user-attachments/files/26635203/gles.patch) applied?

With that patch applied wio-app runs!

The touch points are in the correct place after adding this change:
```diff
diff --git a/src/backends/wio.zig b/src/backends/wio.zig
index 7a56f337..05dab9c3 100644
--- a/src/backends/wio.zig
+++ b/src/backends/wio.zig
@@ -211,8 +211,8 @@ pub fn addEvent(self: *@This(), win: *dvui.Window, event: wio.Event) !bool {
         .scroll_horizontal => |ticks| return try win.addEventMouseWheel(-ticks * dvui.scroll_speed, .horizontal),
         .touch => |touch| {
             const button = touchIdToDvuiButton(touch.id) orelse return false;
-            const xnorm = @as(f32, @floatFromInt(touch.x)) / self.size_physical.w;
-            const ynorm = @as(f32, @floatFromInt(touch.y)) / self.size_physical.h;
+            const xnorm = @as(f32, @floatFromInt(touch.x)) / self.size_natural.w;
+            const ynorm = @as(f32, @floatFromInt(touch.y)) / self.size_natural.h;
             const old = &self.touch[touch.id];
             if (std.math.isInf(old.x)) {
                 old.* = .{ .x = xnorm, .y = ynorm };
```

Does that look okay?  I'm good with all of this - merge when ready!

PR #809, opened 2026-04-10
OpenGL ES support

ypsvlq:
This changes the texture format from GL_UNSIGNED_INT_8_8_8_8 to GL_UNSIGNED_BYTE. I think this is contrary to dvui's documentation:
```
    /// This is the native format of Color.PMA.  On little endian this is the
    /// same as abgr_8_8_8_8, on big endian it is rgba_8_8_8_8.
    rgba_32,
```
but in practice leads to correct rendering (on my little-endian machine).

### DIFF
--- a/src/backends/render/opengl.zig
+++ b/src/backends/render/opengl.zig
@@ -168,7 +168,7 @@ pub fn textureCreate(_: *@This(), pixels: [*]const u8, width: u32, height: u32, 
 
 pub fn textureUpdate(_: *@This(), texture: dvui.Texture, pixels: [*]const u8) !void {
     gl.bindTexture(gl.TEXTURE_2D, @intCast(@intFromPtr(texture.ptr)));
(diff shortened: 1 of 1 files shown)


david-vanderson:
I'm also unsure about the docs, if you have a better way to phrase it please do.  I think it's very unlikely that any of this is correct on big endian machines, but also unlikely that anyone tries to use it on one.  If they do then we can sort it out then.

Merge when ready.  For wio stuff please feel free to commit to main directly, or file a PR if you want me to look at it.  Thank you!

ypsvlq:
I mean that on my little endian machine, textures with format .rgba_32 are layed out bytewise as 0xRR 0xGG 0xBB 0xAA, whilst the docs suggest it would be 0xRRGGBBAA as a u32 (i.e. reversed).

(at least I assume it's correct rendering, the Zig logo is orange like this which matches the SDL backend)

david-vanderson:
You are probably right - and yes the zig logo being orange is correct.

PR #810, opened 2026-04-11
testing: create parent directories for snapshots and image files

### LINKED ISSUE #801: dvui.testing.snapshot doesn't work for nested paths.
Just ran into this while using dvui.testing in a path like `src/subfolder/file.zig`.

Because [here](https://github.com/david-vanderson/dvui/blob/823ee494e7aeb798c7a6db6949ce699ae1275f41/src/testing.zig#L236) the expected `@src().file` may contain path delimiters, you will hit a "no such directory entry error" when [creating the snapshot file](https://github.com/david-vanderson/dvui/blob/823ee494e7aeb798c7a6db6949ce699ae1275f41/src/testing.zig#L258) for a test in a nested path.

Abridged excerpt of problematic code shown below.

```zig
    const filename = try std.fmt.allocPrint(self.allocator, "{s}-{s}-{d}", .{ src.file, src.fn_name, self.snapshot_index });
   //...
   const image_name = try std.fmt.allocPrint(self.allocator, "images/{s}-{s}.png", .{ filename, image_suffix });
  // ...
  var file = try dir.createFile(image_name, .{});
```

I can make a repro if desired. I could also consider contributing a fix once confirmed how we should fix it. Should we recursively create the parent directory for that file? Should we just sanitize the path to remove `/`?

Jamazi4:
fixes #801 

`snapshot()` builds filenames from `src.file` which may contain path separators, as a result it caused "no such directory entry error".

This solves the issue by recursively recreating parent folder structure.

### DIFF
--- a/src/testing.zig
+++ b/src/testing.zig
@@ -249,12 +249,9 @@ pub fn snapshot(self: *Self, src: std.builtin.SourceLocation, frame: dvui.App.fr
     defer widget_hasher = null;
 
     if (@import("build_options").snapshot_image_suffix) |image_suffix| {
-        dir.makeDir("images") catch |err| switch (err) {
-            error.PathAlreadyExists => {},
-            else => return err,
-        };
         const image_name = try std.fmt.allocPrint(self.allocator, "images/{s}-{s}.png", .{ filename, image_suffix });
         defer self.allocator.free(image_name);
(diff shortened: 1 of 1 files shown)


david-vanderson:
Thank you!

PR #814, opened 2026-04-17
reduce motion

david-vanderson:
* add global reduce_motion bool
* add checkbox to debug window

When reduce_motion is true, animation() rewrites the end_time to be start_time+1, making the animation expire the next frame (and causing an immediate next frame).

Timers are exempted.

### DIFF
--- a/src/Backend.zig
+++ b/src/Backend.zig
@@ -186,9 +186,9 @@ pub fn preferredColorScheme(self: Backend) ?dvui.enums.ColorScheme {
     return self.impl.preferredColorScheme();
 }
 
-/// Get the preferredMotionAmount if available
-pub fn preferredMotionAmount(self: Backend) ?dvui.enums.MotionAmount {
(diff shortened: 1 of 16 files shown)


david-vanderson:
@MichaelBelousov @foxnne Rewriting animations to be "instant" seems to work okay, but would love some testing - if you use this branch you can toggle it on and off in the debug window.  Thanks!

foxnne:
> @MichaelBelousov @foxnne Rewriting animations to be "instant" seems to work okay, but would love some testing - if you use this branch you can toggle it on and off in the debug window. Thanks!

I just tested with graphl and it seemed to work perfectly for us, I don't see any oddities in tree widget, any animations I've added, or the autosizing paned widget. Thank you for this!

david-vanderson:
Awesome thanks!  I'll add more comments.  I'm hoping one of you can figure out how to flow the preference from the browser to this.

PR #817, opened 2026-04-17
add `Backend.preferredMotionAmount`

### LINKED ISSUE #813: [suggestion] add ability for backends to query system theme preference and accessibility preferences from DVUI
provide via the backend interface or dvui.App (or some other mechanism) a way to get the system theme as well as system accessibility settings (e.g. prefers reduced motion). As indicated in #638, DVUI could also have a default animation behavior change when the backend indicates that the host prefers reduced motion.

Ideally at the beginning of loading my app I could do something vaguely like:

```zig
const w = dvui.currentWindow();

w.themeSet(switch (w.backend.system_preferences.os_theme_tag) {
  .default, .light => my_light_theme,
  .dark => my_dark_theme,
});

// not sure about this one, please help. I'm thinking of std_options from zig std lib
dvui.global_options.motion_preference = w.backend.system_preferences.motion_preference;
```

This could be implemented for example in the web backend with something like:
 
```js
const themePrefLight = window.matchMedia("(prefers-color-scheme: light)");
const themePrefDark = window.matchMedia("(prefers-color-scheme: dark)");

const motionPrefUnset = window.matchMedia("(prefers-reduced-motion: no-preference)");
const motionPrefReduced = !motionPrefUnset && window.matchMedia("(prefers-reduced-motion: reduced)");

// pardon the inconsistency you get what I mean
const motionPrefEnumVal = motionPrefUnset ? 0 : (motionPrefReduced ? 2 : 1);
const themePrefEnumVal = themePrefLight ? 1 : (themePrefDark ? 2 : 0);
// or reuse an existing dvui wasm init export I didn't check for 
wasm.instance.exports.dvui_backend_host_prefs_init(motionPrefEnumVal, themePrefDark)
```

I'm not sure if we want to cache this on program start like I have in this WebBackend.js suggestion, but for the wasm case I'd rather we not encourage people call out to javascript media queries several times while rendering a frame, if that is somehow used.

MichaelBelousov:
(Doesn't close #813 until the base branch is merged into master)

- based on and merges into reduce_motion branch
- add a `preferredMotionAmount` function all backends
- change `dvui.reduce_motion` to optional `dvui.motion_amount` with type `enum { reduced, all }`, and null state meaning unset.
- call `preferredMotionAmount` during `dvui_init` in WebBackend to initialize `dvui.reduce_motion` for web backends automatically. I am happy to discuss this, I could easily change it to so like `preferredColorScheme` the user is responsible for using it

worth noting, we might be able to add this for other backends:
https://stackoverflow.com/questions/59708960/how-do-i-change-the-prefers-reduced-motion-setting-in-browsers

### DIFF
--- a/src/Backend.zig
+++ b/src/Backend.zig
@@ -186,9 +186,9 @@ pub fn preferredColorScheme(self: Backend) ?dvui.enums.ColorScheme {
     return self.impl.preferredColorScheme();
 }
 
-/// Get the preferredMotionAmount if available
-pub fn preferredMotionAmount(self: Backend) ?dvui.enums.MotionAmount {
-    return self.impl.preferredMotionAmount();
+/// Get the prefersReducedMotion if available
+pub fn prefersReducedMotion(self: Backend) bool {
+    return self.impl.prefersReducedMotion();
 }
 
 /// Called by `dvui.refresh` when it is called from a background
--- a/src/Backend.zig
+++ b/src/Backend.zig
@@ -186,6 +186,11 @@ pub fn preferredColorScheme(self: Backend) ?dvui.enums.ColorScheme {
     return self.impl.preferredColorScheme();
 }
 
+/// Get the preferredMotionAmount if available
+pub fn preferredMotionAmount(self: Backend) ?dvui.enums.MotionAmount {
+    return self.impl.preferredMotionAmount();
+}
+
 /// Called by `dvui.refresh` when it is called from a background
 /// thread.  Used to wake up the gui thread.  It only has effect if you
 /// are using `dvui.Window.waitTime` or some other method of waiting until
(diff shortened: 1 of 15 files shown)


MichaelBelousov:
fixing issue, my bad putting it up this early, please wait. And doing more testing on our web build.

EDIT: done

david-vanderson:
Sounds good.  Is there a situation you can think of where we would treat `null` and `all` differently?  Are you imagining `null` telling dvui to go try and figure it out automatically?

MichaelBelousov:
> Is there a situation you can think of where we would treat `null` and `all` differently? Are you imagining `null` telling dvui to go try and figure it out automatically?

I don't necessarily think we'll treat them differently and I agree it complicates e.g. checkboxes. I personally prefer preserving knowledge of the "unset" state so the application can decide what to do. But I defer to you. A bool can be more convenient so I don't mind changing it.

MichaelBelousov:
I don't see any difference on any platform between `.all` and `null`. I will simplify it to a bool.

MichaelBelousov:
It has been reverted to a bool. Lmk if you have other comments.

david-vanderson:
If we run into something going forward we can always change it.  Thanks!

PR #818, opened 2026-04-18
Improve OpenGL Backend

Sumandora:
I recently worked a little to get dvui running inside another application by hooking its render/event functions.

Here's it running in Minecraft Bedrock Edition (running under Linux through Android compatibility layers) :^)

I ran into 2 major problems, first of all during the `begin` function, the vbo and ebo is not rebound, meaning that since Minecraft binds its own buffers the entire thing just implodes (usually with some error inside nvidias glcore, or whatever the mesa equivalent is for that). Fixing this did not fix the crashes.

After resetting some states it finally didn't crash anymore, but it also didn't render anything.

I found out through sheer luck, that setting `#version 330` was wrong, but instead I needed `#version 330 core`. There was no error log, because dvui currently assumes shaders compile fine. So this PR also adds some nice error logs when shader compilation fails.

### DIFF
--- a/src/backends/render/opengl.zig
+++ b/src/backends/render/opengl.zig
@@ -157,6 +157,7 @@ pub fn begin(self: *@This(), _: std.mem.Allocator) !void {
     gl.disable(gl.DEPTH_TEST);
     gl.disable(gl.STENCIL_TEST);
     gl.disable(gl.CULL_FACE);
+    gl.disable(gl.SCISSOR_TEST);
 
     gl.enable(gl.BLEND);
     gl.blendFunc(gl.ONE, gl.ONE_MINUS_SRC_ALPHA);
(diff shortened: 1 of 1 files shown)


david-vanderson:
Awesome!  Thank you!

I'll do a round of testing on my machines.  Where did you need to change `#version 330` to `#version 330 core`?  (I don't see it in the diff)

Sumandora:
> Awesome! Thank you!
> 
> I'll do a round of testing on my machines. Where did you need to change `#version 330` to `#version 330 core`? (I don't see it in the diff)

Presumably because the game runs through android emulation and thus opts for gles (GL for embedded devices), which doesn't support non-core. I just changed the string that you interpolate into the shader source.

It is not a general change, some people might have setups where putting core there could cause issues, so I wont change that in the project. You can add core to the version string yourself, but if we hardcode it then people cant remove it anymore. So I opt for the more flexible, not-hardcoding it.

Sumandora:
Turns out Minecraft also leaves scissor testing enabled, so we need to disable that as well... I guess the list of things to reset will grow the more people do with this...

david-vanderson:
Thank you!

PR #819, opened 2026-04-18
Fix: cursor not being captured on window leave

mdm5995:
This caused the mouse to be permanently 'held' until a new click event was sent.


david-vanderson:
Thank you!

PR #820, opened 2026-04-18
Fix/dx11 texture bugs

mdm5995:
These crashes were discovered when I was trying to skip the texture cache entirely and just have DVUI render every frame. The DX11 path was bugged, causing a double-free crash.

The texture cache logic was not written with optional bypass in mind, so the system only works with explicit calls to `textureUpdate()`, which was not implemented for dx11.

I verified that adding it to the cache avoids the error, but I also implemented a `textureUpdate()` function for dx11 since it didn't have one before.

### DIFF
--- a/src/Texture.zig
+++ b/src/Texture.zig
@@ -214,6 +214,9 @@ pub const ImageSource = union(enum) {
             if (invalidate == .always) {
                 var tex_mut = cached_texture;
                 try tex_mut.updateImageSource(self);
+                // if you don't add this to the cache, then it will never clean up the prev texture
+                // (it also won't cache this new texture, either, causing it to get it again from the cache, and try to double-free it)
+                dvui.textureAddToCache(key, tex_mut);
                 return tex_mut;
             } else return cached_texture;
         } else {
@@ -340,7 +343,6 @@ pub fn update(tex: *Texture, pma: []const Color.PMA, interpolation: TextureInter
         // texture update not supported by backend, destroy and create texture
         if (err == TextureError.NotImplemented) {
             const new_tex = try create(pma, tex.width, tex.height, interpolation);
-            destroyLater(tex.*);
             tex.* = new_tex;
         } else {
             return err;
--- a/src/backends/dx11.zig
+++ b/src/backends/dx11.zig
@@ -679,7 +679,28 @@ pub fn textureCreate(self: Context, pixels: [*]const u8, width: u32, height: u32
 
     try state.texture_interpolation.put(state.dvui_window.gpa, texture, interpolation);
 
+    log.info("created texture @0x{x}", .{@intFromPtr(texture)});
+
     return dvui.Texture{ .ptr = texture, .width = width, .height = height };
+}
+
+pub fn textureUpdate(
+    self: Context,
+    texture: dvui.Texture,
+    pixels: [*]const u8,
+) dvui.Backend.TextureError!void {
+    const state = stateFromHwnd(hwndFromContext(self));
+    // cast to resource to please the compiler god
+    const tex: *win32.ID3D11Resource = @ptrCast(@alignCast(texture.ptr));
+
+    state.device_context.UpdateSubresource(
+        tex,
+        0, // subresource
+        null, // full region
+        pixels,
+        texture.width * 4,
+        0,
+    );
 }
 
 pub fn textureCreateTarget(self: Context, width: u32, height: u32, interpolation: dvui.enums.TextureInterpolation) !dvui.TextureTarget {
@@ -758,10 +779,14 @@ pub fn textureReadTarget(self: Context, texture: dvui.TextureTarget, pixels_out:
 }
 
 pub fn textureDestroy(self: Context, texture: dvui.Texture) void {
+    log.info("Destroying texture @0x{x}", .{@intFromPtr(texture.ptr)});
     const state = stateFromHwnd(hwndFromContext(self));
     const tex: *win32.ID3D11Texture2D = @ptrCast(@alignCast(texture.ptr));
     if (!state.texture_interpolation.remove(texture.ptr)) {
-        log.err("Destroyed texture that did not have a stored interpolation", .{});
+        log.err(
+            "Destroyed texture @0x{x} that did not have a stored interpolation",
+            .{@intFromPtr(texture.ptr)},
+        );
     }
     _ = tex.IUnknown.Release();
 }
@@ -775,9 +800,13 @@ pub fn textureFromTarget(self: Context, texture: dvui.TextureTarget) !dvui.Textu
     defer state.arena.free(pixels);
     try self.textureReadTarget(texture, pixels.ptr);
 
+    log.info("Destroying texture @0x{x}", .{@intFromPtr(texture.ptr)});
     const tex: *win32.ID3D11Texture2D = @ptrCast(@alignCast(texture.ptr));
     const interpolation = if (state.texture_interpolation.fetchRemove(texture.ptr)) |kv| kv.value else blk: {
-        log.err("Target texture destroyed that did not have a stored interpolation", .{});
+        log.err(
+            "Destroyed texture @0x{x} that did not have a stored interpolation",
+            .{@intFromPtr(texture.ptr)},
+        );
         break :blk .linear;
     };
     _ = tex.IUnknown.Release();
@@ -1603,7 +1632,7 @@ pub fn main() !void {
 
     const window_class = win32.L("DvuiWindow");
 
-    var gpa_instance = std.heap.GeneralPurposeAllocator(.{}){};
+    var gpa_instance: std.heap.DebugAllocator(.{}) = .init;
     const gpa = gpa_instance.allocator();
     defer _ = gpa_instance.deinit();
 
--- a/src/backends/dx11.zig
+++ b/src/backends/dx11.zig
@@ -679,7 +679,7 @@ pub fn textureCreate(self: Context, pixels: [*]const u8, width: u32, height: u32
 
     try state.texture_interpolation.put(state.dvui_window.gpa, texture, interpolation);
 
-    log.info("created texture @0x{x}", .{@intFromPtr(texture)});
+    log.debug("created texture @0x{x}", .{@intFromPtr(texture)});
 
     return dvui.Texture{ .ptr = texture, .width = width, .height = height };
 }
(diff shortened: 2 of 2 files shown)


david-vanderson:
Sorry about that, and thanks very much for the PR!

I think we want to keep the `destroyLater()` in `update()` for people not using the cache.  I've pushed https://github.com/david-vanderson/dvui/commit/8bf0843e784a17f23f3dad0ac8c8e92cc03d3594 - does that fix the crash for you (prior to adding texture updating to the dx11 backend)?

If that looks good to you, can you adjust this PR to only add the textureUpdate() stuff?

mdm5995:
I'm confused, destroyLater uses the cache. It appends it to the texture_cache.trash list, which then gets freed at end of frame. I would expect someone not using the cache to allocate the texture on a per-frame arena, or some other lifetime scoped arena. The texture_cache.add() function adds the prev key to the trash list automatically. I wasn't able to find any way at all to fully bypass the cache because of this deep coupling.

I think a better pattern than 'defer destroy' like `destroyLater` is to allocate the texture explicitly on a frame lifetime memory region. This isn't really possible because of the way backends abstract away the allocations though, since dx11 is managing the texture memory, we can't take advantage of this bulk lifetime cleanup.

The issue with the update function calling destroyLater is that it will probably lead to double free issues at some point, since it's adding it to the cache trash, and if you add something with the same key to the cache, that also adds whatever was there previously to the trash list, potentially duplicating things? Not sure if i'm reproducing this correctly in my head right now, but that's why I chose to go through the cache. 

Would it be possible to have destroyLater not use the cache? create a list of handles to destroy maybe? Is there another way to destroy textures besides appending to trash that i'm missing?

david-vanderson:
> I'm confused, destroyLater uses the cache. It appends it to the texture_cache.trash list, which then gets freed at end of frame. I would expect someone not using the cache to allocate the texture on a per-frame arena, or some other lifetime scoped arena. The texture_cache.add() function adds the prev key to the trash list automatically. I wasn't able to find any way at all to fully bypass the cache because of this deep coupling.

This is confusing, and I'm not totally sure I'm right.  But I think texture_cache.trash being used for destroyLater can be thought of as an implementation detail - it doesn't necessarily mean you are using the texture cache for other stuff.

> The issue with the update function calling destroyLater is that it will probably lead to double free issues at some point, since it's adding it to the cache trash, and if you add something with the same key to the cache, that also adds whatever was there previously to the trash list, potentially duplicating things? Not sure if i'm reproducing this correctly in my head right now, but that's why I chose to go through the cache.
> 
> Would it be possible to have destroyLater not use the cache? create a list of handles to destroy maybe? Is there another way to destroy textures besides appending to trash that i'm missing?

I think that's essentially what is going on, and the confusing part is that the trash list is held inside Texture.Cache.  Maybe we should split it out?

The other way is to directly call `dvui.currentWindow().backend.textureDestroy()` (or `.textureDestroyTarget()`).  The reason that doesn't get used much is because of deferred rendering.  In subwindows all the rendering commands are stored and run later, so you usually don't want to destroy a texture until the end of the frame (it might be referenced by commands that haven't actually been executed yet).

Does that help?

mdm5995:
This is a difficult problem because we can't really access the memory on our own terms. Probably some handle system that accesses the textures indirectly would just let us double free without issue. That's a pattern I've appreciated, since use-after-free and double-free are both handled by the underlying system. 

In my case when I was trying to render a realtime scene, I was very confused by the 'never use cache' setting actually just using the cache anyways. 

An idea: separate the resource destruction from the cache, delegate that to a separate system. could still be called trash, but it's not related to the cache itself. Maybe:
```zig
const textures = dvui.currentWindow().texture_system;
const trash = textures.trash;
const cache = textures.cache;
```
when you add something to the cache, if it evicts a previous key it goes on the trash list just the same. 

As a possible extension, this trash list could have as many duplicates as you want, and just handle double frees internally somehow when it goes an does the actual cleanup iteration.  dvui would send the backend dvui-owned handles for a consistent api, and the backends could completely own the actual memory more directly (they have to anyway). if you free a bad handle, the backend returns an error instead of a segfault.

This would be a pretty large overhaul, but the fundamental issue I see is conflating resource management with caching strategy. Even without the handles, just separating out the trash and cache would solve a lot of these weird edge cases i think. then you could just destroyLater the old pointer and return the new one no problem since it would be totally independent of cache logic.

RedPhoenixQ:
> An idea: separate the resource destruction from the cache, delegate that to a separate system. could still be called trash, but it's not related to the cache itself. Maybe:
> 
> ```zig
> const textures = dvui.currentWindow().texture_system;
> const trash = textures.trash;
> const cache = textures.cache;
> ```
> 
> when you add something to the cache, if it evicts a previous key it goes on the trash list just the same.

With the behaviour you are describing, just renaming [`Texture.Cache`](https://github.com/david-vanderson/dvui/blob/main/src/Texture.zig#L49) to `Texture.System` and `window.texture_cache` to `texture_system` would be enough. The internal fields are already called `cache` and `trash` and work pretty much as you describe, minus the double-free handling. 

Double-free could be handled by the texture trash so you can add the same texture without crashing. This is probably a reason to keep the cache and trash in the same logic to prevent double-frees between the two. Use-after-free should only ever happen if the user explicitly calls `dvui.currentWindow().backend.textureDestroy()`, in which case they've taken responsibility for managing lifetimes. If you want the "free at end of frame" lifetime, you add the texture to the trash list. 

The `Texture` struct and  `Texture.Cache` is the resource management system your talking about that manages the lifetime of your textures. If you use the backend apis directly (`dvui.currentWindow().backend.texture...`) you are in full control of when the textures are freed without any caching, but then you have to manually ensure you follow dvui's expectation of texture lifetimes for subwindows etc.

> I was very confused by the 'never use cache' setting actually just using the cache anyways.

The [doc comment for the option I think you mean](https://github.com/david-vanderson/dvui/blob/main/src/Texture.zig#L180-L181) says that it will generate a texture _each frame_, not more often. So it won't cache across frames (and thus is "real time") but "creating" the exact same texture multiple times per frame doesn't do unnecessary work. This is also only an `InvalidationStrategy`, not a "caching strategy" implying that it's always cached and you are only choosing when it's removed from the cache.

> if you free a bad handle, the backend returns an error instead of a segfault

I don't think this is reasonable. When you are using the "lower level" backend apis directly (and not the destroyLater trash list) you are responsible for properly managing the resource. It's effectively the same as using an `Allocator` directly and double-freeing or giving it a pointer it didn't allocate. The backend should not have to keep track of all active resources within itself. If the external api allows if (dx11, sdl) it should of course return an error and not crash if you give it bad data, but that's up to the external graphics api.

david-vanderson:
I mostly agree, although I'm less sure about whether we should automatically deal with double-frees.

> This would be a pretty large overhaul, but the fundamental issue I see is conflating resource management with caching strategy. Even without the handles, just separating out the trash and cache would solve a lot of these weird edge cases i think. then you could just destroyLater the old pointer and return the new one no problem since it would be totally independent of cache logic.

I think we are almost there.  The trash and cache operate separately, but trash just happens to be located inside the cache.

Did my commit fix your crash?  If so, then let's get the textureUpdate part of this merged, and then I can separate the trash out from the cache.

RedPhoenixQ:
> With the behaviour you are describing, just renaming [`Texture.Cache`](https://github.com/david-vanderson/dvui/blob/main/src/Texture.zig#L49) to `Texture.System` and `window.texture_cache` to `texture_system` would be enough. The internal fields are already called `cache` and `trash` and work pretty much as you describe, minus the double-free handling.

@david-vanderson just as a heads up

mdm5995:
I'm currently working to get everything compiling on 0.16, so I can't quickly verify your patch. As I understand the code it should work, since there aren't two calls to `trash.append()` which was the issue previously. I can separate out the textureUpdate implementation into a separate PR if you like, but I could also just rename the `texture_cache.cache => textures.cache` and `texture_cache.trash => textures.trash`. Although that just fixes the conceptual coupling; the texture_cache.add() function still appends existing values to the cache which was the original cause of the bug. 

Perhaps the cache add function should return existing keys and let the user decide if it's garbage? then the happy path could be a composition:
```zig
const prev: ?Texture = textures.cache.add(...);
if (prev) |tex| {
    textures.trash.append(tex);
}
```
I'm working to get everything compiling on 0.16 for my project, so I should be able to have a build to test your patch soon-ish

david-vanderson:
> Perhaps the cache add function should return existing keys and let the user decide if it's garbage? then the happy path could be a composition:

Interesting will think about it.

> 
> I'm working to get everything compiling on 0.16 for my project, so I should be able to have a build to test your patch soon-ish

I've merged that fix into the zig16-dev branch.

mdm5995:
@david-vanderson Just want to add a comment here that you're patch fixed the bug (tested by reverting to an old build which uses image source (i switched to using raw textures instead of image source in later builds) and commenting out that branch, which causes an instant seg fault. uncommenting it resolves the seg fault. 

I had to get dx11 working with 0.16 to test it, and I decided to align the dx11 backend with the SDL backend code so that there is only one context object instead of a context (=== hwnd) and a WindowState struct that has the actual context. This simplifies the public 'on-top' usage since you don't need to create 2 different things. Would you be interested in a PR for that? I could loop in the textureUpdate function there as well. If not I can update this PR to remove the fix I did for the seg fault and just have the textureUpdate function.

As an aside, could anyone provide insight into the dvui internals that would explain why using a raw texture is orders of magnitude slower than using an image resource with invalidation = .always? my mental model has them as basically the same. If they are actually the same, it's possible there was an implementation difference causing it not to actually be updated, I didn't scrutinize too closely.

david-vanderson:
> @david-vanderson Just want to add a comment here that you're patch fixed the bug (tested by reverting to an old build which uses image source (i switched to using raw textures instead of image source in later builds) and commenting out that branch, which causes an instant seg fault. uncommenting it resolves the seg fault.

Great - thanks!

> 
> I had to get dx11 working with 0.16 to test it, and I decided to align the dx11 backend with the SDL backend code so that there is only one context object instead of a context (=== hwnd) and a WindowState struct that has the actual context. This simplifies the public 'on-top' usage since you don't need to create 2 different things. Would you be interested in a PR for that?

Let's hold off on that for a bit, we are just looking into the implications of having multiple OS windows in https://github.com/david-vanderson/dvui/issues/828 - I don't understand where this will go yet.


> If not I can update this PR to remove the fix I did for the seg fault and just have the textureUpdate function.

Yes please!

> As an aside, could anyone provide insight into the dvui internals that would explain why using a raw texture is orders of magnitude slower than using an image resource with invalidation = .always? my mental model has them as basically the same. If they are actually the same, it's possible there was an implementation difference causing it not to actually be updated, I didn't scrutinize too closely.

I don't have an answer, and it's possible that we are doing something wrong.  If you get a chance, making a reproduction example is probably what we would need to investigate.

mdm5995:
Ok, I just created #848 with the textureUpdate function and some debug logging for texture seg faults :) 

I think this PR can be closed now. I try to look at the ImageSource/Texture discrepancy if I find the time.

david-vanderson:
> Ok, I just created #848 with the textureUpdate function and some debug logging for texture seg faults :)
> 
> I think this PR can be closed now. I try to look at the ImageSource/Texture discrepancy if I find the time.

Thank you very much!

PR #821, opened 2026-04-18
Fix crash by using the correct pitch factor for the pixel formats in SDL backend

DBarbosaDev:
Streaming yv12/iyuv frames into a texture through `backend.textureUpgrade` causes crashes with a segmentation fault. The crash was triggered by an incorrect pitch value passed as the last argument to SDL_UpdateTexture. Additionally, the bytes per pixel for yv12 and iyuv formats had an incorrect value, instead of 1.5 bytes (12 bits), 1 byte was being used.

This PR fixes both issues by introducing the proper function to calculate pitch values and fixing the wrong value!

**Affected version**
v0.4.0

**Operative system**
Arch Linux with wayland

### DIFF
--- a/src/Examples/applets.zig
+++ b/src/Examples/applets.zig
@@ -15,6 +15,9 @@ pub fn applets() void {
     if (tabs.addTabLabel(active_tab.* == 2, "texture", .{})) {
         active_tab.* = 2;
     }
+    if (tabs.addTabLabel(active_tab.* == 3, "sub rect", .{})) {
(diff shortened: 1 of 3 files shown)


david-vanderson:
Thanks very much for the PR!

It prompted me to add a part to the demo to exercise `updateSubRect()`.

I know there was some trouble on the zig16-dev branch around similar stuff (yuv formats).  I will revisit the texture update api once main is synced with that branch.

Thank you!

PR #823, opened 2026-04-21
build.zig: Support custom SDL3 paths via build options

brodo:
The SDL3 lib now supports setting the library paths through build options. This PR allows users of DVUI to set these paths.


david-vanderson:
Thank you!

PR #824, opened 2026-04-23
Adding a unified system for styling and animations

KernelState:
# New features
- `StyleScheme`, which allows you to define default element styling
- `IconTheme`, which is a proper system for dvui.entypo (to be able to do custom Icon Themes)
- `AnimationRunner`, which makes a simpler event  based animator for hover, click, looping and constant animators that work like dvui.Animation but with a more friendly system
# How to use?
- `StyleScheme`: just define yours and use dvui.setStyleScheme on it
- `IconTheme`: just put in the fields, all what it is, is just a collection
- `AnimationRunner`: use the init function for local management or use `dvui.animationRunner` and `dvui.animationRunnerGet` which are almost exactly the same as `dvui.animation` and `dvui.animationGet` in terms of what they do. They  can also  be callback based by setting the frameFn field to something thats not null (you should run `next` for callback based and `value` for value based).


KernelState:
please note that this is just a draft and it doesn't even compile because of one function in struct_ui

KernelState:
btw, it does manually switch to llvm backend, simply because the zig backend is completely incompatible with me and many other people. here is the issue https://codeberg.org/ziglang/zig/issues/30959. it does destabilize the compiler but without it compatibility is broken with many systems like mine.

david-vanderson:
I will take a closer look at the struct_ui stuff.  Some initial thoughts:

Having some icons related to the theme makes some sense, I hadn't thought about that before.  Will think more but it should be a very small list of generic icons that dvui builtin widgets use (like the `windowHeader()` close X).

The style scheme stuff I will review when I go over #781 - the styling stuff is very hard to draw a clean line through.  I like the idea of consolidating all the builtin widget defaults somewhere.

The animation stuff I'm split on.  I'm not sold on the callbacks.  Making it easier to have looping stuff (maybe we should bake this into the low level animations), and helpers for animation on hover/focus sound great.  The hover/focus needs more work to account for subwindows but I can deal with that.

My intention is to keep this open as a placeholder while we extract and refine the parts to commit.

phatchman:
RE: struct_ui

When I compile with the original struct_ui code I get the error shown at the end of this message, including a stack trace. 

It is the below addition causing the issue as the HashMap uses many item pointers internally. These cannot be supported by struct_ui as we don't know how many items to display.
```
/// Extra colors for themes to use away from the default stylesheets
misc: ?std.StringHashMap(Color) = null,
```

The correct way to fix it would be to register a custom display function for that `misc` field to display the data in that hashmap. If that fields ends up being in the final, happy to make the changes to the struct_ui Theme example to support the display of the hashmap.

It seemed pretty clear to me what the issue was from the compile error, i.e. from the line numbers I could tell it was while displaying the Theme struct and from the error message, that it could not display a hashmap of Color and StringContext. But if there are any suggestion for improvement here, let me know. 

In my view the message should either stay as a compile error, or be logged and the field ignored. I know it was probably just changed to things to compile, but I don't want to keep it as a runtime panic. 

```
sdl3-standalone
└─ run exe sdl3-standalone
   └─ compile exe sdl3-standalone Debug native 1 errors
src\struct_ui.zig:1472:9: error: C-style and many item pointers not supported for [*]hash_map.HashMapUnmanaged([]const u8,Color,hash_map.StringContext,80).Metadata.metadata
                                 
        @compileError(std.fmt.comptimePrint("C-style and many item pointers not supported for {s}.{s}\n", .{ @typeName(@TypeOf(field_value_ptr.*)), field_name }));
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
referenced by:
    displayField__anon_224002: src\struct_ui.zig:1161:39
    displayOptional__anon_219743: src\struct_ui.zig:1425:25
    displayField__anon_212427: src\struct_ui.zig:1165:36
    displayStruct__anon_205297: src\struct_ui.zig:1541:29
    displayField__anon_193587: src\struct_ui.zig:1170:42
    displayStruct__anon_175708: src\struct_ui.zig:1541:29
    displayField__anon_158194: src\struct_ui.zig:1170:42
    displayOptional__anon_129673: src\struct_ui.zig:1425:25
    displayField__anon_110508: src\struct_ui.zig:1165:36
    displayStruct__anon_92215: src\struct_ui.zig:1541:29
    themeEditor: src\Examples\struct_ui.zig:219:32
    structUI: src\Examples\struct_ui.zig:139:20
    demo__anon_47868: src\Examples.zig:221:65
    gui_frame: examples\sdl-standalone.zig:251:23
    main: examples\sdl-standalone.zig:80:39
```

david-vanderson:
@phatchman Thanks for running that down.  We will keep it as a compile error, and that `misc` field won't be merged.  It's important that `Theme` be flat to simplify memory management, and we won't add more colors without some serious discussion about why the existing ones aren't enough.

KernelState:
thanks @phatchman btw, if anyone wants to contribute to the pr, js tell me first beforehand, as I am making breaking changes and swapping out the entire StyleScheme backend.

PR #835, opened 2026-04-30
fix/windows: replace std.os.windows.FALSE with std.os.windows.BOOL.FALSE

SimoneFelici:
Hi,
I was trying to build my app and noticed that the build for Windows didn't work with Zig 0.16.0.

Apparently, in Zig 0.16.0 they moved `std.os.windows.FALSE` into the `std.os.windows.BOOL` enum.
This fix replaces `std.os.windows.FALSE` with `std.os.windows.BOOL.FALSE`.

Let me know if you have some questions!
Thank you


foxnne:
I'm getting the same error

david-vanderson:
Thank you!

PR #838, opened 2026-05-04
zig 0.17 compat : remove array multiplication

Yinameah:
`**` operator will be removed in favor of `@splat`, which already works as expected in 0.16.0, so this commit implement required change in the spirit of future proofing.

See https://github.com/ziglang/zig/issues/24738
Interestingly, we have an occurrence of the "rare enough use case that doesn't deserve to be syntax" which requires 3 lines of comptime instead.


david-vanderson:

PR #841, opened 2026-05-04
zig 0.17 compat : remove most @cImport

Yinameah:
@cImport & friends are deprecated in 0.16 and should be integrated in the build system.

After this commit, c headers translation are moved to the build system, with the exception of the AcessKit stuff, partly because I don't know how it works and partly because I was bored to do this.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -74,10 +74,10 @@ pub fn linkSdl3(
             .library_path = opts.sdl3_library_path,
         })) |sdl3| {
             if (opts.target.result.abi.isAndroid()) {
-                sdl_mod.addIncludePath(sdl3.path("include"));
+                sdl_mod.addIncludePath(sdl3.artifact("SDL3").getEmittedIncludeTree());
                 addAndroidLibC(sdl_mod, opts);
             } else {
-                sdl_translate_c.addIncludePath(sdl3.path("include"));
+                sdl_translate_c.addIncludePath(sdl3.artifact("SDL3").getEmittedIncludeTree());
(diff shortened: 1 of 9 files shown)


Yinameah:
Ok, progress. I was fooled by my system headers.

Seems better now, but I still fail on SDL2 and I'm not sure how to solve this,  because the c include says `#include "SDL2/SDL.h"` and indeed my system headers are in `/usr/include/SDL2/SDL.h`.
However, looking in the `zig-pkg` folder, the SDL2 package has `SDL.h` directly inside a `include/` folder.

This means in the `sdl2-c.h` file, I would need something like :
```
#ifdef DVUI_SDL2_SYSTEM
#include "SDL2/SDL.h"
#include "SDL2/SDL_syswm.h"
#else
#include "SDL.h"
#include "SDL_syswm.h"
#endif
```
and pass yet another `-D` flag from the build system for system integration / build from source. This seems very inelegant.

I'm not sure how why it was working with @cImport, maybe I'm missing something stupid,

---
Orthogonal, but note that I realized I never tried the Android build, and I'm unsure how to do so.

david-vanderson:
> I'm not sure how why it was working with @cImport, maybe I'm missing something stupid,

Nope, this was really subtle.  I vaguely remembered something about it, and was finally able to shake it loose.

Merge when ready - I will take care of the accesskit stuff, and I also don't know how to run the Android thing, so we will have to rely on those using it to fix it if needed.

This is excellent, thank you very much!

Issue #828, opened 2026-04-27
Multiple OS windows


Yinameah:
Hello,

I keep coming back to this idea. Wouldn't it be cool if a `floatingWindow` could be an independent OS window ?
It would probably not be supported by all backends, but it would be a nice opt-in for the ones that do.

I realize it's probably a tricky road to take, but for large applications it's a killer feature, if only to take advantage of multiple physical screens.
I don't have a concrete use-case, but I am interested to dig in this direction, at least see what would be possible.

But I wanted to ask about it here first :
- Would this even be in the scope of DVUI ?
- Is there interest from others ?
- Any show stopper that anybody would already know of ?

foxnne:
I also thought about this with regard to my app, mainly for now I have windows such as an export window that allows previewing exports, which would be really cool to be able to keep open outside of the main window.

I've seen this feature supported in Dear IMGUI, here's some info on their implementation:

https://github.com/ocornut/imgui/wiki/Multi-Viewports

david-vanderson:
Multiple OS windows is definitely in scope for dvui, it just got put on the back burner.  The dx11 backend has a proof-of-concept exercised by the dx11-standalone example.

I'm not totally sure how the design should look, but so far the idea has been to allow nested `dvui.Window` runs.  You can see in `Window.begin()` how it remembers `dvui.current_window` and sets it to itself, and then resets in `Window.end()`.  That would fit with trying to offer an option on `floatingWindow` to be a separate OS window.

There are a few unknowns that I know about:
* What is the lifetime of these OS-windows?  The dx11-standalone example has a list (maybe that's not the right approach), but for `floatingWindow` maybe dvui/backend should internally store these somehow?
  * If the `floatingWindow` stops being run, dvui needs to detect this (using dataSetDeinitFunction maybe) and garbage collect the OS window
* How to route events to the correct window?  Maybe this is mostly backend work?
* Does dvui support any kind of cross-OS-window stuff?  Probably we say "no" initially and deal with it later.

Also there is a lot of state in `dvui.Window`, and some of it is really across all OS windows.  So we have to figure out a way to split that stuff out.

I'm very in favor of moving in this direction, so any work here is much appreciated!

Yinameah:
So, I played around a little bit, and it triggers a lot of questions. Maybe I'll throw a bunch here to see if it helps to frame the case.

[...]

PR #844, opened 2026-05-07
Multi OS windows - exploratory draft

Yinameah:
So, I've done some preliminary, dirty work following [#828](https://github.com/david-vanderson/dvui/issues/828). Just to test some hypothesis.

If you checkout this branch and run `zig build sdl3-mini`, you end up with a ugly window that can spawn a second OS window.

It's SDL3 only, and the second window just has a color square drawn directly by SDL.

The idea is that 
```zig
dvui.floatingWindow(@src(),.{ .dedicated_os_win = true },.{});
```
is all it takes to have an OS window.

I think this can work - modulo the obvious ton of details to iron out - and the next step would be to just have dvui swap the `current_window` during the `floatingWindow`.

However, I realized that all drawing actually [goes via this line](https://github.com/Yinameah/dvui/blob/694dd96648fea0aaed98d9a9487851a470d1ef5e/src/render.zig?plain=1#L100) and that the `current_window` is technically **just used to access the backend**. So the "dvui.Window maps to an OS window" is looser than I figured initially. (e.g. 

What we really want to do is access the **correct renderer** (leaving input handling aside for now). 
I'm not experienced with this stuff the slightest, but as I understand it, in SDL you would do need `SDL_Renderer` for each OS window in any case.
So if we want to swap which render dvui elements... well, render to, we need to change a bit the API and/or the reference model, so that a `FloatingWindowWidget` is able to switch the "current renderer" somehow.

I'm writing all this because I'm not sure to understand well the intent behind the [initRenderer](https://github.com/david-vanderson/dvui/blob/694dd96648fea0aaed98d9a9487851a470d1ef5e/src/Backend.zig?plain=1#L21-L29) pattern. As I understand it, the only reason for this is so that it allow to have a single renderer implementation be usable by different backends. (i.e. same opengl renderer for both wio and glfw). 
How does multiple OS windows work with opengl? Would you as well have one _instance_ of `backend/render/opengl.zig` for each OS window ? @ypsvlq, I would appreciate your comment on that very much, because in my current understanding the answer change quite a bit how I approach the problem.


david-vanderson:
> If you checkout this branch and run `zig build sdl3-mini`, you end up with a ugly window that can spawn a second OS window.

This is awesome.

> I think this can work - modulo the obvious ton of details to iron out - and the next step would be to just have dvui swap the `current_window` during the `floatingWindow`.

I think we can accomplish this by calling `Window.begin()` at the start of `floatingWindow` and `Window.end()` in `deinit()` for the separate OS window.

> 
> However, I realized that all drawing actually [goes via this line](https://github.com/Yinameah/dvui/blob/694dd96648fea0aaed98d9a9487851a470d1ef5e/src/render.zig?plain=1#L100) and that the `current_window` is technically **just used to access the backend**. So the "dvui.Window maps to an OS window" is looser than I figured initially. (e.g.
> 
> What we really want to do is access the **correct renderer** (leaving input handling aside for now). I'm not experienced with this stuff the slightest, but as I understand it, in SDL you would do need `SDL_Renderer` for each OS window in any case. So if we want to swap which render dvui elements... well, render to, we need to change a bit the API and/or the reference model, so that a `FloatingWindowWidget` is able to switch the "current renderer" somehow.

Very good point.  Yes the backend needs to have your `ChildOsWindow` for each separate OS window.  Off the top of my head the backend can either query dvui for the "current_window" when trying to route drawing, or we could have dvui send that information down through the `Backend` api.

> I'm writing all this because I'm not sure to understand well the intent behind the [initRenderer](https://github.com/david-vanderson/dvui/blob/694dd96648fea0aaed98d9a9487851a470d1ef5e/src/Backend.zig?plain=1#L21-L29) pattern. As I understand it, the only reason for this is so that it allow to have a single renderer implementation be usable by different backends. (i.e. same opengl renderer for both wio and glfw). How does multiple OS windows work with opengl? Would you as well have one _instance_ of `backend/render/opengl.zig` for each OS window ? @ypsvlq, I would appreciate your comment on that very much, because in my current understanding the answer change quite a bit how I approach the problem.

I think we need better terminology here, once we have multiple OS windows.  As I understand it, the `initRenderer` stuff is to allow a single renderer implementation across multiple backends, like you said.

For each OS window the renderer implementation will probably need a separate internal instance.  For example in the future we'll have a "SDL_Renderer" implementation, but internally it will need a separate `SDL_Renderer` instance for each OS window.

Maybe we call the renderer implementation a "RenderBackend" to distinguish it from instances of "renderers"?

ypsvlq:
> Would you as well have one instance of backend/render/opengl.zig for each OS window ?

Yes.

> `// FIXME : I'm not sure yet about the API, not even that createExtraWindow/destroyExtraWindow is the right move.`

I don't think it is. I think the intended design is that each `dvui.Window` has its own instance of `dvui.Backend` which maps to an OS window. The problem with this is dvui has no "global init" function, so `SDL_Init` is called in the same function as `SDL_CreateWindow`.

Currently backends are expected to set a global `dvui.io` variable, so I think a `dvui.init` function should be added that optionally calls into the backend's `SDL_Init` equivalent.

david-vanderson:
Assuming doing a nested `Window.begin/end` inside `floatingWindow` works, there's the question of the other things we do with a `Window` in the main loop:
* clearing the window before running dvui stuff
* try backend.setCursor(win.cursorRequested());
* try back.textInputRect(win.textInputRequested());
* try back.renderPresent();

I think logically those last 3 should happen in `floatingWindow.deinit()` just after calling `Window.end()`, but they aren't currently part of the Backend api, so probably need to be.

For the id, I think we can call `dvui.Window.init` and pass `.id_extra = floating_window_id`.

david-vanderson:
> > Would you as well have one instance of backend/render/opengl.zig for each OS window ?
> 
> Yes.
> 
> > `// FIXME : I'm not sure yet about the API, not even that createExtraWindow/destroyExtraWindow is the right move.`
> 
> I don't think it is. I think the intended design is that each `dvui.Window` has its own instance of `dvui.Backend` which maps to an OS window. The problem with this is dvui has no "global init" function, so `SDL_Init` is called in the same function as `SDL_CreateWindow`.
> 
> Currently backends are expected to set a global `dvui.io` variable, so I think a `dvui.init` function should be added that optionally calls into the backend's `SDL_Init` equivalent.

I'm unsure where we will have to draw the lines.  You are right we have no global init function, but we might not need that for these "child os windows".

But maybe this is the time to add one in any case.  Thinking ahead, we will likely have to split each backend into process-specific parts, and os-window-specific parts.

For this effort I'm good with "whatever works".

ypsvlq:
> we will likely have to split each backend into process-specific parts, and os-window-specific parts.

I agree, I think this mostly just consists of removing code from `backend.initWindow` that shouldn't be called every time. In my opinion a `dvui.init` function should be added since it needs to set `dvui.io` anyway.

> I'm unsure where we will have to draw the lines. You are right we have no global init function, but we might not need that for these "child os windows".

I don't really have input for the proposed `floatingWindow` <-> OS window API, but dvui does support multiple OS windows as is:

Screenshot_20260507_223224

(dirty patch to wio-app, creating a second `wio.Window`+`WioBackend`+`dvui.Window`)

I don't think introducing `ChildWindow` state to the backends is a good idea (eventually that struct will grow to be more or less the same as SdlBackend, i.e. the things dvui needs to manage an SDL window), but I do think the API around backend init and window creation could use some work.

david-vanderson:
> I don't really have input for the proposed `floatingWindow` <-> OS window API, but dvui does support multiple OS windows as is:
>
> (dirty patch to wio-app, creating a second `wio.Window`+`WioBackend`+`dvui.Window`)

Cool!  In that example does the main loop run each window sequentially?

> I don't think introducing `ChildWindow` state to the backends is a good idea (eventually that struct will grow to be more or less the same as SdlBackend, i.e. the things dvui needs to manage an SDL window), but I do think the API around backend init and window creation could use some work.

I don't know what we'll end up with.  One thing I'm learning from this effort is we are likely to end up with different models for "child OS windows" (tool windows or preview windows) that are conceptually linked to their "parent OS window" (like sharing minimization state) vs "multiple top-level OS windows" (separate documents within the same app?).  Any experimentation here is great.

ypsvlq:
> Cool! In that example does the main loop run each window sequentially?

Yes, this is just duplicating everything in the main loop.

> we are likely to end up with different models for "child OS windows" (tool windows or preview windows) that are conceptually linked to their "parent OS window" (like sharing minimization state) vs "multiple top-level OS windows" (separate documents within the same app?)

I think there doesn't need to be much distinction on dvui's side. In SDL you have `SDL_CreateWindow` and `SDL_CreatePopupWindow` but both return an `SDL_Window *`.

Yinameah:
Thanks a lot for the feedbacks and pointers. Very useful. Very appreciated.

> I don't think introducing ChildWindow state to the backends is a good idea (eventually that struct will grow to be more or less the same as SdlBackend, i.e. the things dvui needs to manage an SDL window), but I do think the API around backend init and window creation could use some work.

This is a valid concern, I will keep in mind while experimenting.

> I don't think it is. I think the intended design is that each `dvui.Window` has its own instance of `dvui.Backend` which maps to an OS window.

Fully agree, this is the assumption I see in the code as well. But so far there is only one of each, so maybe it doesn't **have to** ?

The more I read it, the more I see `dvui.Window` as mostly a global state and `dvui.CurrentWindow()` as a way to access this shared state that manages a stack of `Subwindow`. There is really not that much state dedicated to one OS window, except the reference to `Backend`, hence the ambiguity. The only exception to that I see is the "show dialog from other thread" use case. That part I don't understand well yet.

On the backend side, you have a bunch of functions that is related to an OS window, but not all. I suppose all the texture stuff could be shared for instance, time/sleep, init a single `std.Io` ...

What if the backend functions that actually manage an OS window just requires an `winid` of sorts ? It would be a way for dvui to state it's "preference" for having this part of the "frame" interact with another OS window, but the backend is also free to ignore this.

What I don't see yet is how much DVUI needs to know about whether the backend _did use_ another OS window or not, and how drawing from another thread works in this scenario.

But I have a good feeling about this, I think I'll follow a bit and see what's coming out.

Yinameah:
Ok, I'm gonna close this draft, because something is not right.

Basically I hit the issue that a FloatingWindowWidget register it's drawing calls for later and this means that I ended up calling into the backend with the wrong renderer in the main window `end()`.

I have a feeling this is a problem no matter what we do in the backend<->dvui problem space. 

So, yeah, don't know exactly where to go next, will take a bit of time to think about it.

david-vanderson:
Really appreciate the exploration here!  I will give it a shot when I can - thanks!

Yinameah:
Sure.
But I haven't given up. Just taking a step back... 
Will try to share another approach soon

PR #853, opened 2026-05-14
Fix macOS scrolling speed when refresh rate is over 60

### LINKED ISSUE #851: Scrolling speed seems to be FPS-dependent on new MacBook with ProMotion
Hi!

I recently got a new machine which has ProMotion (120hz vs 60hz) and in my DVUI app and in the demo, when at 120hz on my laptop scrolling on the trackpad is **2x** too fast. 

I've been looking trying to find any mention of this for SDL, but I haven't seen anything. It seems all my other apps scroll fine, so it must be something with SDL scroll speed again.

EDIT: After more investigation it seems that SDL3 on macOS uses CA/VDisplayLink which drives the apps main loop, which we use the callbacks for. The display link maintains the app's main loop at the refresh rate, which is apparently where our input is also being processed. 2x frames = 2x scrolling speed on macOS it seems.

foxnne:
From what I can tell, macOS input is driven by the CADisplayLink/CVDisplayLink. This means 2x as many events will be reported during 120hz display vs 60hz. I believe the correct fix is to just scale on macOS based on refresh rate. Please let me know if this is wrong!

Closes #851


david-vanderson:
Thank you!

PR #855, opened 2026-05-15
Add mouseTypeHint to Window, which apps can use to best-guess which the user is using

foxnne:
This is an attempt to make apps be able to read a heuristic DVUI implements already to determine if the scrolling of the mouse is coming from a trackpad or a mouse wheel. 

I called it just a hint, which may be the wrong wording, but I wanted it to be known it's just a hint and not an absolute certainty. 

Putting up in a draft so I can test.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -106,6 +106,28 @@ pub fn linkSdl3(
     sdl_mod.addOptions("sdl_options", sdl3_options);
 }
 
+/// Resolve the macOS SDK path via `xcrun --show-sdk-path`. Used to wire SDK include
+/// + framework paths into module-scoped C source. Errors if xcrun isn't on PATH or returns non-zero.
+fn resolveMacosSdkPath(b: *std.Build) ![]const u8 {
+    const argv: []const []const u8 = &.{ "xcrun", "--sdk", "macosx", "--show-sdk-path" };
+    const run = std.process.run(b.allocator, b.graph.io, .{
+        .argv = argv,
+        .stdout_limit = std.Io.Limit.limited(4096),
+        .stderr_limit = std.Io.Limit.limited(4096),
+    }) catch return error.MacosSdkPath;
+    defer {
+        b.allocator.free(run.stdout);
+        b.allocator.free(run.stderr);
+    }
+    switch (run.term) {
+        .exited => |code| if (code != 0) return error.MacosSdkPath,
+        else => return error.MacosSdkPath,
+    }
+    const path = std.mem.trimEnd(u8, run.stdout, " \t\r\n");
+    if (path.len == 0) return error.MacosSdkPath;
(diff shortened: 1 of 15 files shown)


foxnne:
After testing this a bit, it seems to work on Windows. But macOS still reports much smaller increments still, even with a mouse wheel, and I haven't yet been able to reliably tell whether its a trackpad or mouse wheel.

foxnne:
If anyone would like to test this, give fizzy a download and open up the settings pane and make sure control scheme is set to Auto. Then scroll, and the Auto tag should specify (mouse) or (trackpad). I believe this currently should work accurately for Windows and Linux, but not for macOS

foxnne:
@david-vanderson, Okay, from my research, the only practical way to determine on macOS is through checking for precise scrolling, which would require a small `.m` file. This also requires AppKit, but SDL3 would already be linking AppKit anyway, so no new dependency added, however it does need to link the module and find the SDK, so there is a bit of complexity added to the build.zig that is macOS specific.

I completely understand if this added complexity and mixing in `.m` files and looking for the SDK is not aligned with DVUI and not worth it. I believe it's not truly supported in any real way, but the benefit of having a mostly-accurate assumption of the user's hardware prevents needing the user to configure the app themselves. 

So I leave it up to you, I have `fizzy` already on my PR branch if you'd like to test (its update is live). I've verified it works on macOS, both touchpad and mouse, and Windows with a mouse (though i don't have a touchpad for Windows or any Linux device currently). If you feel this isn't worth it, I'll just use a default of one or the other in my app and just let the user set it themselves. 

Thanks!

david-vanderson:
Looks and tests good to me via fizzy.  Definitely worth using this - we can always see what shakes out about including the .m file, but this information is useful.

I want to test the other backends, and add a similar debug output to dvui.  I'm also tempted to change the `addEventMouseWheel()` api instead of having a new one.  I'll do a pass soon.  Thank you!

david-vanderson:
Done a first pass.  Let me know if you object to any of the changes.  I need to do one more round of testing on my machines.

foxnne:
This all looks good to me and I'm building with this and everything seems to work well :)

david-vanderson:
I ran into problems trying to get the glfw (and raylib) backends to work with the same heuristic.  Now trying a different approach that pushes the logic a bit into the backends.

I've done Mac, need to see if I can make Windows and Linux work.

david-vanderson:
Everything is working for me across Windows/Mac/Linux, with the exception of:
* wio backend on mac horizontal trackpad scrolling while holding shift
  * macs seem to treat the trackpad totally differently when holding shift - it looks more like a horizontal mouse wheel

david-vanderson:
Thank you!

PR #856, opened 2026-05-17
build: add option to configure the use of X11/Wayland for GLFW

lzj15:
Added an option to enable or disable X11/Wayland on Linux for the GLFW backend.


david-vanderson:
Thank you!

PR #858, opened 2026-05-18
Replace many deprecated functions

squidy239:
This replaces many deprecated functions including the `std.mem.copy` functions, `@intFromFloat`, the `indexOf` functions, and more. I also removed some unnecessary casts in places where the types now coerce and ran zig fmt.


david-vanderson:
Looks great!  I rewrote the `@memmove` changes to use double slicing so it's (hopefully) more obvious at a glance that the lengths are the same.

Thank you!

PR #859, opened 2026-05-19
Multi os window - working POC for sdl3

Yinameah:
Nothing ready to merge, just doing an PR for ease of test.

This is a "userland" implementation of "on the fly" creation/destruction of a second OS Windows. (with minimal tweaks to `sdl.zig` backend)

At least for me this exercise really helped to get a sense of how this could be done.

You can test with `zig build sdl3-mini`


Yinameah:
Tests are failing because I didn't update the other backends, but you can test my "demo" with `zig build sdl3-min` on this branch.

Yinameah:
Ah, note that I'm working on linux / x86-64 / X11. And that I tested on my machine only. Let me know if somebody has trouble with another platform

src/backends/raylib-c.zig
```diff
@@ -125,7 +125,9 @@ pub fn begin(self: *RaylibBackend, arena: std.mem.Allocator) !void {
 
 pub fn end(_: *RaylibBackend) !void {}
 
-pub fn clear(_: *RaylibBackend) void {
+pub fn clearWindow(_: *RaylibBackend) !void {
+    // FIXME multi-win : Should I call BeginDrawing here ?
+    c.BeginDrawing();
```
Yinameah: I don't know about this. Should `BeginDrawing` be part of the "standard" clear, i.e. the one `dvui.Window.begin` do for you by default ?

Yinameah: And note to self : double check consistency between the two Raylib backends in general

src/backends/sdl.zig
```diff
@@ -1861,11 +1921,6 @@ fn appIterate(_: ?*anyopaque) callconv(.c) c.SDL_AppResult {
         return c.SDL_APP_FAILURE;
     };
```
Yinameah: This I didn't get. I tried to compile with `-Dtarget=x86_64-windows` (and amazingly a working app popup to my screen via wine, this is crazy) but it wasn't using this `appIterate` function

Yinameah: (so I left behind debug stuff, need to at least clean that up)

src/backends/raylib-c.zig
david-vanderson: I'm not sure, but raylib doesn't support multiple OS windows in any case right?

src/backends/sdl.zig
david-vanderson: Ah sorry - you can change the line that checks `sdl_options.callbacks` so it doesn't depend on the OS.  Probably that line should change anyway so if callbacks is null we do it on windows/mac, but if callbacks is true we do it regardless?

david-vanderson: That is crazy that cross-compiling and running under wine worked.  I don't know why that didn't get you into the callbacks path.

src/backends/raylib-c.zig
Yinameah: No, indeed. One win only.

But it's more about consistency in the examples. Maybe I rename the `dvui.Window` opts flag to `prepare_window : bool = true` so it encompass better the idea : "A default bunch of stuff to do before drawing, that you can opt-out if you know what you are doing" and we include `BeginDrawing` ?

david-vanderson: Something like that sounds right.  `Window.begin()` already calls `backend.begin()`, so maybe we thread it through that (maybe backend.begin() needs to take a parameter saying if it is a child window or something)?

Yinameah: Ouuuh, this is a good idea. This way the backend just deals internally with that, and it's one less function in the interface, and less mess in the PR.

src/backends/sdl.zig
Yinameah: Yeah, couldn't belive it. Still don't fully.

Anyways, indeed if I remove the check I can use that code path, and it works, and anyways now that I reverted the window.begin() changes it's a non issue

Yinameah:
Ok. The diffs are less all over the place now.

I have a plan for normalizing the `initWindow` calls but I would prefer to continue in another PR, it will make things more reviewable.

To summarize the changes in this PR :
- Added 3 methods to the Backend interface : 
  - `setCursor`
  - `textInputRect`
  - `renderPresent`
  In some backends, some are no-ops but this allow ...

-  Introduction of `dvui.Window.endOptions.manange_rendering`
    - True by default, and makes Window.end() calls the 3 function above.
    - This simplify a bit some exemples
    - But crucially allows dvui to nest a `Window.begin()/end()` smoothly for multi os win
    - "ontop" exemple use `manange_rendering=false` so it makes it clear you can opt-out

- Tweaks in `sdl` backend
  - mainly added a flag that allow calling `SDL_Init` and `SDL_Quit` only once while calling `initWindow`/`deinit` multiple times.
  - This is a bit duck taped, but I'll rework this in the next stage.
- Added `exemples/sdl-multi-win.zig`. 
  - This is temporary playground that will eventually be removed once a convincing show up of multi os win is integrated in the main demo.
  - you can test with `zig build sdl3-multi-win`

Let me know what you think about this in the current state.

david-vanderson:
> Ok. The diffs are less all over the place now.
> 
> I have a plan for normalizing the `initWindow` calls but I would prefer to continue in another PR, it will make things more reviewable.

Sounds good.

>     * Added 3 methods to the Backend interface :
>       
>       * `setCursor`
>       * `textInputRect`
>       * `renderPresent`
>         In some backends, some are no-ops but this allow ...

I think these should not throw errors?  I can't think of what user code would do if they did throw errors, except crash.

> 
>     * Introduction of `dvui.Window.endOptions.manange_rendering`

This is good.  Maybe "manage_backend"?  Your call on naming.

> Let me know what you think about this in the current state.

Looks good - go ahead and merge this and we can follow up with more.  Thank you!

Yinameah:
> This is good. Maybe "manage_backend"? Your call on naming.

Yeah, this makes sense. Done.

> I think these should not throw errors? I can't think of what user code would do if they did throw errors, except crash.

Adding error to the interface followed that theses 3 functions can all error in the current `sdl` backend implementation. 
I guess we could just `toErr` for log and swallow. But then wouldn't you have the same argument for `Window.end()` itself then ?

Let me give it a shot at least for these 3 functions...

Yinameah:
Ok, I removed these error path. Please have a good look in the last commit, in sdl / sdlgpu and dx11, the `setCursor` had a error path via `cursorShow` that I'm not fully convinced is the best way to tackle this.
I think the "there is not much the user can do about" holds, but I wonder about "debuggability" : Isn't crash a better behavior ?
But maybe it's just the `boolToErr`/`toErr` job. This function could like crash on debug build and log on release for exemple.

david-vanderson:
> Adding error to the interface followed that theses 3 functions can all error in the current `sdl` backend implementation. I guess we could just `toErr` for log and swallow. But then wouldn't you have the same argument for `Window.end()` itself then ?

Yes - still unsure whether `Window.end()` should error or not, but we don't have to answer that now.  Thanks!

> Ok, I removed these error path. Please have a good look in the last commit, in sdl / sdlgpu and dx11, the setCursor had a error path via cursorShow that I'm not fully convinced is the best way to tackle this.
> I think the "there is not much the user can do about" holds, but I wonder about "debuggability" : Isn't crash a better behavior ?
> But maybe it's just the boolToErr/toErr job. This function could like crash on debug build and log on release for exemple.

Yeah that's possible - I'm not convinced we have the best error handling strategy.  For now it seems like logging something and continuing is good, and generally keeping the same behavior in debug and release is good?  At least debug mode implies someone might be looking at the logs.

I'm good with you merging this.  Let me know if there's anything else for this PR or you run into problems.

youssless92:
My app broke after this change, the new behaviour is that it looks like, it is always "refreshing" or the window is hidden/shown continuously on MacOS ( not able to test on linux or Windows for now).
I use SDL3 as backend.
the first commit after which it broke is 06d61863338f6e9150907bd3b2dc5f15156d4c08.

Is there any new consideration to take into account after this change?
attached are the before and after.

Thanks!

https://github.com/user-attachments/assets/2b181ff6-2a30-4645-920e-60b895bc9932

https://github.com/user-attachments/assets/52de99e9-cbd9-4aa7-b454-420a50b667fb

Yinameah:
Sorry for the mess. 

Are you using the `dvui.App` mechanism or you are in charge of the main loop "standalone style" ?

Yinameah:
At least I don't see anything obvious in this particular commit. But the one just before I think is the one that introduced the `manage_backend` option in `dvui.Window.end()`, and default value is `true`. 

So I removed in the examples ([here in sdl-standalone](https://github.com/david-vanderson/dvui/blob/715ffada64fdab5c440641d4e7a4e59f2c664edc/examples/sdl-standalone.zig?plain=1#L75-L79))  the three backend calls that are now [done by default](https://github.com/david-vanderson/dvui/blob/715ffada64fdab5c440641d4e7a4e59f2c664edc/src/Window.zig?plain=1#L1646-L1650)

Your video gives me a bit the feeling of double call on `backend.RenderPresent()`.  But maybe it's unrelated, and something else trigger a spurious refresh.

Yinameah:
Also, I don't see it in your video, but is it the full OS Window that flickers or is it just an UI element, or a floatingWindowWidget ?

youssless92:
Thanks for the reply,

No I don't use dvui.App, I manage the main loop.
The UI elements are flickering.
backend.renderPresent() is called once at the end of the main loop.
I removed the transparency and put back the frame to show you better what is happening in the video.

https://github.com/user-attachments/assets/1b3528f2-85c0-4394-874f-abc17e98c8d4

david-vanderson:
> Your video gives me a bit the feeling of double call on `backend.RenderPresent()`. But maybe it's unrelated, and something else trigger a spurious refresh.

Good guess.  I can reproduce on macos with `sdl3-standalone` if I add `backend.renderPresent();` just after `win.end(.{})`.

@youssless92 Can you remove your call to `backend.renderPresent();` and see if that solves it?

@Yinameah Assuming that is the problem, is there a way we could put a check in the backend to detect this case and spit out a debug log/error?

youssless92:
@david-vanderson  Yes removing it fixed it. Thanks!
Not sure why though 😅
It is a mistake to call it ? or is  SDL_RenderPresent called elsewhere?

Yinameah:
> @david-vanderson Yes removing it fixed it. Thanks! Not sure why though 😅 It is a mistake to call it ? or is SDL_RenderPresent called elsewhere?

It's a "mistake" to call it twice. Same for `setCursor` & `textInputRect`. 
Because it's called for you already in `win.end(.{})`

You can either remove these 3 calls from your code, **or** call `win.end(.{.manage_backend=false})`, which will not call that for you. In any cases, these 3 backend calls should be called once per frame.

But the breakage is definitely on me because the default behavior before was "dvui.Window is not in charge of this" now it's "I'll do it except if you ask me not to"

I actually hesitated quite a bit about the default value for `.manage_backend` because I knew this was changing the default behavior, but in general I expect most users not to care too much about this, just copy-paste the demo example, and in that case it's a better default because it's less lines in your code ;-)

Does this make sense ?

david-vanderson:
> @david-vanderson Yes removing it fixed it. Thanks! Not sure why though 😅 It is a mistake to call it ? or is SDL_RenderPresent called elsewhere?

No, this is entirely on us.  @Yinameah has been working on nested separate OS windows, and as part of that `Window.end` started doing some of these backend calls automatically.

I'm pretty sure that's the right thing to do, but it presents a problem for people upgrading because we don't yet have a mechanism to catch this situation.  Thanks very much for the report, it means we definitely need one!

Yinameah:
> > Your video gives me a bit the feeling of double call on `backend.RenderPresent()`. But maybe it's unrelated, and something else trigger a spurious refresh.
> 
> Good guess. I can reproduce on macos with `sdl3-standalone` if I add `backend.renderPresent();` just after `win.end(.{})`.
> 
> @youssless92 Can you remove your call to `backend.renderPresent();` and see if that solves it?
> 
> @Yinameah Assuming that is the problem, is there a way we could put a check in the backend to detect this case and spit out a debug log/error?

Yeah, we could. With a bool flag in the backend set in `begin`. But shall we test the 3 functions ?

```
// debug : detect multi rendering in one frame
has_setCursor: bool = false,
has_textInputRect: bool = false,
has_renderPresent: bool = false,
```
All set to false in `backend.begin`, set to true in their respective function. log.warn if already true in their respective function, and log.warn in `backend.end` if still false ?

A bit mouthful maybe ... And then do the same in all backends ?

youssless92:
Thanks guys!
I see what happened now, actually in my app I manage the cursor in some cases (mainly when some events happen) so I do need to call it. I will check if I can change my code to use the new default behaviour but most likely I will set manage_backend=false. 
Also I am interested in this change because I was looking for a way to have multiple os windows, even looked into doing it via SDL directly, so I definitely approve of this PR ;)

Yinameah:
> but most likely I will set manage_backend=false.

My expectation is that anything remotely not standard would do this, and if this is your case don't be shy. 

If you have a use case for my osWindow impl, give it a try. It's crude but functional and any real life feedback would be appreciated at this stage. Feel free to open issues related to this an tag me, I'll have a look.

youssless92:
> > but most likely I will set manage_backend=false.
> 
> My expectation is that anything remotely not standard would do this, and if this is your case don't be shy.
> 
> If you have a use case for my osWindow impl, give it a try. It's crude but functional and any real life feedback would be appreciated at this stage. Feel free to open issues related to this an tag me, I'll have a look.

Yes I definitely will try your child window impl, I will try to replace the floating window because I want it to be outside of the main window. 

Otherwise removing only `renderPresent` seems to have done the trick because I still call `setCursor` and `textInputRect`
I will leave it this way for now unless I find new issues.

Yinameah:
> Otherwise removing only renderPresent seems to have done the trick because I still call setCursor and textInputRect
I will leave it this way for now unless I find new issues.

This is still doing some useless work but won't necessarily show off indeed.

david-vanderson:
> Yeah, we could. With a bool flag in the backend set in `begin`. But shall we test the 3 functions ?

Maybe just testing for the renderPresent is enough?  The others are less likely to cause a problem.  And in any case this is a temporary problem for people upgrading.  At some point down the road we could remove the test.

> And then do the same in all backends ?

I think so?  Let me know if you want me to tackle it.  Thank you!

Yinameah:
I can propose an implementation

PR #861, opened 2026-05-19
Updates for web to get fizzy building

foxnne:
This is my first time trying to get a web build working, so please forgive me for mistakes. I've been working on this today, and so far this is all I've needed to make changes to to get fizzy running on web preliminarily.

I'll leave this a draft for now until I can review it a bit more with better knowledge.

### DIFF
--- a/src/Font.zig
+++ b/src/Font.zig
@@ -308,22 +308,26 @@ pub const TextSizeOptions = struct {
 pub fn textSizeEx(self: Font, text: []const u8, opts: TextSizeOptions) Size {
     // ask for a font that matches the natural display pixels so we get a more
     // accurate size
-    const ss_raw = dvui.parentGet().screenRectScale(Rect{}).s;
-    // Guard zero scale (can happen before the first real layout pass on web).
-    const ss = if (ss_raw > 0) ss_raw else 1.0;
+    const ss = dvui.parentGet().screenRectScale(Rect{}).s;
     const ask_size = self.size * ss;
+
+    // set some reasonable defaults in case things go bad
+    if (opts.ascent_out) |ao| ao.* = 10;
+    if (opts.end_idx) |endout| endout.* = text.len;
+    if (ask_size == 0.0) {
+        if (opts.ascent_out) |ao| ao.* = 0;
+        return Size{};
+    }
+
     const sized_font = self.withSize(ask_size);
 
     const cw = dvui.currentWindow();
-
-    if (opts.ascent_out) |ao| ao.* = 10;
 
     // might give us a slightly smaller font
     const fce = dvui.fontCacheGet(sized_font) catch return .{ .w = 10, .h = 10 };
 
     // this must be synced with dvui.renderText()
-    const em_height = if (fce.em_height > 0) fce.em_height else 1.0;
-    const target_fraction = if (cw.snap_to_pixels) 1.0 / ss else self.size / em_height;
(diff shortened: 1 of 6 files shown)


david-vanderson:
Very cool.  So far looks pretty good to me!

foxnne:
I'm going to go ahead and mark this as ready for review so we can discuss any needed changes here, or if any of this needs to be removed. The bulk of the code changes were adding support for touch-hold context menus. I know we discussed a bit that that might not be the best user experience, but I've yet to land on better.

Please let me know of any changes you need made here, and I'll go ahead and rebase on main.

david-vanderson:
The web.js and web.zig changes are great - thank you!

The `Font` sizing thing is something I've wobbled back and forth on.  Should we return a zero size if the asked for size is zero?  Can you say more about the problems you hit with this?

For the context touch hold thing:
* I'm good with adding it, people are going to ask for it
* I added forcing frames while waiting for the long press - I found it testing with the debug "Simulate Touch"
* This prevents you from touch-clicking on a widget inside the context area right?  (Also touch-scrolling)
  * I will think about how we can solve this

david-vanderson:
I pushed a change to make the long press more passive.  I'm not totally sure what all the implications are, but it allows pushing buttons and touch scrolling inside the context.

I did remove `tapOccurred()`, but am unsure about it.  Do you still need that if the widgets inside the context area can receive the touch events now?

foxnne:
> The `Font` sizing thing is something I've wobbled back and forth on. Should we return a zero size if the asked for size is zero? Can you say more about the problems you hit with this?

This may be the wrong fix, but it seems there are scenarios on the web build during initialization where natural_scale is 0, which lead to a divide by zero creating the font using scale [here](https://github.com/david-vanderson/dvui/pull/861/changes#diff-eea8737ca96c127d983c22c27aeb15dd6d06f25db29f051ce722bc6179927cfcR313).

So really it was trying to prevent issues where scale might be zero, which should I think never be the actual case?

david-vanderson:
One minor thing - when tracking elapsed time, we should use `dvui.secondsSinceLastFrame` and accumulate the result.  In the rare case that something goes wrong with the underlying clock, it will report reasonable values (10ms - 71.5 minutes)

david-vanderson:
> This may be the wrong fix, but it seems there are scenarios on the web build during initialization where natural_scale is 0, which lead to a divide by zero creating the font using scale [here](https://github.com/david-vanderson/dvui/pull/861/changes#diff-eea8737ca96c127d983c22c27aeb15dd6d06f25db29f051ce722bc6179927cfcR313).
> 
> So really it was trying to prevent issues where scale might be zero, which should I think never be the actual case?

Good info - I will look into if natural_scale can ever be zero, and also tighten up `textSizeEx` when the scale is 0 (which could be from a bad `ScaleWidget` use.  Thanks!

foxnne:
> One minor thing - when tracking elapsed time, we should use `dvui.secondsSinceLastFrame` and accumulate the result. In the rare case that something goes wrong with the underlying clock, it will report reasonable values (10ms - 71.5 minutes)

I believe I have handled this, fixed the conflict and merged main. I also have built fizzy's web build against this branch, and it seems all works fine still, feel free to test if you'd like: http://fizzyed.it/app/

david-vanderson:
Pushed a commit to early exit out of `textSizeEx` - no reason to go through all that logic if the scale or font size is zero.

Can you test one more time to make sure it works with Fizzy?  If that works, then merge it.  Thank you!

foxnne:
All looks good to me, merging!

PR #864, opened 2026-05-21
Render icons as cached triangle meshes instead of rasterized textures

nat3Github:
Note of Caution, most of this PR was vibecoded with Claude

Previously each icon was rasterized by svg2tvg's z2d backend into a CPU pixel buffer, uploaded to the GPU as a Texture, cached, and drawn with renderTexture. 

This PR Replaces dvui's icon rendering path, tessellating once into a triangle mesh, cached, and replayed every frame as geometry, recached on size change. no textures, no CPU rasterization.

the triangulization happens in a branch of tvg2svg. Users who want rasterization can still use svg2tvg (which still has the rasterization code using z2d)
https://github.com/nat3Github/zig-lib-svg2tvg/tree/dvui-render, there is a option gated dvui demo in that branch
```
zig build -Ddemo=true demo
```

Performance:
CPU rasterization is slow compared to calculating triangles, and rendering triangles on modern gpus is cheap. in my measurements ive seen ~10x speed up (widget, did not measure gpu).  when the pixel buffer is cached its faster than the mesh lookup but mesh lookup is still fast in absolute numbers and first paint reduces dramatrically which is the better behavior for immediate mode in my opinion. it should be fast enough to redraw every frame (if someone wants to do crazy animation on tvgs)
 
What changed

  - renderIcon (src/render.zig): builds an svg2tvg_dvui.MeshBuilder mesh on
  cache miss via appendTvg, anchored at (0,0). Each frame the cached mesh is
  duped into the frame arena, scaled/translated to the target rect, rotated
  (opts.rotation), tinted (opts.colormod), and submitted via renderTriangles.
  The z2d raster path for icons is gone.
  - Window (src/Window.zig): new icon_mesh_cache (keyed by hash(bytes, height,
   IconRenderOptions)) + icon_mesh_frame. Stale entries (untouched last frame)
   evicted in begin(); meshes freed on deinit().
  - dvui.zig: re-exports svg2tvg_dvui and adds IconMeshCacheEntry (both
  useTvg-guarded).
  - build.zig: wires the svg2tvg_dvui module — a fresh instance per dvui
  module, since it imports dvui and can't be shared across module graphs.
  - build.zig.zon: svg2tvg dep bumped to the dvui-render branch, which extends
   RenderOptions with fill_color_override / stroke_color_override /
  stroke_width_override / disable_fill. IconRenderOptions maps 1:1 onto these.

 Public icon API  should be unchanged — icon, buttonIcon, iconWidth, renderIcon,
 IconRenderOptions, IconWidget all keep their signatures.


david-vanderson:
Interesting!  I took a quick look at the code, haven't done any testing.

Why does svg2tvg need a dvui import now?  Makes sense locally for testing, but I don't get why the svg2tvg thing that dvui imports needs a dvui import?

nat3Github:
good point, yes it's only for testing (the demo) behind a lazy gate (-Ddemo=true), not needed for the lib. i'm not sure but i think zon still fetches them  currently but does not use them? i don't know if there is a cleaner way to do that without having it out of tree (maybe out of tree is cleaner with local path zon dep)? i'll be back on sunday/monday, will try that.

nat3Github:
sorry did not have time the last weeks, the demo app in svt2tvg is now seperated into a subdir, so dependencies should be clean now. z2d is used lazily.

to try dvui with the change just checkout this dvui-branch.
to run the demo in svg2tvg, checkout svg2tvg, cd into the demo sub directory, then zig build

update: decided to delete the tvg->texture route completly as it removes complexity, uses still can use it as they can copy the old source code or implement themselves. reason: if the gpu triangle route is truly better dont keep the texture route.

david-vanderson:
The `svg2tvg_dvui` module is not needed anymore right?  Does svg2tvg's build.zig just need to create a module for `root_dvui.zig`?

What's the story with antialiasing?  When I run this branch the icons all look correct but have a hard edge on them.

nat3Github:
the svg2tvg_dvui module (created in dvui's build.zig) == svg2tvg+dvui (it uses root_dvui.zig as root module, then adds dvui as import). Sorry if I misunderstood the question.

about antialiasing, z2d supports antialiasing when drawing (CPU rendered)
with this branch we render down to triangles which then get rendered by the dvui backend, so i guess its Backend implementation dependent how the triangles get drawn on the gpu. My understanding is that i.e. buttons also get rendered as triangles, so they should have the same antialiasing properties (is that true)?

im curious how the hard edge looks on your pc. i checked side by side in testing and could not really make out any differences, i didnt pixel peep though.

david-vanderson:
This is what this branch (left) vs main (right) looks like on my low dpi monitor:

vector_vs_raster

So far dvui backends have not done any antialiasing.  What we do is draw extra skinny triangles that fade to transparent at the edges of polygons and strokes.  The "fade" parameter controls the width from full to transparent.

Aside from rendering icons, this implements a general concave polygon renderer right (maybe not self-intersecting)?  My guess is that for each edge of the polygon, we'd want to render the stroke fade on the "outside" of the edge.

Is there a place in the code where it knows the sequence of "outside" (I guess "inside" for holes) edges?

nat3Github:
mmm on closer look i can see the anti aliasing too, it was just that my high dpi monitor did not make it unpleasant enough for me to see it :((. converted to a draft for now till i have time to work on it, right now its not a better solution to the status quo.

on the aliasing topic, i am curious if there is a design decision behind dvui not supporting anti aliasing through backends? 
i dont know much about graphical rendering but from a short search for about < 500 k triangles (i think reasonable dvui guis should never! reach those dimensions), 4xMSAA hardware antialiasing is faster then if you would blow up the triangle count (estimated 3x) with the small triangle method. 
and you would loose that complexity in source code and in cpu runtime.

thoughts on runtime performance:
things like the icons would be only meshed once by cpu
but the cost in rendering is existent in every frame, longer execution time on gpu usually also means higher power draw (which is something i personally really care about). of course performance claims without measuring are a bit pointless...

david-vanderson:
> on the aliasing topic, i am curious if there is a design decision behind dvui not supporting anti aliasing through backends? i dont know much about graphical rendering but from a short search for about < 500 k triangles (i think reasonable dvui guis should never! reach those dimensions), 4xMSAA hardware antialiasing is faster then if you would blow up the triangle count (estimated 3x) with the small triangle method. and you would loose that complexity in source code and in cpu runtime.

This is a place where I'm totally unsure and haven't thought about it much.  The current strategy is basically copied from dear imgui, but I don't know why they do it.

I suspect there are situations/hardware where hardware antialiasing is not present?  Some very quick searching suggests maybe older/embedded/exotic hardware might be the main sticking point here?  That would explain dear imgui.

It's also possible that in some of those situations developers would need to turn off the fade for performance in any case.

I guess for progress here we need more information, but I'm not sure how to get it.

nat3Github:
>This is a place where I'm totally unsure and haven't thought about it much. The current strategy is basically copied from dear imgui, but I don't know why they do it.

i would guess the main reason is portability and simplicity.

imo the topic of antialiasing and performance goes way deeper and its a tradeoff:
- currently we are using triangles and textures for rendering 
- that keeps backend integration simple but it doesnt seperate the `what` of what we are drawing from the `how` and that is ultimately hurting performance.
- for example if the intermediate render Api would have a drawText or a drawPath, each Backend could use the most efficient method it knows to accomplish that (which is different for different platforms, some platforms come with OS native text rendering for example)
- the big question is how much decoupling is actually needed to unblock more efficiency and whats practical
- how would a different intermediate API look like, what would it be composed of (data wise, the `what`)
- we would shift the "buttons do their own antialiasing" to centralized handling (i.e. when rendering to triangles)
- that could also be the migration route: add a Triangulizer which translates different commands to a mesh till a backend decides to adopt a more efficient rendering mechanism for a part of the render Api

since its your project, do you think this is still inline with dvui or against core principles of dvui?
i could see myself exploring more on this topic (maybe not right now but on vacation in a few months)

note: lately, when doing some more graphically advanced stuff like shadowing and arbitrary paths and shapes the thought came to my mind that part of what i was doing belongs to rendering/backend specific code and should rather live behind a more advanced render API

david-vanderson:
> * that keeps backend integration simple but it doesnt seperate the `what` of what we are drawing from the `how` and that is ultimately hurting performance.
> 
> * for example if the intermediate render Api would have a drawText or a drawPath, each Backend could use the most efficient method it knows to accomplish that (which is different for different platforms, some platforms come with OS native text rendering for example)
> 
> * the big question is how much decoupling is actually needed to unblock more efficiency and whats practical

This makes a lot of sense but it all depends on how much benefits are possible vs. the cost of maintaining extra rendering paths.  Multiple rendering paths also make bug reporting harder.

There was an old issue about doing higher-level kinds of drawing commands, but it never went anywhere.

Things like drawPath seem relatively straightforward, but for something like drawText its hard to imagine how all the font information would be communicated.

Given the increased maintenance, I'd want to see some measurable benefit paired with a usecase that requires that benefit.

On the antialiasing front, we could investigate ways to detect if hardware antialiasing is available and if it's on then we can turn off our fade.  That could be good for performance but unfortunately doesn't allow us to rely on it.

nat3Github:
Ahhh, i actually reimplemented this in another branch but force pushed it here so it opened the old PR, anyways ill just leave it here.

This now is the first step towards #923. 
AA issues should be gone / not introduced.
feel free to test and critique before merge.

changes: move the tvg dvui rendering parts into dvui by adding 
- Path.fillTriangles -> handles non-convex shapes, earcut algo under the hood w/ some mods + dvui style AA bolted on
- Path.Builder.addQuadBezier and Path.Builder.addCubicBezier 


main thought here: why have those in a seperate module when (other) widgets could take advantage of these rendering capabilities, see #923. (note: if its central in dvui rendering we can optimize it better, i.e. if at any point a better implementation path is found mbe through backends, pipeline optimization etc.)

note of caution: i ran some benchmarks but dont have meaningful numbers (quite complex to make a valid statement here). not to say its slow, i just cant classify the performance implications here fully (triangles vs texture (we could theoretically also render triangles to a texture, cache that, now we are meshing on first frame and caching mesh)).
for changing icons per frame, rendering triangles obsiously beats z2d+cache trashing.

> i would guess the main reason is portability and simplicity.

i think this still stands, as discussed in #923. i changed my mind a little bit and would agree with david that the easy route is disabling AA if backend does MSAA. all other platform/backend optimizations like gpu-SDFs must show hefty performance upsides (in real benchmarks) to consider more code branching. my hunch right now there is more to be gained in #637 and Texture Atlasses

david-vanderson:
I like this a lot.  I'm going through it but it's pretty dense so is taking me a while.

I'm impressed you were able to add the AA fade thing!

It looks like `ask_height` is part of the hash, meaning if the size of the icon changes the we regenerate the mesh right?  Is is possible to generate the mesh at some "reasonable" size and scale the vertices on the fly for different ask_heights?

nat3Github:
> is possible to generate the mesh at some "reasonable" size and scale the vertices on the fly for different ask_heights?

in general no, atleast i am not sure how. thats because curves are rendered as polygons, if you blow up the size you would start to see edges in the circles so to say and the other way round you would just render unneeded triangles. 

you could do an intermediate step precalculating  an optimized representation like all pivot points a shape is made up, then recalculate from there. but all pivot point making up a shape is basically your tvg file so little to be gained.

| set                | icons | avg (µs) |
|---------------------|------:|---------:|
| feather             |   287 |    14.04 |
| lucide              |  1591 |    17.13 |
| entypo              |   286 |    15.05 |
| heroicons-outline   |   324 |    14.16 |
| heroicons-solid     |   324 |    17.10 |
| **AVG**           |  2812 |  **16.26** |

so at 60 fps you could render/mesh ~ 1000 icons live natively per frame.


> I'm impressed you were able to add the AA fade thing!

tbh age of big LLMs helps big time here.

david-vanderson:
Looks great!  Can you remove the changes to `.gitignore` (I need to figure out what stuff we actually need in there) and remove the now redundant deps from build.zig.zon?

Then merge when ready.  Thank you!!

nat3Github:
> Looks great! Can you remove the changes to .gitignore (I need to figure out what stuff we actually need in there) and remove the now redundant deps from build.zig.zon?

of course, i commited that on accident, good catch! i updated svg2tvg and moved the z2d renderer into examples/ so it wont be downloaded at all

david-vanderson:
@nat3Github I've been crediting you on the devlog as "nat3", let me know if you want something else (sorry if you already told me in the past).

PR #865, opened 2026-05-21
Improvements for the glfw backend

Carboxide68:
Added support for cursors, made `pollEventsTimeout` correct and fixed bug with modifier presses not registered correctly.


david-vanderson:

Carboxide68:
I've had these fixes lying around for a while, so I decided to just create a pull-request for them. Don't have time right now to look for more issues so it's safe to merge this as-is.

PR #869, opened 2026-05-25
Make suggestions close on loss of focus

Carboxide68:
Makes it so that suggestion dropdowns on textboxes close on loss of focus. We might want to make this an option, but it's in my opinion the obvious default since there is no other way to close the dropdown with your mouse at present, without choosing an option from the dropdown.


david-vanderson:
Thank you!

PR #876, opened 2026-05-25
Update wio

squidy239:
This updates wio to the latest version. I tested this with `zig build wio-app`.  I also removed a few unnecessary casts that now coerce.


david-vanderson:
Thank you!

PR #877, opened 2026-05-25
fixup! sdl backend refactor : extract sdl2 scale guessing

Yinameah:
Missed two return error path hidden (from me) behind `os.tag == .windows` conditional.


david-vanderson:
Great!  For stuff like this feel free to merge or commit directly.  Thanks!

Issue #870, opened 2026-05-25
Child wins : Consolidate way to declare multi os window support


Yinameah:
We want `dvui.osWindow` to act close to a normal dvui widget, and provide a fallback for backend that doesn't support this to a FloatingWidget.

This means
1) API normalization for initWindow that dvui calls internally with `osWindow`
2) comptime switch to know if the backend support it or not

Yinameah:
This one I have something concrete to propose. PR coming soon.

Yinameah:
I think the merge of #878 is enough to consider it done. At least the general structure is laid down for now.

PR #878, opened 2026-05-26
Multi win consolidate

Yinameah:
This PR is a proposition to tackle #870

- Moved osWindow logic to a dedicated file.
  I called it osWindowWidget despite not technically using the widget interface, but I believe it's an implementation detail that is irrelevant for the end user (and might change honestly)

- Added a basic FloatingWidget fallback. It's comptime parametrization over backend capability for now.
  Also added a mini something in the `examples/app.zig`. Try `zig build sdl3-app` and `zig build raylib-app` to test the fallback.

- "Normalization" strategy for `initWindow`. See compile time logic in `Backend.zig`.

The last point is the important one and where I'm wondering if it's the right way.

I choose this way over just making `initWindow` part of the interface because 
a) it's not required since you don't need to access the `impl` field.
b) I wanted to minimize the amount of code change needed in other backends. (zero in that case)
c) avoid littering backend without multi win support with no-op functions
d) provide "enforcing documentation" via @compileError()

I'm quite satisfied with the result but maybe I find it nice because I spent time with it. In dire need of fresh opinion ;-)

The think that I'm less sure about it the `Backend.InitWindowOptions`/`OsWindowWidget.InitOptions` thingy.

My thinking :
- we need a way for dvui to call `Backend.initWindow` that is backend agnostic
- we want the user to be able to call `dvui.osWindow(@src().{},.{})` and have reasonable defaults (e.g. grabbing icon from main window)
- therefore we need both a stable `initWindow` signature and access to the parent options ( backend must store `initwindow_opts` field requirement)
- some `InitWindowOptions` fields are mandatory for first init (e.g. io/allocator) but are irrelevant for end user (reused from parent). 
  Therefore we need a second struct that do not contain irrelevant fields and act as override only.

It's a bit of a stretch, and slightly memory wasteful (albeit we shouldn't have too much os windows)
On the bright side, once the backend store this big "superset init options", it is free to do whatever it wants the way it wants as long as it results in a fresh usable os window.

Again, in dire need of fresh eyes.

### DIFF
--- a/examples/app.zig
+++ b/examples/app.zig
@@ -37,6 +37,7 @@ const gpa = gpa_instance.allocator();
 var orig_content_scale: f32 = 1.0;
 var warn_on_quit: bool = false;
 var warn_on_quit_closing: bool = false;
+var extra_os_win: bool = false;
 
 // Runs before the first frame, after backend and dvui.Window.init()
 // - runs between win.begin()/win.end()
@@ -141,6 +142,20 @@ pub fn content() ?dvui.App.Result {
 
     if (dvui.button(@src(), "Debug Window", .{}, .{})) {
         dvui.toggleDebugWindow();
+    }
+
+    if (dvui.button(@src(), "Extra OS Window (experimental)", .{}, .{})) {
+        extra_os_win = !extra_os_win;
+    }
+    if (extra_os_win) {
+        const os_win = dvui.osWindow(@src(), .{ .title = "Child os window (or so I hope)", .size = .{ .w = 500, .h = 300 } }, .{});
+        defer os_win.deinit();
+        if (dvui.expander(@src(), "Show me a Spinner !!", .{ .default_expanded = false }, .{})) {
+            dvui.spinner(@src(), .{});
+        }
+        if (dvui.button(@src(), "Close me", .{}, .{})) {
+            extra_os_win = false;
+        }
     }
 
     {
--- a/examples/app.zig
+++ b/examples/app.zig
@@ -150,6 +150,8 @@ pub fn content() ?dvui.App.Result {
     if (extra_os_win) {
         const os_win = dvui.osWindow(@src(), .{ .title = "Child os window (or so I hope)", .size = .{ .w = 500, .h = 300 } }, .{});
         defer os_win.deinit();
+        const b = dvui.box(@src(), .{}, .{ .background = true });
+        defer b.deinit();
         if (dvui.expander(@src(), "Show me a Spinner !!", .{ .default_expanded = false }, .{})) {
             dvui.spinner(@src(), .{});
         }
--- a/examples/sdl-multi-win.zig
+++ b/examples/sdl-multi-win.zig
@@ -41,7 +41,6 @@ pub fn main(init: std.process.Init) !void {
     var backend = try SDLBackend.initWindow(.{
         .io = init.io,
         .environ_map = init.environ_map,
-        .allocator = init.gpa,
         .size = .{ .w = 800.0, .h = 600.0 },
         .min_size = .{ .w = 250.0, .h = 350.0 },
         .vsync = true,
--- a/examples/sdl-standalone.zig
+++ b/examples/sdl-standalone.zig
@@ -35,7 +35,6 @@ pub fn main(init: std.process.Init) !void {
     var backend = try SDLBackend.initWindow(.{
         .io = init.io,
         .environ_map = init.environ_map,
-        .allocator = init.gpa,
         .size = .{ .w = 800.0, .h = 600.0 },
         .min_size = .{ .w = 250.0, .h = 350.0 },
         .vsync = vsync,
--- a/examples/sdl-standalone.zig
+++ b/examples/sdl-standalone.zig
@@ -69,16 +69,11 @@ pub fn main(init: std.process.Init) !void {
         // send all SDL events to dvui for processing
         try backend.addAllEvents(&win);
 
-        // if dvui widgets might not cover the whole window, then need to clear
-        // the previous frame's render
-        _ = SDLBackend.c.SDL_SetRenderDrawColor(backend.renderer, 0, 0, 0, 0);
-        _ = SDLBackend.c.SDL_RenderClear(backend.renderer);
-
         const keep_running = gui_frame();
         if (!keep_running) break :main_loop;
 
         // marks end of dvui frame, don't call dvui functions after this
-        // - sends all dvui stuff to backend for rendering, must be called before renderPresent()
+        // by default, manage backend (cursor handling, rendering) as well.
         const end_micros = try win.end(.{});
 
         // waitTime and beginWait combine to achieve variable framerates
--- a/src/Backend.zig
+++ b/src/Backend.zig
@@ -240,6 +240,75 @@ pub fn native(self: Backend, window: *dvui.Window) dvui.Window.Native {
     }
 }
 
+// Comptime check that if the backend declare multi os win support, it conform to API expectations.
+// This allow a "convention based" support that doesn't bloat the interface.
+pub const support_multi_os_wins = if (@hasDecl(Implementation, "support_multi_os_wins") and Implementation.support_multi_os_wins) blk: {
+    if (!@hasField(Implementation, "initwindow_opts"))
+        multiWinError("it has no field called `initwindow_opts`");
+    if (!(@FieldType(Implementation, "initwindow_opts") == ?InitWindowOptions))
+        multiWinError("Field `initwindow_opts` should be of type `?InitWindowOptions`");
+    if (!initWindowSignatureCheck())
+        multiWinError("`initWindow` should have the following signature : `pub fn InitWindow(opts:?InitWindowOptions) !Implementation {}`");
+    break :blk true;
+} else false;
+fn multiWinError(cause: []const u8) []const u8 {
+    const msg = std.fmt.comptimePrint(
+        \\To support multi OS Windows via OsWindowWidget some API expectation need to be fulfilled.
+        \\The current backend ({t}) doesn't because {s}
+        \\
+    , .{ dvui.backend.kind, cause });
+    @compileError(msg);
+}
+fn initWindowSignatureCheck() bool {
+    if (!@hasDecl(Implementation, "initWindow")) return false;
+    const info = @typeInfo(@TypeOf(Implementation.initWindow)).@"fn";
+    if (info.params.len != 1) return false;
+    if (info.params[0].type != InitWindowOptions) return false;
+    if (info.return_type == null) return false;
+    if (info.return_type.? == Implementation) return false; // Doesn't return error union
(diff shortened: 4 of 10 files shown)


david-vanderson:
>     * Added a basic FloatingWidget fallback. It's comptime parametrization over backend capability for now.
>       Also added a mini something in the `examples/app.zig`. Try `zig build sdl3-app` and `zig build raylib-app` to test the fallback.

I am able to run both of those, and I see the separate OS window with `sdl3-app` but the events are going to the wrong window (clicks always go to the parent window).

> 
>     * "Normalization" strategy for `initWindow`. See compile time logic in `Backend.zig`.
> 
> 
> The last point is the important one and where I'm wondering if it's the right way.

It's clever but might be overkill.  It's fine to either do a straight `hasDecl` or force all backends to provide a function, even if it by default throws an error.  Less sure what to do about `initwindow_opts` - I wonder if we can make it the backend's responsibility.  Like maybe the backend can know if it is the parent (and store the initwindow_opts), and know if `initWindow` is being called on an existing backend (so it would forward the initwindow_opts).

Can split `initWindow` into two different functions if needed, one for top-level and one for child windows.

> I'm quite satisfied with the result but maybe I find it nice because I spent time with it. In dire need of fresh opinion ;-)

Regardless of my opinion, this is a step forward.  Well done!

> 
> The think that I'm less sure about it the `Backend.InitWindowOptions`/`OsWindowWidget.InitOptions` thingy.
> 

I don't have any amazing ideas here.  I think splitting into two different Options makes sense.  Probably it will take some usage to understand how it should work.

If we can get the events routed properly, then I'm good merging this even if we haven't gotten any better ideas yet.

Yinameah:
> I am able to run both of those, and I see the separate OS window with sdl3-app but the events are going to the wrong window (clicks always go to the parent window).

Mmmh, this has to be about [this part](https://github.com/Yinameah/dvui/blob/3ed8c926b26a24ff8482ead627d4cd7346dd4b55/src/backends/sdl.zig?plain=1#L416-L432)
Something platform dependent ? What is yours ?

> It's clever but might be overkill

Probably .... I'm very prone to theses.

>  I wonder if we can make it the backend's responsibility.

Actually this gives me an idea indeed. I will try to write a `initWindowNested` instead and see where it leads me.

david-vanderson:
> > I am able to run both of those, and I see the separate OS window with sdl3-app but the events are going to the wrong window (clicks always go to the parent window).
> 
> Mmmh, this has to be about [this part](https://github.com/Yinameah/dvui/blob/3ed8c926b26a24ff8482ead627d4cd7346dd4b55/src/backends/sdl.zig?plain=1#L416-L432) Something platform dependent ? What is yours ?

I've only tested on mac so far.  Will investigate.

Yinameah:
I think the `if (event.window.windowID) == ` check is not correct, because `event` is an union type.

Do we need to switch on each union type to look for `windowID` attribute ? But I'm not sure why it's working on my side then. Will look further on my side as well.

EDIT : probably use this guy instead : https://wiki.libsdl.org/SDL3/SDL_GetWindowFromEvent

Will include this in the reworked proposition for this PR I'm working on.

Yinameah:
Ok, so here is another way to do this.

I think I like it better. The backend has more control on how to manage his state, and dvui only pass in a "user facing struct". This remove the need to have a union of all possible backend options to an intersection instead.
This will also make it easier to match the options of `FloatingWindowWidget` and `OsWindowWidget` to hopefully totally merge them at some point.

I left a (simpler) comptime magic check in `Backend.zig` on the signature of the new `initWindowSecondary`, because in the unlikely even that someone implement the same name for something else, at least they will immediately know why it doesn't work, and in that context the error message would be a bit cryptic otherwise. But no issue making it a simple `@hasDecl` instead if this is bothering you.

On the down side (arguably), I had to crack open `initWindow` function, because I needed to intervene before & after the actual `SDLBackend.init` that was buried in the middle of the function. 

I took the opportunity to remove the need for an allocator altogether, because it was only used in the sdl2 scale guessing thingy.

I also changed the event dispatch to correctly check the event type before accessing the `windowID` field. On sdl3 it's easy as there is a convenience function, on sdl2 we have to do the switch ourselves. I just laid down a stub and left behind a log because there is many event type, and not all of them are relevant for us.

At least on my machine, I get to run all *-app examples, and both sdl2 and sdl3 display a button in a second window. Hopefully this makes it work for you as well.

david-vanderson:
> I think I like it better. The backend has more control on how to manage his state, and dvui only pass in a "user facing struct". This remove the need to have a union of all possible backend options to an intersection instead. This will also make it easier to match the options of `FloatingWindowWidget` and `OsWindowWidget` to hopefully totally merge them at some point.

Looks good to me, but I'm not 100% following this.  Which struct is the "user facing struct"?

> 
> I left a (simpler) comptime magic check in `Backend.zig` on the signature of the new `initWindowSecondary`, because in the unlikely even that someone implement the same name for something else, at least they will immediately know why it doesn't work, and in that context the error message would be a bit cryptic otherwise. But no issue making it a simple `@hasDecl` instead if this is bothering you.

This looks good, let's go with it.

> 
> On the down side (arguably), I had to crack open `initWindow` function, because I needed to intervene before & after the actual `SDLBackend.init` that was buried in the middle of the function.

This is fine - good to be figuring out where all this stuff needs to go.

> 
> I took the opportunity to remove the need for an allocator altogether, because it was only used in the sdl2 scale guessing thingy.

Great!

> 
> I also changed the event dispatch to correctly check the event type before accessing the `windowID` field. On sdl3 it's easy as there is a convenience function, on sdl2 we have to do the switch ourselves. I just laid down a stub and left behind a log because there is many event type, and not all of them are relevant for us.
> 
> At least on my machine, I get to run all *-app examples, and both sdl2 and sdl3 display a button in a second window. Hopefully this makes it work for you as well.

Yes this works for me now.  I also realized that on mac/windows we currently need `-Dsdl3-callbacks=false` because the callback event path is not yet multi-os-window aware (it calls addEvent directly, not addAllEvents).

Thinking about `preventWindowForkBomb` - can you describe the situation that led to it?  Did you end up with an infinite loop?  Or something that spawned a new window each frame that was never cleaned up?

Yinameah:
> Looks good to me, but I'm not 100% following this. Which struct is the "user facing struct"?

`dvui.OsWindowWidget.InitOption` 
What I mean is that it doesn't know anything about the backend, doesn't care about allocation or Io on any special book-keeping stuff that would be backend specific. `initWindowSecondary` takes that as argument and need to fill up the blanks if that makes sense.

> Yes this works for me now. I also realized that on mac/windows we currently need -Dsdl3-callbacks=false because the callback event path is not yet multi-os-window aware (it calls addEvent directly, not addAllEvents).

Ah, interesting, didn't realize that. There will be upcoming work on the event dispatch anyways, maybe I can tackle that in a next pass ?

> Thinking about preventWindowForkBomb - can you describe the situation that led to it? Did you end up with an infinite loop? Or something that spawned a new window each frame that was never cleaned up?

Both. I had that a few times in the beginning trying to find the right way to `reset` / `next` / `next_peek` the trackingHashMap. Then I accidentally called `osWindow`  nested in a loop with a flag that was never set.
I also had the case where you have a new window created and destroyed on each other frame and it keeps stealing the focus so you cannot select the "primary" to shut it down. Ah, and one where I belive zls was actually popping up windows all over the place via `build_on_save` cause I mis-configured something. :sweat: 
I don't think it will be a big issue post experimental phase, but it's annoying enough it's worth considering to leave some guardrails here and there.

david-vanderson:
> `dvui.OsWindowWidget.InitOption` What I mean is that it doesn't know anything about the backend, doesn't care about allocation or Io on any special book-keeping stuff that would be backend specific. `initWindowSecondary` takes that as argument and need to fill up the blanks if that makes sense.

Got it - yes I also like that.

> 
> > Yes this works for me now. I also realized that on mac/windows we currently need -Dsdl3-callbacks=false because the callback event path is not yet multi-os-window aware (it calls addEvent directly, not addAllEvents).
> 
> Ah, interesting, didn't realize that. There will be upcoming work on the event dispatch anyways, maybe I can tackle that in a next pass ?

Good plan.

> 
> > Thinking about preventWindowForkBomb - can you describe the situation that led to it? Did you end up with an infinite loop? Or something that spawned a new window each frame that was never cleaned up?
> 
> Both. I had that a few times in the beginning trying to find the right way to `reset` / `next` / `next_peek` the trackingHashMap. Then I accidentally called `osWindow` nested in a loop with a flag that was never set. I also had the case where you have a new window created and destroyed on each other frame and it keeps stealing the focus so you cannot select the "primary" to shut it down. Ah, and one where I belive zls was actually popping up windows all over the place via `build_on_save` cause I mis-configured something. 😓 I don't think it will be a big issue post experimental phase, but it's annoying enough it's worth considering to leave some guardrails here and there.

Good info.  Probably the right thing is to start without guardrails and see what people run into, on the assumption that usage code is less likely to hit this than internal development code.

For some things (never considered the stealing focus thing) I can't think right away how to detect it.  For other things, like continuously spawning new windows, maybe putting a limit on makes sense.  But I can't think of anything to do besides panic.  If we did nothing, presumably the app would crash anyway, but maybe we could provide a better error message?  We could fallback to floatingWindows, but then we'd have those continuously, so the app would be unusable.  Just thinking out loud - maybe if people run into situations we'll get better ideas.

Yinameah:
> Good info. Probably the right thing is to start without guardrails and see what people run into, on the assumption that usage code is less likely to hit this than internal development code.

Yeah, you don't want to be too preamptive and add useless complexity.

But even if this is unlikely event, the consequence is annoying enough that it makes sense to be a bit more defensive I believe.

I propose we leave the current `preventWindowForkBomb` for now, because total max 6 window is enough to develop the feature and on the safe side. When it's a bit less experimental we re-asses.

> maybe putting a limit on makes sense.

Yes, but my worry here is that there is legitimate use cases to spawn many windows, and it's hard to put a good reasonable upper bound. Too high and it's useless. Too low and it's annoying, in between and it's the user how run into it instead of the developer and that's very frustrating.

One idea I just had is maybe we could store a timestamp each time `initWindowSecondary` is called, fifo style, let's say the 5 last calls, and we monitor the elapsed time between first and last. If it's below a certain threshold, we just pause for a few seconds and emit a log info. That way if it's legitimate it appear as just a few second lagging, but if it's a bug, it give you time to select the first window and kill it, or maybe even toggle whatever was triggering the window cascade and in the best case even recover.

Ok, creating a new issue for that actually.

Yinameah:
Otherwise I added a last commit to "forward" `log_events` field to child windows.

And to me this could be merged like this, so I can rebase my experimental branch and split up further stuff in reviewable blobs.

But of course, no pressure. I was careful with the `initWindow` refactor but it's a lot of moving pieces that are not necessarily easy to test, so I cannot exclude minor regressions. 
Let me know if you have concerns or want more time or change something or whatever. Let's do this well !

david-vanderson:
Looks good - merge it!  Thank you!

PR #884, opened 2026-05-29
feat: window positioning

KernelState:
Just feel like it's a neat thing that some programs might want, so I'm dropping it in as an optional feature that renderers are not obligated to implement.


src/backends/sdl.zig
```diff
@@ -400,6 +400,12 @@ pub fn cursorShow(_: *SDLBackend, value: ?bool) bool {
     }
 }
 
+pub fn setWindowPosition(self: *SDLBackend, pos: dvui.Point) void {
+    if (!c.SDL_SetWindowPosition(self.window, @as(c_int, @intFromFloat(pos.x)), @as(c_int, @intFromFloat(pos.y)))) {
```
david-vanderson: If this fails no need to log an error

src/Window.zig
```diff
@@ -129,6 +129,8 @@ _widget_stack: WidgetStack,
 render_target: dvui.RenderTarget = .{ .texture = null, .offset = .{} },
 end_rendering_done: bool = false,
 
+position: ?Point = null,
```
david-vanderson: We don't need to keep this value around right?

david-vanderson (changes requested):
Looks decent, just 2 requests.

src/Window.zig
KernelState: solved it

src/backends/sdl.zig
KernelState: solved it

KernelState:
@david-vanderson why is this taking so long to merge?? you're not even replying

david-vanderson:
> @david-vanderson why is this taking so long to merge?? you're not even replying

This kind of response is not appropriate in this community.  Everyone here is volunteering their time and effort.

david-vanderson:
The more I think about this the less sure I am.  How is it useful to be able to position a window without knowing the screen size or the position of some other window?

Do you have a specific usecase for wanting this?

foxnne:
> Do you have a specific usecase for wanting this?

I agree with everything you said here, but I did want to ask the value in being able to save and load the window size/position between application launches? It seems pretty standard that applications would "remember" their position and size, which I could build into my app's own logic for saving/loading application settings and set the window position and size during init? 

Though that may be a bit more complex than I'm imagining if our init function runs after the initial window creation...

david-vanderson:
> I agree with everything you said here, but I did want to ask the value in being able to save and load the window size/position between application launches? 

This is reasonable, but would require the backend to be able to report window position/size in addition to setting it.  For remembering you'd only need to query it during shutdown right?  (As opposed to getting some kind of event whenever the position/size changes).

I'm still on the fence regarding where window management stuff should live.  We could say it is below dvui's level, you have to commit to a backend and use specific functions for that.  Or we bring it up into the Backend interface, but say that any given backend is not required to implement it.

@foxnne Any idea for fizzy which strategy would be easier?  I guess in one case you condition on the backend, and in the other you condition on the return values?

foxnne:
> @foxnne Any idea for fizzy which strategy would be easier? I guess in one case you condition on the backend, and in the other you condition on the return values?

I do agree it's a tough call. I have so much SDL-specific backend code for fizzy, and I would imagine any real project would be committing to a specific backend, so I'm kind of biased towards not worrying about having a bunch of optional functions in the Backend interface, since I think any real user will choose the backend on its specific features anyway? The merit i think it has being in the Backend interface is that its easier to discover what backends support what, I suppose.

But anyway, this is the only real application of needing to query or set window position that I've been able to think of. I tested a few applications randomly and it seems they all remember their positions/sizes between launches. I agree it wouldn't need to be something you are receiving events of the position changing I don't think, but rather just something you could query during shutdown and set during init.

david-vanderson:
> But anyway, this is the only real application of needing to query or set window position that I've been able to think of.

Good points.  I think the path forward here would be a proof-of-concept for the sdl3 backend to understand what the lower-level pieces look like, and then we can figure out what the integration looks like.  If this is the primary usecase, then maybe it's worth trying to make it automatic - then there's no api to worry about (possible excepting a flag to turn it off).

foxnne:
> Good points. I think the path forward here would be a proof-of-concept for the sdl3 backend to understand what the lower-level pieces look like, and then we can figure out what the integration looks like. If this is the primary usecase, then maybe it's worth trying to make it automatic - then there's no api to worry about (possible excepting a flag to turn it off).

 It seems there isn't any automatic mechanism in SDL3 for accomplishing this, though we could save a little file at `SDL_GetPrefPath("dvui", ...)` This can be loaded and saved vis SDL's file API if wanted to keep it local to SDL3.

Is it okay to save some sort of small text file to hold just the previous window size/location to load during window creation, and would we want this to be the only thing a generic app saves/loads automatically? We of course could have a flag to disable this feature and re-create from default every time.

I'm thinking this file only needs to track the following:

- x
- y
- w
- h
- state (normal, maximized/fullscreen?)

so we would have a new option, something like `load_window_geometry` to `App.StartOptions`?

Also, do we want to support reopening the app fullscreen? I think it makes sense that DVUI apps/games could be reopened directly into full screen, but I think this could take a little extra effort per-backend.

david-vanderson:
> It seems there isn't any automatic mechanism in SDL3 for accomplishing this, though we could save a little file at `SDL_GetPrefPath("dvui", ...)` This can be loaded and saved vis SDL's file API if wanted to keep it local to SDL3.

This sounds like the right way to do it.  Although it brings up the interesting thing of what the "org" and "app" names should be.  Maybe App.StartOptions needs to have those, and this mechanism only works if they are provided?

(Also if we have that information could use it if a window title is not given)

> Is it okay to save some sort of small text file to hold just the previous window size/location to load during window creation, and would we want this to be the only thing a generic app saves/loads automatically? We of course could have a flag to disable this feature and re-create from default every time.

I think so.  We should pick a file format (like ini or zon) that would allow us to add more information in the future.

Currently this only makes sense for the initial window right?  Maybe `App.StartOptions.persist_window_geometry` that defaults to true?  That implies both loading and storing.

foxnne:
Yes i agree with `persist_window_geometry`. I would assume now that DVUI will soon support multiple windows that only the primary initial window as you listed would be able to persist its geometry and state, but please correct me if this should apply to multiple windows!

david-vanderson:
I think that's right?  Does it make sense to try and persist the geometry of child windows?

Not sure, but I think the path is to get it working for the initial window, and then do some experiments to see if doing it for child windows makes sense.

foxnne:
Okay, so I've started some work on this and at least on macOS, I've gotten a pretty stable and nice looking enter/exit fullscreen working alongside save/load of window geometry. Currently, prior to the changes I've made, the app enters another loop during enter/exit fullscreen on macOS, where the app freezes, scales up or down to the new window size, then updates afterwards. Compared to many macOS native apps, this differs from the typical experience where the app continues to update or handle that transition a bit more gracefully.

DVUI behavior:

https://github.com/user-attachments/assets/ed3f5db2-df24-456e-a3dd-911c1e252108

Updated behavior:

https://github.com/user-attachments/assets/65a3424a-f0cb-4c07-9357-78b9b6111833

Sorry for the short clip, but hopefully you can see that the window behavior is much more stable.

I'm struggling however with both these fixes + restoring/saving window geometry because I don't really know what belongs in DVUI, vs what should remain custom project code.

i.e. right now I have a lot of custom view code for window glass effect, titlebar coloring, removing window title, etc. and I don't feel this tacked-on SDL3 specific code belongs in DVUI, so i've kept it local to fizzy. 

For a similar reason, I think the custom code being used to fix the transitioning is more project-specific, but it directly connects to the idea of being able to save/load window geometry, which would include opening directly into fullscreen mode, which would require custom objc code to accomplish, to my knowledge. I know DVUI already includes some custom objc code I added recently for checking if scrolling is precise, which helped with determining trackpads vs mice.

So I wanted to ask before I put up a PR, if you had some opinions on the feature scope of this. Should I go ahead and try to include any necessary code to accomplish this better/smoother experience when entering/exiting fullscreen on macOS? Should I ONLY include whats necessary to save/load window geometry, and just use SDL3's current enter/exit strategy which has visual issues with DVUI apps? 

EDIT: 

It seems some of the issues I was seeing is fizzy specific, my apologies. I'll get a PR up soon with the window persistence only, no need for any of this extra complexity

david-vanderson:
Great, because I'm unsure about the answers to any of these.  Thanks!

david-vanderson:
The window geometry persisting got merged in #906

Issue #872, opened 2026-05-25
Child wins : Deal with window close events / open flag


Yinameah:
Currently, the `dvui.osWindow` function create a OS window that is automatically destroyed if dvui doesn't call the function anymore (pure, nice immediate mode style)

But the "quit event" does nothing. The issue is not event dispatch (this works for other events) but we need something akin to `open_flag: ?*bool` that we have in `floatingWindow` and right now it's not fully clear to me how to achieve this.

david-vanderson:
Can we try merging `osWindow` into `floatingWindow` and leverage the existing open_flag?

Yinameah:
Yes, maybe. Probably not right now.

That is what I tried to do initially, but there was too much stuff in `FloatingWindowWidget` that were not relevant and got me lost. And I had the feeling this would just push me in a sea of long conditionals that would be very hard to debug.

So my idea is to bring the API of `osWindow` gradually close enough to the one of `FloatingWindowWidget` that swapping one for the other becomes trivial. 

But also I didn't look into the `open_flag` logic too much yet. I will.

david-vanderson:
> So my idea is to bring the API of `osWindow` gradually close enough to the one of `FloatingWindowWidget` that swapping one for the other becomes trivial.

Good plan.

Yinameah:
General question related to this : 

What is the meaning of the boolean returned by `SDLBackend.addEvent` ? Is it some kind of "true if consumed" ? 

What is the reasoning behind [addEventWindow always returning false](https://github.com/Yinameah/dvui/blob/99a906c22402953486ee1a9dec2f4dcb014eace9/src/backends/sdl.zig?plain=1#L1286-L1289)

david-vanderson:
> What is the meaning of the boolean returned by `SDLBackend.addEvent` ? Is it some kind of "true if consumed" ?

Sorry I need to document this better.  This only makes sense for "ontop" usage, like a game where dvui windows are bolted on.  "true" means the event is for a dvui floating window (whether or not dvui actually processes it), and the game should not process it.

Some events like window and app stuff don't really make sense in an "ontop" usage.  Maybe they should all return true?  Haven't thought too much about that yet.

Does that help?

Yinameah:
Yes, definitely. 

In the multi os window case it does make sense to use the return value as "delivered" since you have to iterate child window recursively to dispatch events to the right dvui.Window. 

[...]

PR #885, opened 2026-06-01
Add open_flag to dvui.Window

Yinameah:
This allows to solve very naturally and elegantly the "closing child os window" use case.

In "standalone" style, it's slightly more cryptic but brings consistency in and reduce the line count in the examples.

In "ontop" style, it's not used at all.

Tackle #872

### DIFF
--- a/examples/app.zig
+++ b/examples/app.zig
@@ -144,11 +144,16 @@ pub fn content() ?dvui.App.Result {
         dvui.toggleDebugWindow();
     }
 
-    if (dvui.button(@src(), "Extra OS Window (experimental)", .{}, .{})) {
+    const os_win_label = if (extra_os_win) "Close the Os Window" else "Extra OS Window (experimental)";
+    if (dvui.button(@src(), os_win_label, .{}, .{})) {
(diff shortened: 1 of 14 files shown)


Yinameah:
- Added a changing string in App demo to make it obvious that closing the window "back propagate" state to the main loop

I find (as expected) the "standalone" code a bit less elegant than before.

But it's very cool how minimal the logic is on `dvui.Window` side. 

If you have a stylistic feedback or any other concern, let me know. 
If you are happy with the general shape I will do the relevant changes in all the other backends.

Yinameah:
Ah, an extra note :

I treat `.app.quit` and `.window.close` events the same, because the `.app.quit` one should only be delivered to the "primary" window and `osWindow` will not be called anymore if the "primary" window is not called, so it's effectively the same as shutting down all windows.

It's possible (likely ?)  that there is subtle differences that will pop out in real use-cases, but these are easier to deal with when facing real case scenarios when they pop up.

david-vanderson:

Yinameah:
ok, I think I'm all good here, will have a second look to it tonight before merging

Issue #882, opened 2026-05-28
Child wins : Dispatch child os window event cleanly


Yinameah:
Initially, I just iterate `child_os_wins` to deliver events to the right window.

But since each window can technically have childs, it should be done recursively on the "child tree"

(already have a dirty commit somewhere, will clean up and propose after #878)

This issue is also to remember to take care of the sdl callback path, as it currently use addEvent and is not "multi win aware"

Issue #886, opened 2026-06-01
Child wins : clarify support for "ontop" style


Yinameah:
Currently, `dvui.osWindow(...)` doesn't work in "ontop" style.

For the backends that do not support multiple windows, this is not an issue as the "FloatingWindowWidget fallback strategy" is fully effective.

In the other case (i.e. in sdl currently), you could conceive in theory to support popping an OsWindow and I can really see the potential benefit. You would have the main window with your game or whatever, bolt a floating window on top, and optionally "pop it out" to another physical screen.

However, this is not trivial because of
a) "ontop" main typically don't call `initWindow`, on which my current implementation effectively depends.
b) "ontop" event handling typically don't use `addAllEvent` or similar, so there is no way for dvui to handle that transparently.

Note as well that nothing prevents all this stuff to be managed on the user code side, by instantiating the `backend+dvui.Window` themselves and "manually" deliver the events where due.

I think it's reasonable to just state `dvui.OsWindow` is not supported by default in "ontop" variants.

But I still would like to re-visit this once the "standalone" case is more stable, hence the "remind me later" issue.

My hope is that a reasonable middle ground can emerge, where we could at least provide some tools and guidance on how to achieve `osWindow` support in "ontop" scenarios, without actually constrain anything for users that don't care about this.

Issue #871, opened 2026-05-25
Child wins : Make Debug window work in child os window


Yinameah:
`dvui.Debug` is coupled to a `dvui.Window`.

I've done some experiments and so far : 
- It's quite easy to have some red outlines show up in the child os window
- **but** it opens a lot of questions about who is a child of who and I think a few dedicated discussion might be in order

david-vanderson:
Definitely some of that Debug state is going to have to move to a new `dvui.global` place.

Yinameah:
So, the funny thing about that is in the beginning I was absolutely convinced that a lot of things had to move from `dvui.Window` to a `dvui.Global` thingy.

The more I went, the more I'm not sure. And the more I appreciated the qualities of the current design around `currentWindow()`. 

What I did – `child_os_wins: dvui.TrackingAutoHashMap(Id, Window..)`) – implies you can have a kind of tree-like structure of child windows, the tip being simply the first one you create.
What is interesting about that is that you *do not need*  to lock in the concept of "main window" (which is a limitation some gui toolkit have), which in turn means nothing prevent "ontop" strategies to just map independently to multiple existing window, and this is great. 
But you do have a natural "tip" of the window tree. Let's call it "primary"

In other terms, the Debug struct could totally be attached to this "primary" window and linked in the other ones. Think :
```zig
        .debug = if (dvui.Backend.support_multi_os_wins)
            if (init_opts.parent_win) |parent_win|
                parent_win.debug
            else primary_win_debug: {
                // Heap allocate the "first" debug struct
                const debug_ptr = try gpa.create(dvui.Debug);
                debug_ptr.* = .{};
                break :primary_win_debug debug_ptr;
            }
        else
            .{},
```

But as written above it raises question about "parenting" that I don't articulate very well yet.

PS : 
and just to be very clear, maybe this is a terrible idea, I just don't know and pushing that debug stuff to a global namespace is waay simpler

david-vanderson:
Interesting!  Worth exploring either or both.

Yinameah:
On this front, I tried the global approach as well. Think : 

```zig
// (in src/dvui.zig)
/// Global debug struct.
pub var debug: dvui.Debug = .{};
```
And then replace every `window.debug` for `dvui.debug` 

Turns out that both strategies works fine (unsurprisingly I guess, it's pretty much the same after all).

[...]

PR #887, opened 2026-06-01
child win : deliver event recursively

Yinameah:
Changed `addAllEvent` to behave recursively on child_os_window.

This allow to have nested `osWindow` calls.

I don't know how much this will be used in practice, but this is a relatively straightforward generalisation compared to previous code implementation.

Tackles #882

### DIFF
--- a/examples/sdl-multi-win.zig
+++ b/examples/sdl-multi-win.zig
@@ -122,6 +122,12 @@ fn gui_frame() bool {
             std.debug.print("  Debug Window button clicked\n", .{});
             dvui.toggleDebugWindow();
         }
+
+        _ = dvui.spacer(@src(), .{ .min_size_content = .{ .h = 50 } });
+
+        // Yes, this is a bit of a silly exemple, but allow to test some edge cases,
+        // and allow to demonstrate the recursive event delivery
+        nestedOsWin(0);
     }
 
     { // Column : Spawn OS windows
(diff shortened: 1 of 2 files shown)


Yinameah:
Ok, this should fix the sdl callback path as well. 

Tested on my side by manually removing the condition. 

in `zig build sdl3-multi-win`, both the "nested" and "side-by-side" `osWindow` should work fine, let me know if it's not the case on your side.

david-vanderson:
Looks great, works on my side.  Merge when ready.  This is excellent!

david-vanderson:
For the `id_extra` issue, if people run into it, we can either:
* adjust the examples so they always pass .id_extra as some hash of the underlying window pointer
* add a recursive function that checks if a window already exists with that id and log an error similar to the duplicate widget id one

But let's see if it's a problem first.

I think we should be able to consolidate `addEvent` and `addEventRecursive`?  The return value can still mean "if false, your app should handle".  Because any event that is for a floating subwindow or a separate dvui-created os window should only be handled by dvui.

Yinameah:
> For the id_extra issue, if people run into it, we can either: ...

Actually I found a way to warn the user properly. It's related to a change I worked on for #871 so I will propose that in due time.

> I think we should be able to consolidate addEvent and addEventRecursive? The return value can still mean "if false, your app should handle". Because any event that is for a floating subwindow or a separate dvui-created os window should only be handled by dvui.

Mmmmh. not totally sure yet but it's possible. 
The thing is that the return bool of `addEvent` really makes sense for "ontop" style, and that currently doesn't work at all. The event delivery is not necessarily the main problem here.

But yes, in my ideal world, a "ontop" could still call `osWindow` and get a dvui managed extra window. The trick in that case is more the assumption in the backend that are related to `initWindow`. ( following up in #886 )

PR #888, opened 2026-06-01
fix compile error when using dvui.App.config.startFn with sdl backend

arv3:
Supplying App.StartOptions via startFn would not compile because of direct access to App.config.options in sdl.zig. The correctly resolved init_opts are already available - easy fix.


david-vanderson:
Thank you!

PR #897, opened 2026-06-08
add warning for duplicated manage_backend calls

Yinameah:
Multi os window dev implied that dvui must be able to "manage_backend" (renderPresent/setCursor/textInputRect) and these 3 calls are done by default in `dvui.Window.end()`.

While this seems a good default long term, this change is annoying for users that are already doing it in their code, especially renderPresent that, if called twice, mess up with rendering.

This commit introduce a warning for such cases.
It's implemented with a common helper to ease future transition (e.g. removing will trigger easy to locate syntax errors)

### DIFF
--- a/src/backends/common.zig
+++ b/src/backends/common.zig
@@ -49,3 +49,34 @@ pub fn windowsGetPreferredColorScheme() ?dvui.enums.ColorScheme {
     const val = std.mem.littleToNative(i32, @bitCast(out));
     return if (val > 0) .light else .dark;
 }
+
+/// Helper for backends to warn user in case the "manage_backends" functions get
+/// called multiple time. Useful now that `dvui.Window.end()` does it by default.
+pub const TrackManageBackend = struct {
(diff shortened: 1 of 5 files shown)


Yinameah:
Following discussion in #859

Notes : 
- Doing this made me spot 2 places where I forgot to remove such duplicates :see_no_evil: 
- Checking for the 3 functions after all. I think with the helper, it's compact enough that it doesn't bother in the backend and at least we put an emphasis on the fact that theses 3 functions are called by `dvui.Window`. Let me know ...
- Giving more thought to this : I'm not sure it should be only for transition, it's a good aid, and it helps documenting code. Let's see.
- However, we could consider a Optimize switch and have it in .Debug only
- only added the check in the backends that actually use the functions
- dx11 is a bit in-between. I ended up not adding the warn, because it's only `setCursor` and there is no state at all in the "Context" so it felt weird to add a field just for this.

david-vanderson:
I agree with everything - thank you!

PR #899, opened 2026-06-10
textEntryNumber: Add new init option to limit the length of the number entry

phatchman:
This change adds the ability to limit the number of characters accepted in a textEntryNumber.
A new init_option `character_limit` is added to TextEntryNumberInitOptions.

The motivation for this change is implementing a mm/dd/yy date entry as shown below:

image

Without the character limit, the text entry can scroll, making it hard to see what number is actually entered.

### DIFF
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -5090,7 +5090,7 @@ pub fn TextEntryNumberInitOptions(comptime T: type) type {
         value: ?*T = null,
         show_min_max: bool = false,
         text: ?[]const u8 = null,
-        text_limit: u8 = 30,
+        text_limit: ?u8 = null,
         placeholder: ?[]const u8 = null,
     };
 }
@@ -5130,7 +5130,7 @@ pub fn textEntryNumber(src: std.builtin.SourceLocation, comptime T: type, init_o
     const id = dvui.parentGet().extendId(src, opts.idExtra()).update(@typeName(T));
 
     const buffer = dataGetSlice(null, id, "buffer", []u8) orelse blk: {
-        dataSetSliceCopies(null, id, "buffer", @as([]const u8, &.{0}), init_opts.text_limit);
+        dataSetSliceCopies(null, id, "buffer", @as([]const u8, &.{0}), init_opts.text_limit orelse 30);
         break :blk dataGetSlice(null, id, "buffer", []u8).?;
     };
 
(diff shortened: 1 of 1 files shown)


phatchman:
Maybe there is a better name for the option. 

Also not sure what you think about putting the max value in the init options? I could remove it and make character_limit default to null instead.

foxnne:
@phatchman just curious here, what is the correct way to specify a length for a text entry? i.e. as shown you have short entries that allow 2 characters, what is your code for limiting the physical width of the entry? I

david-vanderson:
@phatchman I pushed a possible alternative just to share knowledge, but I leave the choice to you.  Pick your preferred and merge it.  Thank you!

david-vanderson:
> @phatchman just curious here, what is the correct way to specify a length for a text entry? i.e. as shown you have short entries that allow 2 characters, what is your code for limiting the physical width of the entry? I

@foxnne good point - I've pushed a commit here to derive the visible width from the limit (if given).  Does that seem good?

foxnne:
> @foxnne good point - I've pushed a commit here to derive the visible width from the limit (if given). Does that seem good?

Yes i think this is a great solution!

phatchman:
The reason I implemented it as slices on a fixed-size backing buffer was so that if someone changes the char_limit after the first frame, the textEntry buffer will be set to the correct size. Otherwise only the text entry size changes, not the buffer size.

If you test this in widgetpedia, the text entry size shrinks, but the number of characters you can enter stays at the default.

On changing the physical size, I thought about that, but I couldn't really think of any other widget that behaved that way. So, even though it makes a lot of sense, it still seems to special case something that the user could do themselves with min/max_size_content? i.e. we save them typing this `.min_size_content = font.sizeM(text_limit, 1)`

But this way it works a little more intuitively and you can still override the size with min/max_size_content. I'll send through a commit to fix the "resizing" issue. (if you don't beat me to it).

Edit: I think I remember an agreement to prefix the dvui library keys for Data Get/Set with an _? I've changed buffer to use the _ prefix.
And I changed the sizing back to using sizeM so that the placeholder text "mm" can fit.

phatchman:
> @phatchman just curious here, what is the correct way to specify a length for a text entry? i.e. as shown you have short entries that allow 2 characters, what is your code for limiting the physical width of the entry? I

It's just `.min_size_content = dvui.themeGet().font_body.sizeM(text_limit, 1);`
sizeM is really really handy along with font.textSize.

david-vanderson:
> The reason I implemented it as slices on a fixed-size backing buffer was so that if someone changes the char_limit after the first frame, the textEntry buffer will be set to the correct size. Otherwise only the text entry size changes, not the buffer size.

Good point, I was not thinking about it changing.

You are right that we don't have too many widgets that size themselves like this, but I think it makes sense for this kind of "high-level" widget.  Maybe we'll get more things like this over time, or we'll find out I'm wrong!

Excellent - merge when ready.  Thank you!

phatchman:
Yeah. I didn't think of it at first either, until I tested in Widgetpedia. It was quite nice actually. The new option was already available and I could test it immediately.

phatchman:
forgot to limit how big the slice over the buffer can be. Fixed in 5065682077e8d690293a83ad3a6d49165fa7a84b

PR #900, opened 2026-06-10
Alternative Rect/Box Corner Mode

Logickin-Lambda:
# Todo:
- [ ] Evaluate the performance of the additional overhead
- [x] Find a way to make corner radius changeable with a slider
- [ ] Improve the code quality of the current corner path rendering process
- [x] Finalize the naming to the corner mode and functions
- [x] Improve the drawing quality to each of the corner mode at edge cases
- [x] Passing the test cases

# Overview

_Round Corners, Round Corners Everywhere..._

Round corners are common for a lot of good reasons. Some people think it is a more friendly design, some people think it look nice, some people think it is (was) trendy. Though, because of a little experiment from [this discussion](https://github.com/david-vanderson/dvui/discussions/892), I think this would be a good idea to let users defining their own corner styles to fit their needs, to achieve certain aesthetics or purpose (e.g.: drawing flow charts with specific shape); thus, I decided to start this pr, and here is the proof of concept with the prototype (Applied to the Mixed corner mode):

image

Since this is still the early stage to the upgrade, I will show better examples and designs in the later updates.

# What has it changed 
Originally, the corners use the Rect Type where the corners location are indexed by the following values:
```
/// radius values:
/// - x is top-left corner
/// - y is top-right corner
/// - w is bottom-right corner
/// - h is bottom-left corner
```

While it is a fine design, I have found I tended to mix up the corners for w and h which I thought they are x extended with the length of w or h. Thus, I decided to redefine the field with a better naming similar to how Path type define the corners, to reduce the confusion:

``` zig
pub fn CornerRectType(comptime units: dvui.enums.Units) type {
    ...
    tl: CornerType(units) = .{ .auto = 0 },
    tr: CornerType(units) = .{ .auto = 0 },
    bl: CornerType(units) = .{ .auto = 0 },
    br: CornerType(units) = .{ .auto = 0 },
    ...
```

[In the corner type](https://github.com/david-vanderson/dvui/commit/6d09a56634864f3e4bf4d14a7f921f42fc38b5b0#diff-6c8e10ead02eb148cbfa5b1f9ee3167dd111438a829af992b39a780a84009954), instead of a simple radius, it offers a different corner drawing mode, encapsulated inside a union (the drawing mode are not finalized and are subjected to change). The corner mode will be kept as simple as possible, taking 2 parameters at most to prevent too much bloat in memory for simple drawing mode.
``` zig
pub fn CornerType(comptime units: dvui.enums.Units) type {
    return union(CornerKind) {
        const Self = @This();
        none,
        auto: f32,
        arc: f32,
        nudge_x: f32,
        nudge_y: f32,
        angular: struct { x: f32, y: f32 },
        // oval: struct { x: f32, y: f32 },
    ...
```

With the new Corner and CornerRect mode, the addRect() function in Path will draw the corners type based on the given union, and thanks to the latest zig 0.16.0 update, it is possible to rerun the switch statement for the auto mode without specifically writing a recursion. Though, [the current implementation](https://github.com/david-vanderson/dvui/commit/6d09a56634864f3e4bf4d14a7f921f42fc38b5b0#diff-36c675c2688c0a8685e707dd1d97eaa23d57730be15cdcef6c78343b4719ad0a) is verbose and I will find a more generalized solution for the corners.
``` zig
tl: switch (corners.tl) {
  .none => path.addPoint(.{ .x = r.x, .y = r.y }),
  .arc => path.addArc(c_tl, r_tl.x, math.pi * 1.5, math.pi, @abs(c_tl.y - c_bl.y) < 0.5),
  .nudge_x => path.addPoint(.{ .x = r.x + r_tl.x, .y = r.y }),
  .nudge_y => path.addPoint(.{ .x = r.x, .y = r.y + r_tl.y }),
  .angular => {
      path.addPoint(.{ .x = r.x + r_tl.x, .y = r.y });
      path.addPoint(.{ .x = r.x, .y = r.y + r_tl.y });
  },
  .auto => .{switch (default_corner) {
      // sizing information are not necessary because only the enum part of the union are used.
      .none => continue :tl .{ .none = {} },
      .arc => continue :tl .{ .arc = 0 },
      .nudge_x => continue :tl .{ .nudge_x = 0 },
      .nudge_y => continue :tl .{ .nudge_y = 0 },
      .angular => continue :tl .{ .angular = .{ .x = 0, .y = 0 } },
      else => continue :tl .{ .none = {} },
  }},
}
```

At the user level, because of the change, users will need to define each of the corners unions as shown:
``` zig
const mixed_corners = CornerRect{
    .tl = .{ .none = {} },
    .tr = .{ .arc = 2 },
    .br = .{ .nudge_x = 7 },
    .bl = .{ .angular = .{ .x = 100, .y = 25 } },
};

 _ = dvui.button(@src(), "mixed", .{}, opts.override(.{ .corners = mixed_corners }));
```

However, this process can be tedious for users who just want a few simple corners; thus, there are some functions to simplify the process:

``` zig
// the name of the function is not finalized, and I am about to revert the name allAuto() to all() for less painful upgrade
_ = dvui.button(@src(), "100", .{}, opts.override(.{ .corners = .allAuto(100) })); // uses the corner setting from the theme, but limited to symmetric corner type only
_ = dvui.button(@src(), "100", .{}, opts.override(.{ .corners = .allArc(100) })); // all round corner regardless to the theme
_ = dvui.button(@src(), "100", .{}, opts.override(.{ .corners = .all45Cut(100) })); // all 45 degree corner regardless to the theme
```


Logickin-Lambda:
A quick update here, I have replaced the `allAuto()` and `quadAuto` to `all()` and `quad()` for smoother experience to upgrade the existing projects, without the needs to replacing all the existing all() function into something else.

Besides, I have found that angular mode is not ideal for the default corner of the theme since it was intended to be used for drawing diamonds or triangles, and if this mode is used, we will see odd looking diamond shape on the sliders and scroll bars:

image

For that reason, the Corner will have two types of corner mode, primitive and extended. Primitive Corner Mode (`.none`,`.arc` and `.cut45`) are the one can be used in `all()` and `quad()` since they are geometrically stable in all rect size. The remaining modes are the Extended Corner Mode, and they can be drawn only if the user have manually specify them. If the user trying to define a theme with the extended corner mode, it will fall back to one of the primitive corners. (I am thinking of doing a warning message if some users do that with their theme)

david-vanderson:
This is good stuff!

Some random thoughts so far:
* `.none` can be a special case of `.arc` right?
* `.auto` means "use the theme" right?  Maybe we call it `.theme`?
  * I like this because we can nuke `themeOverride` - it only exists to override the corner_radius.  This is a better solution.
* looking at the existing widget defaults shows:
  * they all use the same value (currently radius 5)
  * they only differ in that some (scrollArea, grid) have sharp corners on the top and radius 5 on the bottom
  * probably that means having a single Theme default corner works, and if corners is null it means use that default on all corners
* seems fine to have corner values (your extended ones) that don't work as theme defaults
  * even so, is it better to have absolute sizes, or do percentages of the rect side lengths?

Logickin-Lambda:
Thanks for your thought! These actually give me some deep thoughts to the details, so let's discuss them one by one:

## `.none` can be a special case of `.arc` right?

It could be, except for the win98 theme which it has a rectangular ninepatch texture, and if we use the `.auto` mode while the theme uses `.arc` mode, it creates a weird looking rounded outline clashing onto the rectangular texture. This problem can also be seen in the current version of the web backend demo:

image

Thus, I decided to add a `.none` none mode specifically for the win98 theme, making it forbidden to draw round corners in `.auto` mode.

image

The mixed corner isn't drawn with the `.auto` mode, but with specific corner type, so it is intended have the weird appearance:

image


However, it still have problems since it also overrides the button used with other theme and the radio buttons are no longer round, so I will find a solution for that.

image

## .auto means "use the theme" right? Maybe we call it .theme?

Yes, I think `.theme` also works and `.auto` means "use the theme", 

## I like this because we can nuke themeOverride - it only exists to override the corner_radius. This is a better solution.

There is a subtle difference between `.auto` and `themeOverride()` which the former have no consideration to the default radius of the theme but the corner style, while the later takes both. It might be possible to use the default radius from theme if I modify the union with optional for the primitive mode:

``` zig
// primitive modes
auto: ?f32,
arc: ?f32,
cut45: ?f32,
```

I will need a bit more experiment to see if it is possible to completely replace the `themeOverride()` function, and find out how ergonomic at the user end. (e.g.: do users need to spam a bunch of `.?` to fetch the values?)

## looking at the existing widget defaults shows: ...
> - they all use the same value (currently radius 5)
> - they only differ in that some (scrollArea, grid) have sharp corners on the top and radius 5 on the bottom
> - probably that means having a single Theme default corner works, and if corners is null it means use that default on all corners

Yes, and this was my intention. Although asymmetrically cut corners look fancy, I wish to keep the default setting simple since people who use all round or no corners are still the majority, and asymmetric corners can be highly depending on the alignment and position which don't really have a good default mode, so I will hand this job to the users instead if they want something customized.

## seems fine to have corner values (your extended ones) that don't work as theme defaults; even so, is it better to have absolute sizes, or do percentages of the rect side lengths?

I did have a thought about this due to the existence of the [CSS discussion](https://github.com/david-vanderson/dvui/discussions/148), while I also know that percentage mode can be handy for some situations. For example, when a user want to draw a diamond, they can just specify 50% for all corners instead of doing all the hard coding or fetching the rect dimension from `.data()`.

However, I am puzzling about how it will be implemented efficiently and how to introduce the idea with less confusion to the users.

In the original corner radius rendering function, it assumes that the max radius size on the box is the minimal value to one of the halved sides so that the radius of each corners don't overlapped to each others:

``` zig
var rad = radius;
const maxrad = @min(r.w, r.h) / 2;
rad.x = @min(rad.x, maxrad);
rad.y = @min(rad.y, maxrad);
rad.w = @min(rad.w, maxrad);
rad.h = @min(rad.h, maxrad);
```

During the upgrade, I have kept the same convention to prevent the corner drawing invading other corners to end up with a glitchy looking shape, and if we decided to use percentages for some of the modes, we need to determine:

- what is the definition to the percentages? does 100% mean:
  - half length to each of the side just like the original implementation to prevent two corners colliding each others
  - 100% still means the whole length of a side, but limited to 50% as the maximum
  - taking the whole side, so if both of the corners collide, they will be drawn according to their ratio, similar to how css works.


The first and the second approach is simpler in logic, but users could be confused by the meaning of 100% being the half length to the sides, or they might questioning why can't they use value more than 50%, respectively. The last approach looks a bit more intuitive and it can be more flexible, but it can be much more complicated to calculate the ratio to maximize the corner "radius" without glitching the shape.

What do you think about these approaches for the percentage mode?

david-vanderson:
Good details!

> It could be, except for the win98 theme which it has a rectangular ninepatch texture

I think we say if someone is using ninepatch, it is up to them to ensure that the corners value works with their ninepatch.  They could be using a ninepatch that visually looks like it has rounded/cut corners and want the focus outline to match.  So win98 theme needs to set the default corner to "arc 0".

The problem in the demo with the corner radius 100 button (good catch!) can be stated "if a callsite specifies particular corners, it is up to that code to deal with potential ninepatch issues".  The most likely thing would be forcing the default ninepatch off when doing weird corner stuff, and the demo should do that.

Does that sound reasonable?

> It might be possible to use the default radius from theme if I modify the union with optional for the primitive mode:

Yes, sorry - I thought that was already part of the proposal.  So the adwaita theme would have a default corner of "arc 5" right?

> What do you think about these approaches for the percentage mode?

That is a lot of good insight.  The more I think about it, the less sure I am.  It solves some questions about what happens if the absolute lengths are too big.  On the other hand I'm having trouble imagining a situation where you want the size of the corner to scale with the size of the rectangle, at least for normal UI stuff.

Maybe we start with the absolute stuff, and see if anybody needs the percentages?  If they are not needed widely then they can always be computed locally.

This is great - my recommendation is to move forward with just the obvious parts:
* splitting corner_radius into corners
* having .arc and .theme

And let's keep discussing the others:
* should cut45 and nudge be separate?  Maybe just a single "cut" that has an x and y offset?
* I'm not sure what "angular" is doing?

Logickin-Lambda:
> to set the default corner to "arc 0"... Does that sound reasonable?

Perhaps you are right because I have suddenly found a problem on the `.none` mode at the user's perspective, and they will very likely wondering:

> why is my `.all()` does nothing at all? I Just want to have a border to have round corner where it doesn't need any texture!

Because of that, I will remove the `.none` mode so the original behavior before the upgrade is retained, until the users want this highly restricted mode in the future.

> Yes, sorry - I thought that was already part of the proposal. So the adwaita theme would have a default corner of "arc 5" right?

No worries. I didn't catch that in the proposal since I didn't aware of the function, but I think that plan is excellent due to the apparently overlapped purpose of that function. Thus, even though I have no guarantee the current default corner working as you described, I will find a way to match the expected result in the final release.

> Maybe we start with the absolute stuff, and see if anybody needs the percentages? If they are not needed widely then they can always be computed locally.

Sure! I think this has another advantage because I just realized if we add the percentage mode to the corners, very likely, users will demand us to do more with other components and units, because I can already foresee some potential requests from the users like:

> - Percentage Mode For Other Widgets, Fonts and Texture?
> - Can we have `vw` and `vh` mode for drawing `Rect`?

Thus, by holding this feature, we could have a bit more time to think of a more general and robust solution for all widgets, perhaps `.Percentage` variant besides `.Natural`, `.Physical`, or some form of `UnitConverter{}` Types? I think this feature alone is worth a whole new pr for discussion because it seems to have a lot of potential yet we might not know, and it definitely needs more people to make an agreement to define the units.

> This is great - my recommendation is to move forward with just the obvious parts:

Alright, before doing the modification, I'd like to confirm a bit more:

> splitting corner_radius into corners

Do you mean the original `.corner_radius` field in `Option` type, or the `corner_radiusGet()` function?

If you mean the former, I am on the way doing it, as if you search my commits with "SKREEKH", you will see a lot of `.corner_radius` fields being commented, replaced by `.corners`. The "SKREEKH" tag is done unprofessionally on purpose to distinguish those are my TODO reminding me to cleanse the code before requesting for a merge which they will all be eliminated.

For `corner_radiusGet()`, I haven't rename the function name yet because I haven't finalize the behavior, and it is easier to explain with how the corner mode works and how they behaved in the current version of `corner_radiusGet()`, so let me answer the following answers at once:

> - should cut45 and nudge be separate? Maybe just a single "cut" that has an x and y offset?
> - I'm not sure what "angular" is doing?

Starting from the original `.arc` mode, because of the radius restriction, each corner can move within their respective colored region, and the dot represents the radius of the respective corners:

image

Originally, I thought `.angular` mode would be sufficient for most type of sharp corners, building all the common required geometries, but after I have studied the addRect() corner rending process and how the default widget corners are defined, I have found the two main problems:

The first problem is the current restriction makes trapezoids or triangles impossible to be drawn, if x and y value are the length of the side of the rectangle, we can only get a diamond shape at most:

image

For the second problem, we may now go back to the `corner_radiusGet()` function because this mode have two parameters, x and y offset, and I don't seem to find a good solution for this mode that plays well with the default widget settings. Currently, `corner_radiusGet()` behavior for that `angular` mode is basically getting the longer side of a rect, but this have caused the [previous problem](https://github.com/david-vanderson/dvui/pull/900#issuecomment-4669975215) where some widgets, especially with long bars ended up with some ugly diamond shapes, and I don't think this is something users would ever wanted as the default. 

I have also thought of using the min value or average for the values, but they are either too conservative with ending at smaller corners than the definition or too unpredictable with the average function.

For that two problems, I ended up with the simpler route, by splitting the angular mode with two other variants, with the concept of `Primitive` and `Extended` mode I have mentioned before. Thus, the three corner mods behave like the following:

The original `.angular` mode works with two parameters, x and y offset, both of the parameters controls the offset of two point with the offset x and y respectively:

image

Since `.angular` mode is not stable as a primitive corner which is not stable with `corner_radiusGet()` used for builtin widgets, I decided to have have restricted version, thus cut45 which it only use the radius parameter so that is can be easily swapped with the original `.arc` mode, for some more sci-fi theme where sharps corners are more prominent.

image

To draw parallelogram or trapezoid, we still can't do that with the two mode above due to the restricted half length radius, so this is where the "nudge" mode comes in. Instead of controlling two point given an offset, "nudge" mode moves the point of the corner (only visually), directly plotting a line between the two adjacent corners (I think I should merge both nudge_x and nudge_y into a single mode):

image

If the adjacent corner have other corner mode, it will use the start and the end point of that corner:

image

The `.oval` mode is not here yet because I haven't implemented the bezier curve, but it works in a similar manner as the `.angular` mode, but curved.

Thus, with mixing and matching these corner modes, we may construct most of the shapes we would commonly used, within the limitation of the half length radius restriction: (the rhombus and the right triangle are possible if I have merged both `.nudge_x` and `.nudge_y` mode into just `.nudge`)

image

Though, the naming and the corner modes are not ideal since they are done for the proof of concept; thus, we may discuss what mode and name should be made into the final version, or if we need to change the rendering process.

david-vanderson:
Excellent - thanks very much for the pictures, I now understand, especially the difference between angular and nudge.  Thanks!

> For the second problem, we may now go back to the corner_radiusGet() function because this mode have two parameters, x and y offset, and I don't seem to find a good solution for this mode that plays well with the default widget settings.

This is a bit confusing, it sounds almost like trying to come up with a single number for each corner mode?  Or is the problem more how to override widget-default corners with corners from the callsite?

Probably I don't understand, but my guess is that we would treat "corners" as a single thing - so callsite code can't set just one corner and expect the rest to come from defaults.

As for the long ugly diamonds, we don't need to programmatically deal with it.  If someone writes a theme that way they get what they asked for.

I think there's something I'm still not getting about corner_radiusGet() - seems like it would turn into cornersGet()?

Logickin-Lambda:
> I think there's something I'm still not getting about corner_radiusGet() - seems like it would turn into cornersGet()?

You are right, and I am an idiot lol.

The original problem was related to functions like these:

```
rs.r.stroke(self.options.corner_radiusGet().scale(rs.s, CornerRect.Physical), .{ .thickness = b.x * rs.s, .color = self.options.color(.border) });
```

I thought I have to return one radius value for the scale function, but I have forgotten the fact that the `.corner_radius` parameters in `.stroke` and `.fill` has already been replaced with `.corners`, meaning that I can just pass the corner values instead of the picking either one. 

Everything is now much easier, and the only tricky part now, is this function call, starting from line 3279 in `dvui.zig`:

```
const tl = dvui.Point.Physical{ .x = r.x + cr.tl.getRadius(), .y = r.y + cr.tl.getRadius() };
path.addArc(tl, cr.tl.getRadius(), std.math.pi * 1.5, std.math.pi, false);
```

But this could be solved after I have refactored the corner function, so this shouldn't be a huge issue.

> As for the long ugly diamonds, we don't need to programmatically deal with it. If someone writes a theme that way they get what they asked for.

Then I will ditch the concept of `Extended` and `Primitive` type while they will have no fallback logic for `.theme` mode since the problem of `corner_radiusGet()` is solved as mentioned and I think your approach is more idiomatic to zig which reduce the hidden and magical behavior.

Thank you! This is a crucial hint to my overthought problem.

Logickin-Lambda:
Finally, made some progress for the theme override function (I have changed the corner mode in win98 for testing):

https://github.com/user-attachments/assets/d693f6dd-64db-4c3b-ab6d-cedd0507c921


It was actually a hard task since the combo of `.themeOverride()` and `.override()` added a lot of complexity to the task which results in a lot of potential yoyo code, converting the type back and forth:

image

I nearly gave up since I was stuck at this problem for a couple of days which I thought I couldn't find a good solution to determine the corner style within one function call since with the original setup, there is no way to tell if users use the default corner or the default corner **mode** just with `.theme` alone, while there are some default struct can only accept corner mode in comptime which using `dvui.currentWindow().?.theme` seemed impossible.

However, because of the default structs for the widgets, I was thinking what if I just add a new corner mode specifically used for defining the default setting of the widgets? Thus, `.widgetDefault` is created for this purpose, and since we still need to resolve the `.theme` with a new function, we could relocate the original logic `.themeOverride()` into that corner function. In other words, **we can now nuke the `.themeOverride()`** because of the use of this new mode. You may wondering when to use `.widgetDefault` or `.theme`. Normally, users should always use `.theme` unless they are defining a default option type for a new widget they are developing.

image
 
To override the theme, since it requires both the `Theme` and the `Corner` type, I decided to include the corner determination function in the `.init()` function of the `WidigetData` type because this is where most of the widgets have to go through before rendering, while both of aforementioned struct are presented:

``` zig
pub fn init(src: std.builtin.SourceLocation, init_options: InitOptions, opts: Options) WidgetData {
    const parent = dvui.parentGet();
    const id = parent.extendId(src, opts.idExtra());

    var options = if (dvui.debug.options_override.get(id)) |val| val.@"0" else opts;
    if (options.corners) |*corners| corners.* = corners.finalize(options.theme); // <--- Newly added
```

In addition, I have replaced the original union design with a simpler struct design since there is no clean way to convert an active field of a union into another active field, and I have to reconstruct them case by case, result in some hard to read switch statement which is not good for maintenance. With the direct access to the struct field, it should also be easier to access the data for sliders.

david-vanderson:
Looking good!  I think we can simplify it further - I can't think of a reason you'd want to combine the theme default and the widget default together.

The default for CornerRect should be .all(0).

The `.theme` option doesn't need to take a number - you either get exactly the corner from the theme, or you specify something else.  That should line up with the normal Options.override.

Then the final determination for each corner can be done when it's required:
* if options.corners is null assume .all(0)
* if the specific corner is .theme, look it up, otherwise use directly

Does this help at all?  Let me know if/when you are done with this.  I think I understand it well enough to take over if needed.

Logickin-Lambda:
Unfortunately, I am not sure if that can happen and I think my explanation have left you some confusions.

`.all(r)` is actually uses the `.theme` mode under the hood:

``` zig
pub fn all(r: f32) Self {
    return CornerRectType(units).quad(r, r, r, r);
}

pub fn quad(r_tl: f32, r_tr: f32, r_br: f32, r_bl: f32) Self {
    return .{ .tl = .{ .type = .theme, .rx = r_tl, .y = r_tl }, .tr = .{ .type = .theme, .rx = r_tr, .y = r_tr }, .bl = .{ .type = .theme, .rx = r_bl, .y = r_bl }, .br = .{ .type = .theme, .rx = r_br, .y = r_br } };
}
```

The purpose of the `.theme` mode is to let user create a corner with any arbitrary size, but the style of the corner is based on the theme, similar to the current `.all(r)` but with auto corner style selection. For example, when the default corner of a custom theme is `cut45(10)`, if the users use `.all(100)`, the widget will have a corner radius of 100, using the cut45 mode. If this mode doesn't take a number and it is bind to the corner setting of the theme, the corner will always be the size of the given theme, and the `.all(r)` won't work. This was why it originally called the `.auto` mode.

Nevertheless, it only solves part of the problem since each widget has their own default corner size and we need to preserve the setting if users doesn't override, and if a theme override the default corner, it takes a different logic:

``` zig
// The original themeOverride(theme) function before the upgrade
if (t.max_default_corner_radius) |mdcr| {
    if (ret.corner_radius != null) {
        ret.corner_radius.?.x = @min(ret.corner_radius.?.x, mdcr);
        ret.corner_radius.?.y = @min(ret.corner_radius.?.y, mdcr);
        ret.corner_radius.?.w = @min(ret.corner_radius.?.w, mdcr);
        ret.corner_radius.?.h = @min(ret.corner_radius.?.h, mdcr);
    }
}
```

If I ignore this part of the code and use the `.theme` mode, because the button widget have a default option using the corner radius of 5, the `.theme` will use the corner style of the theme, but not the radius. As a result, the win98 theme will lose its sharp corner. This is why `.widget_default` is added into the design, and it can't take no value either because most of the widget have their default radius defined in their options, and it can't be solved with a single radius value:

``` zig
// The original defaults option struct in ButtonWidget.zig 
pub var defaults: Options = .{
    .name = "Button",
    .role = .button,
    .margin = Rect.all(4),
    .corner_radius = Rect.all(5),  // <-- This needs the `.widget_default` mode
    .padding = Rect.all(6),
    .background = true,
    .style = .control,
};

// The original slider_entry_defaults in dvui.zig
pub var slider_entry_defaults: Options = .{
    .name = "SliderEntry",
    .role = .slider,
    .margin = Rect.all(4),
    .corner_radius = dvui.Rect.all(2), // <-- so is this, and this doesn't have the radius of 5
    .padding = Rect.all(2),
    .background = true,
    // min size calculated from font
    .style = .control,
};

```

Perhaps we might mixed up the corner modes during the discussion since your `.theme` mode works more similar to my `.widget_default`, while `.theme` actually means some other modes in my implementation. Let's see if this could clear things up, and if we need further actions.

>  Let me know if/when you are done with this.  I think I understand it well enough to take over if needed.

I was thinking of adding an `.oval` mode with bezier curves, but perhaps I am going to skip this part since the use case is rare. Though, I will add some new themes to demonstrate the new feature after I have cleaned the code I have tagged with the comments, hopefully done by this week.

david-vanderson:
Aha - thank you, yes I misunderstood.

> The purpose of the `.theme` mode is to let user create a corner with any arbitrary size, but the style of the corner is based on the theme, similar to the current `.all(r)` but with auto corner style selection.

Okay I think I've got it.  The best example I can think of is the windowHeader close button:
* close button has widget default `.theme(1000)` ("Use the theme style but as large as possible")
* Adwaita has default `.arc(5)`, so the close button is a circle
* win98 has default `.none`, so close button is a square
* potential theme has default `.cut45(10)`, so the close button is a diamond

Is this correct?

If so, then we should be able to remove `Theme.max_default_corner_radius`, because:
* if you want square corners, make the theme default by `.none`
* if you want smaller corners than the theme has, you directly edit the theme

Does that make sense?

Logickin-Lambda:
Bingo! Except for the `.none` mode since we have decided to remove it in the earlier discussion since this [makes `.all(r)` doing nothing in a way that might confuse users](https://github.com/david-vanderson/dvui/pull/900#issuecomment-4687625561) which is not something we want to see in `.all(r)`. The win98 theme now uses the `.arc(0)`, but you get the idea. 

The `Theme.max_default_corner_radius` should be replaced by Theme.corners during the development in one of the earlier [commits](https://github.com/david-vanderson/dvui/pull/900/changes/6d09a56634864f3e4bf4d14a7f921f42fc38b5b0#diff-5f2317c02c275b6989a1c464c45b6925237d4b8fa22295b2101aa20bbb99af24), so it should be already gone.

Though, should I rename the corner mode? Do I need to make it more consistent? Currently, I have:

```
widget_default // use the theme's corner, and the default size is replaced by the theme if the theme have the smaller corner size
theme          // follows the theme's corner style, but in any given sizes
arc
cut45
nudge
angular
```

Logickin-Lambda:
btw, let me show you the new builtin theme. The font is [Interceptor](https://www.pixelsagas.com/?download=interceptor), and the font is in SIL Open Font License, and both of the license files are included in the project.

image

david-vanderson:
> ```
> widget_default // use the theme's corner, and the default size is replaced by the theme if the theme have the smaller corner size
> theme          // follows the theme's corner style, but in any given sizes
> arc
> cut45
> nudge
> angular
> ```

Now that I understand what is going on, I suggest:
* combine widget_default into theme by making the radius optional (or using a placeholder like -1)
  * I don't think we need to keep the logic of taking the minimum if we have a placeholder
* `arc` -> `round`
* `.corners = .theme` means all corners default to the theme
* `.corners = .all(5)` means use the theme style but radius 5
* maybe `cut45` -> `cut` but less sure
* `nudge` and `angular` seem good

> btw, let me show you the new builtin theme. The font is [Interceptor](https://www.pixelsagas.com/?download=interceptor), and the font is in SIL Open Font License, and both of the license files are included in the project.

Very cool!  Exciting!

Logickin-Lambda:
Thanks, so most of the modes are settled and I could change it accordingly soon. For the `cut45` mode, since you also doubt a little about the alternative names, I decided to take a look what wood workers describe these kind of corner cut.

What about... `.chamfer` or `.bevel` to replace `.cut45`?

image

> combine widget_default into theme by making the radius optional (or using a placeholder like -1) 

One question remains. If I merge `.widget_default` and `.theme`. How would I handle the different default corner size for `const defaults: Option{}`? The problem here is that some of the widgets use a different corner radius as the default option:

``` zig
// The original defaults option struct in ButtonWidget.zig 
pub var defaults: Options = .{
    .name = "Button",
    .role = .button,
    .margin = Rect.all(4),
    .corner_radius = Rect.all(5),  // <-- This needs the `.widget_default` mode
    .padding = Rect.all(6),
    .background = true,
    .style = .control,
};

// The original slider_entry_defaults in dvui.zig
pub var slider_entry_defaults: Options = .{
    .name = "SliderEntry",
    .role = .slider,
    .margin = Rect.all(4),
    .corner_radius = dvui.Rect.all(2), // <-- so is this, and this doesn't have the radius of 5
    .padding = Rect.all(2),
    .background = true,
    // min size calculated from font
    .style = .control,
};
```

Since `.all()` uses the `.theme` mode, and if I replace that with `.all(-1)` as your suggestion, the distinction of these default corner size will be lost, ending up some corner rendering larger than the original design. Using `.all(5)` and `.all(2)` don't work either because it doesn't use the corner size of the theme, but the style, and this will break the win98 theme by rendering the default rounded corner.

``` zig
// If we have merged .theme and .widget_default mode:
// in ButtonWidget.zig 
pub var defaults: Options = .{
    .name = "Button",
    .role = .button,
    .margin = Rect.all(4),
    .corners = .all(-1),  // Replaced, if there is no theme, use arc(5) by default
    .padding = Rect.all(6),
    .background = true,
    .style = .control,
};

// slider_entry_defaults in dvui.zig
pub var slider_entry_defaults: Options = .{
    .name = "SliderEntry",
    .role = .slider,
    .margin = Rect.all(4),
    .corners = .all(-1), // <-- wait... it is also now... .arc(5)? but the original was .all(2) by theme default
    .padding = Rect.all(2),
    .background = true,
    // min size calculated from font
    .style = .control,
};

// same thing for the checkbox default, your square will be much rounder since the default corner_radius was also 2,
// but with this suggested change, it will default to .arc(5) instead of the original .arc(2)
```

Have I misunderstood your idea? Should we really merge the two modes?

david-vanderson:
> What about... `.chamfer` or `.bevel` to replace `.cut45`?

`.chamfer` is good!

> Since `.all()` uses the `.theme` mode, and if I replace that with `.all(-1)` as your suggestion, the distinction of these default corner size will be lost, ending up some corner rendering larger than the original design. Using `.all(5)` and `.all(2)` don't work either because it doesn't use the corner size of the theme, but the style, and this will break the win98 theme by rendering the default rounded corner.

Excellent - getting to the bottom here. To restate it:
* some widget defaults want the theme style AND size (button)
* some widget defaults want the theme style BUT custom size (sliderEntry)
* some themes (win98) want to override all sizes to zero

Across our widget defaults for radius, we have:
* most use 5
* some use 2
* some use 0
* some use 1000 (as large as possible)
* groupBox uses 3, but maybe that could be 2

Some options:
* for the win98 case, we could bring back the `.none` (or `.square`) corner style, which ignores radius
* we could say widget default radius is a property of the widget, only style is set by the theme (beyond making it square)
* theme could give a single default radius (5), and some widgets default smaller
* theme could have two defaults, one "normal" (5), and one "small" (2)

Does all that make sense?

I'm leaning towards brining back `.square` and having the theme define a single default with radius 5:
* button default would be `.corners = .all(-1)` (could make `.theme` and alias for `.all(-1)` to make it obvious
* sliderEntry default would be `.corners = .all(2)`

How does that sound?

Logickin-Lambda:
> I'm leaning towards brining back `.square` and having the theme define a single default with radius 5:

Me too! That's was the reason why I did a `.none` mode since win98 theme is the only exception here, while other corner modes could follow the same default theme dimension without any issue.

> - we could say widget default radius is a property of the widget, only style is set by the theme (beyond making it square)
> - button default would be .corners = .all(-1) (could make .theme and alias for .all(-1) to make it obvious
> - sliderEntry default would be .corners = .all(2)

Sounds good, and I think it is fair since it is basically means "this is what you get from what you have set for your theme", so I will test that to see if it works well.

Logickin-Lambda:
Alright, I think I am about to finish this iteration:

The suggested solution is applied, and to make it intuitive to use the default theme corner radius, instead of letting users explicitly typing `.all(-1)`, I decided to add a helper to abstract this process with a `.default()` function:

``` zig
pub fn default() Self {
    return .{ .type = .theme, .rx = -1, .y = -1 };
}
pub fn allDefault() Self {
    return .{ .tl = .default(), .tr = .default(), .bl = .default(), .br = .default() };
}
```

They are identical to `.theme(-1) and .all(-1)` respectively, but easier to understand.

With, these changes, there are minor visual differences, but I actually prefer in this new way:

Because of the `.square` mode, win98 theme now have a square slider instead of the original round design, and I find it fit better to the mechnical, blocky theme.

win98 slider change

If .`.default()` is used, since we are now using the exact corner size of the theme instead of finding the minimum, we could draw larger corner for the buttons, and these near hexagonal shape button looks really nice!

image

Furthermore, I have fixed some of the visual glitches, and updated some part of the demo for better demonstration to the new corner mode:

The highlight on the tabs were originally unaligned to the tab because it was drawn only in arc mode while the scaling factor were missing, but because the new corner mode, I have fixed it so that it can be drawn with the proper theme and scaling:

tab corner fix

The original corner radius slider in the theming demo freaked out when user slowly sliding between the largest possible radius and null. I have replaced that with a checkbox to toggle between the default null and other corner mode. The required slider will also be shown accordingly. (Outdated video, the `theme` mode now defaults to `round`)

https://github.com/user-attachments/assets/e7dda467-21e4-46c9-8597-92d04e439617

Corner radius will show some of the common corner modes, and showing how different theme changes the default corner:

image

Lastly, Red and Green variant of the Tech theme are added:

new theme

david-vanderson:
> With, these changes, there are minor visual differences, but I actually prefer in this new way:

Yeah I agree.  Thanks for fixing the tab highlight!

> Lastly, Red and Green variant of the Tech theme are added:

Looks great!

> Alright, I think I am about to finish this iteration:

Awesome!  Let me know when you are close and I'll go through all the changes again.  Can we change the "allRound" things to just "round"?

Logickin-Lambda:
> Thanks for fixing the tab highlight!

No worries, as this is something related to the corner feature, so I wanna change it to work with the corner update.

> Can we change the "allRound" things to just "round"?

The `allXxxx()` convention was done because they are the subtype of `.all(r)` and `.quad(r,r,r,r)`, thus`.allRound()`, `quadChamfers()`, and etc, following the similar convention. If we want to ditch `all` for the CornerRect helpers, to not mistaken with the `Corner` type functions while representing the idea of changing all four corners, perhaps we need them plural.

Though, for the `.quad` case, unlike java or c++, we can't use polymorphism to have multiple parameter profile with the same function name, so we might likely just keep them as `quadXxxx(r,r,r,r)`, which we will have the following for the `CornerRect` type:

``` zig
// .nudges(x,y) is not included since this basically shrinks the rectangle which is meaningless while this is 
// not a common type for drawing a border.
.all(r)
.defaults()
.rounds(r)
.squares()
.chamfers(r)
.angulars(x,y)

// I only offer these three mode to have quad since other modes will takes 8 parameters which 
// is too many for a function, so users should explicitly declare the struct with the `Corner`
// functions instead, while. .square and .default only have one size, so no quad required.
.quad(r,r,r,r)
.quadRounds(r,r,r,r)
.quadChamfers(r,r,r,r)
```

I can see the reason for that change since these all function modes are commonly used, so `rounds(r)` feels more convenient to type and less noisy to `allRound(r)`. This is a simple change, so I can change it right now to see how well it looks, and change it further.

>  Let me know when you are close and I'll go through all the changes again.

I think this iteration is basically done, so you may prepare to go through all the changes. I have some future plans to extend the corner modes, but they won't be in this release.

PS: The name change have already [applied](https://github.com/david-vanderson/dvui/pull/900/commits/5a1bf372ce207746329173e1ae329a87697e49db), and I think I didn't mentioned it clear, so here is the reminder.

Logickin-Lambda:
> Let me know when you are close and I'll go through all the changes again.

Hi mate, how is the review doing? Any changes I need to do?

david-vanderson:
> Hi mate, how is the review doing? Any changes I need to do?

Thanks for the check, sorry I've just been behind, will get to it soon!

david-vanderson:
I'm happy with this, will merge today or tomorrow, unless you have any objections to the naming changes?

I will credit you in the devlog as "Accipiter Nova" (as before).  Let me know if you want something else.  Thank you!!

Logickin-Lambda:
Although you have removed the `.quadxxx()` function, I think this is understandable considering round and chamfer are the only type that can have this mode. Others are just minor name changes, so I am fine with those.

For the devlog, because of some compatibility reasons and people getting confused with my naming, ending up all combinations of `Logickin`, `Lambda`, `Accipiter`, and `Nova`, I decided to revert my profile as just `Logickin`. 

Please use `Logickin` for this and the future contributions.

Sorry for this name change, but this is a necessary evil so that people will have less confusion to my artist name and all the urls and references will remain intact.

david-vanderson:
Will do.  Thanks so much for bearing with me through this!

Logickin-Lambda:
No worries, and thanks for the opportunity to let some of my gui ideas to be in the reality! 

After that, I have a future roadmap to the corner types:

As you can see, the corners below can't be done with the current implementation because they needs more than 2 parameters, but due to the niche use case, it is also not a good idea to include all these variants. It is possible to use `overlay` to hide some of the edges and corners with a different `rect` element, but this doesn't work if the shape are transparent; thus, I am thinking of a generic callback to let users implementing their own Corner mode and bypassing the limitation to the path functions, filling out the niche style corner their their own needs. I call it the "Edge" and "Corner" Case Update. 

dvui future plan, the corner &#39;corner&#39; case

Though, I will pause this future pr for a month or two since:

- I can observe if users can adopt the changes effectively and if they have encountered some bugs or performance issue.
- I need some experiments to find a good way squeeze the function pointer into the `Corner` type (perhaps using `extern union`?)
- I am working on a really complex library from sketch as another project

david-vanderson:
If the shapes become concave, then we'll have to implement a concave polygon tessellator with edge fading.  That would be very useful, but also a good deal of work.

From a high level a custom callback function is similar to requiring the calling code to do all the rendering.  Is there some way that would be different then having the calling code skip the normal rendering and do all custom?

Logickin-Lambda:
> we'll have to implement a concave polygon tessellator with edge fading

hmmm... this already sounds like a great challenge, and may I know what is the limitation to the current design? Though, I only know little about tessellation in graphics programming, so I will need to study the concept before I could impose some ideas and solutions.

> From a high level a custom callback function is similar to requiring the calling code to do all the rendering. Is there some way that would be different then having the calling code skip the normal rendering and do all custom?

Good question and I think you have reminded me of an overlooked detail. If user want to draw a shape that is so unique which requires a custom path, perhaps they should just draw them using `Path` since at this stage, the difference between a custom callback and drawing directly with `Path` are small, just realized.

I thought that approach could solve the scaling problem just like other corner types, but I have forgotten that this approach will require users to store additional dimension details externally (in a struct or an array of their widgets), and these details are still not scaled anyway. As a result, this actually fails to my original expectation and objective.

Thanks for asking that question since I have realized the idea is not working as good as I expected and it might cause more problems.

david-vanderson:
> hmmm... this already sounds like a great challenge, and may I know what is the limitation to the current design?

dvui converts all drawing to triangles.  The existing routines handle convex polygons and stroking line-segment paths.  Both are relatively easy to convert.  The extra wrinkle is that we add thin triangles to the edges that "fade" from the color to transparent to produce a kind of antialiasing effect.

Converting concave polygons is harder (and self-intersecting ones even worse).  There are algorithms to do it, but I haven't looked into them much yet.  I think the svg2tvg branch connected with https://github.com/david-vanderson/dvui/pull/864 uses some of them.

Eventually I hope we can add a general concave path drawing routine.  But I don't know if we can use the existing algorithms as is and still be able to add our fade to the edges.

nat3Github:
yep this is implemented in #864, so all those shapes above are possible, even self intersecting shapes to a degree. cheers!

PR #901, opened 2026-06-11
TableWidget

david-vanderson:
This grew out of a request from @MichaelBelousov to have something like GridWidget that was closer to a spreadsheet.

Thank goodness for GridWidget, and thank you @phatchman - I'm really standing on your shoulders here!

Try the "table" section of the demo here: https://david-vanderson.github.io/demo-table/

* "styling" tab you can arrow-key around the cells, and click or press enter to edit the values.
  * for the demo only visible edits are persisted
* "csv" tab you can click the column headers to sort them.  This part steals heavily from https://codeberg.org/BSF/dvui-editor

If this proceeds, then I'll figure out how this and GridWidget go together.  Either merging somehow or delineating the usecases, haven't thought about it much yet.

This has gotten far enough to warrant getting outside opinions.  Is this going in a good direction?  What are the obvious missing features?


david-vanderson:
@foxnne Your feedback would be appreciated!

MichaelBelousov:
noted potential issues:
- on mobile the table jitters when swiping past the bounds of the table, it's jarring
  - I could not reproduce on my laptop touchpad
- I expected the escape key to unfocus the edit mode of a cell
- I expected the tab key to move rightwards to the next cell (and possibly wrap to the next row)

I assume you can sort editable rows? The demo doesn't show that. 

Otherwise looks great to me, awesome work! I love the ability to scroll through all those rows!

phatchman:
Looks great! Had to copy Examples\with_headers.csv to src\examples for it to build.

Yeah, I can see they are quite similar both in functionality and code and it seems to me we should be able to come up with some way to merge them and have the Table Widget be a specialization of the Grid Widget.

I think the first step could be to refactor the Grid Widget code so that it more closely matches the structure of the Table Widget. e.g. make the Grid Widget a "proper" widget, implement things like autoSize and rename the grid functions so that they are similar to the table ones.

I think then it will be more clear what the differences and similarities are. It would be nice not to have to maintain two lots of "make a sortable header" functions etc.

Happy to work on that in bits and pieces to see how similar we can make them. There are a good set of tests for the grid widget, so that will make life easier. 

Edit: Just re-read your original. Of course, feel free to work out how to merge them yourself. I'm working on another project right now, so it would take me some time to get to it.

david-vanderson:
I'm pretty happy with where this is, sorry for the slow response.

I've uploaded the current version to https://david-vanderson.github.io/demo-table/ (duplicates the table part on the main screen and in the demo window)

Still to be done, probably after initial merge:
* accessibility (will do this pre-merge)
* multi-cell selection
* cell copy/paste
* cell spanning
* row headers

@MichaelBelousov I think I've hit all your feedback:
* mobile touch scrolling jitters should be fixed (thanks for pointing this out!)
* visual difference between a cell having focus and being edited makes unfocus more clear?
* tab moves right (shift-tab left) wrapping rows

The usage code could sort editable rows if it wanted, but I wasn't sure how to put both together into the demo because it's generating the data on the fly.  Do we need a demo specifically to show that?

@phatchman These are the big functionality changes:
* banding and row hover/click are entirely in usage code
* KeyboardNavigation replaced with "layout_only mode", where the table doesn't focus itself or handle events, just positions child elements, and the tab index is left to the usage code
* "checkbox cell" is now in usage code as well - I found that the code can detect if the grid cell has focus and transfer the focus to the checkbox widget

I hope that makes some sense.  Are these changes going to mess you up?  My plan was to swap this in for Grid but if that's terrible for you we can figure out a better transition.

david-vanderson:
Ah - and I need to actually make sorting work in the cars example!

david-vanderson:
(Edited "table" -> "grid")

I think I'm ready to do the switch.  Here's a rough upgrade guide:

Upgrade Guide:
* number columns implied by max column requested last frame
* number of rows is either implied or passed explicitly (when using `rowsVisible` to not run all rows)
* column widths now handled internally
  * expand if grid is expanded horizontally
  * shrink if grid does not have horizontal scrolling
  * cols_rigid for columns (like checkboxes) that don't expand/shrink
* resize_rows/resize_cols -> autoSize()
* row_height_variable -> row heights now always variable
* new `layout_only` for grid to only layout child widgets and skip keyboard navigation
* new `editable()` for static cell text that switches to editable text as needed (like a spreadsheet)
* gridHeading -> grid.colHeader + dvui.label
* gridHeadingSortable -> grid.colHeader + cell.headerSortable
* gridHeadingCheckbox -> grid.colHeader + dvui.checkbox
* gridHeadingSeparator gone, all colHeaders resize at right edge
* VirtualScroller -> rowsVisible
* KeyboardNavigation -> layout_only + tab_index partially solves this, file issue if needed
* banding/row hover -> usage code (see example)

david-vanderson:
I'm ready to merge this.  Will do it tomorrow unless anyone screams.  Can always revert or fixup as needed.

david-vanderson:
I've noticed that the "Styling" demo in Debug mode can drop to low framerates when scrolling around a lot.  It doesn't seem to affect Release modes, so my guess so far is that this is the first time we are doing a lot of textLayouts at the same time which is interacting especially badly with like the DebugAllocator?

Need to do some performance investigation, but not worth holding up merging.

PR #896, opened 2026-06-07
add support for text highlighting/selection to lablelWidget

youssless92:
Add support for text highlighting/selection to lablelWidget using selection fields of TextOptions for usage like in the attached video:

https://github.com/user-attachments/assets/f44f132f-9846-4d4d-a432-ecf7bc4122f9

PR #907, opened 2026-06-13
Fix ellipsis selection for label widget

youssless92:
Fix ellipsis selection for label widget, bug introduced in https://github.com/david-vanderson/dvui/pull/896


david-vanderson:
Haha - good catch.  Thanks!

PR #908, opened 2026-06-15
fancy icontheme stuff

KernelState:
I was trying to add it before with the stylescheme but there were alot of external problems outside of the stylescheme thing.
So, this adds Icon schemes without breaking nor replacing the old API with a toDecls function so the examples and everything can work, issue is that if you really want to just make the iconBrowser and stuff like that take in the IconTheme itself, have fun doing it yourself this is wayyy too messy it's like 50K lines of accumulated stuff that I WILL NOT try to swim through like last time, toDecls is currently quite redundant, I opened a zig issue about it to be able to automate it https://codeberg.org/ziglang/zig/issues/35776 . I did make it undefined fields by default as some icon themes simply do not have all these icons as I'm thinking we can try to load system IconThemes and stuff (I probably wont do it, but I want it to be possible, this is the main point of this for me).

**NOTE FOR REVEIWERS**: this looks like alot but it's just one file that's just entypo copy pasted twice with slight differences so it's all a pattern I literally just did vim find and replace for that.


david-vanderson:
I'm not sure what the goal of this is.  If the idea is to be able to swap in different icons for the stuff that dvui uses internally, then the IconTheme should only include icons we actually use.

KernelState:
this is mainly for configurability in your own applications instead of dvui itself, I don't want to replace all the entypo references because it's very error prone and the codebase is too big for such a drastic change without introducing application crashing bugs

Didas-git:
I don't see the purpose this serves.
A lot of built-in widgets who make use of an icon to show important information already provide a way to change the icon, and the ones who dont (like expander) should just add it to `InitOptions` instead to follow the already established convention.

An "icon theme" is not something that is feasible in any way, to cover every use case you would need well over 10k fields and you still wouldnt be able to cover use cases like using a dot instead of an arrow. Widget-defined configuration is the way for this and already available in some widgets and its not that different from having an "icon theme" only for what its used by dvui, the only difference is that they are attached to where they are actually needed/used, abstractions of this kind should be handled by the user instead (and really unless you want users to be able to change icons, unnecessary at best)

nat3Github:
i agree with the @Didas-git and @david-vanderson
in addition to this imo the pr lacks clear communication
example: title - add additional styleschema for icons, description could include what was done and why it was done (goal + intent). since this is bitrotting anyway i will close it.

note: it sounds a bit like you have an idea for improving dvui at the same time you're not fully sure about how to achieve it. perhaps opening an issue rather would be the right move.

PR #911, opened 2026-06-22
Initial proof of concept of supporting compiled plugins

foxnne:
Hello! 

Please have patience, this is all very new to me and I'm currently still exploring the options I have here, though I think after several days of experimenting, I think I at least know the features I'm looking to support:

### Goals
- [x] Ability to author plugins using DVUI in zig
- [x] Ability to update individual plugins without updating entire app
- [x] Ability to support a plugin "store" of 3rd party plugins

Fizzy concept:

```
        ┌─────────────────────────────────────────────────────────┐
        │                        Shell (Editor)                    │
        │  window · frame loop · menu/sidebar/panel layout · docs  │
        │                                                          │
        │   ┌──────────────┐        ┌──────────────────────────┐   │
        │   │     Host     │◄──────►│        EditorAPI         │   │
        │   │  registries  │  reach │ (shell read/util surface │   │
        │   │  + services  │  back  │  arena, folder, docs, …) │   │
        │   └──────┬───────┘        └──────────────────────────┘   │
        └──────────┼──────────────────────────────────────────────┘
                   │ register(host) + vtable calls
        ┌──────────┴───────────────┐        ┌────────────────────────┐
        │   workbench plugin       │        │   pixelart plugin      │
        │  file tree · tabs/splits │        │  canvas editor         │
        └──────────────────────────┘        └────────────────────────┘
         plugins never import each other — they meet only at the SDK
```

Idea is that we create a `proxy` backend whose only job is to forward calls to the host backend, removing SDL3 from being automatically linked in, which avoids conflicts with unifying SDL context when dispatching render calls (if each plugin has its own individual link to SDL3, `currentWindow.backend` is not unified between app and plugins, and plugins fail with invalid renderers (in my experience).

Instead, we still build using DVUI, but avoid linking SDL3 for the plugins, which then calls the rendering functions through a C-ABI bridge. 

Example of Fizzy running 3 plugins, one for laying out splits and tabs, one for rendering documents, and one for rendering images, in tabs together:

Screenshot 2026-06-22 at 11 46 08 AM

The questions I have are:

1. is there another way to accomplish these goals in a simpler way? I haven't been able to find one, but I'm very new to modularized software.
2. Would it be more helpful for if I could bring my own backend for DVUI and keep this out of it?


david-vanderson:
This is great!

> * is there another way to accomplish these goals in a simpler way? I haven't been able to find one, but I'm very new to modularized software.

You know more than I do - this seems like a promising approach.

> * Would it be more helpful for if I could bring my own backend for DVUI and keep this out of it?

I think it makes sense to add it to dvui?  That way each plugin only needs to reference upstream dvui with the proxy backend, right?  We might run into versioning issues at some point, but I think only if the backend interface changes?

In any case, I'm totally fine with you merging this and continuing to work on it in tree.  I'm pleasantly surprised that it doesn't touch much else, so go for it!

foxnne:
Awesome! Im glad no other changes were necessary :) I'll merge this soon

PR #913, opened 2026-06-28
System font fallback

youssless92:
This is my first attempt to implement an OS Font fallback mechanism to be able to display non latin characters/letters/codepoints.
For emojis I fallback to the added NotoEmoji font which does not seem to support all emojis.

Tested it on MacOS aarch64, Windows x64 and Linux aarch64 X11 (should work fine with Wayland) 
No WASM support

Shaping still need to be implemented for some languages through HarfBuzz (maybe?).
I wasn't able to do it yet because I found it very complex (for me at least).
So in the text layout example I added non Latin text and emojis to demo this PR and implemented a hack to display right to left languages correctly which could be a temp solution before using HarfBuzz.

NOTE: I used LLMs to help me especially in the extra c and objc files in backends. Therefore, I still need to review the code to be sure it doesn't need fixes or optimisations. I am still creating the PR because I don't know when I will be able to work on this for long enough and to be able to see what others think of this approach.

image


david-vanderson:
Thank you very much!  This is something I want to dig into, but unsure when I will get to it.  Thanks for blazing a trail here!

> Shaping still need to be implemented for some languages through HarfBuzz (maybe?).

The current plan is to use kb_text_shape, there is an initial unfinished PR #666 that I will eventually get back to.

youssless92:
Ah didn't know about kb_text_shape, i will look into it to see if I can make it work, although the 30k changes in PR #666 look terrifying 👹

But the idea of this PR was to try and implement a maybe naïve idea of letting the "OS do the work" which worked surprisingly well with minimal changes. 
Although I might have missed some use cases or broken something somewhere which I didn't run into in my tests but from my POV, this a great improvement even though incomplete.

PR #921, opened 2026-07-05
build: add web-serve steps for web examples

Luexa:
As requested on Discord, web examples can now be served locally without external dependencies such as Python or Caddy.

```
zig build serve-web-app -Dbackend=web
zig build serve-web-test -Dbackend=web
```

The web server will output something like this which can be clicked or ctrl+clicked to open the browser in many terminal emulators:
```
info: View the demo at http://localhost:8080
```

This did require adding a new argument to `buildBackend` and `addWebExample`, but it is ignored most of the time due to the lazy nature of the Zig build system. Without using a shared argument like this, each example that used the `serve_web_demo` tool might compile its own tool artifact instead of reusing the same one.

The cache buster tool is still run, but my implementation of the demo-serving tool will send [`ETag`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag) headers and respect [`If-None-Match`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-None-Match) request headers as well to allow the browser itself to query whether to re-cache each file. Additionally, for increased precision of the `performance.now()` API on most browsers, I [cross-origin isolate](https://developer.mozilla.org/en-US/docs/Web/API/Window/crossOriginIsolated) the served website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ Tested with Zig v0.16.0 (for Zig v0.15.2, use DVUI branch zig15 or [tag v0.4.0](
 
 ### Troubleshooting Web
 - To load examples for this backend, they must first be served through a (local) web server using:
+  - Zig `zig build serve-web-app -Dbackend=web`
   - Python `python -m http.server -d ./zig-out/bin/web-app`
   - Caddy `caddy file-server --root ./zig-out/bin/web-app --listen :8000`
   - Any other web server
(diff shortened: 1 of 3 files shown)


david-vanderson:
Excellent - thank you!

david-vanderson:
@Luexa I will credit you in the devlog as "Lue".  Let me know if you want something else!

PR #925, opened 2026-07-09
feat(sdl3): implement frame capture and texture readback

brmassa:
Implements `beginFrameCapture` / `endFrameCapture` on the SDL3 GPU
backend, providing a way to capture a dvui frame's full render output to
CPU-accessible RGBA8 pixels while still presenting the frame to screen.

The offscreen texture is created lazily at window size and reused across
captures. After `endFrameCapture` blits the captured texture back into the
swapchain image, the frame's command buffer is submitted and fenced so
that the subsequent download via `textureReadTarget` executes in the
correct order.

Also replaces the `textureReadTarget` no-op stub with a working
implementation using `SDL_CreateGPUTransferBuffer` / `SDL_DownloadFromGPUTexture`
with a dedicated command buffer and fence sync.

Intended for debug/tooling use (screenshots, regression testing), not hot
paths — each capture stalls on a GPU fence.

### DIFF
--- a/src/backends/sdl3gpu.zig
+++ b/src/backends/sdl3gpu.zig
@@ -1621,6 +1621,39 @@ pub fn textureReadTarget(self: *SDLBackend, texture: dvui.TextureTarget, pixels_
     const target: *BackendTextureTarget = @ptrCast(@alignCast(texture.ptr));
     const n = @as(usize, texture.width) * texture.height * 4;
 
+    // The commands that rendered into `texture` (e.g. via `dvui.Picture`)
+    // may still be sitting unsubmitted in `self.cmd` — rendering is batched
+    // and normally only submitted at end of frame. If we went ahead and
+    // downloaded now, our download command buffer (submitted below) could
+    // run before `self.cmd`'s commands: command buffers execute in
+    // submission order, not recording order, so the read could see stale or
+    // uninitialized GPU memory (this is what caused "Save png" in the Plots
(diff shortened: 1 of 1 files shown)


david-vanderson:

In the sdl3gpu-standalone demo > Plots section, it used to be that "Save png" would crash.  With this change it writes an image, but at least on my mac, the image is entirely magenta.  Also I get this on the console:
`2026-07-10 11:44:31.263 osascript[21020:28975566] Metal API Validation Enabled`

Not a blocker, but just wanted to ask - does "Save png" work for you?  If so what OS are you on?

brmassa:
Thanks for testing! Found it — you were right.

`textureReadTarget` was submitting its own command buffer for the download without first submitting whatever render commands were still batched (unsubmitted) in `self.cmd` — e.g. the commands from `dvui.Picture` that actually draw the plot into the offscreen texture. Command buffers execute in submission order, not recording order, so the download could race ahead of the render and read back stale/uninitialized texture memory. That's almost certainly what you were seeing — Metal's API validation layer fills untouched texture memory with a flag color (hence the solid magenta), whereas on Linux/Vulkan without validation layers it apparently got lucky.

Note: I'm on Linux (Vulkan/RADV), so I can't repro the Mac directly, but the race is real regardless of platform and this should fix it. Would appreciate it if you could re-test "Save png" in the Plots demo on your Mac when you get a chance.

david-vanderson:
Saving the plots demo now works as expected, thank you very much!  Merge it!

PR #766, opened 2026-03-06
coalesce draw calls with same texture/clip

david-vanderson:
Fixes #637 

While running the demo, this generally halves the number of backend.drawClippedTriangles calls.

Need some feedback about whether this actually addresses the performance issues.

PR #635, opened 2025-11-16
new backend: sdl3gpu + sample application sdl3gpu-ontop

peterino2:
My crack at https://github.com/david-vanderson/dvui/issues/233 now ready for review!

- sdl3gpu as a new backend enum, added -Dbackend=sdl3gpu as a switch
- followed imgui's strategy for embedding shaders into the exe, for more info: read buildShaders.py
- Tested on macos + Win64 + linux (bazzite)
- added sdl3gpu-ontop and sdl3gpu-standalone as example programs

TODO: 
- [x] SDL3 gpu standalone app
- [x] use SDL_SetGPUScissor instead
- [x] revert change to interleaving uploads. It's nicer if we dont have a command buffer acquired for the entire duration of begin()
- [x] why is perf so bad on my ser8 mini pc??? at least compared to the sdl3 renderer. (like 25 fps on the big demo window vs 300)
- [x] Bug with text rendering- it just looks... anemic and weird compared to the dx11 or sdl3 backend
- [x] implement create and read render targets
- [x] scrub sdl2 references from the sdl3gpu backend (being a copy-paste job from sdl3 there is a lot of littered references in there).
- [x] scrub for spurious comments/temporary test files
- [x] Rebase based MR, will squash noisy CLs before commiting.
- [ ] Investigate strange issues on intel Xe

PR #929, opened 2026-07-11
Batching draw calls

nat3Github:
on #766, #637, #635
Main problem:
Texture switches (fill→font→fill) are dominant for draw call explosion in dense scenes.
i.e. Button first draws Shape -> Triangles (no texture), draws text (font atlas texture), maybe another font (another texture)
next widget -> draws Shape (no texture).

this PR attempts to reduce this behavior by using global atlas with white texel.

changes
- (Window.zig, render.zig): per-window TriangleQueue, flushes on
  (texture,clip) change / render-target switch / endRendering. No backend API
  change.
- (Font.zig, render.zig, dvui.zig): one shared atlas (all fonts + a
  white texel) instead of one atlas per font. Untextured fills sample the white
  texel → share a texture with text → merge into one batch. Lazy rebuild, zero
  cost after warmup. this removes the fill -> font -> fill switches.

i tested cpu and gpu time in demo -> flat (too little geometry to matter), 
in a synthetic dense benchmark scene its ~10% faster on gpu and reduced 572 to 22 draw calls. 
fewer draw calls means fewer api overhead, but in practice its marginal in my testing.

note: batching is always on, so the checkbox in the demo has to go if this is merged.

what do you think? critique welcome

### DIFF
--- a/src/Examples.zig
+++ b/src/Examples.zig
@@ -786,6 +786,30 @@ test "DOCIMG demo" {
     //}
 }
 
+// ponytail: measurement-only, delete with pre_batch_calls at Phase-5 cleanup.
+// Run: zig build test -Dtest-filter="batching draw-call"
+test "batching draw-call measurement" {
+    var t = try dvui.testing.init(.{ .window_size = .{ .w = 800, .h = 600 } });
+    defer t.deinit();
+
+    dvui.Examples.show_demo_window = true;
+    dvui.Examples.show_widgetpedia_window = true;
+
+    const frame = struct {
+        fn frame() !dvui.App.Result {
+            dvui.Examples.demo(.full);
+            return .ok;
+        }
+    }.frame;
+
+    try dvui.testing.settle(frame);
+    const rs = dvui.currentWindow().renderStats();
+    std.debug.print(
+        "\n[batching] demo full frame: pre_batch={d} draw_calls={d} ratio={d:.1}x (tris={d} verts={d} tex_binds={d})\n",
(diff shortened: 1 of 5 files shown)


nat3Github:
actually after further testing the global atlas needs some reworking, right now it fully rebuilds which is to slow.

david-vanderson:
I'm also worried we might be duplicating effort at other levels of the rendering stack.  I saw when reading some SDL code that it was also batching calls to renderGeometryRaw before sending them to the gpu.

I don't know how often that kind of thing is happening in the different backends, but it would explain why our tests so far don't show larger performance differences.

nat3Github:
>  but it would explain why our tests so far don't show larger performance differences.

regarding batched calls the tests showed that the number of calls must be quite big to show differences. 
but maybe still worth it for users with complex scenes. even without this branch we are typicaly still in the 100s of batch calls for most guis and thats easy money for a modern gpu. so maybe the machine is also a reason for this.

> I'm also worried we might be duplicating effort at other levels of the rendering stack. I saw when reading some SDL code that it was also batching calls to renderGeometryRaw before sending them to the gpu.

yes dvui should take a stance on, what is backend implementation specific and what is not. 
as long as we have such a simple backend api i guess its all dvuis responsibility not the backends.

nat3Github:
update: 
i was away for a few weeks, but before that i experimented quite a lot on this, completing a single texture global font/glyph atlas implementation, actually reducing measured draw calls. 
but: 
- a lot of complexity was added
- it was unclear if performance was improved (or it was only tiny margin)
- unclear if single font glyph global texture beats multi texture font and probably depends on the load also

david-vanderson:
This is excellent information, thanks very much for the investigation!

PR #930, opened 2026-07-11
Allow proxy to refresh, expose layout hover_rect

foxnne:
I have a few changes as adding plugins, I think these are reasonable, please let me know if I need to make any changes.


david-vanderson:
For hover_rect, could we have a richer returned struct from addTextEx so you get it as a return value?

foxnne:
> For hover_rect, could we have a richer returned struct from addTextEx so you get it as a return value?

Okay! I made `addTextEx` return a struct containing the event and rect, would this be okay? Please let me know if you have other ideas.

Also, I went ahead and added the logic to trim text down before `toUtf8` here, I hope thats okay.

david-vanderson:
> Okay! I made addTextEx return a struct containing the event and rect, would this be okay? Please let me know if you have other ideas.

That looks great - thanks!

> Also, I went ahead and added the logic to trim text down before toUtf8 here, I hope thats okay.

Perfect. Merge this when ready.

I don't really understand the wake_requested thing. Can you please lay out the sequence of problematic events that this fixes? I'm good with it being merged but for my understanding. Thank you!

foxnne:
> I don't really understand the wake_requested thing. Can you please lay out the sequence of problematic events that this fixes? I'm good with it being merged but for my understanding. Thank you!

I had been trying to figure out a way for plugins to wake the app, and I still haven't been able to find a reliable way... Still experimenting though I removed the wake_requested code. The use case here is showing a "Just a moment..." tooltip over ZLS hovered terms, and then updating that tooltip when the actual response is retrieved. Currently, we HAVE to have an actual input to wake the app to update the tooltip, as the zig plugin is unable to request the app to refresh.

david-vanderson:
When you show "Just a moment..." does a background thread fire off to go do something?  When the response comes in (I assume on a background thread) calling backend.refresh doesn't work?

david-vanderson:
A stopgap solution is when showing "Just a moment..." fire a timer so you get a refresh in like 50-100ms

foxnne:
> A stopgap solution is when showing "Just a moment..." fire a timer so you get a refresh in like 50-100ms

This is what I'm going to do to solve this for now, just refresh until the request comes back on the other thread.

However, I do think I need to discover some way to wake the app, as it seems backend.refresh doesn't seem to cause the window to refresh (at least on macOS, I haven't tested this on Windows yet). I'll keep experimenting, but might just start a new PR for that. Thanks for the suggestions!

david-vanderson:
That sounds good for now.  If you need another pair of eyes, tell me how you are reproducing it and I'll look into it.

Merge the rest when ready.  Thank you!

foxnne:
I finally just created a minimal example and it seems that the bug looked like the app wouldnt be able to update from the plugin, but I was wrong. I created a minimal example plugin updating a popup and it all worked fine. So I will clean things up and make sure I have only whats necessary here, and then I'll merge :) all seems to work as expected now, sorry for the misinformation!

PR #931, opened 2026-07-12
feat: dockable panels: tabs, drag-and-drop, and floating panels

brmassa:
dvui has no way to let users rearrange panels — split, tab together, float, drag between areas — the way VSCode, Unreal, or Unity do.

A brief technical how-to:

- `DockLayout` is a pure-data binary tree of splits and tabbed leaves, with stable node slots (free-list, never compacted) so a `split_ratio` pointer or a panel's scroll state survives the tree being rearranged around it.
- `dvui.dockspace()` walks that tree, opening a `PanedWidget` per split and a tabbed header + content box per leaf, exposed to the caller as a simple pull loop (`while (dock.panel()) |p|`).
- Dragging a tab computes drop zones (center = new tab in same zone, edges = split that side) live against the mouse and resolves on release; everything (clicks, drags, closes) is queued as a mutation and applied once per frame so the tree stays immutable for the whole walk.
- Floating panels reuse `FloatingWindowWidget`.
- Persistence is a thin `writeJson`/`parseJson` pair on plain `std.json` — no opinion on *how* an app should save it, just a stable schema.

Potential missing features: no OS-level window detachment and no undo for layout changes.

We might adjust API based on review

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -682,6 +682,7 @@ pub fn buildBackend(
                 _ = addExample("sdl3-ontop", b.path("examples/sdl-ontop.zig"), true, example_opts, dvui_opts);
                 _ = addExample("sdl3-app", b.path("examples/app.zig"), test_dvui_and_app, example_opts, dvui_opts);
                 _ = addExample("sdl3-multi-win", b.path("examples/sdl-multi-win.zig"), true, example_opts, dvui_opts);
+                _ = addExample("sdl3-docking", b.path("examples/docking-standalone.zig"), true, example_opts, dvui_opts);
             }
         },
         .raylib => {
--- a/examples/docking-standalone.zig
+++ b/examples/docking-standalone.zig
@@ -1,8 +1,12 @@
-//! Standalone IDE-mock demonstrating `dvui.dockspace` with full layout
-//! persistence: the layout is loaded from `docking-example-layout.json` (next
-//! to the executable's cwd) on startup and saved back on quit, showing
-//! reviewers the complete save/load story `DockLayout.writeJson`/`parseJson`
-//! enable.
+//! Standalone IDE-mock demonstrating `dvui.dockspace`: drag tabs between
+//! panels, split, float, and reset back to the default layout.
+//!
+//! The "..." button after each tab strip shows `drawHeaderExtra` — dvui draws
+//! nothing in that trailing space, so the menu there is entirely this example's.
+//!
+//! Persisting a layout is the app's job: `DockLayout.snapshot` hands back a
+//! plain `Snapshot` value and `DockLayout.fromSnapshot` rebuilds one, so any
+//! serializer (JSON, ZON, msgpack, ...) works. The widget itself picks none.
 const std = @import("std");
 const dvui = @import("dvui");
 const SDLBackend = @import("sdl-backend");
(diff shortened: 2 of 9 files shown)


brmassa:
Note: I'm still implementing it in my game engine, Turian, and I might eventually need to implement a fix or feature. but its currently already working.

david-vanderson:
This is awesome!  I took a quick look and it's looking good so far.  I hope to go through it in depth soon!

brmassa:
added some small features about drag-and-drop and using the "empty tab area"

david-vanderson:
Good catch on the floatingTooltipWidget id_extra - thanks!

When I run `zig build sdl3-docking` and then click the "..." menu and select "Float Panel" (on any panel), I'm getting a crash:
```
thread 29923457 panic: reached unreachable code
/Users/dvanderson/apps/zig-aarch64-macos-0.16.0/lib/std/debug.zig:420:14: 0x103a70e1f in assert (sdl3-docking)
             ^
/Users/dvanderson/code/dvui/src/WidgetData.zig:366:21: 0x103b8c5a3 in validate (sdl3-docking)
    std.debug.assert(self.id != Id.undef); // Indicates a use after deinit() error.
                    ^
/Users/dvanderson/code/dvui/src/widgets/BoxWidget.zig:79:28: 0x103cbc5a7 in data (sdl3-docking)
    return self.wd.validate();
                           ^
/Users/dvanderson/code/dvui/src/Widget.zig:72:28: 0x103b8c4ef in data (sdl3-docking)
    return self.vtable.data(self.ptr).validate();
                           ^
/Users/dvanderson/code/dvui/src/dvui.zig:1129:45: 0x103b94003 in parentReset (sdl3-docking)
    const currentId = cw.current_parent.data().id;
                                            ^
/Users/dvanderson/code/dvui/src/widgets/DockingWidget.zig:539:21: 0x103de2a5b in deinit (sdl3-docking)
    dvui.parentReset(self.data().id, self.data().parent);
                    ^
/Users/dvanderson/code/dvui/examples/docking-standalone.zig:118:30: 0x103b77153 in main (sdl3-docking)
            defer dock.deinit();
                             ^
/Users/dvanderson/apps/zig-aarch64-macos-0.16.0/lib/std/start.zig:737:30: 0x103b7786b in callMain (sdl3-docking)
    return wrapMain(root.main(.{
```

Are you seeing this?

I'm also seeing a similar crash when dragging a tab outside any of the docking drop zones, like above the tab strip.

david-vanderson:
I'm also experimenting with having multiple `dockspace`, but I'm running into "duplicate tag" errors, and a crash when trying to drag a panel from one dockspace to another.

It looks like tags are primarily used for `onTabContextMenu`, is that correct?

brmassa:
you were right. Ive found other bugs too. I'm already doing some changes. probably later today

brmassa:
i did some extra changes. you might test it

david-vanderson:
> i did some extra changes. you might test it

Thanks!  Looks like that fixed the crashes.

Should we support multiple `dockspace` things in the same app?  I can imagine an app that has a bunch of static stuff and a few different places where docking stuff lives?

When I add a second dockspace it's close to working, but:
* I can't drag/drop between the two spaces
* I get "duplicate tag" error logs, like `error(dvui): duplicate tag name "hierarchy" id...`

brmassa:
David, nice observation. I did not designed to accomodate multiple dockspaces. It could be a future feature, for sure.

david-vanderson:
> David, nice observation. I did not designed to accomodate multiple dockspaces. It could be a future feature, for sure.

Sounds good.  With the addition to the demo window, do we need the `docking-standalone` example anymore?

Go ahead and merge this when you are ready.  Thank you!

brmassa:
i believe we can keep the docking-example while we do not address the multi-dockspace feature. it will serve as baseline project. I will later open a issue to document this future feature

PR #932, opened 2026-07-14
Add hover fading (helper + menu and button)

xsawyerx:
(I know this isn't fully complete - other widgets didn't get it, but I needed just this for now, so this is what I implemented. If you want me to implement for others, I can. I figured I'll submit it meanwhile.)

This introduces the ability to have hover fade helper, and I added hover fade ability for both ButtonWidget and MenuItemWidget. The commits have a better description.

It respects `reduce_motion` too.

### DIFF
--- a/src/Examples/styling.zig
+++ b/src/Examples/styling.zig
@@ -19,7 +19,13 @@ pub fn styling() void {
     }
 
     {
-        dvui.label(@src(), "Styling Buttons", .{}, .{});
+        {
+            var hbox = dvui.box(@src(), .{ .dir = .horizontal }, .{});
+            defer hbox.deinit();
+
+            dvui.label(@src(), "Styling Buttons", .{}, .{});
+            _ = dvui.sliderEntry(@src(), "hover fade: {d:0.2}", .{ .value = &dvui.hover_fade_secs, .min = 0, .max = 1.0, .interval = 0.01 }, .{ .min_size_content = .width(200) });
+        }
 
         var hbox = dvui.box(@src(), .{ .dir = .horizontal }, .{});
         defer hbox.deinit();
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -1975,18 +1975,17 @@ pub var hover_fade_secs: f32 = 0.12;
 /// Only valid between `Window.begin` and `Window.end`.
 pub fn hoverFade(id: Id, hovered: bool) f32 {
     const target: f32 = if (hovered) 1 else 0;
-    if (reduce_motion) return target;
+    if (reduce_motion or hover_fade_secs == 0.0) return target;
     // Default 0 (instead of target) so a widget with the first frame already hovered
     // (e.g. a context-menu item that pops up under the cursor) fades in rather
     // than starting fully lit.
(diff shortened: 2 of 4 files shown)


david-vanderson:
Just to double-check - for the menuItem, you are only looking to show the fade back after the mouse leaves it?  At least for now not trying to fade in the active color?

Otherwise, looks great - nicely done!

xsawyerx:
> Just to double-check - for the menuItem, you are only looking to show the fade back after the mouse leaves it? At least for now not trying to fade in the active color?

That was on me. I didn't even notice the fade-in was missing. Added it now.

Honestly, I'm not a big fan of how it came out... Maybe you would have a better idea on how to do it.

Do you want me to implement it for other widgets?

david-vanderson:
> That was on me. I didn't even notice the fade-in was missing. Added it now.

Thanks!

> Honestly, I'm not a big fan of how it came out... Maybe you would have a better idea on how to do it.

This is specifically for the menuItems right?  It looks pretty good to me on the buttons.  For the menuItems I kind of see what you mean at least for the themes "Adwaita Light" (which changes the font color) and "Gruvbox" (which has a totally different hover color than active).  Is that what you mean?

If so, unfortunately I don't have any great ideas.  Maybe we could immediately change the text color and only fade the fill?

But since the effect is quite short, I'm okay with going forward with it.  We can always tweak it later.

> Do you want me to implement it for other widgets?

Only if/when you need it.

david-vanderson:
Let's go with only fading the fill and see what people think.  Thank you!

Anyone reading this - this might need more tweaking later, please reopen or file a new issue if it causes problems for you.

xsawyerx:
Had a busy week so didn't get to this till now.

> > Honestly, I'm not a big fan of how it came out... Maybe you would have a better idea on how to do it.
> This is specifically for the menuItems right? It looks pretty good to me on the buttons. For the menuItems I kind of see what you mean at least for the themes "Adwaita Light" (which changes the font color) and "Gruvbox" (which has a totally different hover color than active). Is that what you mean?

Visually, I think it's good. I was thinking from a coding perspective. I'm still learning Zig and trying to find what "elegance" here means. :)

> > Do you want me to implement it for other widgets?
> Only if/when you need it.

I'll create a ticket so it can be tracked and see if I can put some time into aligning widgets on this. I don't like the idea that this is a capability limited to just a few widgets because I didn't need it in others (yet?). Maybe once all the widgets have it, a better coding pattern for it will emerge? I'll take a stab at it and let you know!

david-vanderson:
> Visually, I think it's good. I was thinking from a coding perspective. I'm still learning Zig and trying to find what "elegance" here means. :)

Got it.  I think the code looks decent, and it might be more of a dvui thing than a zig thing, so don't worry too much about it.

> I'll create a ticket so it can be tracked and see if I can put some time into aligning widgets on this. I don't like the idea that this is a capability limited to just a few widgets because I didn't need it in others (yet?). Maybe once all the widgets have it, a better coding pattern for it will emerge? I'll take a stab at it and let you know!

Sounds good.  There are a lot of widgets, and they probably don't all have the color infrastructure you need yet.  That part of dvui is still a work-in-progress (like many parts!).

PR #933, opened 2026-07-16
tabIndexDirection: directional focus navigation

### LINKED ISSUE #857: Gamepad Navigation Support
### Reason of the feature

Navigating an UI currently requires a mouse or/and basic keyboard. For applications/games running on consoles, handhelds (like the Steam Deck), or other platforms where is not common to see these peripherical , keyboard/mouse-only navigation creates a poor user experience or renders the UI completely inaccessible.
Different Backends have their own way of dealing with input, and DVUI should have some way to interact with it

---

### Proposal
Introduce a gamepad navigation system. Since DVUI is immediate mode, this usually has two stages:
1. Input State Feeding: A way for developers to pass input (including gamepad) each frame.
2. Directional Navigation: An internal system that handles focus shifting based on geometric proximity, layout proximity, Unique IDs, etc. (e.g., pressing D-Pad Down finds the closest widget in the current nearest geometry down, or the next in the same layout).

---

#### 1. Focus Navigation Logic

**Hierarchical Traversal**: Layouts are explicitly nested. Directional movement sequentially increments or decrements the active widget's structural index (Handled via an array?), filtering for visible/enabled components.

**Pros**: Incredibly fast and deterministic; navigation paths match the code declaration, Shouldn't allow spatial bugs to exists.

**Cons**: Strictly limits design to: basic grids, columns, and trees. Cannot naturally resolve focus in complex canvases.

**Geometric Proximity**: Focus relies on 2D visual screen coordinates. Interactive widgets flush their active IDs and Axis-Aligned Bounding Boxes to a spatial tree. Inputs trigger a raycast or angular distance calculation to find the closest target.

**Pros**: Total layout freedom. Supports overlapping canvases, diagonal grids, and offsets regardless of order.

**Cons**: This is expensive to calculate and there are possible edge cases that creates navigation loops or trap focus if bounding boxes align poorly on resolution dependent stuff.

**Explicit Graphic Routing**: Developers hard-code directional links directly inside the UI declaration using identifiers (e.g., moving down from ID_PLAY targets ID_OPTIONS).

**Pros**: Developers design the navigation experience. This bypasses previously mentioned limitations.

**Cons**: A LOT of maintenance. Resizing dynamic lists forces complex runtime ID re-evaluations, which enters directly into conflict with IMGUI paradigms.

#### 2. Analog Stick Processing

**Discretized Accumulator**: Treats analog deflection like a digital button passing through a multiplier. Signals below a deadzone are dropped, valid signals fire, then a time accumulator speeds up repeat intervals the longer the stick is tilted.

**Pros**: Snappy, grid precision. Makes lists responsive and prevents accidental multi-item scrolling.

**Cons**: Feels rigid and jagged when handling continuous visual components like sliders, color wheels, or certain styles of viewports.

**Frame-Independent Delta Impulse**: Analog input acts as a constant force vector integrated against the frame's delta_time. Tilt maps directly to velocity.

**Pros**: Exceptionally smooth performance for viewports, large document scrolls, or dragging sliders.

**Cons**: Difficult to accurately lock onto single line items in a menu unless snapping mechanics are implemented.

#### 3. Modal and Input Interception 

**Dedicated Intercept State Machines**: Widgets dynamically spawn internal listeners. When activated, they intercept incoming signals before they filter upward to the global parent, locking out main traversal until a condition is met.

**Pros**: Parent menus don't need external modal tracking flags.

**Cons**: Modifying input behaviors requires coupling with the rendering of specific widgets.

**Global Intercept**:  Widgets can have a global state lock. Normal routing is skipped, and all raw interaction events dump directly into the locked widget ID.

**Pros**: Unified API which applies identically to any widget (e.g simple buttons, complex modals, or full remap screens).

**Cons**: Makes the state to mantain and is more complex. We get lovely, deadlocked UI elements.

--- 

### References
- [SDL3 Gamepad system ](https://wiki.libsdl.org/SDL3/CategoryGamepad)
- [Gamepad DB for SDL](https://github.com/mdqinc/SDL_GameControllerDB), maintained list of a lot Gamepads systems, this is probably more on the backend, but is data 
- [Nuklear Gamepad](https://github.com/RobLoach/nuklear_gamepad), an extension for nuklear to handle gamepad natively
- [Nuklear Console](https://github.com/RobLoach/nuklear_console), Nuklear Console interface with Support for gamepad navigation

david-vanderson:
fixes #857 

The strategy is (edited):
* start from the center of the focused widget
* look in the given direction (arbitrary angle)
* find closest widget in that direction
  * distance in given direction weighted normally
  * distance in perpendicular direction weighted low when moving vertically, high when moving horizontally
* otherwise do nothing

I've added up/down/left/right key handling by default to do this.  But I ran into some issues that I'm unsure about:

Scroll areas already use up/down/left/right to scroll.  I modified it slightly so ScrollContainer ignores up/down/left/right if no scrolling is possible (all content fits in the viewport).

Focus Groups also use up/down/left/right for focus movement.  I modified it to use tabIndexDirection if `nav_key_dir` is null.  But you still can't escape a focus group via this.

We'll probably need to get more real usage experience to know how to proceed.


david-vanderson:
From discussion with @MichaelBelousov one usecase is a floatingMenu that has a bunch of expanders that each inside have a flexbox with some differently-sized widgets.  Need a way to navigate everything by arrow keys directionally.

david-vanderson:
@MichaelBelousov I think this is ready for merging.  It also includes a bunch of stuff to make `floatingMenu` work like a popup (when passed `.style == .popup`), after I looked at what you were doing to achieve this.  Hopefully after this you shouldn't have to explicitly handle the mouse clicks or checking for escape.

Can you try in this branch to use the demo > Layout > Directional Navigation > Popup and let me know if that is what you were thinking?

MichaelBelousov:
> Can you try in this branch to use the demo > Layout > Directional Navigation > Popup and let me know if that is what you were thinking?

Will do later today!

MichaelBelousov:
Superb work! Works intuitively and meets all criteria I was hoping to have!
I'm excited to integrate it!

The only thing I'd wonder is, and it's not even super relevant for me, is it trivial to customize the exact target for a given cardinal (left/right/up/down/next(tab)/prev(shift-tab)) direction?

One thing I found possibly unintuitive (although correct for my use case) was the four direction buttons (up/down/left/right) where settling on `up` and then clicking "down" brought you to `right` or `left` since they are vertically before `down`. Visually though they are not really "down".

So of course that is a different strategy which I think is more akin to maybe games, but I think could be solved with precise tab index control. So as long as that's easy, I think it works great.

Let me know if you have questions about that but I think this is a great feature, good default, and perfect for my case! Thank you!

david-vanderson:
> Let me know if you have questions about that but I think this is a great feature, good default, and perfect for my case! Thank you!

Awesome thank you!

> One thing I found possibly unintuitive (although correct for my use case) was the four direction buttons (up/down/left/right) where settling on `up` and then clicking "down" brought you to `right` or `left` since they are vertically before `down`. Visually though they are not really "down".

I agree.  My plan here is to wait for someone to be using it that kind of context and explore the options then.  I suspect it will involve both some tweaking/customization to the algorithm plus some way to directly control edge cases.

Anyone reading this - if you are in this situation please file an issue!

PR #941, opened 2026-08-02
Vulkan wio backend

### LINKED ISSUE #191: Vulkan backend
I'm eager to contribute it

Deins:
See: src/backends/render/vulkan/README.md

**Tested on**: linux (high end amd gpu, wayland env) and  windows 10 laptop (integrated intel gpu).  
Verified that validation layers reports no errors. But possible that on some hardware will still uncover bugs.

**Known issues / limitations** 
* the backend layer itself (`renderer/renderer/vulkan.zig`) which is responsible of doing vk context initialization only accommodates basic usage. There are too many options how to set it up. For anything advanced use .custom backend, include rendering as module  and fork and modify the bridge layer to your needs. At least some things that lack: 
  * ~Additional vk.queue selection/creation is not exposed. Useful for async data transfers or compute jobs, video encoding/decoding etc.~ (now through user handled device selection user can request which queues and how many it wants)
  * ~vulkan 1.3+ dynamic rendering is supported by renderer but not by bridge layer itself.~ (implemented)
  * currently everything is set up to directly render to window target over which there much less control - for hdr lighting, MSAA or specific color format anyhow  requires offscreen render target. In which case don't use provided option for depth buffer (its not used by dvui, just for convenience). Do your own `3d content -> renders to offscreen`, `offscreen texture -> rendered/copied to window target`, `dvui renders directly to window target`. To render dvui in custom target fork/modify to accomodate. 
* non vulkan related, on windows as usual resizing window freezes app and looks bad. (can be fixed with ugly resize callbacks later). 
* index/vertex buffers, image handles are pre-allocated, app must specify safe limits, draws above limit get ignored. dynamic growth currently not supported.  

Any suggestions / improvements / criticism  are welcome.


david-vanderson:
This is great!

I made the `wio-standalone` example have the same name when using vulkan, I assume that was an oversight?

Otherwise looks good to me - merge when ready.  Thank you!

david-vanderson:
Sorry should have said everything works on my linux machine and windows machine.  Well done!

src/backends/wio.zig
```diff
@@ -271,39 +271,46 @@ pub fn main(main_init: std.process.Init) !void {
     var events: wio.EventQueue = .empty;
     defer events.deinit();
 
-    const gl_options: ?wio.GlOptions = if (dvui.render_backend.kind == .opengl) .{
+    const gl_options = if (comptime dvui.render_backend.kind == .opengl) wio.GlOptions{
```
ypsvlq, line 274: I would leave this set to `null` for non-OpenGL, which is the default in `wio.CreateWindowOptions` anyway.

src/backends/wio.zig
```diff
@@ -271,39 +271,46 @@ pub fn main(main_init: std.process.Init) !void {
...
-    var window = try wio.Window.create(.{
-        .event_fn_data = &events,
-        .title = config.title,
-        .size = .{ .width = @trunc(config.size.w), .height = @trunc(config.size.h) },
-        .scale = 1,
-        .gl_options = gl_options,
-    });
+    } else {};
+
+    var window = if (comptime dvui.render_backend.kind == .opengl)
```
ypsvlq, line 280: The above change would make this unnecessary.

src/backends/render/vulkan.zig
```diff
@@ -0,0 +1,1445 @@
...
+    if (self.depth_images.len != 0) self.allocator.free(self.depth_images);
+    if (self.depth_memories.len != 0) self.allocator.free(self.depth_memories);
+    if (self.depth_layouts.len != 0) self.allocator.free(self.depth_layouts);
+    self.depth_image_views = &.{};
+    self.depth_images = &.{};
+    self.depth_memories = &.{};
+    self.depth_layouts = &.{};
+}
+
+fn initVulkanResources(allocator: std.mem.Allocator, window: *wio.Window, options: InitOptions) !VulkanResources {
```
ypsvlq, line 928: Eventually this should probably be refactored so it can work with the SDL/GLFW backends.

ypsvlq (commented):
Nice work! Some small comments but nothing merge-blocking, also tested working on my laptop.

I will also mention a "quirk" of Wayland where Vulkan rendering with vsync makes window resizes choppy unless presents are synchronised to the frame callback, which wio exposes through `window.enableDrawAvailableEvents()`.

src/backends/render/vulkan.zig
Deins: agreed, has to be seen whats vulkan common but not backend common and split. But will leave it for now until other backend actually gets implemented 😅

jedugsem:
if i do this:
```
    .target = target,
    .optimize = optimize,
    .backend = .wio,
    .render_backend = .vulkan,
});
```
with an up to date version it complains that render_backend is not and option

could it be that it was missed to wire this up, because the examples work a diffrent way?
i get:
```
error: invalid option: -Drender_backend
```

ypsvlq:
@jedugsem `.@"render-backend"` (I think this should be changed).

Deins:
pushed fix - renamed it to just `renderer`

PR #942, opened 2026-08-02
fix: specify opengl internal texture format as rgba8

agavistas:
this fixes issues for anyone attempting to use more modern opengl features on dvui textures. we already use u8 arrays, so this matches current behavior

### DIFF
--- a/src/backends/render/opengl.zig
+++ b/src/backends/render/opengl.zig
@@ -295,7 +295,7 @@ pub fn renderTarget(self: *@This(), maybe_texture: ?dvui.TextureTarget) !void {
 }
 
 fn createTexture(pixels: ?[*]const u8, options: dvui.Texture.CreateOptions) !u32 {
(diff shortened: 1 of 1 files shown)


src/backends/render/opengl.zig
```diff
@@ -295,15 +295,15 @@ pub fn renderTarget(self: *@This(), maybe_texture: ?dvui.TextureTarget) !void {
 }
 
 fn createTexture(pixels: ?[*]const u8, options: dvui.Texture.CreateOptions) !u32 {
-    if (options.format != .rgba_32) {
+    if (options.format != .RGBA_32) {
```
david-vanderson: This doesn't compile for me, was this change intended?

agavistas: It seems my ZLS instance decided to change that for some reason? Was not intended.

david-vanderson:
Thank you!

PR #955, opened 2026-08-14
Text break Improvements

nat3Github:
- word breaking: break on hyphen/slash in addition to space, and collapse trailing whitespace at wrap points so the next line doesn't start with leading spaces.
- chines-japanese-korean-aware keep-with punctuation: don't start a wrapped line with closing punctuation (、。」) or end one with opening punctuation (「（), matching how text is conventionally wrapped.
- test against a subset of Unicode's LineBreakTest.txt

### DIFF
--- a/src/widgets/TextLayoutWidget.zig
+++ b/src/widgets/TextLayoutWidget.zig
@@ -1238,12 +1238,7 @@ const AddTextExAction = enum {
     hover,
 };
 
-// ponytail: covers the common CJK punctuation blocks (CJK Symbols and
-// Punctuation, fullwidth forms, general typographic quotes), not a full
-// UAX #14 "keep with" class table. CJK ideographs/kana/hangul themselves
-// need no special-casing here: the width cutoff above already breaks
-// between any two codepoints when no space/hyphen/slash is found, which is
-// exactly "break anywhere" for scripts with no interword spacing.
+// NOTE: common CJK punctuation blocks only, not the full UAX #14 class table
 fn isCjkClosingPunct(cp: u21) bool {
     return switch (cp) {
         0x3001,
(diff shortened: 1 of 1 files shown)


nat3Github:
note: whats still missing is correct breaking behavior for emojis glyphs. i have that in a seperate branch but it's not really relevant currently.

nat3Github:
edit: 
@ reviewers: closing this PR, since i am working on a much bigger / broader solution to font engine related issues. 

my verdict is that all font related issues in dvui (arabic/asian lang support, correct shaping, font discovery, bidi, text breaking, styling, var fonts, emoji fonts) cover such an ambious spectrum that it would be more efficient combining effort and outsourcing this into a seperate library.
the alternative is adding it piece by piece causing a lot of churn, managing overhead and have low locality of behavior.

david-vanderson:
Sounds good.  I am hoping that kb_text_shape (see #666) gives us shaping, bidi, breaking, and part of emoji support, but haven't gotten that far yet.

nat3Github:
i am working on a lib with similar scope to kb_text_shape but its a bit more ambitious, will try to integerate / co develop with dvui when its ready, i think it will solve a lot of issues

david-vanderson:
> i am working on a lib with similar scope to kb_text_shape but its a bit more ambitious, will try to integerate / co develop with dvui when its ready, i think it will solve a lot of issues

Sounds awesome!

One thing I've been trying to wrap my head around is how to achieve the current dvui `textSizeEx` semantics.  We want to pass a large amount of text, but only measure/layout until we get to a max width (either for soft line wrap or to find where the mouse click was).  This is easy currently, but I'm not sure how to do it when adding another byte can shorten the total width (either a ligature or some other combination).

kb_text_shape doesn't have that kind of api - you have to pass all the bytes up front (unless I'm totally missing something).

As you are investigating this, any information on how to do this in a streaming fashion would be super useful!

Issue #923, opened 2026-07-08
"Modern" DVUI look - gradients, smooth curves, blur and shadows!


nat3Github:
more thoughts on graphics:
related: #637, #120, #699 

lately my experiments with dvui included custom widgets, tinkering with tvg rendering without z2d dependency / textures -> straight triangles.

<img width="666" height="446" alt="Image" src="https://github.com/user-attachments/assets/231ff51a-35df-4480-98cd-c9e9a1a93a65" />

-----------
1.  Gradient fills, Smooth Curve Helpers (im working on this, #864, #972 )
- merge tvg-specific path into a shared ColorSource/gradient helper both Path and render_tvg call vendored inside dvui.
- Unlocks gradient buttons/panels, shadow/glow rings, gradient borders.

2. Real blur pass (large radius / backdrop-filter,  #956 )
  - Needs backend support, somewhat doable on cpu / backend independent but not pretty
  - unlocks frosted glass panels, real drop shadows on floating elements.

3. Blend modes (havent looked into it probably not worth the effort)
  - Needs backend support
  - unlocks (glow/light effects, rare in modern UI). my weather app used it (currently falling back to alpha layering)

4. Arbitrary/rounded clip (stencil) (wont implement myself)
  - Needs backend support
  - specific but cool usecases
  - probably not worth the effort

-----------------

DVUI Goals (Not specific to this issue but important to guide design decisions):

Portability 
- use dvui anywhere web, desktop, mobile, audio plugins (im using dvui on an iplug2 branch)
- the gui should look the same on each backend
- efficient implementation to be able to run on lower end devices like mobile, PIs, embedded

Nice Graphics
- fidelity / quality, make defaults look good.
- customizable look / style, provide the building blocks for users to build unique looking tailored applications
- (G)UI can have many forms, id love to build 2D games with dvui or use it for all kinds of data visualization.

Ecosystem
- Widget Libraries, vendored by OSS users for different domains
- Icon, Asset libraries
- big palette in dvui itself is nice, but i think for new users overwhelming to create new widgets

Stability
- developement out of the box, stable backends
- ultimately stability fewer changes / updates / codebase churn

david-vanderson:
>     1. Gradient fills, Smooth Curve Helpers (im working on this)

I would love `Path` helpers to add various curves.  Just being able to stroke smooth curves would be a great addition.

This is wonderful to work towards - thank you!

[...]

PR #956, opened 2026-08-15
Feature/blur backdrop

nat3Github:
adds backdrop blur to dvui.
image

i tested in in the web browser (the serve-web-app is great for that!), sdl3, raylib, sdl2
its worth to note that the blur in my testing looked different for sdl2 then the other backends.

sdl2 skews/compresses a quad's UV mapping rather than clamping the edge resulting in a visually different blur.

#923


david-vanderson:
Sorry I am traveling right now, will get to this in a few days, thanks!

david-vanderson:
I love this!
* I'm surprised by the lifecycle of `BlurBackdrop`.  That kind of "store the object in the datastore, it will be automatically collected" isn't something I had thought of before.  When it is garbage collected, the texture is as well right?  If true, `deinit` is not needed?
  * I haven't been able to figure out the right shape for stuff like this - `Picture` has the same problem, and maybe this is the solution.
* The reason you can't use `Picture` is because the drawing might go across the boundary right?  I think your method of recording which render commands is good, sorry it's so fiddly.  If we see another usecase that needs to record/replay render commands we can make a better api.

david-vanderson (approved):

foxnne:
This is awesome, well done!

nat3Github:
> When it is garbage collected, the texture is as well right? If true, deinit is not needed?

thanks for the feedback! afaik yes, but deinit is still needed. actually its wired up false in the example and leaks.
i modified the api to make it more in line with what we already have, should be good to go now


> The reason you can't use Picture is because the drawing might go across the boundary right? 

i am not 100% sure if you could achieve something similar with Picture, but boundary crossing content was the issue.

new api:
```zig
...
const backdrop = dvui.BlurBackdrop.get(@src());
...
backdrop.init(some_rect, .{backdrop.radius_px});
defer backdrop.deinit();
```

david-vanderson:
Excellent, I will chew on that new api and see if `Picture` should do the same thing.

Can you move the new demo part into the "Applets" section?  Then merge it.  Thank you!

Issue #883, opened 2026-05-29
Child wins : Prevent accidental new window fork bomb


Yinameah:
Extracting this particular concern from [this discussion](https://github.com/david-vanderson/dvui/pull/878#issuecomment-4568788009) in #878.

At time of writing, my proposal is to leave a `preventWindowForkBomb()` call inside `initWindowSecondary` that puts a hard limit of 6 OS window, preventing the very unpleasant experience of crashing your computer because of a bad infinite loop of new window popping up.

The plan would be to re-asses this when the child OS window feature goes out of experimental realm.

Yinameah:
For sake of searchability :  

Details about the failure mode : 

> > Thinking about preventWindowForkBomb - can you describe the situation that led to it? Did you end up with an infinite loop? Or something that spawned a new window each frame that was never cleaned up?
> 
> Both. I had that a few times in the beginning trying to find the right way to reset / next / next_peek the trackingHashMap. Then I accidentally called osWindow nested in a loop with a flag that was never set. I also had the case where you have a new window created and destroyed on each other frame and it keeps stealing the focus so you cannot select the "primary" to shut it down. Ah, and one where I belive zls was actually popping up windows all over the place via build_on_save cause I mis-configured something. 😓 I don't think it will be a big issue post experimental phase, but it's annoying enough it's worth considering to leave some guardrails here and there.

and idea about how to tackle this :

> One idea I just had is maybe we could store a timestamp each time `initWindowSecondary` is called, fifo style, let's say the 5 last calls, and we monitor the elapsed time between first and last. If it's below a certain threshold, we just pause for a few seconds and emit a log info. That way if it's legitimate it appear as just a few second lagging, but if it's a bug, it give you time to select the first window and kill it, or maybe even toggle whatever was triggering the window cascade and in the best case even recover.

david-vanderson:
I didn't think about opening so many windows that it screws with the rest of the OS.  I figured just the app would die.

Yeah in that case it makes sense, because we can turn a ton of windows opening into a ton of floatingWindows, which still allows the user to kill the app, like you said.

For that I suggest a global var for the total number of windows, where the default is like 20.  When you try to open the 21st, you get a floatingWindow and logs an error that says how to bump the global var.  Dealing with timing stuff is quite tricky, so let's try a simple total first, easy to test.  It also gives an escape hatch for someone who needs it.

Yinameah:
[...]

PR #959, opened 2026-08-16
Multi OS window : allow configuration of max number of windows

Yinameah:
See #883.

Took the opportunity to make the FloatingWindow fallback dynamic, turn out to be easier than expected.
Moved the check from the backend to the DVUI part so it will cover other backends as well.


david-vanderson:

PR #961, opened 2026-08-16
web: extract shared code into web-common.js

Fristender:
## This refactor is the prerequisite for #962 
Moves all code shared between the in-page backend (`web.js`) and the upcoming Worker backend into a new ES module `web-common.js` (~1060 lines): shaders, `WebRenderer` base class (GL setup, textures, render geometry, import-object builder), and event helpers (`encodeModifiers`, `touchIndex`, `WheelHandler`, `HiddenInputManager`, `getTouchCoords`, `dvui_fetch`). `web.js` shrinks from ~1300 to ~620 lines and now just holds the app-specific `Dvui` subclass.

## Review guide

- [`web-common.js:1-95`](https://github.com/david-vanderson/dvui/blob/58394dda17a7b2fe7616e9cf6d12c698da803a46/src/backends/web-common.js#L1-L95) — shader sources + `dvui_fetch`, moved verbatim; check no drift vs. old `web.js`.
- [`web-common.js:96-150`](https://github.com/david-vanderson/dvui/blob/58394dda17a7b2fe7616e9cf6d12c698da803a46/src/backends/web-common.js#L96-L150) — string/UTF helpers and `encodeModifiers`/`touchIndex`.
- [`web-common.js:400-550`](https://github.com/david-vanderson/dvui/blob/58394dda17a7b2fe7616e9cf6d12c698da803a46/src/backends/web-common.js#L400-L550) — `WebRenderer` class; [`buildImports()`](https://github.com/david-vanderson/dvui/blob/58394dda17a7b2fe7616e9cf6d12c698da803a46/src/backends/web-common.js#L442-L476) — verify the import list is the union both backends need.
- [`web-common.js:1050-1085`](https://github.com/david-vanderson/dvui/blob/58394dda17a7b2fe7616e9cf6d12c698da803a46/src/backends/web-common.js#L1050-L1085) — default no-op host functions; `web.js` overrides the ones it implements.
- [`web.js:5`](https://github.com/david-vanderson/dvui/blob/58394dda17a7b2fe7616e9cf6d12c698da803a46/src/backends/web.js#L5) — imports from `./web-common.js`; [`web.js:58`](https://github.com/david-vanderson/dvui/blob/58394dda17a7b2fe7616e9cf6d12c698da803a46/src/backends/web.js#L58) — `class Dvui extends WebRenderer`. Everything below should be override-only; confirm no duplicated logic remains.
- [`build.zig:1641-1642`](https://github.com/david-vanderson/dvui/blob/58394dda17a7b2fe7616e9cf6d12c698da803a46/build.zig#L1641-L1642) — `web-common.js` added to install set for `web-test`/`web-app`.
- Verify: `zig build web-app` still serves and renders identically (pure refactor, no behavior change intended).

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1615,6 +1615,8 @@ fn addWebExample(
     compile_step.dependOn(&b.addInstallFileWithDir(output, install_dir, "index.html").step);
     const web_js = b.path("src/backends/web.js");
     compile_step.dependOn(&b.addInstallFileWithDir(web_js, install_dir, "web.js").step);
+    const web_common = b.path("src/backends/web-common.js");
+    compile_step.dependOn(&b.addInstallFileWithDir(web_common, install_dir, "web-common.js").step);
     b.addNamedLazyPath("web.js", web_js);
     compile_step.dependOn(&install_wasm.step);
     compile_step.dependOn(&install_noto.step);
--- a/src/backends/web-common.js
+++ b/src/backends/web-common.js
@@ -1,0 +1,1059 @@
+/**
+ * Fetch a url and return its body as bytes.
+ * @param {string} url
+ * @returns {Promise<Uint8Array>}
+ */
+export async function dvui_fetch(url) {
+    let x = await fetch(url);
+    let blob = await x.blob();
+    //console.log("dvui_fetch: " + blob.size);
+    return new Uint8Array(await blob.arrayBuffer());
+}
+
+export const vertexShaderSource_webgl = `
+    precision mediump float;
+
+    attribute vec4 aVertexPosition;
+    attribute vec4 aVertexColor;
+    attribute vec2 aTextureCoord;
+
+    uniform mat4 uMatrix;
+
+    varying vec4 vColor;
+    varying vec2 vTextureCoord;
+
+    void main() {
+      gl_Position = uMatrix * aVertexPosition;
+      vColor = aVertexColor / 255.0;  // normalize u8 colors to 0-1
+      vTextureCoord = aTextureCoord;
+    }
+`;
+
+export const vertexShaderSource_webgl2 = `# version 300 es
+
+    precision mediump float;
+
+    in vec4 aVertexPosition;
+    in vec4 aVertexColor;
+    in vec2 aTextureCoord;
+
+    uniform mat4 uMatrix;
+
+    out vec4 vColor;
+    out vec2 vTextureCoord;
+
+    void main() {
+      gl_Position = uMatrix * aVertexPosition;
+      vColor = aVertexColor / 255.0;  // normalize u8 colors to 0-1
+      vTextureCoord = aTextureCoord;
+    }
+`;
+
+export const fragmentShaderSource_webgl = `
+    precision mediump float;
+
+    varying vec4 vColor;
+    varying vec2 vTextureCoord;
+
+    uniform sampler2D uSampler;
+    uniform bool useTex;
+
+    void main() {
+        if (useTex) {
+            gl_FragColor = texture2D(uSampler, vTextureCoord) * vColor;
+        }
+        else {
+            gl_FragColor = vColor;
+        }
+    }
+`;
+
+export const fragmentShaderSource_webgl2 = `# version 300 es
+
+    precision mediump float;
+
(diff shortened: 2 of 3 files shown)


david-vanderson:
@Fristender please be patient, I will get to this but it might be a week or two.

@MichaelBelousov Can you read this PR description and tell me if splitting the js across multiple files might cause any problems?

MichaelBelousov:
Ofc, will take a look later!

MichaelBelousov:
@david-vanderson I can take over as reviewer for this if you are Ok with that

david-vanderson:
> @david-vanderson I can take over as reviewer for this if you are Ok with that

That would be amazing!  This PR is not supposed to change any behavior, just a refactor.

MichaelBelousov:
Unrelated, @Didas-git it looks like #706 broke web.d.ts for some users (like myself). I didn't realize because I'm still behind on a version (and I don't always run the typecheck rn lol). The web.d.ts should contain the list of public functions from web.js. Why was the exported interface removed? I am a user of that interface.

On second thought, let's move discussion to that PR.

MichaelBelousov:
before I do a complete review, @Fristender, do you have a strong reason that web-standalone can't just (at least initially) just import web.js and reuse exported shaders and etc?

I don't mind the refactor and I know you already put work into this, but I'd generally prefer we do invasive changes as progressively as possible. In this case, separate the `HiddenInputManager` and `WheelHandler` logic _inside_ web.js before we move it to a separate web-common.js. Would you be ok with adding a third PR before this one that does that, so this is easier to review?

MichaelBelousov:
FWIW I can review it in this state but I will be slow. It'll be faster if you split up your changes into more well-scoped PRs that make diffing as usable as possible. I know you already split it once, but it's valuable to do these things in chunks we can reason about.

Fristender:
> before I do a complete review, @Fristender, do you have a strong reason that web-standalone can't just (at least initially) just import web.js and reuse exported shaders and etc?

The idea is to preserve `web-app` only implementation details inside `web.js` and move things that `web-app` and `web-standalone` could share into `web-common.js`. That way it's easier to differentiate between shared and `web-app` only implementations.


> I don't mind the refactor and I know you already put work into this, but I'd generally prefer we do invasive changes as progressively as possible. In this case, separate the `HiddenInputManager` and `WheelHandler` logic _inside_ web.js before we move it to a separate web-common.js. Would you be ok with adding a third PR before this one that does that, so this is easier to review?

I am okay with this, working on this now. Also, do you guys have some tests that verify every feature works? Doesn't have to be code, could be a series of steps and the expected behavior. That way I can verify the refactor doesn't have regressions to save time.

MichaelBelousov:
> Also, do you guys have some tests that verify every feature works? Doesn't have to be code, could be a series of steps and the expected behavior. That way I can verify the refactor doesn't have regressions to save time.

@david-vanderson can you comment? My understanding is there are only some zig tests and the demo window. 

It would be cool @david-vanderson if we had a pixel comparison threshold test for several backends, but I'm sure there are a number of problems with that. Graphl does have something like that with a screen shot diff test gh action. I should open it up

david-vanderson:
> > Also, do you guys have some tests that verify every feature works? Doesn't have to be code, could be a series of steps and the expected behavior. That way I can verify the refactor doesn't have regressions to save time.
> 
> @david-vanderson can you comment? My understanding is there are only some zig tests and the demo window.

That's correct.  I've been hesitant to add too many tests given the high rate of feature changes, but that does seem to be slowing down, so maybe we are transitioning to a more test-friendly phase.

Usually I click through the demo as a smoke test.  For the web, I do it both on a PC and then on a phone (more limited).

But I can totally do that - if you two can get the code in a shape that looks good I can do manual regression testing.

> It would be cool @david-vanderson if we had a pixel comparison threshold test for several backends, but I'm sure there are a number of problems with that. Graphl does have something like that with a screen shot diff test gh action. I should open it up

We tried that at the very beginning of the testing infrastructure but couldn't figure out how to make it work.  I still think there is some kind of way here - like limited widget-tree hashing for normal tests, and specialized pixel comparison tests for low level stuff.  But unsure how to proceed.

PR #963, opened 2026-08-19
featured projects: replace webmaker2000 with zorts

nhanb:
The former is semi-abandoned and very outdated, so I'm replacing it with a more up-to-date project that actually compiles on zig 0.16.


david-vanderson:
Thank you!

PR #964, opened 2026-08-19
Fix C based backends in 0.17.x branch

AndrewKraevskii:
fix regression from #958
Old translate c inited C fields to 0s by default. Some of examples and backends relied on this behavior by omitting some of fields.
I didn't check all examples but it fixed sdl3gpu-standalone at least.

I didn't set dvui.h to use default_init since it doesn't really have much of structs in it.


david-vanderson:
Thank you!

PR #965, opened 2026-08-22
Refactor and fix some fade stuff

xsawyerx:
The commits themselves include the details, but basically:

Commit 1: Refactor so we could fade multiple items in a widget, and to have different speeds amongst them and other widgets.

Commit 2: Fix a calculation mistake for the hover when the app is idle.

### DIFF
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -1944,8 +1944,9 @@ pub fn animationGet(id: Id, key: []const u8) ?Animation {
 /// How long a hover fade (see `hoverFade`) takes to reach full intensity.
 pub var hover_fade_secs: f32 = 0.12;
 
-/// Lets a hover wash fade in and out (instead of snapping the frame at once).
-/// Call at most once per widget per frame.
+/// The longest a single `stateFade` step may count for.
+const state_fade_max_step_secs: f32 = 1.0 / 30.0;
+
 ///
 /// Only valid between `Window.begin` and `Window.end`.
 pub fn stateFade(id: Id, key: []const u8, on: bool, secs: f32) f32 {
@@ -1956,7 +1957,8 @@ pub fn stateFade(id: Id, key: []const u8, on: bool, secs: f32) f32 {
     // in rather than starting fully lit.
     const cur = dataGetPtrDefault(null, id, key, f32, 0);
     if (cur.* != target) {
-        const step = currentWindow().secs_since_last_frame / hover_fade_secs;
+        const dt = @min(currentWindow().secs_since_last_frame, state_fade_max_step_secs);
+        const step = dt / secs;
         cur.* = if (cur.* < target) @min(target, cur.* + step) else @max(target, cur.* - step);
         refresh(null, @src(), id);
     }
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -1947,7 +1947,7 @@ pub var hover_fade_secs: f32 = 0.12;
 /// The longest a single `stateFade` step may count for.
 const state_fade_max_step_secs: f32 = 1.0 / 30.0;
 
-///
+/// Call at most once per widget per frame.
 /// Only valid between `Window.begin` and `Window.end`.
 pub fn stateFade(id: Id, key: []const u8, on: bool, secs: f32) f32 {
     const target: f32 = if (on) 1 else 0;
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -1941,7 +1941,7 @@ pub fn animationGet(id: Id, key: []const u8) ?Animation {
     return currentWindow().animations.get(h);
 }
 
-/// How long a hover fade (see `hoverFade`) takes to reach full intensity.
+/// How long a state fade (see `stateFade`) takes to reach full intensity.
 pub var hover_fade_secs: f32 = 0.12;
 
 /// The longest a single `stateFade` step may count for.
--- a/src/dvui.zig
+++ b/src/dvui.zig
@@ -1948,19 +1948,23 @@ pub var hover_fade_secs: f32 = 0.12;
 /// Call at most once per widget per frame.
 ///
 /// Only valid between `Window.begin` and `Window.end`.
-pub fn hoverFade(id: Id, hovered: bool) f32 {
-    const target: f32 = if (hovered) 1 else 0;
-    if (reduce_motion or hover_fade_secs == 0.0) return target;
-    // Default 0 (instead of target) so a widget with the first frame already hovered
-    // (e.g. a context-menu item that pops up under the cursor) fades in rather
-    // than starting fully lit.
-    const cur = dataGetPtrDefault(null, id, "__hover_fade", f32, 0);
+pub fn stateFade(id: Id, key: []const u8, on: bool, secs: f32) f32 {
+    const target: f32 = if (on) 1 else 0;
+    if (reduce_motion or secs <= 0.0) return target;
+    // Default 0 (instead of target) so a widget with the first frame already in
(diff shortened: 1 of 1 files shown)


david-vanderson:
Looks pretty good.

I don't quite understand the need for the 1/30s cap.  The refresh should be generating continuous frames.  But maybe you mean a spurious pause, like the the whole machine froze for half a second?  I'm unsure what the "right" thing to do there is.  In one sense we "missed" the fade.  But if I set the fps to like 4 fps, then it looks weird that the fade takes so long.  But we also don't want this to be too complicated.

Can you give more details about the focus ring fade issue?  Maybe there is something else going on?

xsawyerx:
> I don't quite understand the need for the 1/30s cap

This entire thing gave me a headache. :D

Here's my thinking: The problem is the idling. If we render frame 1 and then idle for 2 seconds and *then* click a button, the time in between is 2 seconds, so due to idle from frame 1, we get no fade (if the progress is increments of `elapsed / 0.12`, then `2.0/0.12` is way past 1.0, so it just sees it as "paint the end state", fully lit.

The refresh does generate continuous frames but I don't think it works in this case. Frame 1 was already done and painted. `refresh()` runs during the fade but since frame 1 was already rendered, there's no `refresh()` call. The fade hasn't started yet because we haven't pressed the button yet. By the time we press it, the diff in time from frame 1 and frame 2 is enough to fully paint and not fade in.

Instead, we take the either the real elapsed time or `1/30s`, so a 2 second gap counts as 0.033s, the same as any ordinary frame.

> But maybe you mean a spurious pause, like the the whole machine froze for half a second?

This wasn't *the reason* but it's also fixed by this.

> But if I set the fps to like 4 fps, then it looks weird

Fair. At 60 fps, it's 0.0166... and that's under `1/30` so that's OK. At 30 fps it's exactly that. At 4 fps... crappy. I figured isn't likely to happen, but I really have no experience and I figured it would stay at very high fps so it should be okay.

Maybe we should try and only fix the first frame:

```zig
const elapsed = currentWindow().secs_since_last_frame;
const dt = if (cur.* == 0 or cur.* == 1) @min(elapsed, state_fade_max_step_secs) else elapsed;
```

A fade is only not active at exactly `0` or `1`. The app can only sleep when every fade is not active, because a running fade calls `refresh()` every frame and that prevents sleeping. So "this fade is at not active" and "the previous gap might have been an idle sleep" are basically the same condition.

Not sure it's worth it, but arguably it would not redefine `hover_fade_secs` as "seconds, multiplied by 30/fps when it's below 30fps".

What do you think?

> Can you give more details about the focus ring fade issue? Maybe there is something else going on?

The focus ring fade always starts on a "bad frame" (that "frame 1" above). That is hard to see in practice because a mouse creates a slew of mouse events anyway.

The focus ring doesn't have a fade out, but that's another problem not covered here, and maybe we shouldn't do it.

david-vanderson:
> This entire thing gave me a headache. :D

Haha!

Okay I think I see what you mean, but I'm still not quite getting it.  A hover fade should always start on a frame where something changed (the mouse moved on a widget, or the focus moved to a new widget).  When we render that frame I expected the fade t=0, and call refresh.  Only on the next frame should we use the time elapsed.

Alright typing that out I see that would require the code to be a bit more complicated, we need to know if this is the first frame or not.

Does that make sense?  Is there a reason we can't treat the first frame specially?

If that works, it means at a low framerate you never see the fade, because t goes from 0 on the first frame to 1 on the next frame.  I think that makes sense?

xsawyerx:
Yes, that makes sense. It'll be better.

> Is there a reason we can't treat the first frame specially?

No, we just have to store one bit.

I'd hoped to spot the first frame by checking whether `t` is still exactly `0` or `1`, since a fade only rests there. But that only works if the first frame moves `t`. If it leaves `t` at `0`, then the second frame looks identical to the first, takes `dt=0` again, and the fade never moves. Infinite loop. :(

So we store the previous `on` next to `t` and compare:

```zig
const FadeState = struct { t: f32 = 0, on: bool = false };

const s = dataGetPtrDefault(null, id, key, FadeState, .{});
const changed = s.on != on;
s.on = on;

if (s.t != target) {
    if (!changed) {
        const step = currentWindow().secs_since_last_frame / secs;
        s.t = if (s.t < target) @min(target, s.t + step) else @max(target, s.t - step);
    }
    refresh(null, @src(), id);
}
return s.t;
```

4 extra bytes per fade, and the 1/30 constant goes away.

> at a low framerate you never see the fade [...] I think that makes sense?

Agreed. If a frame is longer than the fade there's nothing to render, so snapping is perfectly fine. (If we're okay with 4 extra bytes per fade.)

david-vanderson:
> Agreed. If a frame is longer than the fade there's nothing to render, so snapping is perfectly fine. (If we're okay with 4 extra bytes per fade.)

Can we get away with using like `t = -1` as the "we are starting" flag?  I bit more confusing but no extra bytes?

xsawyerx:
> > Agreed. If a frame is longer than the fade there's nothing to render, so snapping is perfectly fine. (If we're okay with 4 extra bytes per fade.)
> 
> Can we get away with using like `t = -1` as the "we are starting" flag? I bit more confusing but no extra bytes?

Yes, but we'll need another hint and we can use `= 2` for that, I think. It's not very readable, but it works.

```zig
const cur = dataGetPtrDefault(null, id, key, f32, 0);

// -1 and 2 mean "state changed on this frame, hold at 0 / at 1"
// That frame is usually that that woke from idle, so elapsed time is the whole pause, not the frame
// The next frame takes the first real step
const starting = (cur.* == -1 or cur.* == 2);
var t: f32 = if (cur.* == -1) 0 else if (cur.* == 2) 1 else cur.*;

if (t != target) {
    if (!starting and (t == 0 or t == 1)) {
        cur.* = if (t == 0) -1 else 2;
    } else {
        const step = currentWindow().secs_since_last_frame / secs;
        t = if (t < target) @min(target, t + step) else @max(target, t - step);
        cur.* = t;
    }
    refresh(null, @src(), id);
} else {
    cur.* = t; // clear the sentinel once settled, or a later idle re-entry snaps
}

return t;
```

Maybe you have a better way to do this?

david-vanderson:
> Maybe you have a better way to do this?

Sorry, I definitely did not understand the details, so I had to write out longer code to figure it out:
```zig
pub fn hoverFade(id: Id, hovered: bool) f32 {
    const target: f32 = if (hovered) 1 else 0;
    if (reduce_motion or hover_fade_secs == 0.0) return target;
    
    // -1 means we previously hit the target at 0
    // -2 means we previously hit the target at 1
    // >= 0 means we are currently fading 
    // Default 0 (instead of target) so a widget with the first frame already hovered
    // (e.g. a context-menu item that pops up under the cursor) fades in rather
    // than starting fully lit.
    const cur = dataGetPtrDefault(null, id, "__hover_fade", f32, -1);
    
    if ((cur.* == -1 and target == 1) or (cur.* == -2 and target == 0)) {
        // starting a new fade
        cur.* = if (target == 1) 0 else 1;
        refresh(null, @src(), id);
        return cur.*;
    }
    
    // not in a fade
    if (cur.* < 0) return target;
    
    // in the middle of a fade
    const step = currentWindow().secs_since_last_frame / hover_fade_secs;
    cur.* = if (cur.* < target) @min(target, cur.* + step) else @max(target, cur.* - step);
    if (cur.* != target) {
        refresh(null, @src(), id);
        return cur.*;
    } else {
        // finished a fade
        cur.* = if (cur.* == 0) -1 else -2; // mark as done
        return target;
    }
}
```

Does this match your understanding?

PR #966, opened 2026-08-22
Support fade in for floating windows

xsawyerx:
Assumes on #965.

Allows `FloatingWindowWidget` to set fade in from invisible. It reinstates whatever the user had it set to before.

I needed this to avoid an abrupt dialog open on an app.

This has to be inside the widget rather than in calling code, because `dvui.floatingWindow()` paints the modal dimming and the window background during setup, before the caller gets the widget back.

The first frame of a new window is the throwaway one used to measure and centre it, so the fade starts on the second frame, the first one that is actually on screen.

I put this info in the commit too.

### DIFF
--- a/src/WidgetData.zig
+++ b/src/WidgetData.zig
@@ -278,8 +278,9 @@ pub fn focusBorder(self: *const WidgetData) void {
     if (self.visible()) {
         const rs = self.borderRectScale();
         const thick = 2 * rs.s;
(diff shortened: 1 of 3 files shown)


david-vanderson:
The focus fade looks very nice!  Once we settle #965 we can merge this, and then I need to add the floatingWindow fade to the demo somewhere, and also thread it through the `dialog` stuff.  Thank you!

Issue #120, opened 2024-09-08
text outline/shadow/gradient


david-vanderson:
ryuukk on discord asked if we had text outline/shadow/gradient and gave this example:
![image](https://github.com/user-attachments/assets/f6a802a5-0b1e-49ff-8f8e-9c898ec3ee33)

The primary usecase is to make text legible when on top of arbitrary colors.  Secondary is to give more visual options.

Outline means that every glyph appears surrounded by a halo of black (hopefully any color, but at least black).  Shadow is similar, but only in a single direction (most likely bottom-right, but hopefully any direction).

ryuukk gave an implementation strategy where we bake the outline/shadow into the text rasterization process.  We can do that, but I'm worried about thick outlines overlapping the previously drawn character.

The first strategy ("copy triangles") I want to try is having renderText first render all the characters (in a single line) in the outline color multiple times offset in each direction (is up, down, left, right enough?).  Then render the normal characters on top.
Upsides:
* not having a separate rasterization
* can independently color the outline/font at render time
* optionally keep the character spacing the same (so the outlines can all run together)
* can dynamically change the shadow direction
Downsides:
* might not look good, because multiple outlines on top of each other can kill antialiasing
* thick outlines might require even more copies offset by different amounts in each direction

Text gradient is somewhat orthogonal, and is the ability to have a vertical gradient from the top of the text line (not each individual glyph) to the bottom.  I don't think a non-vertical gradient makes sense?

Right now we should be able to do this by manipulating the colors assigned to each vertex when rendering text.  If the "copy triangles" approach for outlines/shadows works, we could also have a separate vertical gradient on the outline/shadow, although I'm not sure how useful that would be.

PR #970, opened 2026-08-26
Feature/Gradient Support - REVERTED

nat3Github:
Related Issues: #923 #120
Add Gradient Support to following items:
- Text
- Icons
- Path draw fns
- Options (fill, border, text):  where supported option fields are transitioning from Color -> ColorOrGradient
```zig
color_fill: ?ColorOrGradient = null,
color_fill_hover: ?ColorOrGradient = null,
color_fill_press: ?ColorOrGradient = null,
color_text: ?ColorOrGradient = null,
color_text_hover: ?ColorOrGradient = null,
color_text_press: ?ColorOrGradient = null,
color_border: ?ColorOrGradient = null,
```
Gradient Features:
- linear Multi Stop Gradients, angle, external rect anchoring
- Radial Multi Stop Gradients (ellipse, circle), angle, anchoring, focal point
- Scattered Multi Color Gradient, individual position

Bildschirmfoto 2026-08-26 um 08 51 23

the example was added under styling where the old gradient demo was.

### DIFF
--- a/examples/raylib-ontop.zig
+++ b/examples/raylib-ontop.zig
@@ -115,7 +115,7 @@ fn colorPicker(result: *dvui.Color) void {
         defer hbox.deinit();
 
         dvui.labelNoFmt(@src(), &color_hex, .{}, .{
-            .color_text = result.*,
+            .color_text = .{ .color = result.* },
             .gravity_y = 0.5,
         });
 
--- a/examples/raylib-zig-ontop.zig
+++ b/examples/raylib-zig-ontop.zig
@@ -124,7 +124,7 @@ fn colorPicker(result: *dvui.Color) void {
         defer hbox.deinit();
 
         dvui.labelNoFmt(@src(), &color_hex, .{}, .{
-            .color_text = result.*,
+            .color_text = .{ .color = result.* },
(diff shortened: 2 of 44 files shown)


david-vanderson:
This is so cool.  I've done a first pass through the code and it looks pretty good.

`Gradient` has slices of stops - who owns that memory?  For example, in `Path.fillConvex` if we are not rendering we do `cw.addRenderCommand` but are not copying the gradient stop slices.  Should we copy them there?  Or am I missing something?

nat3Github:
> Gradient has slices of stops - who owns that memory? For example, in Path.fillConvex if we are not rendering we do cw.addRenderCommand but are not copying the gradient stop slices. Should we copy them there? Or am I missing something?

we should definitely copy them! good catch!

nat3Github:
sorry, my vscode! automatically merged this when i pressed sync (i dont understand why, but it probably was set to the wrong repo or something, edit: it accidently was set to tracking dvui instead of my fork), i reverted it from main for now, new PR is here #972, sorry for the chaos

david-vanderson:
Haha no problem, that's happened to me!

Issue #419, opened 2025-06-28
content offset for a few seconds after event (sdl on mac)


david-vanderson:
Super weird issue on mac with sdl backend:

* `zig build sdl3-app`
* open demo window, change theme to Adwaita Dark
* after an event, there is a thin (1 pixel?) border added to the whole OS window content area
* wait a few seconds and the border goes away

This also causes all the dvui content to be shifted down by a bit (1 pixel?).

nat3Github:
i can confirm the 1 pixel border in the top of the window in the adwaita dark theme (also in a lot of other themes)
but it does not go away after few seconds

david-vanderson:
> i can confirm the 1 pixel border in the top of the window in the adwaita dark theme (also in a lot of other themes) but it does not go away after few seconds

Does it ever go away?  Just retesting I noticed I see a thin (1 pixel?) border around the entire content area of the window, and then it goes away in ~2 seconds if I stop all input.

nat3Github:
we are talking about this one right?:

<img width="604" height="198" alt="Image" src="https://github.com/user-attachments/assets/59047e6c-8b30-4b6a-a0de-48a0fd4f5bfa" />

i only have it on the top, and only appears with some themes. it does not go away

david-vanderson:
That is interesting - it looks like we have a rouge border on something.  I will look to see if any of our container widgets have a border by default.

No that's not what I am seeing, and it's very frustrating because I don't see it when taking a screenshot or recording or even zooming the screen.

nat3Github:
> No that's not what I am seeing, and it's very frustrating because I don't see it when taking a screenshot or recording or even zooming the screen.

thats weird! tbh its hard for me to understand how the issue looks you are dealing with.
besides that, the issue I was experiencing is Appkit drawing a border (it only shows up with system settings dark mode enabled vs light mode) i got a PR fix for that one, but thats not what this issue is about right? if i understand correctly i cannot reproduce your issue on my macbook

david-vanderson:
> > No that's not what I am seeing, and it's very frustrating because I don't see it when taking a screenshot or recording or even zooming the screen.
> 
> thats weird! tbh its hard for me to understand how the issue looks you are dealing with. besides that, the issue I was experiencing is Appkit drawing a border (it only shows up with system settings dark mode enabled vs light mode) i got a PR fix for that one, but thats not what this issue is about right? if i understand correctly i cannot reproduce your issue on my macbook

[...]

PR #971, opened 2026-08-26
disable appkit auto titlebar

nat3Github:
see https://github.com/david-vanderson/dvui/issues/419#issuecomment-5414504154


david-vanderson:
Looks good - merge when ready.  Thanks!

PR #977, opened 2026-08-31
std3gpu: reset manage backend tracker for external cmdbuffer case/lqvqklltsmur

AndrewKraevskii:
In case of externally managed command buffer (sdl3gpu-ontop example) it was never reset which caused it to spam incorrect warnings.

Also removed calls to `renderPresent` for this example since they are already called in `win.end`


AndrewKraevskii:
huh ci is failed because it can't reach zig mirrors. Are they all down somehow?

david-vanderson:
I assume it's because zig 0.17 doesn't exist yet, and our CI doesn't know how to handle it?

PR #980, opened 2026-09-01
Change render_backend selection behaivior

jedugsem:
When implementing my vulkan App i noticed that the wio backend can be used nicely on custom a vulkan Context with few changes. Without this change the wio_backend implementation has to be copied to the app even if it could reuse its code (saves the app 500 Lines of code). The names are a bit confusing but i think they are now more clear than before. The vulkan_wio render_backend creates the vulkan context and is dependent on wio, the vulkan render_backend is the dvui_vulkan_renderer now and can also be used with the wio backend, when provided a vulkan context.

The code that is now possible without creating a custom backend when selecting .wio and .vulkan:
```
   const backend_ptr: *dvui.backend = try self.alloc.create(dvui.backend);
    backend_ptr.* = try dvui.backend.init(
        .{
            .io = self.io,
            .window = self.window,
        },
    );

    const backend_renderer: *dvui.render_backend = try self.alloc.create(dvui.render_backend);
    backend_renderer.* = try dvui.render_backend.init(
        self.alloc,
        .{
            .dev = self.device,
            .gpu_allocator = .{ .builtin = memory },
            .render_pass = .{ .static = .{
                .render_pass = self.render_pass,
                .subpass = 0,
            } },
            .graphics_queue_family_index = self.graphics_queue_index,
            .max_frames_in_flight = 4,
        },
    );

    try self.dvui_win = dvui.Window.init(@src(), self.alloc, backend_ptr.backend(&backend_renderer.*), .{});

```

Would you be interested in a example application which renders a triangle and floating window with wio and a simple vulkan context. It has about 700 lines of code and build.zig with about 170 which runs on at least linux and android.

### DIFF
--- a/src/backends/render/vulkan/renderer.zig
+++ b/src/backends/render/vulkan/renderer.zig
@@ -35,6 +35,8 @@ const fs_spv align(64) = @embedFile("dvui.frag.spv").*;
 
 const Self = @This();
 
+pub const kind: dvui.enums.RenderBackend = .vulkan_external;
+
 /// Vulkan declarations used by this renderer. Custom backends should use this
 /// re-export so their handles have the same generated Zig types.
 pub const vulkan = vk;
@@ -880,7 +882,7 @@ pub fn endFrame(self: *Self) !RecordedFrame {
 //
 const Backend = Self;
 
-pub fn begin(self: *Self, arena: std.mem.Allocator, framebuffer_size: dvui.Size.Physical) void {
+pub fn beginWithSize(self: *Self, arena: std.mem.Allocator, framebuffer_size: dvui.Size.Physical) !void {
     _ = arena; // autofix
     self.active_render_target = null;
     if (self.cmdbuf == .null_handle) @panic("dvui_vulkan_renderer: Command bufer not set! (missing beginFrame()?)");
@@ -906,7 +908,7 @@ pub fn begin(self: *Self, arena: std.mem.Allocator, framebuffer_size: dvui.Size.
     dev.cmdPushConstants(cmdbuf, self.pipeline_layout, .{ .vertex_bit = true }, 0, @sizeOf(PushConstants), &push_constants);
 }
 
-pub fn end(self: *Backend) void {
+pub fn end(self: *Backend) !void {
     defer self.frame_active = false;
     const prepass = self.finishPrepass() catch |err| {
         slog.err("failed to finalize renderer prepass: {}", .{err});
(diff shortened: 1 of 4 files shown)


david-vanderson:
Sorry I'm having trouble understanding this.  I see that currently the vulkan backend depends on wio - is the intention to make that optional?  Or is the problem that you can't currently use wio/vulkan from a dependency?

jedugsem:
There are currently two vulkan backend in the dvui codebase. One depends on the other and provides additonal features (initialization). 

The render_backend in render/vulkan/renderer.zig (simple) is a backend which provide triangles to be used with existing vulkan structures (initialized by the user). (for example in a game)
The render_backend in render/vulkan.zig (wrapper) is a wrapper around the render_backend in render/vulkan/renderer.zig (simple) it initializes the vulkan structures so vulkan can be used in normal apps.

The build system doesn't let me combine the first (simple) render_backend with the wio_backend, even if it would be really useful.

What is the problem:
When i choose wio_backend and .vulkan i get the (wrapper) render_backend.
When i choose custom_backend and .vulkan i get the (simple) render_backend. But i loose the wio_backend.

Because of this problem i had to copy the wio backend implementation to my app and provide it as a custom backend. To combine the wio backend and the (simple) render_backend.

I hope you understand it now. It is really difficult for me to explain because i dont know how to name these two vulkan backends correctly i'm talking about.

src/backends/render/vulkan/renderer.zig
jedugsem, line 1: These changes make this (simple) renderer complete, before it could not be provided as a backend.render_impl

david-vanderson:
Ah thank you very much, that helps a lot!  Give me some time to fully wrap my brain around this.  Can you link/paste the example app you mentioned?

jedugsem:
Yes here it is : https://github.com/jedugsem/vulkan-template second commit is with this dvui branch.
Interesting are mostly the functions init_qui and recordCommandBuffer. It works on android and linux but the flake.nix is really bad. I have not noticed any non working features, the floating window can correctly position itself.

jedugsem:
The app depends on wio itself only because it is needed for building the apk in build.zig.

Deins:
I see. Yeah once you want both wio and vulkan but get in between to do your own stuff there is no good option provided.
But just curious - whats missing from wio-vulkan initialization that you need to make your own context? Or its more frame pacing / swapchain management related? 

Otherwise I think your solution makes sense. 
Haven't thought about it too much, but I think it might be better to swap vulkan and vulkan_wio naming around (as RenderBackend is dvui and backend agnostic), but might be nitpicking.
```
// Dvui initializes and owns Vulkan context
vulkan,
// Application owns Vulkan context and frame lifecycle.
vulkan_external,
```

jedugsem:
I want to learn vulkan and simulate stuff for university (might include vulkan compute). Also the provided vulkan context is vulkan 1.2 or vulkan 1.3 and my phone has vulkan 1.1 where it did not work.
At the moment the (wrapper) render_backend depends on a wio window for initialization, but your naming is more future proof. I like it.
 Making both these vulkan render_backends work with pugl should be easy. The sdl backend is not seperate from the render_backend, thats more work (might look into it). On glfw the (simple) render_backend doesn't make much sense because of the callback event system it's more difficult to integrate. The (wrapper) render_backend should also be easy to make work on glfw.
 I will do the naming changes if David Vanderson doesn't have other changes in mind.

david-vanderson:
> I will do the naming changes if David Vanderson doesn't have other changes in mind.

Yes please.  Sorry I've been swamped and haven't had time to look into this, but if you and @Deins agree that's good enough for me!

Deins (approved):

jedugsem:
Rebased, after the naming change this is a non breaking change. Thank you for your patience this is my first PR.
If there interest in the example just contact me here.

david-vanderson:
@Deins Thank you!

@jedugsem Congrats on the first PR merge!  Well done!

Issue #987, opened 2026-09-15
Supporting blurring underneath windows - window render order, RenderFrontToBack


foxnne:
Hi! 

I just wanted to make an issue for this and see if you had any ideas.

Essentially the issue is wanting to render multiple floating windows, and have a mechanism for rendering whats underneath each window so its "background" can be translucent and blurred. 

I have this working currently, but needed to add a new function for `deferRender`, which just adds the draw command as a "custom" draw command type and renders it afterwards, so we can render those window's rects in-order. 

I think we do this essentially with RenderBackToFront, but that works within a single window, correct? 

Any info on this would be great, I think the effect is pretty cool!

<img width="817" height="731" alt="Image" src="https://github.com/user-attachments/assets/ea4c769d-9290-4943-8e49-903f82a0902d" />

<img width="1800" height="1169" alt="Image" src="https://github.com/user-attachments/assets/07db696f-9540-4217-8892-3bbb6fcf75a8" />

Code using this in fizzy below, let me know if you need any more info:

```zig
pub const RenderCommand = struct {
    clip: Rect.Physical,
    alpha: f32,
    snap: bool,
    kerning: bool,
    cmd: Command,

    pub const Command = union(enum) {
        text: TextOptions,
        texture: struct {
            tex: Texture,
            rs: RectScale,
            opts: TextureOptions,
        },
        pathFillConvex: struct {
            path: Path,
            opts: Path.FillConvexOptions,
        },
        pathFill: struct {
            contours: []const Path,
            opts: Path.FillOptions,
        },
        pathStroke: struct {
            path: Path,
            opts: Path.StrokeOptions,
        },
        triangles: struct {
            tri: Triangles,
            tex: ?Texture,
        },
        /// Code that runs when the queue is replayed, with rendering immediate and the clip,
        /// alpha, snap and kerning of the moment it was queued. For work that must see what
        /// the queues before it drew
        custom: Custom,
    };

    pub const Custom = struct {
        ctx: ?*anyopaque,
        draw: *const fn (ctx: ?*anyopaque) void,
    };
};
```

```zig
pub fn deferRender(ctx: ?*anyopaque, draw: *const fn (ctx: ?*anyopaque) void) void {
    const cw = dvui.currentWindow();
    if (!cw.render_target.rendering) {
        cw.addRenderCommand(.{ .custom = .{ .ctx = ctx, .draw = draw } }, false);
        return;
    }
    draw(ctx);
}
```

```zig
[...]

PR #988, opened 2026-09-15
Add custom draw command - How to render underneath windows/order 

foxnne:
See #987.

My understanding:

dvui doesn't render floating windows immediately; each subwindow accumulates commands during the frame and they're replayed in subwindow order at `Window.end`

whats underneath the window hasnt yet been drawn when we ask for the window, so we need to defer to after everything has been placed and rendered.

I'm just going to leave this as a draft as we had spoke a little about potentially doing this a different way with RenderBackToFront


david-vanderson:
Thanks for writing that, I don't think I understood previously.  `RenderBackToFront` is about reordering the drawing calls, but this is different.

The sequence is:
* render draw calls for main window
* read pixels from main window, blur, draw as background for floating window
* render draw calls for floating window

Reordering these doesn't help.  Now that I understand that, I just need to think about whether to have a dedicated "blur" render command or use the fully-general "custom" render command.

Can you think of any other situations where you'd want to run arbitrary code at render time?

foxnne:
I can't think of any other scenarios right now, I just couldn't think of a good reason to specialise it for blur, but honestly that's probably less confusing, as it really is just a callback for timing for blur.

PR #992, opened 2026-09-18
textLayout: Long lines performance

david-vanderson:
UNFINISHED, also currently has a bunch of debug stuff turned on

But far enough along that I think I understand most of the api.  The plan is to generalize `cacheLayoutBytes()` into `cacheLayoutNext()` that functions kind of like an iterator.

The idea is it returns the "current visible region" until enough bytes have been send to `addText` to go past the end of the current region.  Then in the next call it advances to the next visible region.

For long lines, I am going to put markers down every 500 pixels (or maybe 1000).  So we can detect and skip over parts of a line that are not visible.

```zig
/// Return start/end bytes for the current visible region.
///
/// Bytes to `addText` before the current region are counted and dropped.
///
/// The current region will be returned until bytes received >= region end.
/// The next call to this function will advance to the next visible region.
///
/// Note that if bytes past the current region are given to `addText`, it will
/// internally call this to advance the region.
///
/// After the last visible region, returns maxInt(usize) as both start and end.
///
/// When cache_layout is false or can't be used in a frame, returns 0,maxInt(usize).
pub fn cacheLayoutNext(self: *TextLayoutWidget) Region {
```

Usage is like this:
```zig
const normal_opts = tl.data().options.strip();
outer: while (true) {
    const cln = tl.cacheLayoutNext();
    iter.setByteRange(cln.start, cln.end); // byte range we want syntax highlighted
    while (iter.next()) |h| {
        tl.addText(h.text, h.opts orelse normal_opts);
        if (tl.bytes_seen >= cln.end) continue :outer; // need next byte range
    } else {
        break :outer; // ran out of text
    }
}
```

One part I'm less sure about is how to use when the usage code doesn't want to supply all the bytes, but only the bytes in the visible regions.  Need some way to "skip" to the start of the next region.  Can simply `tl.bytes_seen += skip` but could be better.

@foxnne How does this api look to you?


david-vanderson:
This is an alternate solution to https://github.com/david-vanderson/dvui/pull/945

foxnne:
@david-vanderson this looks good to me! Thank you so much for taking the time to look into this, I greatly appreciate it!

nat3Github:
cool stuff! showing only visible content is probably the same solution to grid's performance problems with a lot of Rows

david-vanderson:
I noticed some bugs related to varying line ascents.  Will follow up separately for those.

